PR #3, opened 2025-04-24
Adding CODEOWNERS

svij-sc:
Adding `CODEOWNERS` file


kmontemayor2-sc (commented):
LGTM - I feel like we probably want more fine-grained controls for e.g this file but we can get around to that later.

PR #4, opened 2025-04-24
OSS Initial Release

svij-sc:
Releasing code for active OSS development.

Will point public over from https://github.com/snap-research/GiGL to here soon


Makefile
```diff
@@ -0,0 +1,478 @@
+include dep_vars.env
+
+SHELL := /bin/bash
+CONDA_ENV_NAME=gnn
```
kmontemayor2-sc, line 4: btw should we name this `gigl`? to match with what we do in our cicd?

.dockerignore
```diff
@@ -0,0 +1,66 @@
...
+.bloop
+.metals
+.cache
+.lib
+tmp
+.tmp
+large_scale_gnn.egg-info/
+gigantic_graph_learning.egg-info/
+gigl.egg-info/
+gigl_dataflow_setup.egg-info/
```
kmontemayor2-sc, line 27: wdyt about removing these legacy ignores now?

python/pyproject.toml
```diff
@@ -0,0 +1,204 @@
...
+    "nbsphinx==0.9.3",
+    "sphinx-hoverxref==1.3.0",
+    "mistune>=2.0.3",
+    "nbconvert>=7.16.2",
+    "sphinx-tabs==3.4.5",
+    "sphinx-design==0.6.0"
+]
+
+[project.urls]
+Homepage = "https://github.com/research/GiGL"
```
kmontemayor2-sc, line 153: update?

kmontemayor2-sc (approved):
Thanks Shubham!

yliu2-sc (approved):

Makefile
svij-sc: we need to rename docker images too - prolly worth doing as a separate change

.dockerignore
svij-sc: done

PR #6, opened 2025-04-25
Setup labeled tensors

kmontemayor2-sc:
We do this so we can put "batches" of input data for anchor nodes, positive labels, and negative labels into GLT. 

We also need to update `NodeSamplerInput` but I have a separate PR for that.

http://go/internal-gigl-pull/492


mkolodner-sc (approved):

python/gigl/utils/data_splitters.py
```diff
@@ -258,6 +262,168 @@ def __call__(
...
+    num_negative_labels: Optional[int] = None
+
+
+def setup_labeled_input_from_dataset(
+    dataset: Dataset,
+    node_ids: torch.Tensor,
+    supervision_edge_types: Optional[Union[EdgeType, tuple[EdgeType, EdgeType]]] = None,
+    return_label_info: bool = False,
+) -> Union[torch.Tensor, tuple[torch.Tensor, LabelInfo]]:
+    """Setups up a tensor of labeled data containing anchor nodes, positive node ids, and negative negative based on provided node_ids and dataset.
```
yliu2-sc, line 277: The doc string needs a bit of clean up for readability

python/gigl/utils/data_splitters.py
```diff
@@ -258,6 +262,168 @@ def __call__(
...
+
+    Each tensor that is returned will be a 2d tensor of form:
+        [[node_id_i, [positive_node_ids_i], [negative_node_ids_i]?], ...]
```
yliu2-sc, line 280: Description doesn't match the return type is `Union[torch.Tensor, tuple[torch.Tensor, LabelInfo]]`

python/gigl/utils/data_splitters.py
```diff
@@ -258,6 +262,168 @@ def __call__(
...
+            If None, then the dataset must be homogeneous, and the edges in the dataset will be used.
+            If a single edge type is provided, then the dataset must be heterogeneous,
+            and the edge type provided will be used to generate positive labels.
+            If two edge types are provided, then the dataset must be heterogeneous,
+            the first edge type will be used to generate positive labels, and the second edge type will be used to generate negative labels.
+        return_label_info (bool): Whether to return the label info or not. Defaults to False.
+
+    Returns:
+        The returned tensor is always of shape [Number of nodes, max number of positive labels + max number negative labels].
+        If return_label_info:
```
yliu2-sc, line 315: On another thought, it might be more natural to return separate tensors for positive and negative, if you can have them in the order of node_ids. Or also return a node_ids tensor that's in order. This avoids the introduction of custom Labelnfo type and pos, neg labels can directly be used separately.

python/gigl/utils/data_splitters.py
```diff
@@ -258,6 +262,168 @@ def __call__(
...
+    if supervision_edge_types is None:
+        if isinstance(dataset.graph, Mapping):
+            raise ValueError(
+                f"No supervision edge types provided, but dataset is heterogeneous."
+            )
+        positive_node_topo = dataset.graph.topo
+        negative_node_topo = None
+    elif isinstance(supervision_edge_types, EdgeType):
+        if not isinstance(dataset.graph, Mapping):
+            raise ValueError(
```
yliu2-sc, line 329: in Homogeneous cases, we don't allow users do provide supervision edge type?

python/gigl/utils/data_splitters.py
```diff
@@ -258,6 +262,168 @@ def __call__(
...
+    # Note that GLT defaults to CSR under the hood, if this changes, we will need to update this.
+    indptr = topo.indptr  # [N]
+    indices = topo.indices  # [M]
+    starts = indptr[anchor_node_ids]  # [N]
+    ends = indptr[anchor_node_ids + 1]  # [N]
+
+    max_range = int(torch.max(ends - starts).item())
+    # Sample all labels based on the CSR start/stop indices.
+    # Then pad so the tensor can be of correct structure.
+    padded_labels = torch.stack(
```
yliu2-sc, line 413: you can probably use 
https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pad_sequence.html
or just https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html
to avoid the torch.cat

kmontemayor2-sc: Sure - though actually it's possible to vectorize the whole thing :)

Benchmarking against 100M edges with 10M anchor nodes gives:

```python benchmark.py
created dataset
using pad...
  0%|▏                                                                                                                                                                                                                                                                                                                                                                                                                                                                      | 4474/10000000 [00:00<03:43, 44728.74it/s]
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10000000/10000000 [03:56<00:00, 42240.05it/s]
pad took 299.53 seconds
using mask...
mask took 2.67 seconds
```

source: 

```py
import time
import torch
from graphlearn_torch.data import Dataset, Topology
import tqdm
from collections import deque

PADDING_NODE: torch.Tensor = torch.tensor(-1, dtype=torch.int64)


def pad(anchor_node_ids: torch.Tensor, topo: Topology) -> torch.Tensor:
    """Returns the padded labels and the max range of labels.

    Given anchor node ids and a topology, this function returns a tensor
    which contains all of the node ids that are connected to the anchor node ids.
    The tensor is padded with `PADDING_NODE` to the maximum number of labels.

    Args:
        anchor_node_ids (torch.Tensor): The anchor node ids to use for the labels. [N]
        topo (Topology): The topology to use for the labels.
    Returns:
        The shape of the returned tensor is [N, max_range].
        tuple[torch.Tensor, int]: The labels and the max range of labels.
    """
    # indptr is the ROW_INDEX of a CSR matrix.
    # and indices is the COL_INDEX of a CSR matrix.
    # See https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format)
    # Note that GLT defaults to CSR under the hood, if this changes, we will need to update this.
    indptr = topo.indptr  # [N]
    indices = topo.indices  # [M]
    starts = indptr[anchor_node_ids]  # [N]
    ends = indptr[anchor_node_ids + 1]  # [N]

    max_range = int(torch.max(ends - starts).item())
    # Sample all labels based on the CSR start/stop indices.
    # Then pad so the tensor can be of correct structure.
    padded_labels = torch.stack(
        [
            torch.nn.functional.pad(
                indices[start:end],
                (0, max_range - (end - start)),
                "constant",
                PADDING_NODE,
            )
            for start, end in tqdm.tqdm(zip(starts, ends), total=len(starts))
        ]
    )
    return padded_labels


def mask(anchor_node_ids: torch.Tensor, topo: Topology) -> torch.Tensor:
    """Returns the padded labels and the max range of labels.

    Given anchor node ids and a topology, this function returns a tensor
    which contains all of the node ids that are connected to the anchor node ids.
    The tensor is padded with `PADDING_NODE` to the maximum number of labels.

    Args:
        anchor_node_ids (torch.Tensor): The anchor node ids to use for the labels. [N]
        topo (Topology): The topology to use for the labels.
    Returns:
        The shape of the returned tensor is [N, max_range].
        tuple[torch.Tensor, int]: The labels and the max range of labels.
    """
    indptr = topo.indptr  # [N]
    indices = topo.indices  # [M]
    starts = indptr[anchor_node_ids]  # [N]
    ends = indptr[anchor_node_ids + 1]  # [N]

    max_range = int(torch.max(ends - starts).item())
    # Sample all labels based on the CSR start/stop indices.
    # Then pad so the tensor can be of correct structure.
    # Create a tensor of indices for gathering, handling varying lengths.

    # Creates "indices" for us to us, e.g [[0, 1], [2, 3]]
    ranges = starts.unsqueeze(1) + torch.arange(max_range)  # [N, max_range]
    # Clamp the ranges to be valid indices into `indices`.
    ranges.clamp_(min=0, max=ends.max().item() - 1)
    # Mask out the parts of "ranges" that are not applicable to the current label
    # filling out the rest with `PADDING_NODE`.
    mask = torch.arange(max_range) >= (ends - starts).unsqueeze(1)
    labels = torch.where(mask, torch.full_like(ranges, PADDING_NODE), indices[ranges])
    return labels


if __name__ == "__main__":
    edges = torch.stack(
        [
            torch.arange(0, 10_000_000, dtype=torch.int64).repeat(10),
            torch.randint(0, 10_000_00, (100_000_000,), dtype=torch.int64),
        ]
    )
    ds = Dataset()
    ds.init_graph(
        edge_index=edges, edge_ids=torch.arange(100_000_000), graph_mode="CPU"
    )
    print("created dataset")

    anchor_nodes = torch.arange(0, 10_000_000, dtype=torch.int64)
    print("using pad...")
    start = time.perf_counter()
    pad(anchor_nodes, ds.graph.topo)
    end = time.perf_counter()
    print(f"pad took {end - start:.2f} seconds")

    print("using mask...")
    start = time.perf_counter()
    mask(anchor_nodes, ds.graph.topo)
    end = time.perf_counter()
    print(f"mask took {end - start:.2f} seconds")
    ```

File: python/gigl/utils/data_splitters.py [hunk 6]
```diff
@@ -258,6 +262,128 @@ def __call__(
...
+            If a single edge type is provided, then the dataset must be heterogeneous,
+            and the edge type provided will be used to generate positive labels.
+            If two edge types are provided, then the dataset must be heterogeneous,
+            the first edge type will be used to generate positive labels, and the second edge type will be used to generate negative labels.
+    Returns:
+        Tuple of (positive labels, negative_labels?)
+        negative labels may be None depending on supervision_edge_types.
+        The returned tensors are of shape N x M where N is the number of nodes and M is the max number of labels, per type.
+    """
+    if supervision_edge_types is None:
```
mkolodner-sc: Had some discussion offline here about whether we should support this functionality (i.e. passing `None` in `supervision_edge_types`, or if we should instead always require users to pass in some `supervision_edge_type` here. For now, since we are converting labels to their own edge type in both the UDL and SSL cases, we will always have a heterogeneous graphs for our currently use cases, so perhaps we can be more restrictive here and only require `supervision_edge_types` to be passed in. This way, we only need to handle `dataset.graph` as an abc.Mapping, which may simplify some of our logic after calling this function as well.

python/gigl/utils/data_splitters.py
```diff
@@ -258,6 +262,128 @@ def __call__(
...
+    indptr = topo.indptr  # [N]
+    indices = topo.indices  # [M]
+    starts = indptr[anchor_node_ids]  # [N]
+    ends = indptr[anchor_node_ids + 1]  # [N]
+
+    max_range = int(torch.max(ends - starts).item())
+
+    # Sample all labels based on the CSR start/stop indices.
+    # Creates "indices" for us to us, e.g [[0, 1], [2, 3]]
+    ranges = starts.unsqueeze(1) + torch.arange(max_range)  # [N, max_range]
```
mkolodner-sc, line 369: Very interesting vectorized solution here, thanks!

mkolodner-sc (approved):
Thanks Kyle!

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -513,6 +517,133 @@ def test_check_edge_index(self, _, edges):
...
+        )
+        ds = Dataset()
+        ds.init_graph(
+            edge_index=to_heterogeneous_edge(edges),
+            edge_ids=to_heterogeneous_edge(torch.arange(edges.shape[1])),
+            graph_mode="CPU",
+        )
+
+        # TODO(kmonte): Update to use a splitter once we've migrated splitter API.
+        node_ids = torch.tensor([9, 10])
```
yliu2-sc, line 535: why not do [9, 10, 11, 12]  to make sure more cases are covered?

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -513,6 +517,133 @@ def test_check_edge_index(self, _, edges):
...
+        a_to_b = EdgeType(_NODE_A, _TO, _NODE_B)
+        a_to_c = EdgeType(_NODE_A, _TO, _NODE_C)
+        edges = {
+            a_to_b: torch.tensor(
+                [
+                    [10, 11, 12, 13, 10],
+                    [10, 10, 11, 11, 12],
+                ]
+            ),
+            a_to_c: torch.tensor(
```
yliu2-sc, line 556: do indices overlap between node types?

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -513,6 +517,133 @@ def test_check_edge_index(self, _, edges):
...
+            edge_index=edges,
+            edge_ids={e: torch.arange(edges[e].shape[1]) for e in edges},
+            graph_mode="CPU",
+        )
+        # TODO(kmonte): Update to use a splitter once we've migrated splitter API.
+        node_ids = torch.tensor([10])
+        positive, negative = get_labels_for_anchor_nodes(
+            dataset=ds, node_ids=node_ids, supervision_edge_types=a_to_b
+        )
+        expected = torch.tensor([[10, 11]])
```
yliu2-sc, line 574: shouldn't this be [10, 12] ?

yliu2-sc (approved):
Overall logic looks ok, minor comments to address before merge.

PR #8, opened 2025-04-25
Support batching input in our neighbor loader

kmontemayor2-sc:
This allows us to sample against multiple nodes in the same "batch".

Moving over http:://go/internal-gigl-pull/496

### DIFF
--- a/python/gigl/distributed/distributed_neighborloader.py
+++ b/python/gigl/distributed/distributed_neighborloader.py
@@ -58,6 +58,8 @@ class DistNeighborLoader(DistLoader):
         channel_size: str = "4GB",
         process_start_gap_seconds: int = 60,
         num_cpu_threads: Optional[int] = None,
+        shuffle: bool = False,
+        drop_last: bool = False,
         _main_inference_port: int = DEFAULT_MASTER_INFERENCE_PORT,
         _main_sampling_port: int = DEFAULT_MASTER_SAMPLING_PORT,
     ):
@@ -105,6 +107,8 @@ class DistNeighborLoader(DistLoader):
             num_cpu_threads (Optional[int]): Number of cpu threads PyTorch should use for CPU training/inference
                 neighbor loading; on top of the per process parallelism.
                 Defaults to `2` if set to `None` when using cpu training/inference.
+            shuffle (bool): Whether to shuffle the input nodes. (default: ``False``).
+            drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
             _main_inference_port (int): WARNING: You don't need to configure this unless port conflict issues. Slotted for refactor.
                 The port number to use for inference processes.
                 In future, the port will be automatically assigned based on availability.
@@ -200,8 +204,8 @@ class DistNeighborLoader(DistLoader):
             sampling_type=SamplingType.NODE,
             num_neighbors=num_neighbors,
             batch_size=batch_size,
-            shuffle=False,
-            drop_last=False,
+            shuffle=shuffle,
+            drop_last=drop_last,
             with_edge=True,
             collect_features=True,
             with_neg=False,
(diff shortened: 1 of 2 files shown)


python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -66,6 +76,56 @@ def test_distributed_neighbor_loader(self):
...
+            local_process_world_size=1,
+        )
+        count = 0
+        for datum in loader:
+            self.assertIsInstance(datum, HeteroData)
+            count += 1
+
+        self.assertEqual(count, 1)
+        assert_close(datum[nt].node.msort(), torch.tensor([10, 11, 12, 14]))
+        assert_close(datum[nt].batch.msort(), torch.tensor([10, 12]))
```
Mkolodne, line 127: Nit: Can we use `assert_tensor_equality` here [instead](https://github.com/Snapchat/GiGL/blob/main/python/tests/test_assets/distributed/utils.py#L6)? Functionality-wise would be the same but may add some more consistency across our codebase

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -66,6 +76,56 @@ def test_distributed_neighbor_loader(self):
         # https://paperswithcode.com/dataset/cora
         self.assertEqual(count, 2708)
 
+    def test_distributed_neighbor_loader_batched(self):
+        nt = DEFAULT_HOMOGENEOUS_NODE_TYPE
```
Mkolodne, line 80: nit: can we use node_type instead of nt here? Ditto elsewhere for abbreviated vars

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -66,6 +76,56 @@ def test_distributed_neighbor_loader(self):
...
+        positive_edge = EdgeType(nt, POSITIVE_LABEL_RELATION, nt)
+        negative_edge = EdgeType(nt, NEGATIVE_LABEL_RELATION, nt)
```
Mkolodne, line 82: nit: positive_edge_type, negative_edge_type?

Mkolodne (approved):
Thanks a lot Kyle! LGTM :)

python/tests/unit/distributed/distributed_neighborloader_test.py
kmontemayor2-sc: Done, thanks.

python/tests/unit/distributed/distributed_neighborloader_test.py
kmontemayor2-sc: done.

mkolodner-sc (approved):
Thanks!

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -179,17 +196,28 @@ def __init__(
...
             with_edge=True,
-            edge_dir=dataset.edge_dir,
             collect_features=True,
-            to_device=device,
-            worker_options=worker_options,
+            with_neg=False,
+            with_weight=False,
+            edge_dir=dataset.edge_dir,
+            seed=None,  # it's actually optional - None means random.
         )
```
zfan3-sc, line 211: would be good to pass `**kwargs` straight to the base class for parameters that's transparent to the child class

kmontemayor2-sc: I've had bad experiences with `**kwargs` in the past silently dropping stuff due to typos and I prefer to avoid them.

Also in this specific case we already pass all args to `DistLoader`. Xiaobin mentioned offline that we can pass more stuff in to `SamplingConfig` which I've added some of - but IME we should make those explicit params in the future as needed.

zfan3-sc: okay if we keep inheritance shallow it's good. otherwise when base class introduces a new param we need to touch every class along the inheritance chain.

PR #9, opened 2025-04-26
Update VAI envvar assumptions, add torch convenience env vars

nshah-sc:
- Add convenience functions to grab `LOCAL_RANK` and `LOCAL_WORLD_SIZE` from torch distributed launchers
- Update env vars to gracefully fallback to reasonable values when not set by VAI (these don't get set in single-replica jobs)


python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -62,7 +62,7 @@ def get_world_size() -> int:
     It is the user's responsibility to create multiple processes per machine.
     Throws if not on Vertex AI.
     """
-    return int(os.environ["WORLD_SIZE"])
+    return int(os.environ.get("WORLD_SIZE", 1))
```
kmontemayor2-sc, line 65: hmmm iirc we had some intention of this throwing if called when not on VAI, though maybe that behavior is odd and we should have it also work locally/etc...

@svij-sc do you remember if we were planning to do this? Or just have some unified `get_trianing_context` function that'd return a `DistContext` depending if on VAI/Local/KFP/etc and we get users to use *that*

kmontemayor2-sc (approved):
Generally LGTM but I'm curious what @svij-sc has to say about my other comment

python/gigl/common/utils/vertex_ai_context.py
svij-sc: IMO, if these are not set then something has gone wrong and we should probably error out instead.

@nshah-sc , is the intention here solely to tackle problem w/ trainng/infering on VAI with world_size of 1? Or is this for local training/inference? 

If its the later case then an alternative solution we should consider, as @kmontemayor2-sc pointed out, is providing first class support for other types of train/infer runners i.e. Local with some "get_dist_context" utility.

svij-sc (changes requested):
Blocking till above comment has a consensus.

python/gigl/common/utils/vertex_ai_context.py
kmontemayor2-sc: A quick solution here is to make these functions work like:

```
def get_rank():
   if not on_vai():
      raise Error()
   return os.env.get("RANK," 1)
```

nshah-sc: none of the above functions currently throw if not on VAI.  we can add this throwing logic if necessary but the current functions will break if run on VAI with a single-node job, hence the updates made here.  

> IMO, if these are not set then something has gone wrong and we should probably error out instead
@svij-sc these variables are not set by VAI if running a single node job, which causes this os.environ[] access pattern to break.  There was an implicit assumption in our code beforehand that this would get set on all VAI jobs, but that is not true.

kmontemayor2-sc: > adding throwing logic in the VAI functions -- we can add this check, but none of the functions in vertex_ai_context.py currently throw if not on VAI today to be clear.

My understanding was that we used `os.environ[]` to implicitly check if something was on VAI - the assumption being if it wasn't then we'd error out when these functions were called (which is it's own problem - it'd be a descriptor error message).

Again the code I suggested [here](https://github.com/Snapchat/GiGL/pull/9#discussion_r2064042407) gets us to both throwing when. not on VAI and solving the one node VAI issue.  

Note we could make the return types `Optional` here - `get_leader_hostname` doesn't really have a sensible "default" for us to default to if we're not on VAI, but then we need to check downstream if the return values are None which is a bit of a hassle.

python/gigl/common/utils/torch_training.py
```diff
@@ -26,6 +26,24 @@ def get_rank() -> int:
...
+
+def get_local_world_size() -> int:
+    """
+    This is automatically set by Kubeflow PyTorchJob launcher
+    Returns:
+        int: Total number of local (within-node) processes involved in distributed training
+    """
+    return int(os.environ.get("LOCAL_WORLD_SIZE", 1))
+
+
```
svij-sc, line 46: Does this actually get set by something automatically?

IIUC right now we expect the user to manage the local rank and world size since we are not automatically spinning up those proceses for them using torch run, et al (which would traditionally set those variables), thus I would imagine these are are probably not getting populated by default and thus setting default values here is likely a vector of future bugs?

For exposing this info, @mkolodner-sc and I were thinking about potentially adhering to https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_context.py#L33

Similar to pyg: https://pytorch-geometric.readthedocs.io/en/2.5.0/modules/distributed.html#torch_geometric.distributed.DistContext

If we want to introduce these maybe we need something like `torchrun_context.py` to avoid overloading?

svij-sc (approved):
approving to unblock since some consensus was reached offline re: adding assertions for `vertex_ai_context`.
Added a followup question for `torch_training.py`.

python/gigl/common/utils/torch_training.py
nshah-sc: Some launchers do set this property.  e.g. as you said, [torchrun](https://pytorch.org/docs/stable/elastic/run.html#environment-variables).  The challenge is not all do, e.g. VAI.  Hence having graceful fallbacks when calling this util under an uncertain launcher environment is important IMO.  In this case for example, I don't believe we make strong assumptions that the caller of this function (`torch_training.get_local_world_size`) is running in VAI context or some other context.

IIUC your concern is that if the user intends to run on VAI, and something that should 100% of the time be set on VAI isn't set, it should fail.  I agree with this concern. But this shouldn't be the responsibility of a "generic" torch distributed utility which exposes the environment variables used for distributed setup (I think torch_training.py falls under this category).  I'm OK with this behavior in vertexai_context.py though.  I think we're aligned.

re: the kind of context you're talking about, I could see this as a convenience layer exposed in `torch_training.py` since I think inferring this kind of context would be done from environment variables anyway, and if we intend to call this on environments which might be VAI, torchrun, etc... we'll have to account for the disparity in behaviors there.

svij-sc: >  Hence having graceful fallbacks when calling this util under an uncertain launcher environment is important IMO.

This is the exact problem.
If implemented someone will use it in VAI context and expect it to be populated correctly but in that case it will always be incorrect since it will default to `0`, `1` respectively. Having a generic utility work for some cases and not expected to work for others is an anti-pattern.

> re: the kind of context you're talking about, I could see this as a convenience layer exposed in torch_training.py since I think inferring this kind of context would be done from environment variables anyway, and if we intend to call this on environments which might be VAI, torchrun, etc... we'll have to account for the disparity in behaviors there.

correct.
We could have `vai.get_dist_context`, `torch_run.get_dist_context` , et al. Or even a "get_dist_context" that returns the relevent context based on where it might be running - this needs to be thought through more and designed respectively. Although, in this case the user is responsible for using the right function to build the right context based on how their dist training is setup. What we are not doing is populating a value that is expected to be incorrect w/ VAI but correct w/ torchrun, et al.

nshah-sc: > generic utility work for some cases and not expected to work for others is an anti-pattern
Here is where we are misaligned -- I believe this is a very typical fallback for all distributed launcher settings.  

setting this in VAI does not give me pause because this is the implicit setting when the variables are not set (i.e. single process DDP implies `RANK=0 WORLD_SIZE=1`)

PR #5, opened 2025-04-25
[FIX] GiGL Component Documentation Table

shubhamvij:
**Scope of work done**

The Component table was broken after OSS migration - this fixes the formatting.
Update: I found out this is due to our recently introduced `mdformat` linter - by default it doesn't consider tables and breaks their formatting by inserting new lines.

As a result, I had to add a mdformat plugin for correctly formatting markfown tables.

Prior:
Screenshot 2025-04-24 at 7 17 34 PM

W/ Fix:

Screenshot 2025-04-24 at 7 17 25 PM

Where is the documentation for this feature?: N/A

Did you add automated tests or write a test plan?

***Updated Changelog.md?*** NO

***Ready for code review?:*** NO

PR #10, opened 2025-04-28
Lock pip version to fix broken pip-compiles. 

svij-sc:
**Scope of work done**

New `pip` [version](https://pypi.org/project/pip/25.1/) `25.1` seems to have introduced a regression where `pip-compile` is failing.
Example: https://github.com/Snapchat/GiGL/actions/runs/14714954906/job/41296087148

Locking the pip version allows `pip-compile` to continue to work i.e. https://github.com/Snapchat/GiGL/actions/runs/14714954906/job/41296087498

This is change is currently blocking https://github.com/Snapchat/GiGL/pull/5

### DIFF
--- a/containers/Dockerfile.builder
+++ b/containers/Dockerfile.builder
@@ -40,7 +40,7 @@ RUN echo 'export PATH="/tools/google-cloud-sdk/bin:/usr/lib/jvm/java-1.11.0-open
 RUN echo 'export JAVA_HOME="/usr/lib/jvm/java-1.11.0-openjdk-amd64"' >> /root/.bashrc
 
 # Create the environment:
-# TODO: (svij) Build env using single entrypoing for better maintainability `make initialize_environment` 
+# TODO: (svij) Build env using single entrypoint `make initialize_environment` for better maintainability
 RUN conda create -y --name gigl python=3.9 pip
 
 # Update path so any call for python executables in the built image defaults to using the gnn conda environment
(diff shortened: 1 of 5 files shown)


containers/Dockerfile.builder
```diff
@@ -40,6 +40,7 @@ RUN echo 'export PATH="/tools/google-cloud-sdk/bin:/usr/lib/jvm/java-1.11.0-open
 RUN echo 'export JAVA_HOME="/usr/lib/jvm/java-1.11.0-openjdk-amd64"' >> /root/.bashrc
 
 # Create the environment:
+# TODO: (svij) Build env using single entrypoint `make initialize_environment` for better maintainability
 RUN conda create -y --name gigl python=3.9 pip
```
mkolodner-sc, line 44: dumb q: what's the reason why we don't need to specify pip version here?

mkolodner-sc (approved):
Thanks!

containers/Dockerfile.builder
svij-sc: AFAIK, only pip-compile is broken but the real reason is I want to refactor this anyways soon as per TODO - thus left it as a future exercise.

svij-sc:
> Approval, also do you think we should file a bug against pip tools?

Yea., thats the plan. There are some related bugs I found reported already but none specific to this.
I want to create a min repro example and file one, but don't want to block us interim.

PR #12, opened 2025-04-30
Swap to use 'CoraFromGCS'

kmontemayor2-sc:
Do this so we don't get rate limited downloading in CI/CD


python/gigl/src/applied_tasks/test_tasks/academic.py
```diff
@@ -1,11 +1,10 @@
-from torch_geometric.datasets.planetoid import Planetoid
-
 from gigl.common.logger import Logger
+from gigl.src.mocking.lib.pyg_datasets_forks import CoraFromGCS
 
 logger = Logger()
 
 
-def log_stats_for_pyg_planetoid_dataset(dataset: Planetoid):
+def log_stats_for_pyg_planetoid_dataset(dataset: CoraFromGCS):
```
svij-sc, line 7: CoraFromGCS subclasses `Planetoid` ; any reason we can't keep the typing information there?

python/gigl/src/applied_tasks/test_tasks/academic.py
```diff
@@ -34,7 +33,7 @@ def log_stats_for_pyg_planetoid_dataset(dataset: Planetoid):
     logger.info(f"Is undirected: {data.is_undirected()}")
 
 
-def get_pyg_cora_dataset(store_at: str = "/tmp/Cora") -> Planetoid:
+def get_pyg_cora_dataset(store_at: str = "/tmp/Cora") -> CoraFromGCS:
```
svij-sc, line 36: Same here re: keeping typing information to Planetoid

python/gigl/src/applied_tasks/test_tasks/academic.py
kmontemayor2-sc: Sure - done in both cases :)

python/gigl/src/applied_tasks/test_tasks/academic.py
kmontemayor2-sc: Done.

svij-sc (approved):
Thanks

PR #14, opened 2025-04-30
Implement range-based partitioner

mkolodner-sc:
**Scope of work done**

- Implements the range-based partitioner class and updates corresponding tests

### DIFF
--- a/python/gigl/distributed/dist_link_prediction_range_partitioner.py
+++ b/python/gigl/distributed/dist_link_prediction_range_partitioner.py
@@ -1,0 +1,358 @@
+import gc
+import time
+from typing import Dict, Optional, Union
+
+import torch
+from graphlearn_torch.distributed.rpc import all_gather
+from graphlearn_torch.partition import PartitionBook, RangePartitionBook
+from graphlearn_torch.utils import convert_to_tensor
+
+from gigl.common.logger import Logger
+from gigl.distributed.dist_link_prediction_data_partitioner import (
+    DistLinkPredictionDataPartitioner,
+)
+from gigl.distributed.utils.partition_book import get_ids_on_rank
+from gigl.src.common.types.graph_data import EdgeType, NodeType
+from gigl.types.graph import FeaturePartitionData, GraphPartitionData, to_homogeneous
+
+logger = Logger()
+
+
+class DistLinkPredictionRangePartitioner(DistLinkPredictionDataPartitioner):
+    """
+    This class is responsible for implementing range-based partitioning. Rather than using a tensor-based partition
+    book, this approach stores the upper bound of ids for each rank. For example, a range partition book [4, 8, 12]
+    stores edge ids 0-3 on the 0th rank, 4-7 on the 1st rank, and 8-11 on the 2nd rank. While keeping the same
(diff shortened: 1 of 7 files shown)


python/gigl/distributed/dist_link_prediction_range_partitioner.py
```diff
@@ -0,0 +1,358 @@
...
+        tuple[GraphPartitionData, Optional[FeaturePartitionData], PartitionBook],
+        tuple[
+            Dict[EdgeType, GraphPartitionData],
+            Optional[Dict[EdgeType, FeaturePartitionData]],
+            Dict[EdgeType, PartitionBook],
+        ],
+    ]:
+        """
+        Partitions edges of a graph, including edge indices and edge features. If heterogeneous, partitions edges
+        for all edge type. Must call `partition_node` first to get the node partition book as input. The difference
```
kmontemayor2-sc, line 280:
```suggestion
        for all edge types. You must call `partition_node` first to get the node partition book as input. The difference
```

nit.

python/gigl/distributed/dist_link_prediction_range_partitioner.py
```diff
@@ -0,0 +1,358 @@
...
+            sorted_node_ids_indices
+        ]
+        return FeaturePartitionData(feats=partitioned_node_features, ids=None)
+
+    def _partition_edge_index_and_edge_features(
+        self,
+        node_partition_book: Dict[NodeType, PartitionBook],
+        edge_type: EdgeType,
+    ) -> tuple[GraphPartitionData, Optional[FeaturePartitionData], PartitionBook]:
+        r"""Partition graph topology of a specific edge type. For range-based partitioning, we partition
```
kmontemayor2-sc, line 142: btw I don't think this needs to be a raw string but non-blocking. '

Though let's be consistent with doc comments like:

```
"""comment on first line
"""

vs.

"""
comment on second line
"""
```

python/gigl/distributed/dist_link_prediction_range_partitioner.py
```diff
@@ -0,0 +1,358 @@
...
+        Returns:
+            FeaturePartitionData: Ids and Features of input nodes
+        """
+        features_partition_data = super()._partition_node_features(
+            node_partition_book=node_partition_book, node_type=node_type
+        )
+        # The parent class always returns ids in the feature_partition_data, but we don't need to store the partitioned node feature ids for
+        # range-based partitioning, since this is available from the node partition book.
+        assert features_partition_data.ids is not None
+        _, sorted_node_ids_indices = torch.sort(features_partition_data.ids)
```
kmontemayor2-sc, line 131: btw why not `msort` here that way we don't generate the index tensor?

kmontemayor2-sc (approved):

python/gigl/distributed/dist_link_prediction_range_partitioner.py
mkolodner-sc: IIUC the index tensor is what we need, rather than the sorted `features_partition_data.ids`, since this is how we re-order the node features.

PR #16, opened 2025-04-30
Setup our loader to "automatically" support supervised edges -> homogeneous data

kmontemayor2-sc:
Final API here is still moderately up in the air - e.g.  should we auto-infer `supervision_edge_types`?


python/gigl/utils/data_splitters.py
```diff
@@ -258,6 +262,115 @@ def __call__(
             return splits[DEFAULT_HOMOGENEOUS_NODE_TYPE]
 
 
+def get_labels_for_anchor_nodes(
+    dataset: Dataset,
+    node_ids: torch.Tensor,
+    supervision_edge_types: Union[PyGEdgeType, tuple[PyGEdgeType, PyGEdgeType]],
```
kmontemayor2-sc, line 268: Since we only intend for this to get called by the loader, which uses pyg edge types, it's best to swap to them here too.

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -40,6 +43,118 @@ def __getitem__(
...
+        # Then we strip out the anchor node ids, which are the first column of the per_batch_view
+        # And get:
+        # [
+        #   [positive_label_0, negative_label_0],
+        #   [positive_label_1, negative_label_1]
+        # ]
+        without_anchors = per_batch_view[:, 1:]
+        homogeneous_data.y = without_anchors
+        # Which become our labels.
+        return homogeneous_data
```
mkolodner-sc, line 155: Looking back at the code now, it seems that we don't even really use the `self._positive_label_edge_type` and `self._negative_label_edge_type` in this code, other than to filter down to only the message passing edge types. Initially, my thinking was that we'd use these edge types to select and store the positive node ids and negative node ids per anchor node. However, if we're relying on the `node_ids` to specify the positive and negative labels, this makes we wonder if we need to store the `edge_types` in the `dataset` at all after we do [this step](https://github.com/Snapchat/GiGL/blob/c8fd5be3b74da6b23c198d267af9af6cf8f67244/python/gigl/distributed/distributed_neighborloader.py#L361-L364) for the UDL case, making it so that we can instead just use a Data object instead of HeteroData. I suppose it comes down to how this edge_type would be used in the `SSL` case if we want to have a unified interface between UDL and SSL. Maybe we can chat more on this tomorrow and see if there's any simplification we can make here for UDL.

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -163,6 +172,107 @@ def test_distributed_neighbor_loader_heterogeneous(self):
...
+            num_neighbors=[2],
+            input_nodes=(node_type, torch.tensor([10])),
+            context=self._context,
+            local_process_rank=0,
+            local_process_world_size=1,
+            supervision_edge_types=supervision_edge_types,
+        )
+
+        count = 0
+        for datum in loader:
```
mkolodner-sc, line 261: qq: What is `datum` abbreviated for?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -163,6 +172,107 @@ def test_distributed_neighbor_loader_heterogeneous(self):
...
+            count += 1
+
+        self.assertEqual(count, 1)
+        assert_tensor_equality(
+            datum.node,
+            expected_node,
+            dim=0,
+        )
+        assert_tensor_equality(datum.y, expected_y)
+        dsts, srcs, *_ = datum.coo()
```
mkolodner-sc, line 272: nit: Is there a reason for the `*` here

Also did we ever figure out the reason why its reversed? If we change `edge_dir="in"`, what is the behavior in that case?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -163,6 +172,107 @@ def test_distributed_neighbor_loader_heterogeneous(self):
...
+            partitioned_node_features=None,
+            partitioned_negative_labels=None,
+            partitioned_positive_labels=None,
+        )
+        dataset = DistLinkPredictionDataset(rank=0, world_size=1, edge_dir="out")
+        dataset.build(partition_output=partition_output)
+
+        loader = DistNeighborLoader(
+            dataset=dataset,
+            num_neighbors=[2],
```
mkolodner-sc, line 252: Will this perform a fanout for each anchor node from the `positive_label` and `negative_label` edges? For UDL, I imagine we wouldn't want this behavior.

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -163,6 +172,107 @@ def test_distributed_neighbor_loader_heterogeneous(self):
...
+            input_nodes=(node_type, torch.tensor([10])),
```
mkolodner-sc, line 253: Does a user have to provide a `node_type` here? I imagine for a homogeneous dataset the preference is to not have to worry about `node` or `edge` types altogether if possible

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -109,6 +225,16 @@ def __init__(
                 Defaults to `2` if set to `None` when using cpu training/inference.
             shuffle (bool): Whether to shuffle the input nodes. (default: ``False``).
             drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
+            supervision_edge_types (Union[EdgeType, Tuple[EdgeType, EdgeType]]): The edge types to use as labels for supervised training.
```
mkolodner-sc, line 228: Referencing the current PR description, my view is we shouldn't have users pass this in and should auto-infer this instead. Referencing PyG's API [[1](https://pytorch-geometric.readthedocs.io/en/2.6.0/_modules/torch_geometric/distributed/dist_neighbor_loader.html)], they don't seem to have any notion of an `EdgeType` argument

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -219,10 +277,56 @@ def __init__(
...
+                ):
+                    positive_label_edge_type = edge_type
+                elif (
+                    edge_type[1] == NEGATIVE_LABEL_RELATION
+                    and edge_type[0] == message_passing_edge_type[0]
+                    and edge_type[2] == message_passing_edge_type[2]
+                ):
+                    negative_label_edge_type = edge_type
+            supervision_edge_types = (
+                (positive_label_edge_type, negative_label_edge_type)
```
kmontemayor2-sc, line 303: This part is a bit hacky - I'm working on a follow up to abstract these things to util functions :)

PR #17, opened 2025-04-30
Address Partitioner Tech Debt

mkolodner-sc:
**Scope of work done**

- Rename partitioner `DistLinkPredictionDataPartitioner` -> `DistLinkPredictionPartitioner` for easier readability and better conformity to GLT's naming
- Make calling `share_memory` on a `RangePartitionBook` a no-op, since moving the `partition_bounds` to shared memory doesn't actually work in practice, as GLT's native `RangePartitionBook` has no `ForkingPickler` method.

### DIFF
--- a/python/gigl/utils/share_memory.py
+++ b/python/gigl/utils/share_memory.py
@@ -24,18 +24,20 @@ def share_memory(
     Calling `share_memory_()` on an empty tensor may cause processes to hang, although the root cause of this is currently unknown. As a result,
     we opt to not move empty tensors to shared memory if they are provided.
 
+    When calling `share_memory` on a RangePartitionBook, we don't need to move the partition bounds to shared memory, since GLT doesn't natively
+    provide a ForkingPickler registration method for the `RangePartitionBook`, and the cost of not moving this to shared memory is minimal,
+    since the size of this tensor is very small, being equal in length to the number of machines.
+
     Args:
-        entity (Optional[Union[torch.Tensor, Dict[_KeyType, torch.Tensor]]]):
+        entity (Optional[Union[torch.Tensor, PartitionBook, Dict[_KeyType, torch.Tensor], Dict[_KeyType, PartitionBook]]]):
             Homogeneous or heterogeneous entity of tensors which is being moved to shared memory
     """
 
-    if entity is None:
+    if entity is None or isinstance(entity, RangePartitionBook):
         return None
     elif isinstance(entity, abc.Mapping):
         for entity_tensor in entity.values():
             share_memory(entity_tensor)
-    elif isinstance(entity, RangePartitionBook):
-        share_memory(entity.partition_bounds)
     else:
         # If the tensor has a dimension which is 0, it is an empty tensor. As a result, we don't move this
         # to shared_memory, since share_memory_() is unsafe on empty tensors, which may cause processes to hang.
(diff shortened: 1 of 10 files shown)


python/gigl/utils/share_memory.py
```diff
@@ -24,21 +24,23 @@ def share_memory(
...
         for entity_tensor in entity.values():
             share_memory(entity_tensor)
-    elif isinstance(entity, RangePartitionBook):
-        share_memory(entity.partition_bounds)
     else:
         # If the tensor has a dimension which is 0, it is an empty tensor. As a result, we don't move this
         # to shared_memory, since share_memory_() is unsafe on empty tensors, which may cause processes to hang.
         if 0 in entity.shape:
-            return None
+            return
```
kmontemayor2-sc, line 45: qq why `return None` -> `return`? I feel like the former is more accurate given our function signature returns `None`.

python/tests/test_assets/distributed/run_distributed_dataset.py
```diff
@@ -1,11 +1,9 @@
-from typing import Literal, MutableMapping, Optional, Type
+from typing import Literal, MutableMapping, Optional
```
kmontemayor2-sc: Don't we still need `Type` here?

python/tests/unit/utils/share_memory_test.py
```diff
@@ -47,7 +47,7 @@ def test_share_memory(
         if isinstance(entity, torch.Tensor):
             self.assertTrue(entity.is_shared())
         elif isinstance(entity, RangePartitionBook):
-            self.assertTrue(entity.partition_bounds.is_shared())
+            self.assertFalse(entity.partition_bounds.is_shared())
```
kmontemayor2-sc, line 50: probably best to add a comment here for why this is - otherwise it'd be a bit odd reading this in the future if you don't have context.

kmontemayor2-sc (approved):
Thanks Matt!

python/gigl/utils/share_memory.py
mkolodner-sc: Functionally they both are the same, the difference is only semantic. It was brought up to me that return `None` in a line of code may feel confusing, implying that we may return something other than `None` elsewhere, which is why I've opted to switch to just `return`, making it explicit here that there is no return value in this function.

python/tests/test_assets/distributed/run_distributed_dataset.py
mkolodner-sc: Yup, added it back, whoops 😄

python/gigl/utils/share_memory.py
kmontemayor2-sc: > It was brought up to me that return None in a line of code may feel confusing

This is surprising to me, `return None` explicitly returning `None` seems straightforward to me. In other languages (e.g. java) `return null` vs `return` *are* semantically different right, so I feel like since ultimately the return value *is* `None` we should be explicit about that

kmontemayor2-sc: meta comment -did you close this before or after my second comment? If I comment on a closed thread - does it stay closed? Github is strange...

mkolodner-sc: > meta comment -did you close this before or after my second comment? If I comment on a closed thread - does it stay closed? Github is strange...

Closed this before second comment, seems like its open now though, so commenting on closed thread does open it then?

PR #18, opened 2025-05-01
swap to generating label edge types based on message passing edge types

kmontemayor2-sc:
As you've mentioned prior, we should do this eventually - I think this will make our supervised HeteroData -> Data conversion easier.

### DIFF
--- a/python/gigl/types/graph.py
+++ b/python/gigl/types/graph.py
@@ -22,6 +22,7 @@ DEFAULT_HOMOGENEOUS_EDGE_TYPE = EdgeType(
 _POSITIVE_LABEL_TAG = "gigl_positive"
 _NEGATIVE_LABEL_TAG = "gigl_negative"
 
+_EdgeType = TypeVar("_EdgeType", EdgeType, tuple[str, str, str])
 # TODO(kmonte, mkolodner): Move SerializedGraphMetadata and maybe convert_pb_to_serialized_graph_metadata here.
 
 
@@ -146,7 +147,9 @@ class LoadedGraphTensors:
         self.negative_label = None
 
 
-def message_passing_to_positive_label(message_passing_edge_type: EdgeType) -> EdgeType:
+def message_passing_to_positive_label(
+    message_passing_edge_type: _EdgeType,
+) -> _EdgeType:
     """Convert a message passing edge type to a positive label edge type.
 
     Args:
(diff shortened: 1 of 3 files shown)


python/gigl/types/graph.py
```diff
@@ -155,6 +146,68 @@ def treat_labels_as_edges(self) -> None:
         self.negative_label = None
 
 
+def message_passing_to_positive_label(message_passing_edge_type: EdgeType) -> EdgeType:
```
mkolodner-sc, line 149: Quick question -- I notice here we are using GiGL edge Type, but in the previous PR we were using `torch_geometric`'s edge type of `Tuple[str, str, str]`. Should we strive for better consistency here? At least for me I wouldn't mind using `GiGL` edge type across the PRs for now, and we can scope the migration from GiGL `EdgeType` to Torch_geometric `EdgeType` as a follow up once a lot of the trainer stuff is more scoped out. WDYT?

mkolodner-sc (approved):
Thanks a lot Kyle!

python/gigl/types/graph.py
kmontemayor2-sc: We should probably migrate to only GiGL EdgeType - though I strongly suspect it will be very hard to actually fully migrate (due to `GiGLEdgeType.__repr__` and us putting it into format strings for like, bq tables [1] and the default tuple str representation is not a valid bq table name.

I think we should migrate to only GiGL edge type for now including in our loader input [2] which is imo why we have this mixup currently, though how about we do so as a follow up PR? 

And the migration is even more complicated because we can't rely on MyPy to check the types due to us ignoring all pyg imports so we we use the actual pyg type def it wont be checked.


[1]: https://github.com/Snapchat/GiGL/blob/ba46b542ffc3452c2468b451d1293d9b6f2d75b5/python/gigl/src/data_preprocessor/lib/enumerate/utils.py#L63

[2]: https://github.com/Snapchat/GiGL/blob/ba46b542ffc3452c2468b451d1293d9b6f2d75b5/python/gigl/distributed/distributed_neighborloader.py#L47

kmontemayor2-sc: Actually - since the migration is going to be such a struggle - and we really shouldn't be exposing `GiGLEdgeType` anymore than we should, I can swap this to PyG edge type (or both - it doesn't really matter in this case right?)

mkolodner-sc: Sounds good, I imagine at some point we'll need to do a pass through the `distributed` folder and migrate to use `torch_geometric` types when possible (I have a [TODO](https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/distributed_neighborloader.py#L18) for this here), so no need to be blocked on this PR for this, although it'd be easier I imagine to swap to PyG's typing if this is being used subsequently by this [PR](https://github.com/Snapchat/GiGL/pull/16). 

Thanks!

PR #20, opened 2025-05-01
E2E Heterogeneous GLT Inference [2/2]: Inference Loop

mkolodner-sc:
**Scope of work done**

- Sets up heterogeneous inference logic and relevant logic in the homogeneous inference example
- Updates default values of DistNeighborLoader for `process_start_gap_seconds`

We opt to infer on each node type + corresponding dataloader in their own inference process. This enables a cleaner solution overall and minimal memory/compute competition between dataloaders in the same inference job. 

Example Successful Test Run
- [VAI Pipeline](https://console.cloud.google.com/vertex-ai/pipelines/locations/us-central1/runs/dblp-glt-test-on--78818?inv=1&invt=AbxVGAl)
- [VAI Inference Job](https://console.cloud.google.com/vertex-ai/locations/us-central1/training/765065084236988416/cpu?inv=1&invt=AbxVGQ)


examples/distributed/homogeneous_inference.py
```diff
@@ -169,9 +170,6 @@ def _inference_process(
 
     log_every_n_batch = int(inferencer_args.get("log_every_n_batch", "50"))
```
svij-sc, line 171: Nit:
```
# Users can feel free to parse this argument from `inferencer_args` however they want if they want more
# customizability for their fanout strategy
    
    -->
    
# Users can feel free to parse this argument from `inferencer_args` 
```
The extra line is not really adding any value and contriversal as there are other ways to specify CNS in our codebase too

examples/distributed/heterogeneous_inference.py
```diff
@@ -141,16 +141,16 @@ def _inference_process(
...
     """
-
     fanout_per_hop = int(inferencer_args.get("fanout_per_hop", "10"))
-    # This fanout is defaulted to match the fanout provided in the CORA UDL E2E Config:
-    # `python/gigl/src/mocking/configs/e2e_udl_node_anchor_based_link_prediction_template_gbml_config.yaml`
+    # This fanout is defaulted to match the fanout provided in the DBLP E2E Config:
+    # `python/gigl/src/mocking/configs/dblp_node_anchor_based_link_prediction_template_gbml_config.yaml`
     # Users can feel free to parse this argument from `inferencer_args` however they want if they want more
-    # customizability for their fanout strategy.
+    # customizability for their fanout strategy for the current node type.
```
svij-sc, line 153: Probably dont need the last line: `however they want if they want more
    # customizability for their fanout strategy for the current node type.` tbh, its a little confusing as customizability can be achieved through config too. Is it easier this way? yes.
    But based on above code somewhat obvious that this can be parsed from config level too

examples/distributed/heterogeneous_inference.py
```diff
@@ -169,35 +169,43 @@ def _inference_process(
...
 
-    # This value defines the `node_type` tag that will be used for writing to GCS and BQ. We default to "user".
-    embedding_type = inferencer_args.get("embedding_type", "user")
-
     device = gigl.distributed.utils.get_available_device(
         local_process_rank=process_number_on_current_machine,
     )  # The device is automatically inferred based off the local process rank and the available devices
 
+    # We split the node ids on the current machine across each of the processes on the current machine
+    node_type_to_input_node_ids = dataset.node_ids
```
svij-sc, line 177: lets add type information  so its easier to read
`node_type_to_input_node_ids: type = dataset.node_ids`

svij-sc: Re:
> We split the node ids on the current machine across each of the processes on the current machine

has this already happened to you by the time you get `. node_ids` ?

svij-sc: btw one general comment - outside of scope of this PR.

Reading this again its hard to wrap my mind around that node_ids are not all node_ids but rather ones that exist for this specific partition of the dataset. I wonder if there is a way to make this more explicit.

examples/distributed/heterogeneous_inference.py
```diff
@@ -169,35 +169,43 @@ def _inference_process(
...
-    model: nn.Module = _init_example_gigl_model(
+    model: nn.Module = _init_example_gigl_heterogeneous_model(
         state_dict=model_state_dict,
-        node_feature_dim=node_feature_dim,
-        edge_feature_dim=edge_feature_dim,
+        node_type_to_feature_dim=node_type_to_feature_dim,
+        edge_type_to_feature_dim=edge_type_to_feature_dim,
         inferencer_args=inferencer_args,
         device=device,
     )
```
svij-sc, line 211: RE following below:
> if num_files_at_gcs_path > 0:

Instead of warning we should probably also either 1. delete the files or 2. error out
Btw, why would this be the case?
The entry point to inferencer should delete this by default.

examples/distributed/heterogeneous_inference.py
```diff
@@ -379,44 +384,63 @@ def _run_example_inference(
...
+            load_embedding_start_time = time.time()
+
+            load_embeddings_to_bigquery(
+                gcs_folder=embedding_output_gcs_folder,
+                project_id=bq_project_id,
+                dataset_id=bq_dataset_id,
+                table_id=bq_table_name,
+            )
+            logger.info(
+                f"Finished loading embeddings to BigQuery, which took {time.time()-load_embedding_start_time:.2f} seconds for node type {inference_node_type}"
```
svij-sc, line 441: should this be a background thread?
Otherwise we are blocking on the next inferencer starting.

svij-sc (commented):
Looking great, thanks for the work!

Btw, do you have examples of this running and any cpu/gpu profiling to share as compared w/ homogeneous profiling?

examples/distributed/heterogeneous_inference.py
mkolodner-sc: You're right -- currently inferencer will clean this up by default, so this wouldn't happen for current use cases in GiGL orchestration. I'll change this to cleanup these files in this case.

examples/distributed/heterogeneous_inference.py
mkolodner-sc: Makes sense, I'll update this to run in the background, thanks!

mkolodner-sc:
@svij-sc I've attached a successful VAI pipeline and inference link in the PR description, thanks!

examples/distributed/heterogeneous_inference.py
```diff
@@ -379,52 +388,74 @@ def _run_example_inference(
...
-    )
+        bq_project_id, bq_dataset_id, bq_table_name = BqUtils.parse_bq_table_path(
+            bq_table_path=output_bq_table_path
+        )
 
-    # After inference is finished, we use the process on the Machine 0 to load embeddings from GCS to BQ.
-    if distributed_context.global_rank == 0:
-        logger.info("--- Machine 0 triggers loading embeddings from GCS to BigQuery")
-        load_embedding_start_time = time.time()
+        embedding_output_gcs_folder = InferenceAssets.get_gcs_asset_write_path_prefix(
```
svij-sc, line 403: nit: Lets add a comment why we write to this path?
i.e. we need it for unenumeration I believe?

examples/distributed/heterogeneous_inference.py
```diff
@@ -379,52 +388,74 @@ def _run_example_inference(
...
+                gcs_folder=embedding_output_gcs_folder,
+                project_id=bq_project_id,
+                dataset_id=bq_dataset_id,
+                table_id=bq_table_name,
+                should_run_async=should_run_async,
+            )
+
     logger.info(
         f"--- Program finished, which took {time.time()-program_start_time:.2f} seconds"
     )
```
svij-sc, line 454: do we need to call any cleanup code anywhere to gracefully shutdown everything?
Also, do we need torch.barriers to make different parts of the pieline i.e. data reads from gcs; partitioning; training loop more distinct?

svij-sc (commented):
Still working through, will have full review by EOD
But some minor comments interim.

examples/distributed/heterogeneous_inference.py
svij-sc: Revisiting this, deleting in two places from a duplicity perspective is not ideal - but for an example I guess its fine.

svij-sc (approved):
LGTM, minor comments

examples/distributed/heterogeneous_inference.py
mkolodner-sc: > do we need to call any cleanup code anywhere to gracefully shutdown everything?

We don't -- each of the `dataset_building` and `inference` processes are spawned in separate processes, and any necessary shutdown happens there. In this parent process, there are no items to be shutdown.

> Also, do we need torch.barriers to make different parts of the pipeline i.e. data reads from gcs; partitioning; training loop more distinct?

Yes, there are already barriers in place between these different parts of the pipeline.

PR #22, opened 2025-05-02
Add `DistABLPLoader` and transform to convert to homogeneous

kmontemayor2-sc:
Add `DistABLPLoader` to GiGL as part one of "homogeneous ABLP" support.

Since we want to move towards treating all label edges as full edges in GiGL, we had some request for when there's just one message passing edge type to treat the graph as "homogeneous" and this is the first step towards that.

I will have a follow-up PR to setup the labels properly on the output data object.


python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -109,6 +138,7 @@ def __init__(
                 Defaults to `2` if set to `None` when using cpu training/inference.
             shuffle (bool): Whether to shuffle the input nodes. (default: ``False``).
             drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
+            message_passing_edge_type (EdgeType): The edge type to use for message passing.
```
mkolodner-sc, line 141: nit: Optional[EdgeType]

Also, do you think its worth renaming this to `training_message_passing_edge_type`? i see its a bit lengthy, but IIUC we don't want to be setting this if we are doing inference, or else we'll be introducing all the label logic

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -167,6 +164,105 @@ def test_distributed_neighbor_loader_heterogeneous(self):
...
+        # 11 -> {13, 17}
+        # 15 -> {13, 14}
+        # 16 -> {12, 14}
+        # Positive labels
+        # 10 -> 15
+        # 15 -> 16
+        # Negative labels
+        # 10 -> 16
+        # 11 -> 14
+        # https://dreampuf.github.io/GraphvizOnline/?engine=dot#digraph%20G%20%7B%0A%0A%20%20%20%2010%20-%3E%20%7B11%2C%2012%7D%0A%20%20%20%2011%20-%3E%20%7B13%2C%2017%7D%0A%20%20%20%2015%20-%3E%20%7B13%2C%2014%7D%0A%20%20%20%2016%20-%3E%20%7B12%2C%2014%7D%0A%20%20%20%2010%20-%3E%2015%20%5Bcolor%3D%22blue%22%5D%0A%20%20%20%2015%20-%3E%2016%20%5Bcolor%3D%22blue%22%5D%0A%20%20%20%2010%20-%3E%2016%20%5Bcolor%3D%22red%22%5D%0A%20%20%20%2011%20-%3E%2014%20%5Bcolor%3D%22red%22%5D%0A%7D
```
mkolodner-sc, line 209: Thanks for this visualization! I think we can use an abbreviated link here -- I followed the link and clicked share -- would https://is.gd/w2oEVp work?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -167,6 +164,105 @@ def test_distributed_neighbor_loader_heterogeneous(self):
...
+        [
+            param(
+                "Positive and Negative edges",
+                labeled_edges={
+                    _POSITIVE_EDGE_TYPE: torch.tensor([[10, 15], [15, 16]]),
+                    _NEGATIVE_EDGE_TYPE: torch.tensor([[10, 11], [16, 14]]),
+                },
+                expected_node=torch.tensor([10, 11, 12, 13, 14, 15, 16, 17]),
+                expected_srcs=torch.tensor([10, 10, 15, 15, 16, 16, 11, 11]),
+                expected_dsts=torch.tensor([11, 12, 13, 14, 12, 14, 13, 17]),
```
mkolodner-sc, line 177: For my learning, why is the order of the src and dsts the same every time?

mkolodner-sc (commented):
Thanks Kyle, left some comments + questions

mkolodner-sc (approved):
Thanks a lot Kyle! LGTM provided comments are addressed.

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -60,6 +88,7 @@ def __init__(
         num_cpu_threads: Optional[int] = None,
         shuffle: bool = False,
         drop_last: bool = False,
+        message_passing_edge_type_to_get_labels_from: Optional[EdgeType] = None,
```
yliu2-sc, line 91: why not just call it `supervision_edge_type`? or `label_edge_type`?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -200,6 +233,62 @@ def __init__(
             pin_memory=device.type == "cuda",
         )
 
+        # Eventually - we should expose this as an arg to be passed in.
+        self._transforms: list[
+            Callable[[Union[Data, HeteroData]], Union[Data, HeteroData]]
+        ] = []
+
+        if isinstance(curr_process_nodes, torch.Tensor):
```
yliu2-sc, line 241: add comments that this determines homogeneous vs heterogeneous

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -200,6 +233,62 @@ def __init__(
...
+        # TODO(kmonte): Once GLT sampling works more generally, we should only sample against message_passing_edge_type for this code path.
+        if message_passing_edge_type_to_get_labels_from is not None:
+            if len(node_ids.shape) != 1:
+                raise ValueError(
+                    f"node_ids must be a 1D tensor when supervision_edge_types are provided, got {node_ids.shape}."
+                )
+
+            if not isinstance(dataset.graph, abc.Mapping):
+                raise ValueError(
+                    "When `message_passing_edge_type` is provided, the dataset must be heterogeneous."
```
yliu2-sc, line 256: nit: `message_passing_edge_type` is not a defined variable name

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -200,6 +233,62 @@ def __init__(
...
+                )
+            (
+                positive_label_edge_type,
+                negative_label_edge_type,
+            ) = select_label_edge_types(
```
yliu2-sc, line 261: Do we only do this for homogeneous cases?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -200,6 +233,62 @@ def __init__(
...
+
+            positive_labels, negative_labels = get_labels_for_anchor_nodes(
+                dataset, node_ids, positive_label_edge_type, negative_label_edge_type
+            )
+            if negative_labels is not None:
+                node_ids = torch.cat(
+                    [node_ids.unsqueeze(1), positive_labels, negative_labels], dim=1
+                )
+            else:
+                node_ids = torch.cat([node_ids.unsqueeze(1), positive_labels], dim=1)
```
yliu2-sc, line 273: can probably just keep these separate and not cat

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -200,6 +233,62 @@ def __init__(
...
+                    "When `message_passing_edge_type` is provided, the dataset must be heterogeneous."
+                )
+            (
+                positive_label_edge_type,
+                negative_label_edge_type,
+            ) = select_label_edge_types(
+                message_passing_edge_type_to_get_labels_from, dataset.graph.keys()
+            )
+
+            positive_labels, negative_labels = get_labels_for_anchor_nodes(
```
yliu2-sc, line 265: Is the purpose of the addition here to load labels? In that case this should be part of the ABLP loader and not DistNeighborLoader?

yliu2-sc (commented):
Could you provide more context in the PR description?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -40,6 +44,30 @@ def __getitem__(
         return _BatchedNodeSamplerInput(self.node[index].view(-1), self.input_type)
 
 
+class _UDLToHomogeneous:
```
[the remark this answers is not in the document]
kmontemayor2-sc: Yeah, WDYT about `_SupervisedToHomogeneous`?

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: This is the edge type that the labels are based on, I'm not sure it's "right" to call it a label edge type, this is like `user-to-user` and then we'd have `user-positive-user` and `user-negative-user` which are inferred in the graph. 

I guess this edge type is "supervised" but I feel like if I saw "supervision_edge_type" I'd assume it was the edge type that contains the labels.

WDYT?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -109,6 +138,10 @@ def __init__(
                 Defaults to `2` if set to `None` when using cpu training/inference.
             shuffle (bool): Whether to shuffle the input nodes. (default: ``False``).
             drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
+            message_passing_edge_type_to_get_labels_from (Optional[EdgeType]): The edge type to use for message passing, only needs to be passed in for training.
+                If passed in, then the dataset must be heterogenneous.
```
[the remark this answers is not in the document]
kmontemayor2-sc: thanks!

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -109,6 +138,10 @@ def __init__(
...
+                If passed in, supervision edge types will be automatically selected based on the message passing edge type,
```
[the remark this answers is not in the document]
kmontemayor2-sc: Hmmm, I think that ideally users shouldn't have to worry about this at all, I've left a todo to add more details about how the inferred edge types get there once this is more fleshed out.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: Done.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: good catch, done

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: hmmm WDYM homogeneous? We can really only do this if the dataset is heterogeneous, and I guess this should still be valid if there are multiple sets of supervision edge types.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: I wanted to keep us with one loader - but if you feel strongly that we should split up the loaders to be more task specific we can do that.

kmontemayor2-sc: I don't think that PyG really provides guidance here right, it's just about what we're sampling from, and in this case we are sampling from nodes in either case.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: In that case we'd do the `cat` in the `_BatchedNodeSamplerInput` instead? We could do that but I'm not sure that's much cleaner?

python/gigl/utils/data_splitters.py
```diff
@@ -288,37 +290,25 @@ def get_labels_for_anchor_nodes(
...
+    if not isinstance(dataset.graph, Mapping):
         raise ValueError(
-            f"Got unknown type for supervision_edge_types: {type(supervision_edge_types)}"
+            "The dataset must be heterogeneous to select labels for anchor nodes."
         )
+    positive_node_topo = dataset.graph[positive_label_edge_type].topo
+    if negative_label_edge_type is not None:
+        negative_node_topo = dataset.graph[negative_label_edge_type].topo
+    else:
+        negative_node_topo = None
```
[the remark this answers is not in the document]
kmontemayor2-sc: won't we throw when we do the `[]` access?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -109,6 +138,11 @@ def __init__(
                 Defaults to `2` if set to `None` when using cpu training/inference.
             shuffle (bool): Whether to shuffle the input nodes. (default: ``False``).
             drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
+            message_passing_edge_type_to_get_labels_from (Optional[EdgeType]): The edge type to use for message passing, only needs to be passed in for training.
+                If passed in, then the dataset must be heterogeneous.
+                If passed in, supervision edge types will be automatically selected based on the message passing edge type,
```
yliu2-sc, line 143: I don't think users need to care about this, this homogeneous and heterogeneous is our internal representation and not what the users see, right?

python/gigl/distributed/distributed_neighborloader.py
yliu2-sc: Oh I see this is what goes into the sampler, but what about the return from the distNeighborloader? Is it stored all in batch.y?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -109,6 +145,7 @@ def __init__(
                 Defaults to `2` if set to `None` when using cpu training/inference.
             shuffle (bool): Whether to shuffle the input nodes. (default: ``False``).
             drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
+            transforms (Sequence[Callable[[Union[Data, HeteroData]], Union[Data, HeteroData]]]): A sequence of transforms to apply to the data.
```
mkolodner-sc, line 148: What transforms do you imagine we'll need to be passing in the future to the `DistNeighborLoader`? Should we make the transform input arg and collate specific to the ABLP dataloader, rather than relying on an inheritance pattern here?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -167,6 +167,103 @@ def test_distributed_neighbor_loader_heterogeneous(self):
...
+            param(
+                "Positive edges",
+                labeled_edges={_POSITIVE_EDGE_TYPE: torch.tensor([[10], [15]])},
+                expected_node=torch.tensor([10, 11, 12, 13, 14, 15, 17]),
+                expected_srcs=torch.tensor([10, 10, 15, 15, 11, 11]),
+                expected_dsts=torch.tensor([11, 12, 13, 14, 13, 17]),
+            ),
+        ]
+    )
+    def test_distributed_neighbor_loader_with_supervision_edges(
```
mkolodner-sc, line 191: nit: can we call this `test_ablp_dataloader`?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -220,15 +282,175 @@ def __init__(
...
+        local_process_world_size: int,  # TODO: Move this to DistributedContext
+        input_nodes: Optional[
+            Union[torch.Tensor, Tuple[NodeType, torch.Tensor]]
+        ] = None,
+        num_workers: int = 1,
+        batch_size: int = 1,
+        pin_memory_device: Optional[torch.device] = None,
+        worker_concurrency: int = 4,
+        channel_size: str = "4GB",
+        process_start_gap_seconds: int = 60,
```
mkolodner-sc, line 310: Small note: I have a PR https://github.com/Snapchat/GiGL/pull/39/files#diff-e751a9cd787e206721cd05c941effd36bae4e56559ac269889b7e9fdb27eb240L59 which includes an update to change this to a float for the NeighborLoader, can we aim to do the same here once that lands?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -220,15 +282,175 @@ def __init__(
...
+        self._shutdowned = True
+        # TODO(kmonte): Remove these checks and support properly heterogeneous NABLP.
+        if isinstance(input_nodes, tuple):
+            raise ValueError(
+                "Heterogeneous NABLP is not supported yet. Provide node_ids as a tensor."
+            )
+        if isinstance(num_neighbors, abc.Mapping):
+            raise ValueError(
+                "Heterogeneous NABLP is not supported yet. Provide num_neighbors as a list of integers."
+            )
```
mkolodner-sc, line 383: nit: update to ABLP, ditto elsewhere

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -207,6 +255,20 @@ def __init__(
             pin_memory=device.type == "cuda",
         )
 
+        # Eventually - we should expose this as an arg to be passed in.
```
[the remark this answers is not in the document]
kmontemayor2-sc: done.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: PyG takes in `transform` [1] so I don't think it's too odd to add this? I think that if we want to implement pre-fetching in the future then we'd also want to pre-run the transforms here right? So it makes sense to store them as part of the class.

Is the concern here that it's confusing why the DistABLPLoader is using transforms?

[1]: https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/loader/node_loader.html#NodeLoader

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: updated.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: Done.

python/tests/unit/distributed/distributed_neighborloader_test.py
kmontemayor2-sc: sure, done.

python/gigl/distributed/distributed_neighborloader.py
yliu2-sc: I'd vote for keeping DistNeighborLoader intact from the glt version, since it's going to be used for other cases say random neg, node classification, link prediction, unless we'd expect transforms in those cases. Anything specific to ABLP can go into ABLP loader.

mkolodner-sc: Gotcha, makes sense to conform to PyG here. I think my confusion here is that it feels weird that `DistNeighborLoader` is currently accepting `transform` as an argument but we don't have an immediate use case for that class right now, while the `DistABLPLoader`, which we do have a `transform` use-case ready for, doesn't take this as an argument to `__init__`. I feel like if we have `transform` accepted as an argument here for `DistNeighborLoader`, we should be consistent and also have `DistABLPLoader` do the same.

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -220,15 +281,175 @@ def __init__(
...
+            drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
+        """
+
+        # Set self._shutdowned right away, that way if we throw here, and __del__ is called,
+        # then we can properly clean up and don't get extraneous error messages.
+        # We set to `True` as we don't need to cleanup right away, and this will get set
+        # to `False` in super().__init__()` e.g.
+        # https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_loader.py#L125C1-L126C1
+        self._shutdowned = True
+        # TODO(kmonte): Remove these checks and support properly heterogeneous NABLP.
```
yliu2-sc, line 374: For het cases, node ids would still just be a tensor of anchor nodes (no mapping), since we'd likely have separate loaders for different supervision edge types(that has different anchor nodes). Dataset will be heterogeneous tho. We can already support that in this PR?

But we do need EdgeType mapping for num neighbors. If we can't support it rn it's ok to leave as TODO.

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -220,15 +281,175 @@ def __init__(
...
+                    f"input_nodes must be provided for heterogeneous datasets, received node_ids of type: {dataset.node_ids.keys()}"
+                )
+            input_nodes = dataset.node_ids
+        if len(input_nodes.shape) != 1:
+            raise ValueError(
+                f"input_nodes must be a 1D tensor, got {input_nodes.shape}."
+            )
+        if not isinstance(dataset.graph, abc.Mapping):
+            raise ValueError(
+                f"The dataset must be heterogeneous for ABLP. Recieved dataset with graph of type: {type(dataset.graph)}"
```
yliu2-sc, line 399: What do we do for homogeneous SSL cases? We'd create the labels and het dataset in partitioner?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -220,15 +281,175 @@ def __init__(
...
+        logger.info(
+            f"Converted input nodes to tuple of ({DEFAULT_HOMOGENEOUS_NODE_TYPE}, {input_nodes[1].shape})."
+        )
+        transforms = [
+            _SupervisedToHomogeneous(
+                message_passing_edge_type=DEFAULT_HOMOGENEOUS_EDGE_TYPE,
+            )
+        ]
+        # TODO(kmonte): stop setting fanout for positive/negative once GLT sampling is fixed.
+        zero_samples = [0] * len(num_neighbors)
```
yliu2-sc, line 428: does this mean that we don't sample for positive/negative edge types? ex. we have a user-story graph. Pos edge type is user-story. For user nodes we want sample user-story-user. Positive story nodes we want to sample story-user-story. How would the zero samples work in this case?

yliu2-sc (commented):
Thanks Kyle! I think this structuring makes more sense now.  `to_heterogeneous_edge` and `select_label_edge_types` are not included in this PR btw.

python/gigl/distributed/distributed_neighborloader.py
mkolodner-sc: My understanding is that this is only setting the `zero_sample` for the `user -> positive_to -> story` relation type. We would still have `user->to->story` and `story->to->user` edges with non-zero fanouts, which would provide the subgraphs for `user` and `story` nodes. This is done so that when we are hopping out from a user, we don't accidentally use the supervision edge, which should only be used as label metadata IIUC.

python/gigl/distributed/distributed_neighborloader.py
mkolodner-sc: Correct this is done in the `dataset_factory` [here](https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/dataset_factory.py#L149-L177)

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: I feel like "transforms" are pretty generic, no? if we feel strongly I can move all of this to the `DistABLPLoader` but I feel like it's not a big deviation from PyG to have these (or even torch if we thing of `collate_fn` as a "transform".


We actually don't have a usecase for `DistABLPLoader` to take in transform as an argument, just for it to automatically generate one transform if needed.

I prefer to not add `transforms` as an arg to `DistABLPLoader` until we actually have a usecase for user-provided transforms.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: for het cases users need to provide the node type to iterate over right, like `(UserNodeType, UserNodeIds)`. Otherwise, how do we know which type of anchor nodes to fan out from? Node ids are not unique across node types, we can have `User, 0` and `Item, 0` as valid `NodeType, NodeId` tuples.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: My understanding is the same as Matt's here.

python/gigl/distributed/distributed_neighborloader.py
mkolodner-sc: > We actually don't have a usecase for DistABLPLoader to take in transform as an argument, just for it to automatically generate one transform if needed.

Hmm, I feel like we should be consistent between either both `DistABLPLoader` and `DistNeighborLoader` taking a `transform` argument, or neither take a `transform` argument. If neither `DistABLPLoader` or `DistNeighborLoader` require the user to provide `self._transforms` at this time and GLTNeighborLoader doesn't have this [[1](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/distributed/dist_neighbor_loader.py#L75-L87)], can we remove this as an argument altogether?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -40,6 +49,30 @@ def __getitem__(
         return _BatchedNodeSamplerInput(self.node[index].view(-1), self.input_type)
 
 
+class _SupervisedToHomogeneous:
```
zfan3-sc, line 52: Do we want to expose this class to the user and let user explicitly provide it? If not how do we know when to treat the input graph as a homogeneous graph with labels v.s. a true heterogeneous graph?

kmontemayor2-sc: Hmmm, for now we are doing this is we see that there's one positive label, and (optionally) a negative label type. 

We can have this be explicitly exposed instead - if you're ok with needing to use this yourself maybe that'd be simpler. 

Ultimately depends how little users want o be exposed to the "heterogeneous" representation I guess :)

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -220,15 +281,175 @@ def __init__(
...
+            )
+        if not isinstance(dataset.graph, abc.Mapping):
+            raise ValueError(
+                f"The dataset must be heterogeneous for ABLP. Recieved dataset with graph of type: {type(dataset.graph)}"
+            )
+        if DEFAULT_HOMOGENEOUS_EDGE_TYPE not in dataset.graph:
+            raise ValueError(
+                f"With Homogeneous ABLP, the graph must have {DEFAULT_HOMOGENEOUS_EDGE_TYPE} edge type, received {dataset.graph.keys()}."
+            )
+        positive_label_edge_type, negative_label_edge_type = select_label_edge_types(
```
zfan3-sc, line 405: Not necessarily need to be addressed in this PR. Just want to call out ideally we should be able to have multiple positive and negative edge types.

python/gigl/distributed/distributed_neighborloader.py
zfan3-sc: > we are doing this is we see that there's one positive label, and (optionally) a negative label type.

I see, then how do we deal with heterogeneous graph trained with supervised labels? In that case we have positive label do not need this transformation

kmontemayor2-sc: For a truly heterogeneous graph (containing e.g. User and Content node types), then we don't want to use this right?

My understanding is that this should *only* be used for "psuedo-heterogeneous" graphs where there's only one node type, and one message passing edge type. 

For cases where we have multiple sets of label edge types I think we can extend this.

If there are multiple "true" node or edge types in what case would we want to convert to homogeneous Data?.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: Sure, I guess we can remove for now (though I feel like it makes our code a bit more complicated).

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: ack yes, we can do so as a follow up :)

python/gigl/distributed/distributed_neighborloader.py
zfan3-sc: > If there are multiple "true" node or edge types in what case would we want to convert to homogeneous Data?

Ah no. My understanding was this dataloader is gonna be used by both "true" het graph and homogeneous graph with supervision labels. If this is only for latter I think it's fine as is.

zfan3-sc (approved):

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: We do intend to use this dataloader for "true" het graphs - just that this transform shouldn't be applied to "true" graphs

yliu2-sc (approved):
The new changes in structure looks fine to me. We should probably rename `_BatchedNodeSamplerInput` --> `_LabeledNodeSamplerInput` or sth more fitting.

PR #24, opened 2025-05-02
Add docker config to install_dev_deps

mkolodner-sc:
**Scope of work done**

- Update `install_dev_deps` script to also configure docker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,12 +70,14 @@ install_dev_deps: check_if_valid_env
 	bash ./requirements/install_scala_deps.sh
 	pip install -e ./python/
 	pre-commit install --hook-type pre-commit --hook-type pre-push
+	gcloud auth configure-docker us-central1-docker.pkg.dev
 
 # Production environments, if you are developing use `make install_dev_deps` instead
 install_deps:
 	bash ./requirements/install_py_deps.sh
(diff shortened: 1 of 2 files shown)


scripts/setup_environment.sh
```diff
@@ -28,3 +28,6 @@ CURR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 make install_dev_deps
 
 source ~/.profile
+
+# Setting up docker with `us-central1-docker.pkg.dev`
```
svij-sc, line 32: this comment is not adding a lot of value

svij-sc: If we add this to `install_dev_deps` maybe we don't need to add it anywhere else?

mkolodner-sc: Updated, thanks!

Makefile
```diff
@@ -66,13 +66,15 @@ check_if_valid_env:
 # if developing, you need to install dev deps instead
 install_dev_deps: check_if_valid_env
 	@# Install docker driver that will allow us to build multi-arch images
```
svij-sc, line 68: nit: Remove this outdated comment?

PR #25, opened 2025-05-02
Makerule to build documentation website

svij-sc:
**Scope of work done**

Notably, I have migrated the old `conf.py` and `index.rst` moved to top level dir so the path resolution is more obvious - as before it kept breaking.

Includes other  `conf.py` and `index.rst` updates that now allow us to build the documentation using the new make rules and the paths / navigation resolve adequately.

Following, I have also migrated the doc files so they are not deeply nested and more discoverable: https://github.com/Snapchat/GiGL/pull/26


Makefile
```diff
@@ -477,3 +477,7 @@ release_gigl:
 
 publish_docs:
 	@echo "This needs to be implemented"
+
+build_docs:
+	sphinx-build -M clean . gh_pages_build
+	sphinx-build -M html . gh_pages_build
```
kmontemayor2-sc, line 483: It seems like this build rule doesn't get used here, will this (eventually) be used as part of the doc publishing pipeline?

kmontemayor2-sc (approved):
I assume this is some "setup" change and we haven't updated the docs/etc yet?

Ideally, I'd like to be able to see the docs we end up generating, but this change LGTM.

Makefile
svij-sc: Correct, it is being used here: https://github.com/Snapchat/GiGL/pull/28/files#diff-42447f270c10809d699b2b598e8dbcdfc0321edb5ff8982b4b102a56a39d8af3R50

svij-sc:
Thanks for the approval @kmontemayor2-sc 

> Ideally, I'd like to be able to see the docs we end up generating

The docs need some work i.e. alot of links are broken, et al.
I wanted to limit the scope here so I made limited changes needed to start building the docs automatically. One this and https://github.com/Snapchat/GiGL/pull/28 are merged we should be in a better situation here.
For reference on what might the generation look like you can see a successful results of [deployment](https://github.com/snap-research/GiGL/actions/runs/14806387587/job/41575241673?pr=21) in our snap-research repo: https://snap-research.github.io/GiGL/

PR #26, opened 2025-05-02
Migrate documentation to docs folder

svij-sc:
**Scope of work done**

Prior to this documentation was hidden; nested deep inside `docs/sphix/sourcs/....` ; migrating them to docs folder so they have better discoverability.


docs/user_guide/overview/components.md
```diff
@@ -0,0 +1,55 @@
+# GiGL Components
```
svij-sc, line 1: note: this file actually changed as the prior was using a table plugin to render these nicely.
But the table plugin was not versioned and now is not playing nicely with other plugins.

table was used strictly as a formatting nicety.

gh_pages_source/_static/custom.css
```diff
@@ -0,0 +1,6 @@
+.center-image {
```
svij-sc, line 1: self note: remove this

kmontemayor2-sc (approved):

PR #30, opened 2025-05-05
Update early stop logic

mkolodner-sc:
**Scope of work done**

Updates early stop to be:
- Decoupled from the modeling task spec
- Decoupled from checkpointing logic

### DIFF
--- a/python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
+++ b/python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
@@ -436,11 +436,11 @@ class NodeAnchorBasedLinkPredictionModelingTaskSpec(
                         num_batches=self.num_val_batches,
                     )
 
-                    self._early_stopper.step(
+                    has_metric_improved, should_early_stop = self._early_stopper.step(
                         value=eval_metrics[self._early_stop_criterion]
                     )
                     # Check if we need to early stop based on patience
-                    if self._early_stopper.should_early_stop():
+                    if should_early_stop:
                         break
 
                 profiler.step() if profiler else None  # type: ignore
--- a/python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
+++ b/python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
@@ -314,12 +314,6 @@ class NodeAnchorBasedLinkPredictionModelingTaskSpec(
         self.tasks = model.tasks
         self._graph_backend = model.graph_backend
 
-        self._early_stopper = EarlyStopper(
-            early_stop_patience=self._early_stop_patience,
-            should_maximize=self._should_maximize,
-            model=self.model,
-        )
-
         return self.model
 
     # function for setting up things like optimizer, scheduler, criterion etc.
@@ -358,6 +352,12 @@ class NodeAnchorBasedLinkPredictionModelingTaskSpec(
             gbml_config_pb_wrapper=gbml_config_pb_wrapper,
             graph_backend=self._graph_backend,
             device=device,
+        )
+
+        self._early_stopper = EarlyStopper(
+            early_stop_patience=self._early_stop_patience,
+            should_maximize=self._should_maximize,
+            model=self.model,
         )
 
         main_data_loader = data_loaders.train_main
--- a/python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
+++ b/python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
@@ -1,5 +1,4 @@
 from contextlib import ExitStack
-from copy import deepcopy
 from distutils.util import strtobool
 from time import time
 from typing import Any, Dict, Optional, OrderedDict, Type
@@ -176,15 +175,10 @@ class NodeAnchorBasedLinkPredictionModelingTaskSpec(
         self._early_stop_criterion = EvalMetricType[
             kwargs.get("early_stop_criterion", "loss")
         ]
-        early_stop_patience = int(kwargs.get("early_stop_patience", 3))
-        should_maximize = (
+        self._early_stop_patience = int(kwargs.get("early_stop_patience", 3))
+        self._should_maximize = (
             False if self._early_stop_criterion == EvalMetricType.loss else True
         )
-        self._early_stopper = EarlyStopper(
-            early_stop_patience=early_stop_patience,
-            should_maximize=should_maximize,
-        )
-        self._best_val_model: Optional[Dict[str, Any]] = None
 
         # Loading Task
         self.tasks = NodeAnchorBasedLinkPredictionTasks()
@@ -320,6 +314,12 @@ class NodeAnchorBasedLinkPredictionModelingTaskSpec(
         self.tasks = model.tasks
         self._graph_backend = model.graph_backend
 
+        self._early_stopper = EarlyStopper(
+            early_stop_patience=self._early_stop_patience,
+            should_maximize=self._should_maximize,
+            model=self.model,
+        )
+
         return self.model
 
     # function for setting up things like optimizer, scheduler, criterion etc.
@@ -436,12 +436,9 @@ class NodeAnchorBasedLinkPredictionModelingTaskSpec(
                         num_batches=self.num_val_batches,
                     )
 
-                    has_improved = self._early_stopper.step(
+                    self._early_stopper.step(
                         value=eval_metrics[self._early_stop_criterion]
                     )
-                    if has_improved:
-                        self._best_val_model = deepcopy(self.model.state_dict())
-
                     # Check if we need to early stop based on patience
                     if self._early_stopper.should_early_stop():
                         break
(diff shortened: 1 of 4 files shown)


python/gigl/src/common/modeling_task_specs/utils/early_stop.py
```diff
@@ -1,56 +1,60 @@
...
-        value = metrics[self.criterion]
-        if self.has_improved(value=value):
-            self.early_stop_counter = 0
+    def step(self, value: float) -> bool:
+        """
+        Steps through the early stopper provided some criterion. Returns whether the provided criterion improved over the previous best criterion.
+        Args:
+            value (float): Criterion used for stepping through early stopper
+        Returns:
+            bool: Whether there was improved over previous best criteiron
```
nshah-sc, line 33: this comment has some grammar/typo issues

python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
```diff
@@ -429,21 +436,27 @@ def train(
                         num_batches=self.num_val_batches,
                     )
 
+                    has_improved = self._early_stopper.step(
+                        value=eval_metrics[self._early_stop_criterion]
+                    )
+                    if has_improved:
+                        self._best_val_model = deepcopy(self.model.state_dict())
```
kmontemayor2-sc, line 443: btw is this line expensive? we'd be storing an entire copy of the model in memory right?

python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
```diff
@@ -429,21 +436,27 @@ def train(
...
+
                     # Check if we need to early stop based on patience
-                    if self.early_stopper.should_early_stop(eval_metrics, self.model):
+                    if self._early_stopper.should_early_stop():
```
kmontemayor2-sc, line 446: nit. add log?

python/gigl/src/common/modeling_task_specs/utils/early_stop.py
```diff
@@ -1,56 +1,60 @@
...
         self.prev_best = float("-inf") if self._should_maximize else float("inf")
-        self.early_stop_counter = 0
-        self.early_stop_patience = early_stop_patience
-        self.best_val_model: Dict[str, Any] = {}
 
-    def has_improved(self, value: float):
+    def _has_metric_improved(self, value: float) -> bool:
         return (self._should_maximize and value > self.prev_best) or (
             not self._should_maximize and value < self.prev_best
         )
```
kmontemayor2-sc, line 25: btw is this more readable as `if should_maximize: return value > prev_best else return value < prev_best`?

python/gigl/src/common/modeling_task_specs/utils/early_stop.py
```diff
@@ -1,56 +1,60 @@
...
-                f"Found invalid early stop criterion {early_stop_criterion.name}. Please make sure to supply one of {supported_early_stop_criteria}."
-            )
-        self.criterion = early_stop_criterion
-
-        self._should_maximize: bool = self.criterion != EvalMetricType.loss
+    """
+    Handles early stopping logic, keeping track of the best performing model provided some criterion
+    """
+
+    def __init__(self, early_stop_patience: int, should_maximize: bool):
```
nshah-sc, line 11: thanks for this change! I think the only point of consideration is that most small-scale uses of early stopping can likely save some code by just tracking the best state dict in memory, and it would be a small convenience to be able to have a `should_track_best_model` which can default to False.  In such cases, it saves the user a few lines to manage the state preservation themselves without really losing any flexibility.  i have no strong opinions on this, but it is a consideration given the point of this class is convenience anyway

nshah-sc (approved):
some thoughts in PR -- thanks!

kmontemayor2-sc (commented):
Oh oops didn't realize I wasn't (formally) tagged here, still getting used to github :P

python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
```diff
@@ -172,12 +173,18 @@ def __init__(self, **kwargs) -> None:
...
-        early_stop_criterion = kwargs.get("early_stop_criterion", "loss")
+        self._early_stop_criterion = EvalMetricType[
+            kwargs.get("early_stop_criterion", "loss")
+        ]
         early_stop_patience = int(kwargs.get("early_stop_patience", 3))
-        self.early_stopper = EarlyStopper(
-            early_stop_criterion=EvalMetricType[early_stop_criterion],
+        should_maximize = (
+            False if self._early_stop_criterion == EvalMetricType.loss else True
+        )
```
yliu2-sc, line 182: already default in the function call

python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
yliu2-sc: We can move it to CPU and not store and copy in GPU. Also do some gc to get rid of the old copied model params

python/gigl/src/common/modeling_task_specs/utils/early_stop.py
svij-sc: Probably better if we just pass in the model intro the constructor as an optional.
If provided we can track best model for the user automatically.

svij-sc: FWIW, its fine if this "deepcopy" of model weights, et al is opinionated.
Given that this is an optional field people can just do their own thing.

python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
mkolodner-sc: We already have log in `self.-early_stopper.should_early_stop()`

python/gigl/src/common/modeling_task_specs/node_anchor_based_link_prediction_modeling_task_spec.py
mkolodner-sc: We'll still need to pass this in if `should_maximize` is true, will aim to leave this as is

python/gigl/src/common/modeling_task_specs/utils/early_stop.py
mkolodner-sc: Sounds good! As mentioned, we'll make this field optional by optionally passing the model into the constructor rather than using a flag, thanks for the discussion here :)

python/gigl/src/common/modeling_task_specs/utils/early_stop.py
```diff
@@ -1,57 +1,92 @@
...
             )
-            self.prev_best = value
-            self.best_val_model = deepcopy(model.state_dict())
+            self._prev_best = value
+            if self._model is not None:
+                # Making a deep copy of the best model and moving to CPU to save GPU memory
+                self._best_model = {}
+                for identifier, layer in self._model.state_dict().items():
+                    self._best_model[identifier] = deepcopy(layer).cpu()
+                gc.collect()
```
nshah-sc, line 63: This seems a little complex for the operation we are considering -- i'm surprised there is not something simpler.  But, given this passes testing and it looks correct, this LGTM

python/gigl/src/common/modeling_task_specs/utils/early_stop.py
```diff
@@ -1,57 +1,87 @@
...
+            self._early_stop_counter = 0
             logger.info(
-                f"Validation {self.criterion.name} improved to {value:.4f} over previous best {self.prev_best}. Resetting early stop counter."
+                f"Validation criteria improved to {value:.4f} over previous best {self._prev_best}. Resetting early stop counter."
             )
-            self.prev_best = value
-            self.best_val_model = deepcopy(model.state_dict())
+            self._prev_best = value
+            if self._model is not None:
+                self._best_model = deepcopy(self._model.state_dict())
```
mkolodner-sc, line 62: It seems like there's some more complexity here with regards to how this interplays with the model being wrapped with DDP and trying to move the state_dict() to cpu tensor-by-tensor. As a result, I think it's fine to leave this as a deepcopy of the whole state dict, since we have enabled in this PR custom checkpointing strategies if `deepcopy(self._model.state_dict())` is too memory intensive for GPU.

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -65,6 +64,9 @@ def test_distributed_neighbor_loader(self):
             local_process_rank=0,
             local_process_world_size=1,
             pin_memory_device=torch.device("cpu"),
+            # TODO (mkolodner-sc): Dynamically infer free ports and remove this argument
```
mkolodner-sc, line 67: These tests are unrelated to this PR, but were failing when trying to merge this -- most likely due to port overlap between the new `test_distributed_neighbor_loader_batched` and `test_distributed_neighbor_loader`. For now, I have manually specified ports for each of these tests -- we can deprecate this in the future when we are able to dynamically infer free ports rather than relying on the `DEFAULT_INFERENCE_PORT` and `DEFAULT_SAMPLING_PORT`

cc @kmontemayor2-sc for vis

svij-sc: Should we try to free ports when the test finishes?

mkolodner-sc: Yeah, I can work on that as a follow-up.

PR #34, opened 2025-05-06
Demote downloading log from info to debug

nshah-sc:
Downloading from GCS currently creates very polluted logs -- this is evident in cases where we download batches of files to train or run inference on through GcsUtils.  This should likely be a debug statement anyway.


nshah-sc:
> BTW do we save debug logs at all?

we don't -- this is currently defaulted to INFO in Logger but it wouldn't be very hard to expose it in one way or another via an envvar

PR #37, opened 2025-05-07
Set _shutdowned right away

kmontemayor2-sc:
As discussed offline, if we don't do this then we can get extra messages error messages if we throw during __init__ which is confusing.

### DIFF
--- a/python/gigl/distributed/distributed_neighborloader.py
+++ b/python/gigl/distributed/distributed_neighborloader.py
@@ -119,6 +119,13 @@ class DistNeighborLoader(DistLoader):
                 Currently defaults to: gigl.distributed.constants.DEFAULT_MASTER_SAMPLING_PORT
         """
 
+        # Set self._shutdowned right away, that way if we throw here, and __del__ is called,
+        # then we can properly clean up and don't get extraneous error messages.
+        # We set to `True` as we don't need to cleanup right away, and this will get set
+        # to `False` in super().__init__()` e.g.
+        # https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_loader.py#L125C1-L126C1
+        self._shutdowned = True
+
         if input_nodes is None:
             if dataset.node_ids is None:
                 raise ValueError(


python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -119,6 +119,13 @@ def __init__(
                 Currently defaults to: gigl.distributed.constants.DEFAULT_MASTER_SAMPLING_PORT
         """
 
+        # Set self._shutdowned right away, that way if we throw here, and __del__ is called,
+        # then we can properly clean up and don't get extraneous error messages.
+        # We set to `True` as we don't need to cleanup right away, and this will get set
+        # to `False` in super().__init__()` e.g.
+        # https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_loader.py#L125C1-L126C1
+        self._shutdowned = True
```
mkolodner-sc, line 127: Will setting this to `True` early have any unwanted side effects?

kmontemayor2-sc: It shouldn't, we don't want to shutdown until everything is setup and setting this to true will skip attempting to shutdown. 

https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_loader.py#L270-L271

mkolodner-sc: Hmm, would setting this to True early though make it so that this logic will never get called from [here](https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_loader.py#L272-L282)
```
    if self._is_collocated_worker:
      self._collocated_producer.shutdown()
    elif self._is_mp_worker:
      self._mp_producer.shutdown()
    else:
      if rpc_is_initialized() is True:
        for server_rank, producer_id in zip(self._server_rank_list, self._producer_id_list):
          request_server(
            server_rank, DistServer.destroy_sampling_producer,
            producer_id
          )
```
I understand that we don't want this `shutdown` logic to happen if we throw early, but wouldn't we want this to be called during the intentional destruction of this class?

kmontemayor2-sc: yes that will not get called provided if we have not set  `_is_mp_worker` and friends (which we *also* have not set prior to `__init__` being finished. 

So, if this is set to `False` then we will *still* throw when we try to shutdown since all of the other fields are not set yet.

Since it `_is_shutdowned` gets set to `False` in the super class [1] we still still safely shutdown once everything is instatialized. (thoguh really it should be set to False later in __init__ but that's a different issue.

[1]: [817258](https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_loader.py#L125)

mkolodner-sc (approved):
Thanks Kyle!

PR #39, opened 2025-05-08
Improve Range Based Partitioning

mkolodner-sc:
**Scope of work done**

Changes to improve the performance of range-based partitioning and TFRecord Dataloading
- Restructure `_partition_by_single_chunk` in partitioner so that the rank indices are only sent to RPC if we are generating a partition book, reducing RPC communication load
- Slight restructure of range-based partitioning to make cleanup happen earlier
- Delete reference to `input_data` so that its assets are cleaned up as well through reduced reference count
- Change `process_start_gap_seconds` in the neighbor loader to be float instead of int type, conforming better to params for `init_neighbor_loader_worker`
- Change dataset factory to pass `node_tf_dataset_options` and `edge_tf_dataset_options`, rather than just one singular `tf_dataset_options`. This is important to tune separately during load testing.

### DIFF
--- a/python/gigl/common/data/load_torch_tensors.py
+++ b/python/gigl/common/data/load_torch_tensors.py
@@ -291,11 +291,10 @@ def load_torch_tensors_from_tf_record(
     else:
         # In this setting, we start and join each process one-at-a-time in order to achieve sequential tensor loading
         logger.info("Loading Serialized TFRecord Data in Sequence ...")
-        # Sequentially launch node, edge, and label loading processes to
-        # reduce the peak memory usage. Here we launch edge loading process
-        # first since edge data is larger than node data, so launching edge
-        # before launching node reduces peak memosry usage compared with
-        # first launching node loading and then launching edge loading.
+        # Here we launch edge loading process first since experimentally
+        # we have found that, since edge data is larger than node data,
+        # launching edge before launching node reduces peak memory usage
+        # compared with first launching node loading and then launching edge loading.
         edge_data_loading_process.start()
         edge_data_loading_process.join()
         node_data_loading_process.start()
--- a/python/gigl/common/data/load_torch_tensors.py
+++ b/python/gigl/common/data/load_torch_tensors.py
@@ -291,6 +291,11 @@ def load_torch_tensors_from_tf_record(
     else:
         # In this setting, we start and join each process one-at-a-time in order to achieve sequential tensor loading
         logger.info("Loading Serialized TFRecord Data in Sequence ...")
+        # Sequentially launch node, edge, and label loading processes to
+        # reduce the peak memory usage. Here we launch edge loading process
+        # first since edge data is larger than node data, so launching edge
+        # before launching node reduces peak memosry usage compared with
+        # first launching node loading and then launching edge loading.
         edge_data_loading_process.start()
         edge_data_loading_process.join()
         node_data_loading_process.start()
@@ -330,6 +335,11 @@ def load_torch_tensors_from_tf_record(
             f"Rank {rank} has finished loading data in {time.time() - start_time:.2f} seconds. Wait for other ranks to finish loading data from tfrecords"
         )
         barrier()
+
+    # Shutdown manager process to remove dependencies on large tensors,
+    # so that when we delete them in partitioner they will be garbage
+    # collected. Note that just deleting or clearing these output dicts here is not
+    # enough, and we have to shutdown the manager process.
 
     del node_output_dict, edge_output_dict, error_dict
 
--- a/python/gigl/common/data/load_torch_tensors.py
+++ b/python/gigl/common/data/load_torch_tensors.py
@@ -1,4 +1,3 @@
-import gc
 import time
 import traceback
 from dataclasses import dataclass
@@ -334,17 +333,6 @@ def load_torch_tensors_from_tf_record(
             f"Rank {rank} has finished loading data in {time.time() - start_time:.2f} seconds. Wait for other ranks to finish loading data from tfrecords"
         )
         barrier()
-
-    # Shutdown manager process to remove dependencies on large tensors,
-    # so that when we delete them in partitioner they will be garbage
-    # collected. Note that just deleting or clearing these output dicts here is not
-    # enough, and we have to shutdown the manager process.
-
-    del node_output_dict, edge_output_dict, error_dict
-
-    manager.shutdown()
-
-    gc.collect()
 
     logger.info(
         f"All ranks have finished loading data from tfrecords, rank {rank} finished in {time.time() - start_time:.2f} seconds"
--- a/python/gigl/common/data/load_torch_tensors.py
+++ b/python/gigl/common/data/load_torch_tensors.py
@@ -1,3 +1,4 @@
+import gc
 import time
 import traceback
 from dataclasses import dataclass
@@ -290,10 +291,10 @@ def load_torch_tensors_from_tf_record(
     else:
         # In this setting, we start and join each process one-at-a-time in order to achieve sequential tensor loading
         logger.info("Loading Serialized TFRecord Data in Sequence ...")
-        node_data_loading_process.start()
-        node_data_loading_process.join()
         edge_data_loading_process.start()
         edge_data_loading_process.join()
+        node_data_loading_process.start()
+        node_data_loading_process.join()
         if serialized_graph_metadata.positive_label_entity_info is not None:
             positive_label_data_loading_process.start()
             positive_label_data_loading_process.join()
@@ -329,6 +330,12 @@ def load_torch_tensors_from_tf_record(
             f"Rank {rank} has finished loading data in {time.time() - start_time:.2f} seconds. Wait for other ranks to finish loading data from tfrecords"
         )
         barrier()
+
+    del node_output_dict, edge_output_dict, error_dict
+
+    manager.shutdown()
+
+    gc.collect()
 
     logger.info(
         f"All ranks have finished loading data from tfrecords, rank {rank} finished in {time.time() - start_time:.2f} seconds"
--- a/python/gigl/common/data/load_torch_tensors.py
+++ b/python/gigl/common/data/load_torch_tensors.py
@@ -174,7 +174,8 @@ def load_torch_tensors_from_tf_record(
     serialized_graph_metadata: SerializedGraphMetadata,
     should_load_tensors_in_parallel: bool,
     rank: int = 0,
-    tf_dataset_options: TFDatasetOptions = TFDatasetOptions(),
+    node_tf_dataset_options: TFDatasetOptions = TFDatasetOptions(),
+    edge_tf_dataset_options: TFDatasetOptions = TFDatasetOptions(),
 ) -> LoadedGraphTensors:
     """
     Loads all torch tensors from a SerializedGraphMetadata object for all entity [node, edge, positive_label, negative_label] and edge / node types.
@@ -188,7 +189,8 @@ def load_torch_tensors_from_tf_record(
         serialized_graph_metadata (SerializedGraphMetadata): Serialized graph metadata contained serialized information for loading tfrecords across node and edge types
         should_load_tensors_in_parallel (bool): Whether tensors should be loaded from serialized information in parallel or in sequence across the [node, edge, pos_label, neg_label] entity types.
         rank (int): Rank on current machine
-        tf_dataset_options (TFDatasetOptions): The options to use when building the dataset.
+        node_tf_dataset_options (TFDatasetOptions): The options to use for nodes when building the dataset.
+        edge_tf_dataset_options (TFDatasetOptions): The options to use for edges when building the dataset.
     Returns:
         loaded_graph_tensors (LoadedGraphTensors): Unpartitioned Graph Tensors
     """
@@ -222,7 +224,7 @@ def load_torch_tensors_from_tf_record(
             "entity_type": _NODE_KEY,
             "serialized_tf_record_info": serialized_graph_metadata.node_entity_info,
             "rank": rank,
-            "tf_dataset_options": tf_dataset_options,
+            "tf_dataset_options": node_tf_dataset_options,
         },
     )
 
@@ -235,7 +237,7 @@ def load_torch_tensors_from_tf_record(
             "entity_type": _EDGE_KEY,
             "serialized_tf_record_info": serialized_graph_metadata.edge_entity_info,
             "rank": rank,
-            "tf_dataset_options": tf_dataset_options,
+            "tf_dataset_options": edge_tf_dataset_options,
         },
     )
 
@@ -249,7 +251,6 @@ def load_torch_tensors_from_tf_record(
                 "entity_type": _POSITIVE_LABEL_KEY,
                 "serialized_tf_record_info": serialized_graph_metadata.positive_label_entity_info,
                 "rank": rank,
-                "tf_dataset_options": tf_dataset_options,
             },
         )
     else:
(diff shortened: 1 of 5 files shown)


python/gigl/common/data/load_torch_tensors.py
```diff
@@ -290,10 +291,10 @@ def load_torch_tensors_from_tf_record(
     else:
         # In this setting, we start and join each process one-at-a-time in order to achieve sequential tensor loading
         logger.info("Loading Serialized TFRecord Data in Sequence ...")
-        node_data_loading_process.start()
-        node_data_loading_process.join()
         edge_data_loading_process.start()
         edge_data_loading_process.join()
+        node_data_loading_process.start()
+        node_data_loading_process.join()
```
kmontemayor2-sc, line 297: QQ why did we make this change?

python/gigl/common/data/load_torch_tensors.py
```diff
@@ -330,6 +331,12 @@ def load_torch_tensors_from_tf_record(
         )
         barrier()
 
+    del node_output_dict, edge_output_dict, error_dict
```
kmontemayor2-sc, line 334: for my own learning, why do we need to do this? It seems like we're still keeping references to the data in the dicts around so are we saving a lot of memory here?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -581,22 +582,27 @@ def _partition_single_chunk_data(
...
         """
         # chunk_res is a list where index `i` corresponds to Tuple[input_data_on_i, rank_indices_on_i]
-        chunk_res: List[Tuple[Optional[Tuple[torch.Tensor, ...]], torch.Tensor]] = []
+        chunk_res: List[
+            Tuple[Optional[Tuple[torch.Tensor, ...]], Optional[torch.Tensor]]
+        ] = []
         chunk_length = chunk_end_pos - chunk_start_pos
         chunk_rank = partition_function(rank_indices, (chunk_start_pos, chunk_end_pos))
 
+        input_indices = torch.arange(chunk_length, dtype=torch.int64)
```
kmontemayor2-sc, line 596: can we save some memory by not always setting this to int64? I suppose this in practice is pretty small (<100k) so maybe it doesn't matter too much?

python/gigl/distributed/dist_range_partitioner.py
```diff
@@ -191,7 +191,7 @@ def edge_partition_fn(rank_indices, _):
             partition_function=edge_partition_fn,
         )
 
-        del edge_index, target_indices, edge_feat
+        del input_data, edge_index, target_indices, edge_feat
```
kmontemayor2-sc, line 194: it seems like `input_data` is some reference to `self._edge_index[edge_type]` which we keep around, does adding this delete actually help with memory?

python/gigl/distributed/dist_range_partitioner.py
```diff
@@ -216,11 +216,20 @@ def edge_partition_fn(rank_indices, _):
...
+        if edge_feat_dim is not None:
+            if len(res_list) == 0:
+                partitioned_edge_features = torch.empty(0, edge_feat_dim)
+            else:
+                partitioned_edge_features = torch.cat([r[2] for r in res_list])
+
+        res_list.clear()
+
+        gc.collect()
+
```
kmontemayor2-sc, line 228: Was this where we actually saw the perf gains?

kmontemayor2-sc (commented):
Hi Matt, generally LGTM but in the future it'd be helpful if you could mention why some of these changes were made :) thanks!

python/gigl/common/data/load_torch_tensors.py
mkolodner-sc: Added a comment, its because loading edges first and then nodes data second reduces peak memory usage since edge data is larger than node data.

python/gigl/common/data/load_torch_tensors.py
mkolodner-sc: Without this step and particularly the `manager.shutdown()`, the references to these objects may stick around by the time we are doing partitioning, which means they won't get cleaned up when they should, leading to duplicate copies of the same data. These lines here ensure that we are not keeping any references around that don't need to be here by the end of the TFRecordLoading process.

Added a comment in the file as well

python/gigl/distributed/dist_partitioner.py
mkolodner-sc: This can be large, we hard-code the number of chunks to 8, so this number be as large as the `num_edges / 8`, which warrants using `torch.int64` here. In the setting where the number of edges is small to begin with, I imagine we aren't under a lot of memory pressure in the first place so it'd be fine to set it to  `torch.int64` in that case as well

python/gigl/distributed/dist_range_partitioner.py
mkolodner-sc: We delete this in the next line `del self._edge_index[edge_type]`

python/gigl/distributed/dist_range_partitioner.py
mkolodner-sc: Not specifically from this (most of the gains were from the `manager.shutdown()`, but this makes sense still to move up since we should be trying to clear the `res_list` as soon as possible (in particular, prior to generating the edge partition book and getting its ids), since this is very memory-expensive to keep around.

python/gigl/common/data/load_torch_tensors.py
kmontemayor2-sc: Wait why does changing the order here impact things? It seems to me like we'd still have the same overall memory cure right?

What about loading the larger data into memory first saves on peak usage? Is it possible some of the memory from the edge loading process can get reused by the node loading process? If so, why is the reverse not true?

python/gigl/common/data/load_torch_tensors.py
kmontemayor2-sc: hmmm isn't part of the reason we have `build_dataset` happen in a separate process so that all of this stuff gets cleaned up when the process ends? Or is it that the manager uses memory from outside the process? 

If it's true that we aren't actually cleaning up all resources by using a process, should we stop doing so?

mkolodner-sc: both this TFRecord loading logic and partitioning happen inside of the same spawned process [[1](https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/dataset_factory.py#L89-L146)] from `build_dataset`, which is why we still need to handle the cleanup here

python/gigl/common/data/load_torch_tensors.py
mkolodner-sc: These changes should be based on performance improvement observations from experimentation, cc @xgao4-sc who I believe may have some additional context from observing the performance improvement here

One possibility could be that the cleared memory from the edge loading process can be safely repurposed without concern for the node loading process, since the memory footprint of node loading is a strict subset of edge loading. However, if we try to repurpose the node loading memory, there may not be enough space at the memory address for the entire edge loading process (potentially that contiguous memory address only had 15 GB free for a 10 GB node loading process, but may not have 100 contiguous GB free for edge loading), resulting it in needing to be allocated somewhere else and perhaps not reusable in this case. It is true the overall memory is the same here, but perhaps these semantics might make a difference at a low-level somewhere.

python/gigl/common/data/load_torch_tensors.py
```diff
@@ -290,10 +291,15 @@ def load_torch_tensors_from_tf_record(
     else:
         # In this setting, we start and join each process one-at-a-time in order to achieve sequential tensor loading
         logger.info("Loading Serialized TFRecord Data in Sequence ...")
-        node_data_loading_process.start()
-        node_data_loading_process.join()
+        # Sequentially launch node, edge, and label loading processes to
+        # reduce the peak memory usage. Here we launch edge loading process
+        # first since edge data is larger than node data, so launching edge
+        # before launching node reduces peak memosry usage compared with
```
kmontemayor2-sc, line 297:
```suggestion
        # before launching node reduces peak memory usage compared with
```

nit.

python/gigl/common/data/load_torch_tensors.py
kmontemayor2-sc: Hmmm, in that case should we note that we found this experimentally?

kmontemayor2-sc (approved):
LGTM generally - I feel like there are some open questions wrt the processes and memory usage that we should note down as future work though.

python/gigl/common/data/load_torch_tensors.py
mkolodner-sc: > yes, but my own understanding is that we launch processes so that we automatically clean up resources when the processes close, right?

The item that we are manually cleaning up in this case is the manager and its corresponding `manager.dict()`, which are passed into each spawned TFRecordLoading process, rather than generated inside the spawned process. It remains true that everything that is created within each process is properly cleaned up with the exception of what we provide to the manager.dict(). The main point of using this manager in the first place is so that items we want from the spawned process are **not** cleaned up automatically and able to be leveraged outside this spawned process. That being said, there is still value in ensuring that all other intermediate objects created inside each TFRecord loading job is cleaned up using the spawned process here.

> Maybe my point is that if we're loading sequentially, why start new processes at all? Wouldn't it be faster to just do them all the loading in the same process?

We could, but the benefits here are 1) a unified interface between parallel and sequential loading and 2) stronger guarantees that any assets which are created within each spawned process are properly cleaned up properly upon exit.

python/gigl/common/data/load_torch_tensors.py
mkolodner-sc: Sounds good -- I'll add that note, thanks!

python/gigl/common/data/load_torch_tensors.py
kmontemayor2-sc: > The main point of using this manager in the first place is so that items we want from the spawned process are not cleaned up automatically and able to be leveraged outside this spawned process. The objective of this PR is to ensure that we cleanup the manager's references as well once we no longer need them, which share the same process as the partitioner.


So IOW we still suspect that all of the resources owned by the manager are cleanup when the dataloading processes ends, and this is just to free up memory *while* the dataloading process is running?

> We could, but the benefits here are 1) a unified interface between parallel and sequential loading

I'm not really sure there's a benefit here if we can get perf gains the other way, launching new processes isn't free. But again this is something we should look into as a follow up.

> 2) stronger guarantees that any assets which are created within each spawned process are properly cleaned up properly upon exit, other than what we provide back to the manager

Do we expect that there are other resources we want to clean up? I'm curious what we expect we are cleaning up in this case.

mkolodner-sc: > So IOW we still suspect that all of the resources owned by the manager are cleanup when the dataloading processes ends, and this is just to free up memory while the dataloading process is running?

Correct, this is to free up memory within the `build_dataset` process.

> I'm not really sure there's a strong benefit here if we can get perf gains the other way, launching new processes isn't free. But again this is something we should look into as a follow up.

Sounds good, thanks!

> Do we expect that there are other resources we want to clean up? I'm curious what we expect we are cleaning up in this case.

In this case it would be any remaining resources from using the TFRecordDataloader -- although I'm not sure if there are even any items here which are cleaned up with this spawned process, which provides guarantees of the cleanup. I'll leave this performance tradeoff of launching these processes in the sequential case to the subsequent investigation, thanks!

xgao4-sc: When manager and partitioner are in the **same function**, we need to delete dict and shutdown manager. Otherwise manager may hold references to the tensor causing the tensor cannot be deleted and garbage collected in the partitioner. In that case, when the function returns we observed a memory usage drop, due to the manager is destructed and the references to the tensor are removed, so that the tensor can be garbage collected.

Here manager and partitioner are in different functions, so I am wondering why we still need this. The manager will be shut down when this function returns.

How did we evaluate the gain?

python/gigl/distributed/dataset_factory.py
```diff
@@ -331,6 +331,10 @@ def build_dataset(
 
     output_dataset: DistLinkPredictionDataset = output_dict["dataset"]
 
+    del output_dict
+
+    manager.shutdown()
```
xgao4-sc, line 336: Same question here: does partitioning happen in this function?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -33,7 +33,7 @@ class _DistLinkPredicitonPartitionManager(DistPartitionManager):
     """
 
     def __init__(
-        self, world_size: int, total_val_size: int = 1, generate_pb: bool = True
+        self, world_size: int, total_val_size: int = 0, generate_pb: bool = False
```
xgao4-sc, line 36: Just curious, why do we need to change the default value?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -581,22 +582,27 @@ def _partition_single_chunk_data(
...
 
+        input_indices = torch.arange(chunk_length, dtype=torch.int64)
+
         for rank in range(self._world_size):
             # Filtering for items in chunk which are on the current partition `rank`
             current_rank_mask = chunk_rank == rank
-            per_rank_indices = torch.masked_select(
-                torch.arange(chunk_length, dtype=torch.long), current_rank_mask
-            )
+            per_rank_indices = torch.masked_select(input_indices, current_rank_mask)
```
xgao4-sc, line 601: Nice! We shouldn't repeatedly call `torch.arrange`. Thanks a lot!

python/gigl/distributed/dist_partitioner.py
```diff
@@ -581,22 +582,27 @@ def _partition_single_chunk_data(
...
             chunk_res.append(
                 (
                     index_select(input_data, per_rank_indices),
-                    rank_indices[per_rank_indices],
+                    rank_indices[per_rank_indices] if generate_pb else None,
```
xgao4-sc, line 605: Just curious, can `None` be sent over rpc, and why?

xgao4-sc (commented):
Thanks for the work! Please see my comments and questions below.

python/gigl/common/data/load_torch_tensors.py
xgao4-sc: The drop in memory usage matches what I observed when reversing the order of loading node and loading edge.

Did we add `time.sleep(480)` between different stages? If not, it is non-trivial to tell from the dashboard whether a peak is due to loading tf record, partitioning, or starting GLT loaders. We could roughly tell that

- The first peak is edge loading
- The last four peaks are starting 4 GLT data loaders.

However, the other peaks are really hard to tell. We need to add `time.sleep(480)` and compare dashboard with logs to tell which is which.

> is there any issue with being explicit about the manager shutdown in this case?
Recently I learned that for some multiprocessing libraries, `shutdown` is only a sub-step of destruction, and may not be enough for a graceful exit. I do not recommend having it everywhere unless we know it is very safe (like `gc.colllect()`), or we know we definitely need it (tested by adding `time.sleep(480)` and logs).

python/gigl/distributed/dist_partitioner.py
xgao4-sc: Sounds good! Why do we need to change the default value of `generate_pb`?

python/gigl/distributed/dist_partitioner.py
xgao4-sc: Sounds great, thanks!

python/gigl/common/data/load_torch_tensors.py
mkolodner-sc: Sounds good, in that case I'll go ahead and remove the `shutdown` calls for now and rerun the load test to ensure that the peak memory is still around `70%`. I'll add the sleeps in as well to ensure the peaks make sense.

xgao4-sc (approved):

PR #40, opened 2025-05-09
Add `_SetLabel` transform and `trainerArgs.should_convert_to_heterogeneous`. flag to enable labeled heterogeneous -> homogeneous conversion 

kmontemayor2-sc:
This should conclude the efforts (on the dataloader side) to super "homogeneous" supervised graphs and treat the outputs as homogeneous Data. The `test_cora_supervised` should be a sufficient "MVP training loop", right?

Changes:

* Add `_batch_store` to `DistNeighborLoader` to enable interprocess sharing of "full batches" (e.g. the specially constructed tensors). The batch store let's us preserve the full input batch with sentinels etc, which can be retrieved by the de-duped node ids.
* Add `_SetLabel` which uses some sentinel values to build `y_positive` and `y_negative` on the pyg Data.
* Modify `build_dataset_from_task_config_uri` to split the dataset when `trainerArgs.should_convert_to_heterogeneous = True`.


python/gigl/distributed/dataset_factory.py
```diff
@@ -372,13 +385,28 @@ def build_dataset_from_task_config_uri(
...
+        should_convert_to_heterogeneous = bool(
+            strtobool(args.get("should_convert_to_heterogeneous", "False"))
+        )
 
-    # Should be a string which is either "in" or "out"
-    sample_edge_direction = args.get("sample_edge_direction", "in")
+    if should_convert_to_heterogeneous:
+        # TODO(kmonte): Read train/val/test split counts from config.
+        splitter = HashedNodeAnchorLinkSplitter(
+            sampling_direction=sample_edge_direction,
```
zfan3-sc, line 402: It seems here we coupling message passing `sample_edge_direction` with which side is the anchor node when grouping by node to form anchor node based samples. In production this is decoupled, we tune "in" or "out" sampling by always let `src` in `src->dst` to be the anchor node. cc @xgao4-sc for vis too.

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -87,11 +116,123 @@ def __call__(self, data: HeteroData) -> Data:
...
+        batch_store: abc.MutableMapping[tuple[NodeType, tuple[int, ...]], torch.Tensor],
+        anchor_node_sentinel: int,
+        positive_label_sentinel: int,
+        negative_label_sentinel: Optional[int] = None,
+    ):
+        """
+        Sets the labels for nodes in Data.y_positive and Data.y_negative.
+        The labels are set based on the anchor node and positive label sentinels.
+        We assume that the input batch is a 1D tensor of node ids, or fornat:
+            `[<anchor node>, <anchor node sentinel>, <posittuple in theive labels>, <positive label sentinel>, <negative label>, <negative label sentinel>]`
```
zfan3-sc, line 136: > posittuple in theive labels

typo?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -209,6 +350,13 @@ def __init__(
                     f"num_neighbors must be a dict of edge types with the same number of hops. Received: {num_neighbors}"
                 )
 
+        if not hasattr(self, "_batch_store"):
```
zfan3-sc, line 353: I'm generally confused by the concept of batch_store. Could you add more context on why we want to introduce it in the PR description?

python/gigl/distributed/dataset_factory.py
xgao4-sc: Yes, in the partitioner we partition labels by source node. All label edges with the same source node are guaranteed to be on the same machine, and then we group labels by source node. @mkolodner-sc

xgao4-sc: >but if the label edge dir is not consistent with the rest of the graph then we have more network traffic when sampling right?

Why does direction affect network traffic? Label edge contain two nodes, and we need to sample subgraph of these two nodes. No matter whether we partition by source node or by destination node, we can only guarantee one node is on current machine.

kmontemayor2-sc: > we need to sample subgraph of these two nodes
Oh that's true, nvm.

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -254,15 +266,44 @@ def test_ablp_dataloader(
...
+            assert_tensor_equality(datum.y_negative, expected_negative_labels)
+        else:
+            self.assertFalse(hasattr(datum, "y_negative"))
         dsts, srcs, *_ = datum.coo()
         assert_tensor_equality(datum.node[srcs], expected_srcs)
         assert_tensor_equality(datum.node[dsts], expected_dsts)
 
+    def test_cora_supervised(self):
+        dataset = build_dataset_from_task_config_uri(
+            str(Path(__file__).parent / "cora_ssl_task_config.yaml"),
```
mkolodner-sc, line 287: Hmm what's the reason we choose to use the cora dataset here rather than relying on mocked assets for `cora_ssl_task_config`?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -406,39 +372,81 @@ def __init__(
...
         )
         positive_labels, negative_labels = get_labels_for_anchor_nodes(
             dataset, input_nodes, positive_label_edge_type, negative_label_edge_type
         )
-
+        extracted_input_nodes = input_nodes.unsqueeze(1)
         if negative_labels is not None:
+            anchor_sentinel = -2
+            positive_sentinel = -3
+            negative_sentinel = -4
```
yliu2-sc, line 403: this feels hacky and hard coded. I wonder if there is a cleaner way to do this.

iiuc the goal here is to identify the positive and negative labels when setting them to `.y_pos` `.y_neg`?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -209,6 +163,13 @@ def __init__(
                     f"num_neighbors must be a dict of edge types with the same number of hops. Received: {num_neighbors}"
                 )
 
+        if not hasattr(self, "_batch_store"):
+            logger.info("Setting batch store")
+            node_batches_by_sampled_nodes: abc.MutableMapping[
+                tuple[NodeType, tuple[int, ...]], torch.Tensor
+            ] = torch.multiprocessing.Manager().dict()
+            self._batch_store = node_batches_by_sampled_nodes
```
mkolodner-sc, line 171: Why does `DistNeighborLoader` need a `_batch_store`? Shouldn't this be something that is specific to `DistABLPLoader`?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -282,7 +243,9 @@ def __init__(
         else:
             node_type, node_ids = curr_process_nodes
 
-        input_data = _LabeledNodeSamplerInput(node=node_ids, input_type=node_type)
+        input_data = _LabeledNodeSamplerInput(
+            node=node_ids, input_type=node_type, batch_store=self._batch_store
+        )
```
mkolodner-sc, line 248: Similarly, can we just use `NodeSamplerInput` in this case? We shouldn't need to be concerned with labels for the `DistNeighborLoader` when doing inference, for example

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -406,39 +372,81 @@ def __init__(
...
         )
-
+        extracted_input_nodes = input_nodes.unsqueeze(1)
         if negative_labels is not None:
+            anchor_sentinel = -2
+            positive_sentinel = -3
+            negative_sentinel = -4
             input_nodes = torch.cat(
-                [input_nodes.unsqueeze(1), positive_labels, negative_labels], dim=1
+                [
```
zfan3-sc, line 405: nit: mind adding a comment here to illustrate what does `input_nodes` look like after the transformation?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -406,39 +372,81 @@ def __init__(
...
+                f"input_nodes must be a 1D tensor, got {input_nodes.shape}."
+            )
         positive_label_edge_type, negative_label_edge_type = select_label_edge_types(
             DEFAULT_HOMOGENEOUS_EDGE_TYPE, dataset.graph.keys()
         )
         positive_labels, negative_labels = get_labels_for_anchor_nodes(
             dataset, input_nodes, positive_label_edge_type, negative_label_edge_type
         )
-
+        extracted_input_nodes = input_nodes.unsqueeze(1)
```
mkolodner-sc, line 399: Adding more comments would be helpful for explaining what is happening here

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -406,39 +372,81 @@ def __init__(
...
+                [
+                    extracted_input_nodes,
+                    torch.full_like(extracted_input_nodes, anchor_sentinel),
+                    positive_labels,
+                    torch.full_like(extracted_input_nodes, positive_sentinel),
+                ],
+                dim=1,
+            )
+
+        node_batches_by_sampled_nodes: abc.MutableMapping[
```
zfan3-sc, line 429: In `DistNeighborLoader` we has a similar logic to initialize `self._batch_store`, is it okay to let `super().__init__` set it up?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -490,3 +498,200 @@ def shard(nodes: torch.Tensor) -> torch.Tensor:
...
+    """
+
+    def __init__(
+        self,
+        node: torch.Tensor,
+        input_type: Optional[Union[str, NodeType]],
+        batch_store: Optional[
+            abc.MutableMapping[tuple[NodeType, tuple[int, ...]], torch.Tensor]
+        ] = None,
+    ):
```
mkolodner-sc, line 535: Can we add descriptions of these parameters?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -490,3 +498,200 @@ def shard(nodes: torch.Tensor) -> torch.Tensor:
...
+    ) -> "_LabeledNodeSamplerInput":
+        if not isinstance(index, torch.Tensor):
+            index = torch.tensor(index, dtype=torch.long)
+        index = index.to(self.node.device)
+        full_batch = self.node[index].view(-1)
+        nodes_to_sample = full_batch[full_batch >= 0]
+        if self._batch_store is not None:
+            # Dedup the nodes to sample.
+            # We use a `dict` here to dedup the nodes, as `set` does not preserve
+            # insertion order, which need to keep we can differentiate betwee:
```
mkolodner-sc, line 553: nit: between

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -490,3 +498,200 @@ def shard(nodes: torch.Tensor) -> torch.Tensor:
...
+        # 4. Set the labels in the data object.
+        # TODO(kmonte): Since the labels are padded, we should be able to vectorize this.
+        for node_type in node_types:
+            full_batch: torch.Tensor
+            # data.batch is already de-duped.
+            if is_heterogeneous:
+                batch = tuple(data[node_type].batch.tolist())
+            else:
+                batch = tuple(data.batch.tolist())
+            full_batch = self._batch_store.pop((node_type, batch))
```
zfan3-sc, line 657: It looks like _batch_store will have big keys as batch can be as big as batch_size? Does it introduce extra memory overhead compared with the PoC implementation?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -490,3 +498,200 @@ def shard(nodes: torch.Tensor) -> torch.Tensor:
...
+        # 1. Get the "full batch" of node ids, containing the sentinels.
+        # e.g. [<anchor node>, <anchor node sentinel>, <positive labels>, <positive label sentinel>, <negative labels>, <negative label sentinel>]
+        # or [0, -1, 1, 2, -3, 3, -4]
+        # Where 0 is the anchor node, -1 is the anchor node sentinel, 1 and 2 are the positive labels,
+        # -3 is the positive label sentinel, 3 is the negative label, and -4 is the negative label sentinel.
+        # 2. Select the indices of all the sentinels.
+        # 3. Get the labels between the sentinels.
+        # 4. Set the labels in the data object.
+        # TODO(kmonte): Since the labels are padded, we should be able to vectorize this.
+        for node_type in node_types:
```
zfan3-sc, line 650: nit: The code block contains a lot of tensor operations, would be good to add comments on the expected tensor shapes for better readability

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -490,3 +498,200 @@ def shard(nodes: torch.Tensor) -> torch.Tensor:
...
+        negative_labels: dict[NodeType, torch.Tensor] = {}
+        if is_heterogeneous:
+            node_types = data.node_types
+        else:
+            node_types = [DEFAULT_HOMOGENEOUS_NODE_TYPE]
+
+        # The approach here is:
+        # For each node type:
+        # 1. Get the "full batch" of node ids, containing the sentinels.
+        # e.g. [<anchor node>, <anchor node sentinel>, <positive labels>, <positive label sentinel>, <negative labels>, <negative label sentinel>]
```
mkolodner-sc, line 642: qq: do we need to put the `negative_label_sentinel` in here if its at the end? If we are just getting node ids which are in between, wouldn't just doing all the nodes past the positive label sentinel be sufficient?

mkolodner-sc (commented):
Thanks Kyle, did an initial pass, left some comments

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: because we need to give `_batch_store` to `_LabeledNodeSamplerInput` which is created by `DistNeighborLoader`.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: We could use `NodeSamplerInput` but I think it's easier to always use `_LabeledNodeSamplerInput` since it behaves the same for the non-label use case.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: I can make the field optional though - that way we don't use the dict when we don't need to.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: sure, let me know if I've cleared this up.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: We have two goals here:

1. Sample anchor nodes and their labels together
2. Later, build `y_pos` and `y_neg`

I agree building this special tensor structure isn't ideal - and is a bit hacky.

Do you have an alternative suggestion here for how to accomplish our goals?

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: Done.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: actually modified a bit :P

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: thanks!

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: If we don't add `negative_label_sentinel` then it gets harder to tell when the negative labels end and the next anchor node is.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: There may be extra memory overhead - let's see how this scales and if we need to optimize for memory then we can?

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: Added some comments

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -490,3 +514,229 @@ def shard(nodes: torch.Tensor) -> torch.Tensor:
...
+        Sets the labels for nodes in Data.y_positive and Data.y_negative.
+        The labels are set based on the anchor node and positive label sentinels.
+        We assume that the input batch is a 1D tensor of node ids, or fornat:
+            `[<anchor node>, <anchor node sentinel>, <positive labels>, <positive label sentinel>, <negative label>, <negative label sentinel>]`
+        Will also set anchor node ids.
+
+
+        The returned data will have the following additional attributes:
+            * `y_positive`: A mapping from anchor node id -> positive label node ids.
+            * (Optionally)`y_negative`: A mapping from anchor node id -> negative label node ids.
```
kmontemayor2-sc, line 624: FYI I'm updating this as I think it makes our downstream use cases easier.

We do need to store `anchor_nodes` here somehow anyways.

python/gigl/distributed/dataset_factory.py
```diff
@@ -363,13 +376,27 @@ def build_dataset_from_task_config_uri(
...
+        sample_edge_direction = args.get("sample_edge_direction", "in")
         args_path = "inferencerConfig.inferencerArgs"
+        should_convert_to_heterogeneous = False
     else:
         args = dict(gbml_config_pb_wrapper.trainer_config.trainer_args)
+        sample_edge_direction = args.get("sample_edge_direction", "in")
         args_path = "trainerConfig.trainerArgs"
+        should_convert_to_heterogeneous = bool(
+            strtobool(args.get("should_convert_to_heterogeneous", "False"))
+        )
```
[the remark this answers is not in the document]
mkolodner-sc: Provided we have a way to convert SSL case to have labels (such as [here](https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/dataset_factory.py#L152-L180)), if we are using a training dataset (and thus our ABLPLoader), we should always have labels IIUC. 

> but I'm not sure it's correct to?

What reason would it not be correct for? At the very least, maybe we can set the default to `True` here instead of `False`

python/gigl/distributed/distributed_neighborloader.py
mkolodner-sc: Hmm, it seems like theres a lot of label/training coupling right now between the `DistNeighborLoader` and `DistABLPLoader`. From my perspective, we should have `DistNeighborLoader` not be responsible for any label-specific information, meaning we shouldn't require `DistNeighborLoader` to be aware of any `_batch_store` or `_LabeledNodeSamplerInput` logic (it should just use `_NodeSamplerInput)`. This is so that when we use a `DistNeighborLoader` for inference, we have much stronger guarantees that no trainer or label specific logic is accidentally interfering or causing performance issues with inference-specific code. Instead, we should have `_batch_store` and `_LabeledNodeSamplerInputs` only exist in the `DistABLPLoader`, and we should override the necessary fields/functions to achieve this from the `DistNeighborLoader`.

python/gigl/distributed/distributed_neighborloader.py
mkolodner-sc: Makes sense, thanks!

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -406,39 +366,99 @@ def __init__(
...
         )
-        transforms = [
+        transforms: Sequence[
+            Callable[[Union[Data, HeteroData]], Union[Data, HeteroData]]
+        ] = [
             _SupervisedToHomogeneous(
                 message_passing_edge_type=DEFAULT_HOMOGENEOUS_EDGE_TYPE,
-            )
+            ),
+            _SetLabels(
```
yliu2-sc, line 456: Here you're able to pass in some information to the `_SetLabels` transform. Can we just pass in the separate pos/neg information such label dim, or set pos/neg keys in the batch store to avoid having to set sentinels?

python/gigl/distributed/distributed_neighborloader.py
yliu2-sc: What would be the alternative to having batch store, in the POC solution?

python/gigl/distributed/dataset_factory.py
```diff
@@ -55,6 +61,7 @@ def _load_and_build_partitioned_dataset(
     partitioner_class: Optional[Type[DistPartitioner]],
     node_tf_dataset_options: TFDatasetOptions,
     edge_tf_dataset_options: TFDatasetOptions,
+    treat_ablp_graph_as_heterogeneous: bool = False,
```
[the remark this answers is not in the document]
kmontemayor2-sc: Sure, done.

python/gigl/distributed/distributed_neighborloader.py
yliu2-sc: possible to add more NodeType as key in batch store, instead of concatenating the sentinels?

python/gigl/distributed/distributed_neighborloader.py
mkolodner-sc: > Again, it's totally safe to use _LabeledNodeSamplerInput with unlabled data and is essentially a no-op.

This is the case now, but I am concerned about in the future if this class evolves whether this assumption will always be true I think achieving stronger guarantees through isolation that there is no performance drop in inference due to logic introduced in trainer is very important for us. 

> Do you have suggestions for how we could restructure the code here?

One option that comes to mind right now could be to introduce a private function `_get_node_sampler_input` which in the case of the `DistNeighborLoader` returns

```
def _get_node_sampler_input(self, node_ids, node_type):
   return NodeSamplerInput(node_ids, node_type)
```
and for `DistABLPLoader` returns
```
def _get_node_sampler_input(self, node_ids, node_type):
   return _LabeledNodeSamplerInput(node_ids, node_type, self._batch_store)
```

and in `DistNeighborLoader.__init__()` we just call:

```
input_data = self._get_node_sampler_input(node_ids, node_type)
``` 

This way, we can remove `self._batch_store` from the `DistNeighborLoader` altogether and don't need to have `DistNeighborLoader` responsible for any label-specific logic, which the DistABLPLoader can now be strictly responsible for.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: Hmmm what would that look like?  we could do something like `dict[NodeType, dict[tuple[int,...], Tensor]]` but I'm not sure that's cleaner?

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: excellent idea, thanks!

python/gigl/distributed/dataset_factory.py
```diff
@@ -372,13 +388,33 @@ def build_dataset_from_task_config_uri(
     )
     if is_inference:
         args = dict(gbml_config_pb_wrapper.inferencer_config.inferencer_args)
+        sample_edge_direction = args.get("sample_edge_direction", "in")
```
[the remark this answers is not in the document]
kmontemayor2-sc: yeah done (may have not commented this before).

python/gigl/distributed/distributed_neighborloader.py
yliu2-sc: I think another option to avoid sentinels, you could expand the tensor to 3D sparse or ragged tensor.

mkolodner-sc (approved):
Thanks Kyle! Generally this LGTM -- will be interested to see how this does when we have our training loop example ready. Appreciate the discussion here.

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: Ack, I've left this as a TODO to revist. I think the current solution works for the moment and if we run into performance/etc issues in the future we should be able to re-work it.

I agree that there are some things that aren't ideal here (specifically, _batch_store) but I am not sure that moving to ragged tensor will make this much simpler.

python/gigl/distributed/dataset_factory.py
```diff
@@ -96,6 +104,8 @@ def _load_and_build_partitioned_dataset(
         node_tf_dataset_options=node_tf_dataset_options,
         edge_tf_dataset_options=edge_tf_dataset_options,
     )
+    if should_convert_labels_to_edges:
```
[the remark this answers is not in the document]
kmontemayor2-sc: We will throw.

python/gigl/distributed/dataset_factory.py
```diff
@@ -351,6 +367,7 @@ def build_dataset_from_task_config_uri(
     task_config_uri: str,
     distributed_context: DistributedContext,
     is_inference: bool = True,
+    tfrecord_uri_pattern: str = ".*-of-.*\.tfrecord(\.gz)?$",
```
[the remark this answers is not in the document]
kmontemayor2-sc: I feel like we're probably only going to need this for our tests (famous last words :P).

Since `build_dataset_from_task_config_uri` is called by the users anyways I think they can do with it what they want and we shouldn't need to do some magic inference here.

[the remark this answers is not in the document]
mkolodner-sc: Hmm, if we'd only use this for testing, what do you think about using `build_dataset` instead of `build_dataset_from_task_config_uri` in this case, so that the public facing `build_dataset_from_task_config` can stay simpler in this case? Otherwise, we'd potentially have duplicate usage patterns between specifying the `tfrecord_uri_pattern` here and through some `inferencerArg` or `trainerArg`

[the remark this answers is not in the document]
kmontemayor2-sc: Hmmm, I guess it' s possible that there are different patterns in the future - I don't think it really hurst to add this I'm not sure if it complicates our APIs too much?

And again some of the "magic" is happening here in `build_dataset_from_task_config` which we want to test.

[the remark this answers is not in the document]
mkolodner-sc: Hmm, I'd prefer not to add a parameter here that would just be used for tests. More generally, the idea behind this function was to only rely on `task_config_uri` and `distributed_context` and the `is_inference` flag, the latter two of which are not available in the `task_config_uri`. I'm concerned adding this argument here may open up a possibility for making this function more complex, which is designed to be as simple as possible.

> And again some of the "magic" is happening here in build_dataset_from_task_config which we want to test.

Is it possible for us to move this "magic" to the `build_dataset` function? I think the `build_dataset_from_task_config` should just be a convenience to get fields from our orchestration, but we should probably ensure any important logic lives in `build_dataset`.

mkolodner-sc (approved):
Thanks a lot Kyle!

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -254,7 +254,7 @@ def test_ablp_dataloader(
         assert_tensor_equality(datum.node[dsts], expected_dsts)
 
     # TODO: (mkolodner-sc) - Re-enable this test once ports are dynamically inferred
-    @unittest.skip("Failing due to ports being already allocated - skiping for now")
+    # @unittest.skip("Failing due to ports being already allocated - skiping for now")
```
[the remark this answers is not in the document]
kmontemayor2-sc: sure, will wait :)

PR #42, opened 2025-05-13
Upgrade type annotations for graph conversions

kmontemayor2-sc:
Doing this as I ran into some "bugs" with `to_homogeneous` when I was working on the dataloader stuff...

Documented a bit in the comment over `_GraphEntity` but essentially the issue is that when we had:

```py
def to_homogeneous(x: _T | dict[NodeType | EdgeType, _T]) -> _T:
```

The first '_T` would capture all dicts/etc, and the return type would be ambiguous (I ran into it being deduced as `object`, confusingly.

Ideally, we'd say something like `_T = TypeVar("_T", not=dict)` but that is not supported so we need to enumerate all the types it *may* be instead.

Not so many buisness logic changes so I expect this to be a safe change 🤞

### DIFF
--- a/python/gigl/types/graph.py
+++ b/python/gigl/types/graph.py
@@ -245,6 +245,10 @@ _GraphEntity = TypeVar(
     SerializedTFRecordInfo,
     Optional[SerializedTFRecordInfo],
     list,
+    # TODO(kmonte): Add GLT Partition book here
+    # We cannot at the moment as we mypy ignore GLT
+    # And adding it as a type here will break mypy.
+    # PartitionBook
 )
 
 
--- a/python/gigl/types/graph.py
+++ b/python/gigl/types/graph.py
@@ -233,15 +233,17 @@ def select_label_edge_types(
 # e.g. `reveal_type(to_homogeneous(d: Tensor | dict[..., Tensor] | None]))` is `object`
 # Instead, we enumerate these types, as MyPy does not allow "not" in a TypeVar.
 # We should extend this as necessary, just make sure *never* add any Mapping types.
+# NOTE: We have `Optional[SerializedTFRecordInfo]` in the type,
+# As adding `None` and `SerializedTFRecordInfo` separately do not accomplish the equivalent thing.
+# I believe this is due to the fact that the contraints on a `TypeVar` are not
+# are not treated as a union of the types, but rather each as their own case.
 _GraphEntity = TypeVar(
     "_GraphEntity",
     torch.Tensor,
     GraphPartitionData,
     FeaturePartitionData,
     SerializedTFRecordInfo,
-    Optional[
-        SerializedTFRecordInfo
-    ],  # Adding `None` here doesn't work for some reason...
+    Optional[SerializedTFRecordInfo],
     list,
 )
 
--- a/python/gigl/types/graph.py
+++ b/python/gigl/types/graph.py
@@ -5,6 +5,7 @@ from typing import Optional, TypeVar, Union, overload
 import torch
 from graphlearn_torch.partition import PartitionBook
 
+from gigl.common.data.dataloaders import SerializedTFRecordInfo
 from gigl.common.logger import Logger
 
 # TODO(kmonte) - we should move gigl.src.common.types.graph_data to this file.
@@ -168,10 +169,12 @@ def message_passing_to_positive_label(
         str(message_passing_edge_type[2]),
     )
     if isinstance(message_passing_edge_type, EdgeType):
+        print(f"returning {edge_type} as GiGL EdgeType")
         return EdgeType(
             NodeType(edge_type[0]), Relation(edge_type[1]), NodeType(edge_type[2])
         )
     else:
+        print(f"returning {edge_type} as PyG EdgeType")
         return edge_type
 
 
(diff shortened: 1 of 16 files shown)


python/gigl/types/graph.py
```diff
@@ -5,6 +5,7 @@
 import torch
 from graphlearn_torch.partition import PartitionBook
 
+from gigl.common.data.dataloaders import SerializedTFRecordInfo
```
kmontemayor2-sc, line 8: Note: This is (a little) concerning - we should consider migrating this eventually.

python/gigl/types/graph.py
```diff
@@ -266,24 +286,26 @@ def to_heterogeneous_edge(x: None) -> None:
...
+    x: Union[_GraphEntity, dict[EdgeType, _GraphEntity]]
+) -> dict[EdgeType, _GraphEntity]:
     ...
 
 
 def to_heterogeneous_edge(
-    x: Optional[Union[_T, dict[EdgeType, _T]]]
-) -> Optional[dict[EdgeType, _T]]:
+    x: Optional[Union[_GraphEntity, dict[EdgeType, _GraphEntity]]]
+) -> Optional[dict[EdgeType, _GraphEntity]]:
```
kmontemayor2-sc, line 297: We should move these to `abc.MutableMapping` eventually too but that requires more changes and I'd rather this PR not get to big.

python/gigl/types/graph.py
```diff
@@ -225,7 +226,24 @@ def select_label_edge_types(
...
+# We should extend this as necessary, just make sure *never* add any Mapping types.
+_GraphEntity = TypeVar(
+    "_GraphEntity",
+    torch.Tensor,
+    GraphPartitionData,
+    FeaturePartitionData,
+    SerializedTFRecordInfo,
+    Optional[
+        SerializedTFRecordInfo
+    ],  # Adding `None` here doesn't work for some reason...
```
mkolodner-sc, line 244: What's the error we are seeing when we try to include `None` here? Is the `Optional[SerializedTFRecordInfo]` the workaround to get None type added in this case?

python/gigl/types/graph.py
```diff
@@ -225,7 +226,24 @@ def select_label_edge_types(
...
+# As if we type `to_homogeneous(x: _T | dict[NodeType, _T] | dict[EdgeType, _T]) -> _T`,
+# then `_T` captures the "dict" types, and the output type is not correctly narrowed.
+# e.g. `reveal_type(to_homogeneous(d: Tensor | dict[..., Tensor] | None]))` is `object`
+# Instead, we enumerate these types, as MyPy does not allow "not" in a TypeVar.
+# We should extend this as necessary, just make sure *never* add any Mapping types.
+_GraphEntity = TypeVar(
+    "_GraphEntity",
+    torch.Tensor,
+    GraphPartitionData,
+    FeaturePartitionData,
```
mkolodner-sc, line 240: Should we add `PartitionBook` here as well?

python/gigl/types/graph.py
```diff
@@ -298,13 +320,29 @@ def to_homogeneous(x: None) -> None:
 
 
 @overload
-def to_homogeneous(x: Union[_T, dict[Union[NodeType, EdgeType], _T]]) -> _T:
```
mkolodner-sc, line 301: do we still need 
```
@overload
def to_homogeneous(x: None) -> None:
```
if None is included in `_GraphEntity `? Same question for other relevant functions

kmontemayor2-sc: None is not included in `_GraphEntity`, it's part of the `Optional[SerializedTFRecordInfo]`. 

This is a bit surprising but each of the constraints (e.g. the types in the TypeVar) are treated independently, it's not treated as a big `Union`.

python/gigl/types/graph.py
mkolodner-sc: There seem to be existing cases where `PartitionBook` gets used with `to_homogeneous` [[1](https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/dist_range_partitioner.py#L351)], I imagine the reason its passing though is because GLT has no type stub 😆 . Although I'm not sure what will happen if we put a `type: ignore`'d type like `PartitionBook` into the `TypeVar` here.

python/gigl/types/graph.py
mkolodner-sc: Hmm, I wonder if adding `NoneType` instead of `None` work in this case?

mkolodner-sc: If that doesn't work, then sounds good to leave it as is, thanks :)

mkolodner-sc (approved):
Generally LGTM provided tests pass, thanks for the work Kyle :)

python/gigl/types/graph.py
kmontemayor2-sc: I tried `None, type(None), and type[None]` and none were successful - the precise error here is:

```
python/gigl/distributed/utils/serialized_graph_metadata_translator.py:165: error: Value of type variable "_GraphEntity" of "to_homogeneous" cannot be "Optional[SerializedTFRecordInfo]"  [type-var]
python/gigl/distributed/utils/serialized_graph_metadata_translator.py:165: error: Argument 1 to "to_homogeneous" has incompatible type "dict[EdgeType, Optional[SerializedTFRecordInfo]]"; expected "Mapping[NodeType, Optional[SerializedTFRecordInfo]]"  [arg-type]
python/gigl/distributed/utils/serialized_graph_metadata_translator.py:166: error: Value of type variable "_GraphEntity" of "to_homogeneous" cannot be "Optional[SerializedTFRecordInfo]"  [type-var]
python/gigl/distributed/utils/serialized_graph_metadata_translator.py:166: error: Argument 1 to "to_homogeneous" has incompatible type "dict[EdgeType, Optional[SerializedTFRecordInfo]]"; expected "Mapping[NodeType, Optional[SerializedTFRecordInfo]]"  [arg-type]
```

python/gigl/types/graph.py
kmontemayor2-sc: hmmm actually adding `PartitionBook` doesn't work (probably because we mypy ignore it.

```
python/gigl/types/graph.py:253: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
python/gigl/types/graph.py:287: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
```

When I copy over the definition to our codebase it WAI - I'll add a todo to add it as a type here once accepted.

PR #45, opened 2025-05-13
Update config populator and validation for GLT pipelines

mkolodner-sc:
**Scope of work done**

Generally, this PR aims to clean up the unused fields in the template and populated frozen configs for GiGL V2 (GLT) pipelines. Specifically, we:

- Remove `flattened_graph_metadata` (v1 split_gen outputs) and `dataset_metadata` (v1 sgs outputs) fields created in `config_populator` for GLT Pipelines
- Remove validation check requirement that SGS fields be provided in template config
- Rename `gbml_config_pb_wrapper` field `should_run_experimental_glt_backend` -> `should_run_glt_backend`
- Allow `gbml_config_pb_wrapper.flattened_graph_metadata_pb_wrapper` and `gbml_config_pb_wrapper.dataset_metadata` to be `None` provided the above change for GLT pipelines, error if pb wrapper is accessed when this is not set
- Remove unused fields in example config for GLT 

**Did you add automated tests or write a test plan?**
Frozen config for CORA <> GLT E2E test after changes from [e2e job
](https://console.cloud.google.com/vertex-ai/pipelines/locations/us-central1/runs/cora-glt-udl-test-on--74047?project=external-snap-ci-github-gigl&inv=1&invt=AbxUYQ)
```
datasetConfig:
  dataPreprocessorConfig:
    dataPreprocessorArgs:
      mocked_dataset_name: cora_homogeneous_node_anchor_edge_features_user_defined_labels
    dataPreprocessorConfigClsPath: gigl.src.mocking.mocking_assets.passthrough_preprocessor_config_for_mocked_assets.PassthroughPreprocessorConfigForMockedAssets
featureFlags:
  data_preprocessor_num_shards: '2'
  should_run_glt_backend: 'True'
graphMetadata:
  condensedEdgeTypeMap:
    '0':
      dstNodeType: paper
      relation: cites
      srcNodeType: paper
  condensedNodeTypeMap:
    '0': paper
  edgeTypes:
  - dstNodeType: paper
    relation: cites
    srcNodeType: paper
  nodeTypes:
  - paper
inferencerConfig:
  command: python -m examples.distributed.homogeneous_inference
  inferenceBatchSize: 512
  inferencerArgs:
    log_every_n_batch: '50'
sharedConfig:
  inferenceMetadata:
    nodeTypeToInferencerOutputInfoMap:
      paper:
        embeddingsPath: external-snap-ci-github-gigl.gigl_embeddings.embeddings_paper_cora_glt_udl_test_on__74047
  postprocessedMetadata:
    postProcessorLogMetricsUri: gs://gigl-cicd-perm/cora_glt_udl_test_on__74047/post_processor/post_processor_metrics.json
  preprocessedMetadataUri: gs://gigl-cicd-perm/cora_glt_udl_test_on__74047/data_preprocess/preprocessed_metadata.yaml
  shouldSkipModelEvaluation: true
  shouldSkipTraining: true
  trainedModelMetadata:
    evalMetricsUri: gs://gigl-cicd-perm/cora_glt_udl_test_on__74047/trainer/models/trainer_eval_metrics.json
    scriptedModelUri: gs://gigl-cicd-perm/cora_glt_udl_test_on__74047/trainer/models/scripted_model.pt
    tensorboardLogsUri: gs://gigl-cicd-perm/cora_glt_udl_test_on__74047/trainer/tensorboard_logs/
    trainedModelUri: gs://public-gigl/mocked_assets/2024-07-15--21-30-07-UTC/cora_homogeneous_node_anchor_edge_features_user_defined_labels/trainer/models/model.pt
taskMetadata:
  nodeAnchorBasedLinkPredictionTaskMetadata:
    supervisionEdgeTypes:
    - dstNodeType: paper
      relation: cites
      srcNodeType: paper
```

### DIFF
--- a/python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
+++ b/python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
@@ -148,7 +148,6 @@ class SupervisedNodeClassificationDatasetDataloaders:
         gbml_config_pb_wrapper: GbmlConfigPbWrapper,
         data_loader_types: List[DataloaderTypes],
     ) -> Dict[DataloaderTypes, str]:
-        assert gbml_config_pb_wrapper.dataset_metadata_pb_wrapper is not None
         dataset_pb: dataset_metadata_pb2.SupervisedNodeClassificationDataset = (
             gbml_config_pb_wrapper.dataset_metadata_pb_wrapper.dataset_metadata_pb.supervised_node_classification_dataset
         )
--- a/python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
+++ b/python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
@@ -148,6 +148,7 @@ class SupervisedNodeClassificationDatasetDataloaders:
         gbml_config_pb_wrapper: GbmlConfigPbWrapper,
         data_loader_types: List[DataloaderTypes],
     ) -> Dict[DataloaderTypes, str]:
+        assert gbml_config_pb_wrapper.dataset_metadata_pb_wrapper is not None
         dataset_pb: dataset_metadata_pb2.SupervisedNodeClassificationDataset = (
             gbml_config_pb_wrapper.dataset_metadata_pb_wrapper.dataset_metadata_pb.supervised_node_classification_dataset
         )
@@ -340,6 +341,7 @@ class NodeAnchorBasedLinkPredictionDatasetDataloaders:
         gbml_config_pb_wrapper: GbmlConfigPbWrapper,
         data_loader_types: List[DataloaderTypes],
     ) -> Dict[DataloaderTypes, Union[str, Dict[str, str]]]:
+        assert gbml_config_pb_wrapper.dataset_metadata_pb_wrapper is not None
         dataset_pb: dataset_metadata_pb2.NodeAnchorBasedLinkPredictionDataset = (
             gbml_config_pb_wrapper.dataset_metadata_pb_wrapper.dataset_metadata_pb.node_anchor_based_link_prediction_dataset
         )
--- a/python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
+++ b/python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
@@ -340,7 +340,6 @@ class NodeAnchorBasedLinkPredictionDatasetDataloaders:
         gbml_config_pb_wrapper: GbmlConfigPbWrapper,
         data_loader_types: List[DataloaderTypes],
     ) -> Dict[DataloaderTypes, Union[str, Dict[str, str]]]:
-        assert gbml_config_pb_wrapper.dataset_metadata_pb_wrapper is not None
         dataset_pb: dataset_metadata_pb2.NodeAnchorBasedLinkPredictionDataset = (
             gbml_config_pb_wrapper.dataset_metadata_pb_wrapper.dataset_metadata_pb.node_anchor_based_link_prediction_dataset
         )
--- a/python/gigl/src/common/types/pb_wrappers/gbml_config.py
+++ b/python/gigl/src/common/types/pb_wrappers/gbml_config.py
@@ -46,19 +46,21 @@ class GbmlConfigPbWrapper:
 
     gbml_config_pb: gbml_config_pb2.GbmlConfig
 
-    _dataset_metadata_pb_wrapper: Optional[DatasetMetadataPbWrapper] = None
+    _dataset_metadata_pb_wrapper: Optional[DatasetMetadataPbWrapper] = field(
+        default=None, init=False
+    )
     _graph_metadata_pb_wrapper: GraphMetadataPbWrapper = field(init=False)
     _preprocessed_metadata_pb_wrapper: PreprocessedMetadataPbWrapper = field(init=False)
     _flattened_graph_metadata_pb_wrapper: Optional[
         FlattenedGraphMetadataPbWrapper
-    ] = None
+    ] = field(default=None, init=False)
     _task_metadata_pb_wrapper: TaskMetadataPbWrapper = field(init=False)
     _trained_model_metadata_pb_wrapper: TrainedModelMetadataPbWrapper = field(
         init=False
     )
     _subgraph_sampling_strategy_pb_wrapper: Optional[
         SubgraphSamplingStrategyPbWrapper
-    ] = None
+    ] = field(default=None, init=False)
 
     def __post_init__(self):
         # Populate the _preprocessed_metadata_pb_wrapper field
--- a/python/gigl/src/common/types/pb_wrappers/gbml_config.py
+++ b/python/gigl/src/common/types/pb_wrappers/gbml_config.py
@@ -46,19 +46,19 @@ class GbmlConfigPbWrapper:
 
     gbml_config_pb: gbml_config_pb2.GbmlConfig
 
-    _dataset_metadata_pb_wrapper: Optional[DatasetMetadataPbWrapper] = field(init=False)
+    _dataset_metadata_pb_wrapper: Optional[DatasetMetadataPbWrapper] = None
     _graph_metadata_pb_wrapper: GraphMetadataPbWrapper = field(init=False)
     _preprocessed_metadata_pb_wrapper: PreprocessedMetadataPbWrapper = field(init=False)
     _flattened_graph_metadata_pb_wrapper: Optional[
         FlattenedGraphMetadataPbWrapper
-    ] = field(init=False)
+    ] = None
     _task_metadata_pb_wrapper: TaskMetadataPbWrapper = field(init=False)
     _trained_model_metadata_pb_wrapper: TrainedModelMetadataPbWrapper = field(
         init=False
     )
     _subgraph_sampling_strategy_pb_wrapper: Optional[
         SubgraphSamplingStrategyPbWrapper
-    ] = field(init=False)
+    ] = None
 
     def __post_init__(self):
         # Populate the _preprocessed_metadata_pb_wrapper field
@@ -68,9 +68,12 @@ class GbmlConfigPbWrapper:
         # Populate the _dataset_metadata_pb_wrapper field
         if self.gbml_config_pb.shared_config.HasField("dataset_metadata"):
             dataset_metadata_pb = self.gbml_config_pb.shared_config.dataset_metadata
+            self.__load_dataset_metadata_pb_wrapper(dataset_metadata_pb)
         else:
-            dataset_metadata_pb = None
-        self.__load_dataset_metadata_pb_wrapper(dataset_metadata_pb)
+            logger.info(
+                "Skipping populating dataset_metadata_pb as the message is missing from the input config"
+            )
+
         # Populate the _graph_metadata_pb_wrapper field
         self.__load_graph_metadata_pb_wrapper(
             graph_metadata_pb=self.gbml_config_pb.graph_metadata
@@ -80,9 +83,12 @@ class GbmlConfigPbWrapper:
             flattened_graph_metadata_pb = (
                 self.gbml_config_pb.shared_config.flattened_graph_metadata
             )
+            self.__load_flattened_graph_metadata_pb_wrapper(flattened_graph_metadata_pb)
         else:
-            flattened_graph_metadata_pb = None
-        self.__load_flattened_graph_metadata_pb_wrapper(flattened_graph_metadata_pb)
+            logger.info(
+                "Skipping populating flattened_graph_metadata_pb_wrapper as the message is missing from the input config"
+            )
+
         # Populate the _task_metadata_pb_wrapper field
         self.__load_task_metadata_pb_wrapper(
             task_metadata_pb=self.gbml_config_pb.task_metadata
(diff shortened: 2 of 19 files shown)


python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -46,37 +46,43 @@ class GbmlConfigPbWrapper:
...
+        FlattenedGraphMetadataPbWrapper
+    ] = field(init=False)
     _task_metadata_pb_wrapper: TaskMetadataPbWrapper = field(init=False)
     _trained_model_metadata_pb_wrapper: TrainedModelMetadataPbWrapper = field(
         init=False
     )
-    _subgraph_sampling_strategy_pb_wrapper: SubgraphSamplingStrategyPbWrapper = field(
-        init=False
-    )
+    _subgraph_sampling_strategy_pb_wrapper: Optional[
```
mkolodner-sc, line 59: This should have always been `Optional`, not sure why mypy didn't catch this but have adjusted this now

python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
```diff
@@ -148,6 +148,7 @@ def _get_uri_prefix_map(
         gbml_config_pb_wrapper: GbmlConfigPbWrapper,
         data_loader_types: List[DataloaderTypes],
     ) -> Dict[DataloaderTypes, str]:
+        assert gbml_config_pb_wrapper.dataset_metadata_pb_wrapper is not None
```
kmontemayor2-sc, line 151: These asserts are a bit messy, can we have `the dataset_metadata_pb_wrapper` be some default wrapper (e.g. for an empty proto) if not provided? We can add an addition `has_dataset_metadata_pb_wrapper` to represent if it is intentionally empty or was just unset.

This goes along with how py protos handle nested messages https://protobuf.dev/reference/python/python-generated/#embedded_message

python/tests/unit/src/config_populator/config_populator_functionality_test.py
```diff
@@ -17,18 +19,50 @@
 
 logger = Logger()
 
+_TEMPLATE_V1_GBML_CONFIG_PB = gbml_config_pb2.GbmlConfig(
+    task_metadata=gbml_config_pb2.GbmlConfig.TaskMetadata(
+        node_based_task_metadata=gbml_config_pb2.GbmlConfig.TaskMetadata.NodeBasedTaskMetadata()
+    ),
+)
+
+_TEMPLATE_V2_GBML_CONFIG_PB = gbml_config_pb2.GbmlConfig(
```
kmontemayor2-sc, line 28: nit. Let's name this `"TEMPLATE_CONFIG_FOR_SGS` and `TEMPLATE_CONFIG_FOR_GLT` or equivalent? "V1" and "V2" aren't too descript imo.

python/tests/unit/src/config_populator/config_populator_functionality_test.py
```diff
@@ -39,32 +73,64 @@ def test_config_population_is_accurate(self):
...
+                gbml_config_pb_wrapper.shared_config.HasField("dataset_metadata")
+            )
+
+            # GBML Config Pb Wrapper Checks
+            self.assertTrue(gbml_config_pb_wrapper.should_use_glt_backend)
+            self.assertIsNone(
+                gbml_config_pb_wrapper.flattened_graph_metadata_pb_wrapper
+            )
+            self.assertIsNone(gbml_config_pb_wrapper.dataset_metadata_pb_wrapper)
+        else:
```
kmontemayor2-sc, line 92: I think we should avoid branches in tests as much as possible, in this case I'd have a (minor preference) for having two test cases with some duplicate logic. But this is a very small point, I just wanted to bring up the branching thing generally.

python/gigl/src/common/types/pb_wrappers/dataset_metadata_utils.py
mkolodner-sc: Hmm, I had initially thought about approaching it this way, but my concern here was that a `pb_wrapper` which is provided an empty pb, and thus becomes an empty pb wrapper, feels a bit weird to me, in the sense that the pb wrapper itself becomes only partially functional, where some of its properties/functions may either be empty or just error out [[1](https://github.com/Snapchat/GiGL/blob/main/python/gigl/src/common/types/pb_wrappers/flattened_graph_metadata.py#L72-L80)]. This leads me to believe that generally, we shouldn't be populating a `pb_wrapper` if its internal `pb` is empty, and even if we do, we should probably still have some sort of `assert` in place in these locations whenever its used to prevent unforeseen behavior with using an empty pb.

Provided we'd need these `asserts` anyway, my reasoning was that it would be easier just to set the `pb_wrapper` to None and remove this state of limbo altogether. WDYT?

kmontemayor2-sc: > Hmm, I had initially thought about approaching it this way, but my concern here was that a pb_wrapper which is provided an empty pb, and thus becomes an empty pb wrapper, feels a bit weird to me, in the sense that the pb wrapper itself becomes only partially functional, where some of its properties/functions may either be empty

What's wrong with this? We don't require all fields to be set anyways so they be empty anyways - it should be the consuming code's responsibility to check, right?  I suppose the asserts are checking but they don't really provide any actionable advice to the user when we hit them, right?

I'd prefer if we could adopt more of things like:

```
if blah:
   raise ValueError("SGS expect TaskMetadata to be set, we did not receive that field. Was this config inteded to be run for the GLT backend?")
```

in our codebase which I feel like is much more "use friendly"

> or just error out [[1](https://github.com/Snapchat/GiGL/blob/main/python/gigl/src/common/types/pb_wrappers/flattened_graph_metadata.py#L72-L80)]

That's fine? I mean if the code breaks it breaks?

> Provided we'd need these asserts anyway, my reasoning was that it would be easier just to set the pb_wrapper to None and remove this state of limbo altogether. WDYT?

What state of limbo are we worried about here?

FWIW having empty messages as default values is nice as you can do stuff like `top_level.nested_message.some_field` without needing to do:

```
maybe_nested = top_level.nested_message
if nested is not None:
   foo = maybe_nested.some_field
```

Again I think that if we're trying to "conform to known" then we should align our pb wrapper with the pb behavior.

mkolodner-sc: Discussed offline, similar to above, we'll opt to throw an error when trying to access a `pb_wrapper` with an empty underlying proto so that can reduce the amount of asserts here while ensuring that an incomplete pb wrapper isn't used mistakenly.

python/gigl/src/common/types/pb_wrappers/gbml_config.py
kmontemayor2-sc: should we have this throw when not set too, instead of being optional?

python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -273,11 +287,15 @@ def get_gbml_config_pb_wrapper_from_uri(
...
-        Allows access to a dataset_metadata_pb_wrapper
+        Allows access to a dataset_metadata_pb_wrapper. If the underlying pb is empty, calling this property will instead throw an error.
 
         Returns:
             DatasetMetadataPbWrapper: The dataset metadata pb wrapper
         """
+        if self._dataset_metadata_pb_wrapper is None:
+            raise ValueError(
+                "Attempted to access the dataset metadata pb wrapper when it is not set. Please ensure this property is only accessed when the underlying proto is set"
+            )
```
kmontemayor2-sc, line 298: Let's add a note that this may not be set if using the GLT backend, and is probably being read if using the SGS backend?

python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -313,13 +331,19 @@ def task_metadata_pb_wrapper(self) -> TaskMetadataPbWrapper:
...
-        Allows access to a flattened_graph_metadata_pb_wrapper
+        Allows access to a flattened_graph_metadata_pb_wrapper.  If the underlying pb is empty, calling this property will instead throw an error.
 
         Returns:
-            FlattenedGraphMetadataPbWrapper: The flattened graph metadata pb wrapper
+            FlattenedGraphMetadataPbWrapper: The flattened graph metadata pb wrapper or None if it doesn't exist
         """
+        if self._flattened_graph_metadata_pb_wrapper is None:
+            raise ValueError(
+                "Attempted to access the flattened graph metadata pb wrapper when it is not set. Please ensure this property is only accessed when the underlying proto is set"
```
kmontemayor2-sc, line 345: ditto on why this may happen

python/gigl/src/validation_check/libs/template_config_checks.py
```diff
@@ -366,6 +366,12 @@ def check_if_subgraph_sampler_config_valid(
     Check if subgraphSamplerConfig is valid.
     """
     logger.info("Config validation check: if subgraphSamplerConfig is valid.")
+    gbml_config_wrapper = GbmlConfigPbWrapper(gbml_config_pb=gbml_config_pb)
+    if gbml_config_wrapper.should_use_glt_backend:
+        logger.warning(
+            "Skipping subgraph sampler validation check since GLT Backend is being used."
```
kmontemayor2-sc, line 372: skipping SGS and split gen validation checks?

python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -46,37 +46,43 @@ class GbmlConfigPbWrapper:
...
-        self.__load_flattened_graph_metadata_pb_wrapper(
-            flattened_graph_metadata_pb=self.gbml_config_pb.shared_config.flattened_graph_metadata
-        )
+        if self.gbml_config_pb.shared_config.HasField("flattened_graph_metadata"):
+            flattened_graph_metadata_pb = (
+                self.gbml_config_pb.shared_config.flattened_graph_metadata
+            )
+        else:
+            flattened_graph_metadata_pb = None
+        self.__load_flattened_graph_metadata_pb_wrapper(flattened_graph_metadata_pb)
```
kmontemayor2-sc, line 85:
```suggestion
        if self.gbml_config_pb.shared_config.HasField("flattened_graph_metadata"):             self.__load_flattened_graph_metadata_pb_wrapper(self.gbml_config_pb.shared_config.flattened_graph_metadata)

        else:
            logger.info("Skipping populating flattened_graph_metadata_pb_wrapper as the message is missing from the input config")
```

WDYT about?

kmontemayor2-sc: IF we do it this way we don't need to make the argument to `__load_flattened_graph_metadata_pb_wrapper` optional.

Ditto for the other fields like this.

python/gigl/src/validation_check/libs/template_config_checks.py
mkolodner-sc: Split_gen validation check and skipping is handled separately in the `check_if_split_generator_config_valid` function

python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -178,10 +197,8 @@ def __load_flattened_graph_metadata_pb_wrapper(
...
             flattened_graph_metadata_pb (flattened_graph_metadata_pb2.FlattenedGraphMetadata): The flattened graph metadata proto
         """
-        flattened_graph_metadata_pb_wrapper: FlattenedGraphMetadataPbWrapper = (
-            FlattenedGraphMetadataPbWrapper(
-                flattened_graph_metadata_pb=flattened_graph_metadata_pb
-            )
+        flattened_graph_metadata_pb_wrapper = FlattenedGraphMetadataPbWrapper(
+            flattened_graph_metadata_pb=flattened_graph_metadata_pb
         )
         object.__setattr__(
```
kmontemayor2-sc, line 203: can we do `self._flattened_graph_metadata_pb_wrapper = flattened_graph_metadata_pb_wrapper` here? (ditto generally)

python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -313,13 +332,20 @@ def task_metadata_pb_wrapper(self) -> TaskMetadataPbWrapper:
         return self._task_metadata_pb_wrapper
 
     @property
-    def flattened_graph_metadata_pb_wrapper(self) -> FlattenedGraphMetadataPbWrapper:
+    def flattened_graph_metadata_pb_wrapper(
+        self,
+    ) -> FlattenedGraphMetadataPbWrapper:
         """
-        Allows access to a flattened_graph_metadata_pb_wrapper
+        Allows access to a flattened_graph_metadata_pb_wrapper.  If the underlying pb is empty, calling this property will instead throw an error.
```
kmontemayor2-sc, line 339: mega nit: I think it may get confusing we use `pb` to refer to `partition book` some times and `protobuf` other times. WDYT about using `proto` here generally?

kmontemayor2-sc (commented):
Thanks Matt!

python/gigl/src/common/types/pb_wrappers/gbml_config.py
kmontemayor2-sc: > it seems object.setattr is used to get around this in the scope of the __post_init__() function call specifically

I see, thanks for checking. For my own knowledge can we use `setattr(self, ..., ...)` instead of the `object.__setattr__`?

python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -46,37 +46,51 @@ class GbmlConfigPbWrapper:
...
+        default=None, init=False
+    )
     _graph_metadata_pb_wrapper: GraphMetadataPbWrapper = field(init=False)
     _preprocessed_metadata_pb_wrapper: PreprocessedMetadataPbWrapper = field(init=False)
-    _flattened_graph_metadata_pb_wrapper: FlattenedGraphMetadataPbWrapper = field(
-        init=False
-    )
+    _flattened_graph_metadata_pb_wrapper: Optional[
+        FlattenedGraphMetadataPbWrapper
+    ] = field(default=None, init=False)
```
kmontemayor2-sc, line 56: wait do we need to do `init = False` here still? not just `...: Optiona[X] = None`?

kmontemayor2-sc (approved):
I'd prefer if we can clean up the `init=False`/etc bits here, please try them out but if they don't work this LGTM

python/gigl/src/common/types/pb_wrappers/gbml_config.py
mkolodner-sc: Yeah we do, otherwise the dataclass will expect to receive these fields explicitly when constructing the dataclass with `__init__` (currently we only provide the `gbml_config_pb` when constructing this). We don't want to have `_flattened_graph_metadata_pb` passed explicitly here for example, and thus need it to be with `field(init=False)`.

python/gigl/src/common/types/pb_wrappers/gbml_config.py
mkolodner-sc: This seems to pass type check this time, but still gives error in unit test: `dataclasses.FrozenInstanceError: cannot assign to field '_task_metadata_pb_wrapper'`

PR #46, opened 2025-05-14
Update dataset_factory.py

xgao4-sc:
### Features
Increase the number of rpc threads for partitioning. Both GLT ([reference](https://github.com/alibaba/graphlearn-for-pytorch/blob/v0.2.5/graphlearn_torch/python/distributed/rpc.py#L242)) and PyTorch ([reference](https://docs.pytorch.org/docs/2.5/rpc.html#torch.distributed.rpc.TensorPipeRpcBackendOptions)) use 16 by default.

### DIFF
--- a/python/gigl/distributed/dataset_factory.py
+++ b/python/gigl/distributed/dataset_factory.py
@@ -253,7 +253,7 @@ def _build_dataset_process(
     init_rpc(
         master_addr=distributed_context.main_worker_ip_address,
         master_port=dataset_building_port,
-        num_rpc_threads=4,
+        num_rpc_threads=16,
     )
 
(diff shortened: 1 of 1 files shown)


mkolodner-sc (approved):
Thanks Xiaobin! LGTM with one question -- from your experience do you think this is worth exposing as a parameter, or do you think 16 should be sufficient for most use cases?

xgao4-sc:
> Thanks Xiaobin! LGTM with one question -- from your experience do you think this is worth exposing as a parameter, or do you think 16 should be sufficient for most use cases?

Good question! I have tested 16 worker threads for up to 60 machines (for training) and it works fine. I guess we could hard code it, similar to what we did for partition chunks.

PR #47, opened 2025-05-14
Add test timeouts

kmontemayor2-sc:
Mostly I want to do this so our unittests timeout quicker than 3hours but might as well address all of them.

I'm being a big aggressive here - the unit tests seem to take ~25 minutes, the integration tests take ~50 and the e2e tests like 110 minutes.

I figure we can bump the times (especially the e2e time) back if we run into issues.


svij-sc:
Maybe we can use the same multiplier for all tests?
For unit tests the proposed changes to 35 minutes is using a 1.4 multiplier
1.4 = 35/25
Thus, integration maybe becomes: 70 
and, e2e is 154

Probably also worth leaving a comment why the tests timeout after these values.

kmontemayor2-sc:
> Thoughts on using the same multiplier for all tests? For unit tests the proposed changes to 35 minutes is using a 1.4 multiplier 1.4 = 35/25 Thus, integration maybe becomes: 70 and, e2e is 154

Done.


> Probably also worth leaving a comment why the tests timeout after these values.

Done.

svij-sc (approved):

.github/workflows/on-pr-comment.yml
```diff
@@ -73,8 +76,9 @@ jobs:
           make run_all_e2e_tests
 
   lint-test:
-    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/lint_test') }} 
+    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/lint_test') }}
     runs-on: ubuntu-latest
+    timeout-minutes: 20 # Not sure how long these take but if it's more than 20 minutes something has gone very wrong...
```
svij-sc, line 81: nit: we probably don't need to write "not sure how long these take"

kmontemayor2-sc: Sure done.

PR #49, opened 2025-05-14
Automatically Infer Ports for Distributed Training/Inference

mkolodner-sc:
**Scope of work done**

- Rather than hard-code ports, we now automatically infer the ports based on the ports that are free on the leader (rank 0) machine. 
- Refactor `connect_worker_pool` to use `init_process_group` to communicate information, such as ip address, across machines, rather than relying on GCS + ping. 
- Set `LOCAL_WORLD_SIZE` as an environment variable when launching GLT inference job
- Add port parameters to `DistributedContext`
- Corresponding tests and usage updates


python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -110,69 +120,49 @@ def connect_worker_pool() -> DistributedContext:
...
+    assert main_worker_fields is not None
+    (
+        main_worker_ip_address,
+        master_partitioning_port,
+        master_worker_ports,
+        master_sampling_ports,
+    ) = main_worker_fields
+
+    # Tears down the process group, since we no longer need it for establishing communication between machines
+    dist.destroy_process_group()
```
kmontemayor2-sc, line 160: BTW do we need to do this?

python/gigl/distributed/dist_context.py
```diff
@@ -1,4 +1,11 @@
 from dataclasses import dataclass
+from typing import List
```
kmontemayor2-sc, line 2: nit. can we just do `list`?

python/gigl/distributed/dist_context.py
```diff
@@ -17,3 +22,49 @@ class DistributedContext:
...
+
+    # Master ports for sampling, where master_sampling_ports[i] indicates the master sampling port for the ith local process rank
+    master_sampling_ports: List[int]
+
+    def __post_init__(self):
+        if len(self.master_worker_ports) != len(self.master_sampling_ports):
+            raise ValueError(
+                f"Number of worker ports: {len(self.master_worker_ports)} and number of sampling ports: {len(self.master_sampling_ports)} on machine {self.global_rank} must be the same."
+            )
+        logger.info(f"Got distributed context: {self} ")
```
kmontemayor2-sc, line 40: Will this condition always be the case?  Is it possible that to have different sized process groups for different tasks (dataset building vs sampling).

python/gigl/distributed/dist_context.py
```diff
@@ -17,3 +22,49 @@ class DistributedContext:
...
+
+    # Total number of training or inference processes per machine
+    @property
+    def local_world_size(self):
+        return len(self.master_worker_ports)
+
+
+def get_free_ports(main_worker_ip_address: str, local_world_size):
+    num_ports_required = local_world_size * 2 + 1
+    free_ports: list[int] = []
```
kmontemayor2-sc, line 62: btw I feel like this should take in `num_ports` as an arguement and we call it multiple times as needed?

Additionally can we have it do the torch distributed stuff inside the function?

python/tests/test_assets/distributed/decorator.py
```diff
@@ -0,0 +1,40 @@
+import functools
+import multiprocessing
+
+
+def run_in_separate_process(func):
```
kmontemayor2-sc, line 5: why do we need this function?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -137,7 +161,6 @@ def test_distributed_neighbor_loader_batched(self):
     # TODO: (svij) - Figure out why this test is failing on Google Cloud Build
     @unittest.skip("Failing on Google Cloud Build - skiping for now")
```
kmontemayor2-sc, line 162: can we re-add this n ow too?

python/gigl/distributed/dist_context.py
```diff
@@ -17,3 +22,49 @@ class DistributedContext:
...
     global_world_size: int
+
+    # Master port for partitioning
+    master_partitioning_port: int
+
+    # Master ports for training or inference, where master_worker_ports[i] indicates the master worker port for the ith local process rank
+    master_worker_ports: List[int]
+
+    # Master ports for sampling, where master_sampling_ports[i] indicates the master sampling port for the ith local process rank
+    master_sampling_ports: List[int]
```
kmontemayor2-sc, line 33: should we do `main` like we have for `main_woker_ip_address`?

python/gigl/common/utils/vertex_ai_context.py
xgao4-sc: I guess if we do not call it maybe the port will still be occupied?

python/gigl/distributed/dist_context.py
```diff
@@ -17,3 +22,49 @@ class DistributedContext:
...
+        return len(self.master_worker_ports)
+
+
+def get_free_ports(main_worker_ip_address: str, local_world_size):
+    num_ports_required = local_world_size * 2 + 1
+    free_ports: list[int] = []
+    while len(free_ports) < num_ports_required:
+        candidate_port = glt.utils.get_free_port(main_worker_ip_address)
+        if candidate_port not in free_ports:
+            free_ports.append(candidate_port)
```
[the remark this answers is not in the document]
mkolodner-sc: It could -- the get_free_port() doesn't actually allocate the port -- my concern is that it may provide a duplicate port as a result.

python/tests/unit/distributed/distributed_neighborloader_test.py
mkolodner-sc: I tried -- this still fails on Gcloud though, so leaving this skipped for now since its not in scope of any of the changes.

python/tests/test_assets/distributed/decorator.py
mkolodner-sc: See the docstirng comment below:

```
"""
    Decorator to run a test function in a separate process. This is needed to ensure that certain operations, such as torch.set_num_interop_threads and init_rpc(),
    don't interfere with each other when calling this function across multiple tests. For example, if we try to call torch.set_num_interop_threads without this
    decorator, we may observe:
    ```
    RuntimeError: Error: cannot set number of interop threads after parallel work has started or set_num_interop_threads called
    ```
    """
```

mkolodner-sc:
Closing since this is addressed elsewhere

PR #50, opened 2025-05-15
Add experimental deps

nshah-sc:
Support addition of experimental deps for experimental use (in this case, hydra).

Did you add automated tests or write a test plan?  Tested environment generated by this is solvable.  New experimental deps are importable.  Unit tests run.  


kmontemayor2-sc (commented):
I have a minor preference for just adding Hydra to `dev` deps, would it be a lot of work to do so instead?

I feel like "dev" is already kind of "experimental" and unless hydra+deps are very large I'm not sure there's harm in doing so?

Though if we have a use for hydra as buisness logic I guess it'd make more sense to do it in the core deps?

kmontemayor2-sc (approved):
Synced offline - this lgtm :)

Makefile
```diff
@@ -120,6 +128,20 @@ generate_dev_linux_cuda_hashed_requirements:
 	--extra torch25-cuda-121 --extra transform --extra dev \
 	./python/pyproject.toml
 
+# Can only be run on linux, otherwise generated hashed req file will be wrong
+generate_experimental_linux_cpu_hashed_requirements:
```
svij-sc, line 132: Since its automated for us now, could we also enable this for MACs?

kmontemayor2-sc:
> Though if we have a use for hydra for application code doesn't it need to be a part of our docker images (and therefor "core" deps)?

BTW thoughts on this? Am I confused here?

Makefile
nshah-sc: done

kmontemayor2-sc:
FYI I'm merging https://github.com/Snapchat/GiGL/pull/51, you may need to rebuild reqs after.

Maybe we need to add a check for if the reqs are correct somehow? cc:@svij

mkolodner-sc (approved):

PR #52, opened 2025-05-16
Disable NeighborLoader tests until ports are dynamically allocated

mkolodner-sc:
**Scope of work done**

We are seeing fairly consistent unit test failures due to the hard-coded ports. While I am working on a solution to enable dynamic port inference, we disable these tests here to enable other PRs to be landed swiftly in the meantime.

### DIFF
--- a/python/tests/unit/distributed/distributed_neighborloader_test.py
+++ b/python/tests/unit/distributed/distributed_neighborloader_test.py
@@ -50,8 +50,6 @@ class DistributedNeighborLoaderTest(unittest.TestCase):
             global_world_size=self._world_size,
         )
 
-    # TODO: (mkolodner-sc) - Re-enable this test once ports are dynamically inferred
-    @unittest.skip("Failing due to ports being already allocated - skiping for now")
     def test_distributed_neighbor_loader(self):
         master_port = glt.utils.get_free_port(self._master_ip_address)
         manager = Manager()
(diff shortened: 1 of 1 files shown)


python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -83,6 +85,8 @@ def test_distributed_neighbor_loader(self):
         # https://paperswithcode.com/dataset/cora
         self.assertEqual(count, 2708)
 
+    # TODO: (mkolodner-sc) - Re-enable this test once ports are dynamically inferred
+    @unittest.skip("Failing due to ports being already allocated - skiping for now")
```
kmontemayor2-sc, line 89: btw do we know if boths tests fail? (minor preference to disable as few tests as possible) and based on my n=2 looking its just this test that fails.

Which also might mean we're not freeing up the port :P

Can we try to only disable this one and if cicd fails we can re-disable the other one?

kmontemayor2-sc (approved):
Please see my comment but other than that LGTM

python/tests/unit/distributed/distributed_neighborloader_test.py
mkolodner-sc: Sounds good, will aim to do that, thanks!

PR #54, opened 2025-05-19
Update

mkolodner-sc:
**Scope of work done**

### DIFF
--- a/python/tests/test_assets/distributed/decorator.py
+++ b/python/tests/test_assets/distributed/decorator.py
@@ -1,0 +1,40 @@
+import functools
+import multiprocessing
+
+
+def run_in_separate_process(func):
+    """
+    Decorator to run a test function in a separate process. This is needed to ensure that certain operations, such as torch.set_num_interop_threads and init_rpc(),
+    don't interfere with each other when calling this function across multiple tests. For example, if we try to call torch.set_num_interop_threads without this
+    decorator, we may observe:
+
+    ```
+    RuntimeError: Error: cannot set number of interop threads after parallel work has started or set_num_interop_threads called
+    ```
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        def target(q: multiprocessing.Queue, *args, **kwargs):
+            try:
+                result = func(*args, **kwargs)
+                q.put((True, result))
+            except Exception as e:
+                q.put((False, e))
+
+        q: multiprocessing.Queue = multiprocessing.Queue()
+        process = multiprocessing.Process(target=target, args=(q, *args), kwargs=kwargs)
+        process.start()
+        process.join()
+
+        if not q.empty():
+            success, result_or_exception = q.get()
+            if not success:
+                raise result_or_exception
+
+        if process.exitcode != 0:
+            raise RuntimeError(
+                f"Test {func.__name__} failed in a separate process with exit code {process.exitcode}"
+            )
+
+    return wrapper
--- a/python/tests/unit/distributed/distributed_neighborloader_test.py
+++ b/python/tests/unit/distributed/distributed_neighborloader_test.py
@@ -29,6 +29,7 @@ from gigl.types.graph import (
     message_passing_to_positive_label,
     to_heterogeneous_node,
 )
+from tests.test_assets.distributed.decorator import run_in_separate_process
 from tests.test_assets.distributed.run_distributed_dataset import (
     run_distributed_dataset,
 )
@@ -50,6 +51,7 @@ class DistributedNeighborLoaderTest(unittest.TestCase):
             global_world_size=self._world_size,
         )
 
+    @run_in_separate_process
     def test_distributed_neighbor_loader(self):
         master_port = glt.utils.get_free_port(self._master_ip_address)
         manager = Manager()
@@ -83,8 +85,7 @@ class DistributedNeighborLoaderTest(unittest.TestCase):
         # https://paperswithcode.com/dataset/cora
         self.assertEqual(count, 2708)
 
-    # TODO: (mkolodner-sc) - Re-enable this test once ports are dynamically inferred
-    @unittest.skip("Failing due to ports being already allocated - skiping for now")
+    @run_in_separate_process
     def test_distributed_neighbor_loader_batched(self):
         node_type = DEFAULT_HOMOGENEOUS_NODE_TYPE
         edge_index = {
@@ -190,8 +191,7 @@ class DistributedNeighborLoaderTest(unittest.TestCase):
             ),
         ]
     )
-    # TODO: (mkolodner-sc) - Re-enable this test once ports are dynamically inferred
-    @unittest.skip("Failing due to ports being already allocated - skiping for now")
+    @run_in_separate_process
     def test_ablp_dataloader(
         self,
         _,


kmontemayor2-sc (commented):
Hey Matt, how much time does this change add to our unit tests?

Additionally, if it turns out even with port picking that we run into conflict then do we really need to resolve the port picking anyways?

I am slightly worried that we aren't cleaning up the ports from the neighborloader (somehow) - and maybe we should be looking into resolving that instead of this as a workaround.

Alternatively, once we actually can pick the ports dynamically, maybe we should just build a new `DistContext` for every test in `distributed_neighborloader_test` so they all use different ports?

mkolodner-sc:
> Hey Matt, how much time does this change add to our unit tests?

Here's a [unit test](https://github.com/Snapchat/GiGL/actions/runs/15123827264) from this PR, it seems that it took 20 minutes 59 seconds and here's a [unit test](https://github.com/Snapchat/GiGL/actions/runs/15026355204/job/42227974753) from before we disabled the above tests, and it seems that took 22 minutes 4 seconds, so we don't need to be concerned about the additional time here since it falls within the natural variation of the unit testing.

> Additionally, if it turns out even with port picking that we run into conflict then do we really need to resolve the port picking anyways?


Yes, generally it is not safe to hard-code ports. I have seen some cases [[1](https://console.cloud.google.com/vertex-ai/locations/us-central1/training/5934657596249079808)] recently in e2e jobs (outside of the unit tests which this PR addresses) where the job fails due to ports already being allocated, so we still should find a way to resolve this. 


> I am slightly worried that we aren't cleaning up the ports from the neighborloader (somehow) - and maybe we should be looking into resolving that instead of this as a workaround.

This is a good point, I believe the root fix in this case would be doing `shutdown_rpc()` at the end of each test in `distributed_neighbor_loader`. However, based on my previous experiments, this will still fail, since we'll get an error in a separate test signaling that we can't do `init_rpc` again after `shutdown_rpc` has been called. As a result, we may still need this decorator. More generally, I don't see an issue with ensuring isolation between multiple tests here, since they don't regress the time for running unit tests. Still, I'll aim to reproduce that behavior with `shutdown_rpc` to double check my observations.

> Alternatively, once we actually can pick the ports dynamically, maybe we should just build a new DistContext for every test in distributed_neighborloader_test so they all use different ports?

Correct, once we dynamically select ports, we should ensure that each test uses different ports if possible. However, using this decorator to ensure isolation extends beyond just the port selection issue. In this [PR](https://github.com/Snapchat/GiGL/pull/49), I observed issues with `torch.set_num_interop_threads` being called multiple times (even with the lru_cache) that were mitigated by using this decorator, which is why I introduced it in the first place.

mkolodner-sc:
@kmontemayor2-sc 

>  Still, I'll aim to reproduce that behavior with shutdown_rpc to double check my observations.

I was able to reproduce this behavior locally by commenting out the decorator and adding `shutdown_rpc()`. From this, I was able to avoid the port allocation concern, but observed a different failure where the `test_partitioning_failure` function was failing, since we call `init_rpc` there and it says that we cannot re-init rpc after shutdown. One possibility could be because `test_partitioning_failure` may be called and trying to init_rpc while there is still some rpc communication shutting down in the neighbor loader test. 

It makes sense to me to add the `shutdown_rpc` in this case to these tests to do our due diligence of cleaning up the ports here and resolving the port allocation issue, but more broadly, my preference then is to provide stronger guarantees of test isolation, and fix the `shutdown_rpc` issue, by also using the decorator in these cases.

kmontemayor2-sc:
> It makes sense to me to add the shutdown_rpc in this case to these tests to do our due diligence of cleaning up the ports here and resolving the port allocation issue, but more broadly, my preference then is to provide stronger guarantees of test isolation, and fix the shutdown_rpc issue, by also using the decorator in these cases. I've updated the PR to reflect this.

Sorry I didn't see then when I was typing my other comment.

Again my concern here is that if we're running into issues with multiple dataloaders in one process than wont our users as well?

mkolodner-sc:
@kmontemayor2-sc 

> Again my concern here is that if we're running into issues with multiple dataloaders in one process than wont our users as well?

Had some discussion offline, we can aim to fix the underlying issue by using `shutdown_rpc()` whenever we call `init_rpc` so that the ports aren't interfering with each other despite being hardcoded. This allows us to re-enable these previously disabled tests.

Similarly, we can aim to keep the decorator since it provides stronger guarantees of test isolation, handles the load testing observation that GLT loader has to run in a spawned process but cannot run in main process, and allows `shutdown_rpc()` to be called without interfering with another tests' `init_rpc`. If we want to evaluate multiple dataloaders in one process, we should have an explicit test for it in the future. That being said, there shouldn't be issues with having each dataloader in one process use different ports, and we only call one `shutdown_rpc()` at the end of that process.

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -83,8 +86,9 @@ def test_distributed_neighbor_loader(self):
         # https://paperswithcode.com/dataset/cora
         self.assertEqual(count, 2708)
 
-    # TODO: (mkolodner-sc) - Re-enable this test once ports are dynamically inferred
-    @unittest.skip("Failing due to ports being already allocated - skiping for now")
+        shutdown_rpc()
```
kmontemayor2-sc, line 89: out of curiosity what would happen if we put this in `tearDown()`?

kmontemayor2-sc (commented):
I do think we should have a test for multiple dataloaders in one process - since I feel like it's an important usecase that was previously "covered" and now isn't.

Can we add that in this PR?

python/tests/unit/distributed/distributed_neighborloader_test.py
mkolodner-sc: tearDown() gets called at the end of this batch of unit tests after all test cases have been run IIUC. However, since each function is in a forked process, which automatically gets cleaned up in that new process, this shutdown_rpc() wouldn't do anything. Additionally, only having one shutdown_rpc() in tearDown and no forked process decorator will not resolve the port conflict issues we were seeing, since the ports won't be cleaned up between each test case.

mkolodner-sc:
> I do think we should have a test for multiple dataloaders in one process - since I feel like it's an important use case that was previously "covered" and now isn't.
Can we add that in this PR?

I don't think this is in scope of this PR, which is just aiming to ensure isolation between tests and make sure we call shutdown_rpc. I've raised a follow-up [PR](https://github.com/Snapchat/GiGL/pull/59) to this one which does this instead.

python/tests/test_assets/distributed/decorator.py
kmontemayor2-sc, line 1: Nit. WDYT about putting this in `python/tests/test_assets/decorators.py` instead to reduce our overall nesting?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -267,6 +274,8 @@ def test_ablp_dataloader(
         assert_tensor_equality(datum.node[srcs], expected_srcs)
         assert_tensor_equality(datum.node[dsts], expected_dsts)
 
+        shutdown_rpc()
```
kmontemayor2-sc, line 277: For my own learning, is calling `shutdown_rpc` here required? Or will the ports be cleaned up when the process ends?

Can we leave comments about it?

kmontemayor2-sc:
> I don't think this is in scope of this PR, which is just aiming to ensure isolation between tests and make sure we call shutdown_rpc. I've raised a follow-up https://github.com/Snapchat/GiGL/pull/59 to this one which does this instead.

Sure, sg, thanks!

kmontemayor2-sc (approved):

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -41,6 +46,146 @@
 _NEGATIVE_EDGE_TYPE = message_passing_to_negative_label(DEFAULT_HOMOGENEOUS_EDGE_TYPE)
 
 
+# GLT requires subclasses of DistNeighborLoader to be run in a separate process. Otherwise, we may run into segmentation fault
+# or other memory issues. Calling these functions in separate proceses also allows us to use shutdown_rpc() to ensure cleanup of
+# ports, providing stronger guarantees of isolation between tests.
+
+
+def _spawn_new_process(func, *args):
+    mp.spawn(fn=func, args=args)
```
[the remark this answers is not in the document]
mkolodner-sc: That's a good point, I'll remove it and just call mp.spawn directly in the unit tests. Thanks!

PR #55, opened 2025-05-19
Update Connect Worker Pool

mkolodner-sc:
**Scope of work done**

- Updates `connect_worker_pool` so that we are using `torch.distributed.init_process_group` to communicate the main_worker_ip_address across the machines, rather than writing the IP address to GCP and using `ping`
- Decrease time for `connect_worker_pool` from 3 minutes to ~5 seconds

Did you add automated tests or write  a test plan?
[VAI Inference Job
](https://console.cloud.google.com/vertex-ai/locations/us-central1/training/848264029599170560)Screenshot 2025-05-19 at 6 04 37 PM

### DIFF
--- a/python/gigl/common/utils/vertex_ai_context.py
+++ b/python/gigl/common/utils/vertex_ai_context.py
@@ -2,13 +2,12 @@
 
 import os
 import subprocess
-import time
+from typing import List, Optional
 
-from gigl.common import GcsUri
+import torch.distributed as dist
+
 from gigl.common.logger import Logger
-from gigl.common.services.vertex_ai import LEADER_WORKER_INTERNAL_IP_FILE_PATH_ENV_KEY
-from gigl.common.utils.gcs import GcsUtils
-from gigl.distributed import DistributedContext
+from gigl.distributed.dist_context import DistributedContext
 
 logger = Logger()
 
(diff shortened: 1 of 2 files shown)


python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -2,13 +2,12 @@
 
 import os
 import subprocess
-import time
+from typing import List, Optional
```
kmontemayor2-sc, line 5: nit. can we just use `list`?

python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -2,13 +2,12 @@
...
+
+import torch.distributed as dist
 
-from gigl.common import GcsUri
 from gigl.common.logger import Logger
-from gigl.common.services.vertex_ai import LEADER_WORKER_INTERNAL_IP_FILE_PATH_ENV_KEY
-from gigl.common.utils.gcs import GcsUtils
-from gigl.distributed import DistributedContext
+from gigl.distributed.dist_context import DistributedContext
```
kmontemayor2-sc, line 10: Any reason for the update, are we running into circular dependency issues?

python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -110,69 +109,27 @@ def connect_worker_pool() -> DistributedContext:
...
-        if attempt_num >= max_retries:
-            logger.info(
-                f"Failed to ping HOST_IP after {max_retries} attempts. Exiting."
-            )
-            raise Exception(f"Failed to ping HOST_IP after {max_retries} attempts.")
+        broadcast_list = [None]
+
+    dist.broadcast_object_list(object_list=broadcast_list, src=0)
+    main_worker_ip_address = broadcast_list[0]
+    assert main_worker_ip_address is not None
```
kmontemayor2-sc, line 127: nit. is it easier if this is just `list[str]` and when we're not the leader we use `broadcast_list = [f'follower-{global_rank}']`?

python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -110,69 +109,27 @@ def connect_worker_pool() -> DistributedContext:
     to get the leader worker's internal IP address and to ensure that the workers
     can all communicate with the leader worker.
     """
-
     global_rank = get_rank()
     global_world_size = get_world_size()
+    # Uses the VAI-set environment variables for `RANK`, `WORLD_SIZE`, `MASTER_IP_ADDRESS`, and `MASTER_PORT` for setting up the process group
+    dist.init_process_group(backend="gloo")
```
kmontemayor2-sc, line 115: is `init_process_group` idempotent? What happens if a user has already instantiated a process group before we reach this code?

WDYT about requiring this function to be called in a valid process group already?

python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -110,69 +109,27 @@ def connect_worker_pool() -> DistributedContext:
...
-            )
-            raise Exception(f"Failed to ping HOST_IP after {max_retries} attempts.")
+        broadcast_list = [None]
+
+    dist.broadcast_object_list(object_list=broadcast_list, src=0)
+    main_worker_ip_address = broadcast_list[0]
+    assert main_worker_ip_address is not None
 
+    # Tears down the process group, since we no longer need it for establishing communication between machines
+    dist.destroy_process_group()
```
kmontemayor2-sc, line 130: ditto on what if there is a user-created process group already.

kmontemayor2-sc (commented):
As mentioned offline, can we do some e2e tests to see if this causes any regressions from the previous implementation?

python/gigl/common/utils/vertex_ai_context.py
xgao4-sc: Could we potentially check whether the group is initialized, if so, just broadcast rank 0's ip address?

python/gigl/common/utils/vertex_ai_context.py
xgao4-sc: Could we potentially check whether the group is initialized, and use a boolean to keep it. If it is created outside, then skip this line?

python/gigl/common/utils/vertex_ai_context.py
xgao4-sc: For example, `torch.distributed.is_initialized()` ([documentation](https://docs.pytorch.org/docs/stable/distributed.html#initialization)), but it only checks whether default group is initialized or not. 

`broadcast_object_list` by default use default group ([documentation](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.broadcast_object_list))

xgao4-sc: `torch.distributed.init_process_group` ([documentation](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.init_process_group)) says it "Initialize the default distributed process group."

I am wondering whether user can create some other group or not, or even if they create it is still the default process group.

xgao4-sc: There seems to be some reading here: https://docs.pytorch.org/docs/stable/distributed.html#groups

python/gigl/common/utils/vertex_ai_context.py
xgao4-sc: In addition, it would be good to understand the mechanism of `destroy_process_group()`. I am reading the doc here: https://docs.pytorch.org/docs/stable/distributed.html#shutdown but look forward to more details.

Basically, after calling this, does it guarantee freeing port? If not, there is a risk of port conflict. For users who do not use `get_free_port` and specify port by themselves, calling `connect_worker_pool` occupy the port they specified for other purposes.

xgao4-sc: Is it possible to test the above? For example, getting all available ports before and after this API, and check whether they are the same?

mkolodner-sc:
Closing, as this has already been addressed in https://github.com/Snapchat/GiGL/pull/109

PR #56, opened 2025-05-19
Add local world size to Distributed Context

mkolodner-sc:
**Scope of work done**

- Adds `local_world_size` to `DistributedContext`. This is a step towards making `DistributedContext` more conformant with the PyG [DistContext](https://pytorch-geometric.readthedocs.io/en/2.5.0/_modules/torch_geometric/distributed/dist_context.html) and simplifies training/inference apis more generally
- Sets `local_world_size` automatically in the `os.environ` of the spawned VAI jobs for GLT inference
- Simplifies `init_neighbor_loader_worker` to be more reliant on `distributed_context`

### DIFF
--- a/python/gigl/common/utils/vertex_ai_context.py
+++ b/python/gigl/common/utils/vertex_ai_context.py
@@ -2,12 +2,13 @@
 
 import os
 import subprocess
-from typing import List, Optional
+import time
 
-import torch.distributed as dist
-
+from gigl.common import GcsUri
 from gigl.common.logger import Logger
-from gigl.distributed.dist_context import DistributedContext
+from gigl.common.services.vertex_ai import LEADER_WORKER_INTERNAL_IP_FILE_PATH_ENV_KEY
+from gigl.common.utils.gcs import GcsUtils
+from gigl.distributed import DistributedContext
 
 logger = Logger()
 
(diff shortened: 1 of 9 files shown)


python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -74,6 +74,17 @@ def get_leader_port() -> int:
...
 
+def get_local_world_size() -> int:
+    """
+    The total number of processes spun up on each VAI Machine. This is currently is manually set upon launching a VAI job manually for GLT processes.
+    We should deprecate this in the future if we migrate VAI jobs to be spun up with torchrun instead.
+    Throws if not on Vertex AI.
+    """
+    if not is_currently_running_in_vertex_ai_job():
+        raise _VAI_EXCEPTION
+    return int(os.environ.get("LOCAL_WORLD_SIZE", 1))
```
nshah171, line 85: @mkolodner-sc you may wanna take a look at the discussion in https://github.com/Snapchat/GiGL/pull/9#discussion_r2068846578 which was the initial reason we removed this there.  

@svij-sc mentioned the introduction of this default especially in a file like `torch_training.py` could be confusing as it might obfuscate a variable that was set by one of several possible launchers like torchrun, torchx, etc.  

However I think this situation is a bit different as we are explicitly setting this on VAI, where this is not run.  @mkolodner-sc i think in general if someone calls this on VAI context they may rather expect None to be returned rather than 1.  maybe the fallback logic of resolving None to 1 should be done in user-layer rather than the choice made on library-layer?

mkolodner-sc:
Closing since we are deprecating distributed context

PR #59, opened 2025-05-21
Add test for multiple dataloaders and update `init_neighbor_loader_worker`

mkolodner-sc:
**Scope of work done**

- Adds unit test for instantiating multiple neighbor loaders at once
- Updates `init_neighbor_loader_worker` to handle exception where `set_num_interop_threads` has already been called to support multiple dataloader use case


python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -88,6 +88,55 @@ def test_distributed_neighbor_loader(self):
...
+            _main_sampling_port=40000,
+        )
+
+        count = 0
+        for datum_one, datum_two in zip(loader_one, loader_two):
+            count += 1
+
+        # Cora has 2708 nodes, make sure we go over all of them.
+        # https://paperswithcode.com/dataset/cora
+        self.assertEqual(count, 2708)
```
kmontemayor2-sc, line 136: btw can we also have a third dataloader run after, sequentially?

python/gigl/distributed/utils/init_neighbor_loader_worker.py
```diff
@@ -111,16 +111,28 @@ def init_neighbor_loader_worker(
...
+        except RuntimeError as e:
+            if (
+                "cannot set number of interop threads after parallel work has started"
+                in str(e)
+            ):
+                logger.info(
+                    "Logical CPU cores has already been set for current process."
+                )
+            else:
+                raise e
```
kmontemayor2-sc, line 134: I am worried that when we `torch.set_num_interop_threads(num_cpu_threads)` and it throws some other torch env stuff gets reset?

Should we have some other cache that checks for this instead? It can just be a `set[int]` for the `local_process_rank`, I think

kmontemayor2-sc (commented):
Thanks for the quick follow up, Matt!

python/gigl/distributed/utils/init_neighbor_loader_worker.py
mkolodner-sc: Makes sense, I've opted to use a process - global boolean `_is_cpu_env_initialized` here for this purpose.

python/gigl/distributed/utils/init_neighbor_loader_worker.py
```diff
@@ -113,16 +117,21 @@ def init_neighbor_loader_worker(
...
+            torch.set_num_interop_threads(num_cpu_threads)
+            torch.set_num_threads(num_cpu_threads)
+
+            # Set the logical cpu cores the current process shoud run on. Note
+            # that the sampling process spawned by the process will inherit
+            # this setting, meaning that sampling process will run on the same group
+            # of logical cores. However, the sampling process is network bound so
+            # it may not heavily compete resouce with model training or inference.
+            p = psutil.Process()
+            p.cpu_affinity(logical_cores)
```
kmontemayor2-sc, line 130: it's only `set_num_interop_threads` that throws right? Any reason we chose to not do any of this?

kmontemayor2-sc (approved):
Thanks Matt!

python/gigl/distributed/utils/init_neighbor_loader_worker.py
mkolodner-sc: My understanding is that while the `set_num_interop_threads` is the only call that throws, more generally these all of these lines are responsible for preparing the setting/env for cpu, and this should only be done once per process, which is why they are all included in this conditional.

PR #60, opened 2025-05-21
Correct sequential numbering in markdown formatter

svij-sc:
**Scope of work done**

Prior to this markdown formatter was incorrectly numbering lists.

### DIFF
--- a/.mdformat.toml
+++ b/.mdformat.toml
@@ -7,3 +7,5 @@
 wrap = 120
 
 extensions = [ "tables", ]
+
+number = true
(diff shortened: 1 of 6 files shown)


kmontemayor2-sc (commented):
You know, I kind of like it the way it was before - that way when we need to add a point in the middle/beginning of the list you don't need to update the subsequent ones.

Or will `mdformat` do that for us in this case?

kmontemayor2-sc:
I don't think that's a bug - it means we (the users) don't need to worry about re-number if we go from e.g.

```
1. itema
1. itemb
1. itemc

to 

1. itema
1. itemd
1. itemb
1. itemc
```

I personally like this behavior.

Did `mdformat` re-enumerate all of these lists correct? In which case having the formatter take care of numbering is fine imo.

kmontemayor2-sc (approved):
Synced offline, approval

yliu2-sc (approved):

PR #69, opened 2025-06-02
ignore pickle files

kmontemayor2-sc:
Ignore are pickled files from git. We should never put these into git as a best practice reason (pickled objects are generally unsafe to execute, are very sensitive to python versioning, and are large binaries we don't want in our repo).

However, I do occasionally pickle stuff locally to save on time loading datasets, see https://github.com/Snapchat/GiGL/blob/20c81a0168e9700dc44684ac0509442526478805/examples/distributed/homogeneous_training.py#L357-L370 and would like to avoid needing to make sure I don't accidentally commit them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ graphlearn_torch.egg-info/
 fossa
 fossa.bundle
 fossa*.zip
+
+# Ignore *all* pickled files, we should *never* put these in git.
+*.pkl


.gitignore
```diff
@@ -42,3 +42,6 @@ graphlearn_torch.egg-info/
 fossa
 fossa.bundle
 fossa*.zip
+
+# Ignore *all* pickled files, we should *never* put these in git.
+*.pkl
```
svij-sc, line 47: Thanks!
Probably we should add this to `.Dockerignore` as well.
Raised a PR: https://github.com/Snapchat/GiGL/pull/70

PR #71, opened 2025-06-02
Update GiGL Splitter

mkolodner-sc:
**Scope of work done**

Updated the split API and functionality in GiGL.
- Moves `should_convert_labels_to_edge` to the splitter class and make `supervision_edge_types` param more explicit
- Update SSL building logic to infer edge types based on the supervision edges, rather than all edges
- Update `treat_labels_as_edges` to correctly handle labeled edges in both "in" and "out" edge directions
- Update corresponding tests
 

### DIFF
--- a/python/gigl/distributed/dataset_factory.py
+++ b/python/gigl/distributed/dataset_factory.py
@@ -133,10 +133,7 @@ def _load_and_build_partitioned_dataset(
 
         loaded_graph_tensors.positive_label = positive_label_edges
 
-    if (
-        isinstance(splitter, HashedNodeAnchorLinkSplitter)
-        and splitter.should_convert_labels_to_edges
-    ):
+    if splitter is not None and splitter.should_convert_labels_to_edges:
         loaded_graph_tensors.treat_labels_as_edges(edge_dir=edge_dir)
 
     should_assign_edges_by_src_node: bool = False if edge_dir == "in" else True
@@ -146,11 +143,11 @@ def _load_and_build_partitioned_dataset(
 
     if should_assign_edges_by_src_node:
         logger.info(
-            f"Initializing {type(partitioner_class)} instance while partitioning edges to its source node machine"
+            f"Initializing {partitioner_class.__name__} instance while partitioning edges to its source node machine"
         )
     else:
         logger.info(
-            f"Initializing {type(partitioner_class)} instance while partitioning edges to its destination node machine"
+            f"Initializing {partitioner_class.__name__} instance while partitioning edges to its destination node machine"
         )
     partitioner = partitioner_class(
         should_assign_edges_by_src_node=should_assign_edges_by_src_node
--- a/python/gigl/distributed/dataset_factory.py
+++ b/python/gigl/distributed/dataset_factory.py
@@ -135,7 +135,7 @@ def _load_and_build_partitioned_dataset(
 
     if (
         isinstance(splitter, HashedNodeAnchorLinkSplitter)
-        and splitter._should_convert_labels_to_edges
+        and splitter.should_convert_labels_to_edges
     ):
         loaded_graph_tensors.treat_labels_as_edges(edge_dir=edge_dir)
 
--- a/python/gigl/distributed/dataset_factory.py
+++ b/python/gigl/distributed/dataset_factory.py
@@ -38,12 +38,6 @@ from gigl.distributed.utils.serialized_graph_metadata_translator import (
 )
 from gigl.src.common.types.graph_data import EdgeType
 from gigl.src.common.types.pb_wrappers.gbml_config import GbmlConfigPbWrapper
-from gigl.types.graph import (
-    DEFAULT_HOMOGENEOUS_EDGE_TYPE,
-    GraphPartitionData,
-    message_passing_to_negative_label,
-    message_passing_to_positive_label,
-)
 from gigl.utils.data_splitters import (
     HashedNodeAnchorLinkSplitter,
     NodeAnchorLinkSplitter,
@@ -61,7 +55,6 @@ def _load_and_build_partitioned_dataset(
     partitioner_class: Optional[Type[DistPartitioner]],
     node_tf_dataset_options: TFDatasetOptions,
     edge_tf_dataset_options: TFDatasetOptions,
-    should_convert_labels_to_edges: bool = False,
     splitter: Optional[NodeAnchorLinkSplitter] = None,
     _ssl_positive_label_percentage: Optional[float] = None,
 ) -> DistLinkPredictionDataset:
@@ -77,7 +70,6 @@ def _load_and_build_partitioned_dataset(
             DistPartitioner or subclass of it. If not provided, will initialize use the DistPartitioner class.
         node_tf_dataset_options (TFDatasetOptions): Options provided to a tf.data.Dataset to tune how serialized node data is read.
         edge_tf_dataset_options (TFDatasetOptions): Options provided to a tf.data.Dataset to tune how serialized edge data is read.
-        should_convert_labels_to_edges (bool): Whether to convert labels to edges in the graph. If this is set to true, the output dataset will be heterogeneous.
         splitter (Optional[NodeAnchorLinkSplitter]): Optional splitter to use for splitting the graph data into train, val, and test sets. If not provided (None), no splitting will be performed.
         _ssl_positive_label_percentage (Optional[float]): Percentage of edges to select as self-supervised labels. Must be None if supervised edge labels are provided in advance.
             Slotted for refactor once this functionality is available in the transductive `splitter` directly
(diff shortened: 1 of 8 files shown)


python/gigl/distributed/dataset_factory.py
```diff
@@ -104,8 +96,46 @@ def _load_and_build_partitioned_dataset(
...
     )
-    if should_convert_labels_to_edges:
-        loaded_graph_tensors.treat_labels_as_edges()
+
+    # TODO (mkolodner-sc): Move this code block (from here up to start of partitioning) to transductive splitter once that is ready
+    if _ssl_positive_label_percentage is not None:
+        assert (
+            loaded_graph_tensors.positive_label is None
+            and loaded_graph_tensors.negative_label is None
+        ), "Cannot have loaded positive and negative labels when attempting to select self-supervised positive edges from edge index."
```
kmontemayor2-sc, line 105: nit. this should probably be `raise ValueError` to be consistent?

python/gigl/distributed/dataset_factory.py
```diff
@@ -104,8 +96,46 @@ def _load_and_build_partitioned_dataset(
...
+        positive_label_edges: Union[torch.Tensor, Dict[EdgeType, torch.Tensor]]
+        if isinstance(loaded_graph_tensors.edge_index, abc.Mapping):
+            # This assert is required while `select_ssl_positive_label_edges` exists out of any splitter. Once this is in transductive splitter,
+            # we can remove this assert.
+            assert isinstance(
+                splitter, HashedNodeAnchorLinkSplitter
+            ), f"GiGL only supports {HashedNodeAnchorLinkSplitter.__name__} currently, got {type(splitter)}"
+            positive_label_edges = {}
+            for supervision_edge_type in splitter._supervision_edge_types:
+                assert supervision_edge_type in loaded_graph_tensors.edge_index
```
kmontemayor2-sc, line 115: why do we need this assert? the below code will throw anyways right?

python/gigl/distributed/dataset_factory.py
```diff
@@ -104,8 +96,46 @@ def _load_and_build_partitioned_dataset(
...
+                    positive_label_percentage=_ssl_positive_label_percentage,
+                )
+        elif isinstance(loaded_graph_tensors.edge_index, torch.Tensor):
+            positive_label_edges = select_ssl_positive_label_edges(
+                edge_index=loaded_graph_tensors.edge_index,
+                positive_label_percentage=_ssl_positive_label_percentage,
+            )
+        else:
+            raise ValueError(
+                "Found no partitioned edge index when attempting to select positive labels"
```
kmontemayor2-sc, line 129: nit. it's more like we don't know what to do with this type of `edge_index` right? Maybe we should say so and add the type?

python/gigl/distributed/dataset_factory.py
```diff
@@ -328,27 +323,8 @@ def build_dataset(
...
-                edge_types=[
-                    message_passing_to_positive_label(DEFAULT_HOMOGENEOUS_EDGE_TYPE),
-                    message_passing_to_negative_label(DEFAULT_HOMOGENEOUS_EDGE_TYPE),
-                ],
-            )
-        logger.info(
-            "Will be treating the ABLP labels as heterogeneous edges in the graph."
-        )
+    if splitter is not None:
+        logger.info(f"Received splitter {str(splitter.__class__)}.")
```
kmontemayor2-sc, line 327: nit. why not?
```suggestion
        logger.info(f"Received splitter {type(splitter)}.")
```

python/gigl/types/graph.py
```diff
@@ -107,13 +108,17 @@ class LoadedGraphTensors:
     # Unpartitioned Negative Edge Label
     negative_label: Optional[Union[torch.Tensor, dict[EdgeType, torch.Tensor]]]
 
-    def treat_labels_as_edges(self) -> None:
+    def treat_labels_as_edges(self, edge_dir: Literal["in", "out"]) -> None:
         """
         Convert positive and negative labels to edges. Converts this object in-place to a "heterogeneous" representation.
+        If the edge direction is "in", we must reverse the supervision edge type.
```
kmontemayor2-sc, line 114: let's say why we need to do this?

Comments should say why the code is a certain way, what the code does should be interpretable from reading the code.

python/gigl/types/graph.py
```diff
@@ -148,21 +160,37 @@ def treat_labels_as_edges(self) -> None:
...
+                    positive_label_edge_type = reverse_edge_type(
+                        positive_label_edge_type
+                    )
+                    edge_index_with_labels[
+                        positive_label_edge_type
+                    ] = positive_label_tensor.flip(0)
+                else:
+                    edge_index_with_labels[
+                        positive_label_edge_type
+                    ] = positive_label_tensor
```
kmontemayor2-sc, line 173: nit. this code block is kind of duplicated a lot, wdyt about breaking it out?

kmontemayor2-sc: maybe some helper function with a signature like `def _get_label_edges(edge_dir, edge_type edges: Tensor) -> tuple[EdgeType, Tensor]` or would that not really help too much?

python/gigl/types/graph.py
```diff
@@ -232,6 +271,23 @@ def message_passing_to_negative_label(
         return edge_type
 
 
+# TODO: add tests for this.
```
kmontemayor2-sc, line 274: can we address this TODO?

python/gigl/utils/data_splitters.py
```diff
@@ -115,7 +126,8 @@ def __init__(
         num_val: Union[float, int] = 0.1,
         num_test: Union[float, int] = 0.1,
         hash_function: Callable[[torch.Tensor], torch.Tensor] = _fast_hash,
-        edge_types: Optional[Union[EdgeType, Sequence[EdgeType]]] = None,
+        supervision_edge_types: Optional[list[EdgeType]] = None,
```
kmontemayor2-sc, line 129: btw this could be `supervision_edge_types: Sequence[EdgeType] =(DEFAULT_HOMOGENEOUS_EDGE_TYPE,)`,

Not sure if you think that's cleaner.

python/gigl/utils/data_splitters.py
```diff
@@ -167,26 +180,10 @@ def __call__(
         Mapping[NodeType, Tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
     ]:
         if isinstance(edge_index, torch.Tensor):
-            if self._supervision_edge_types != [DEFAULT_HOMOGENEOUS_EDGE_TYPE]:
-                logger.warning(
-                    f"You provided edge-types {self._supervision_edge_types} but the edge index is homogeneous. Ignoring edge types."
-                )
```
kmontemayor2-sc, line 173: why did we remove this?

python/gigl/utils/data_splitters.py
```diff
@@ -136,12 +149,12 @@ def __init__(
         self._num_val = num_val
         self._num_test = num_test
         self._hash_function = hash_function
+        self._should_convert_labels_to_edges = should_convert_labels_to_edges
```
kmontemayor2-sc, line 152: btw why can't we assume this is always `True`?

kmontemayor2-sc (commented):
Thanks Matt!

python/gigl/utils/data_splitters.py
```diff
@@ -126,8 +143,9 @@ def __init__(
             num_test (Union[float, int]): The percentage of nodes to use for validation. Defaults to 0.1 (10%).
                                           If an integer is provided, than exactly that number of nodes will be in the test split.
             hash_function (Callable[[torch.Tensor, torch.Tensor], torch.Tensor]): The hash function to use. Defaults to `_fast_hash`.
-            edge_types: The supervision edge types we should use for splitting.
-                        Must be provided if we are splitting a heterogeneous graph.
+            supervision_edge_types (Optional[list[EdgeType]]): The supervision edge types we should use for splitting. Must be provided if we are splitting a heterogeneous graph.
+                If None, uses the default homogeneous edge type
```
kmontemayor2-sc, line 147:
```suggestion
                If None, uses the message passing edge type in the graph. 
```

kmontemayor2-sc: is this more descript?

python/gigl/utils/data_splitters.py
```diff
@@ -126,8 +143,9 @@ def __init__(
...
+            should_convert_labels_to_edges (bool): Whether label should be converted into an edge type in the graph.
```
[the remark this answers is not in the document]
kmontemayor2-sc: I think we should document the interplay between this and `supervision_edge_types`

python/gigl/distributed/dataset_factory.py
```diff
@@ -408,17 +385,27 @@ def build_dataset_from_task_config_uri(
...
+            gbml_config_pb_wrapper.task_metadata_pb_wrapper.get_supervision_edge_types()
+        )
+        sample_edge_direction = args.get("sample_edge_direction", "in")
+        args_path = "trainerConfig.trainerArgs"
+        # TODO(kmonte): Maybe we should enable `should_convert_labels_to_edges` as a flag?
+        splitter = HashedNodeAnchorLinkSplitter(
+            sampling_direction=sample_edge_direction,
+            supervision_edge_types=supervision_edge_types,
+            should_convert_labels_to_edges=True,
+        )
```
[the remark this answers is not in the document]
mkolodner-sc: Yeah I moved this out of `build_dataset` for now since all the details with initializing some default splitter felt a bit unclean to me, since that would also require us to accept `supervision_edge_types` and `should_convert_labels_to_edges` to `build_dataset`, and we run into the coupling again between these parameters again. I imagine it isn't too much work to ask customers to initialize the splitter themselves if they want to use `build_dataset` directly.

PR #73, opened 2025-06-02
Update ABLP Loader

mkolodner-sc:
**Scope of work done**

- Moved ABLP Dataloader to its own file and updates logic to use the subclassed GLT classes from https://github.com/Snapchat/GiGL/pull/72
- Update __init__ function and transforms based on updated NeighborSampling logic 
- Add DBLP unit test


python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -421,13 +466,17 @@ def test_cora_supervised(self):
             graph_metadata_pb_wrapper=gbml_config_pb_wrapper.graph_metadata_pb_wrapper,
             tfrecord_uri_pattern=".*.tfrecord(.gz)?$",
         )
-        expected_data_count = 2161
```
mkolodner-sc, line 424: When testing, I was observing 2168 items in the CORA dataset. When printing the number of nodes in `dataset.train_node_ids`, this also yields `2168`. As such, I am updating the correct count here to be 2168.

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,521 @@
...
+            with_edge=True,
+            collect_features=True,
+            with_neg=False,
+            with_weight=False,
+            edge_dir=dataset.edge_dir,
+            seed=None,  # it's actually optional - None means random.
+        )
+
+        # Code below this point is taken from the GLT DistNeighborLoader.__init__() function. We do this so that we may override
+        # the DistSamplingProducer that is used with the GiGL implementation.
```
kmontemayor2-sc, line 316: let's link to the file directly?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,521 @@
...
+        self.worker_options = worker_options
+
+        # We can set shutdowned to false now
+        self._shutdowned = False
+
+        self._is_mp_worker = True
+        self._is_collocated_worker = False
+        self._is_remote_worker = False
+
+        if self.data is not None:
```
kmontemayor2-sc, line 339: it should never be `None` here right?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,521 @@
...
+        )
+        self._mp_producer.init()
+
+    def _get_labels(
+        self, msg: SampleMessage
+    ) -> Tuple[
+        SampleMessage, dict[int, torch.Tensor], Optional[dict[int, torch.Tensor]]
+    ]:
+        """
+        Gets the labels from the output SampleMessage and removes them from the metadata
```
kmontemayor2-sc, line 395: Let's comment why we need to do this?

kmontemayor2-sc: IIRC there's some GLT thing here right? Let's also add a TODO that we can remove this later in that case?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,521 @@
...
+        for i in range(positive_labels.shape[0]):
+            positive_example = positive_labels[i]
+            output_positive_labels[i] = positive_example[
+                positive_example != PADDING_NODE
+            ]
+            if negative_labels is not None:
+                negative_example = negative_labels[i]
+                output_negative_labels[i] = negative_example[
+                    negative_example != PADDING_NODE
+                ]
```
kmontemayor2-sc, line 427: Can we vectorize this?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,521 @@
...
+                    for positive_label in positive_labels[local_anchor_node_id]
+                ]
+            ).to(self.to_device)
+            if negative_labels is not None:
+                negative_labels[local_anchor_node_id] = torch.tensor(
+                    [
+                        global_node_to_local_supervision_node_map[negative_label.item()]
+                        for negative_label in negative_labels[local_anchor_node_id]
+                    ]
+                ).to(self.to_device)
```
kmontemayor2-sc, line 502: can we vectorize this?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,521 @@
...
+
+    def _set_labels(
+        self,
+        data: Data,
+        positive_labels: dict[int, torch.Tensor],
+        negative_labels: Optional[dict[int, torch.Tensor]],
+    ) -> Data:
+        """
+        Sets the labels and relevant fields in the torch_geometric Data object, converting the global node ids for labels to their
+        local index. Removes inserted supervision edge type from the Data variables.
```
kmontemayor2-sc, line 456: can we add notes for why we delete/etc?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -242,513 +235,24 @@ def __init__(
...
-    def _get_node_sampler_input(
-        self, node: torch.Tensor, input_type: Optional[Union[NodeType, str]]
-    ) -> NodeSamplerInput:
-        return _LabeledNodeSamplerInput(node, input_type, self._batch_store)
-
-
-def _shard_nodes_by_process(
-    input_nodes: Union[torch.Tensor, Tuple[str, torch.Tensor]],
+def shard_nodes_by_process(
+    input_nodes: torch.Tensor,
```
kmontemayor2-sc, line 246: nit. this should probably go in it's own util function now, but I guess we can wait until we have `utils.py` here...

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -138,18 +137,33 @@ def _run_distributed_ablp_neighbor_loader(
         expected_node,
         dim=0,
     )
-    assert datum.y_positive.keys() == expected_positive_labels.keys()
+    global_node_to_local_node_map = datum.global_node_to_local_node_map
     for anchor in expected_positive_labels.keys():
+        assert anchor in global_node_to_local_node_map
```
kmontemayor2-sc, line 142: why do we need this assert?

python/tests/unit/distributed/distributed_neighborloader_test.py
kmontemayor2-sc: weird that it changed - I think cora actually has like 2708 or something, and this should be 80% of that...

Was the test not passing/not running before?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,506 @@
...
+                local_process_rank=local_process_rank
+            )
+        )
+
+        # TODO(kmonte): stop setting fanout for positive/negative once GLT sampling is fixed.
+        if isinstance(num_neighbors, dict):
+            num_hop = len(list(num_neighbors.values())[0])
+        else:
+            num_hop = len(num_neighbors)
+        zero_samples = [0] * num_hop
```
kmontemayor2-sc, line 218: I know I wrote this but I think it should actually be `[0 for _ in range(num_hop)]`.

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,506 @@
...
+                metadata[meta_key] = msg[k].to(self.to_device)
+                del msg[k]
+
+        positive_labels = metadata["positive_labels"]
+        negative_labels = (
+            metadata["negative_labels"] if "negative_labels" in metadata else None
+        )
+        return (msg, positive_labels, negative_labels)
+
+    def _supervised_to_homogeneous(self, data: Data) -> Data:
```
kmontemayor2-sc, line 419:
```suggestion
    def _supervised_to_homogeneous(self, data: HeteroData) -> Data:
```

nit.

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,506 @@
...
+            data (Data): Homogeneous graph with the labeled edge type removed
+        """
+        homogeneous_data = data.edge_type_subgraph(
+            [self._supervision_edge_type]
+        ).to_homogeneous(add_edge_type=False, add_node_type=False)
+        return homogeneous_data
+
+    def _set_labels(
+        self,
+        data: Union[Data, HeteroData],
```
kmontemayor2-sc, line 435: sorry I mean we use a TypeVar here to indicate that the returned type is the same as the input but for now this is fine.

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -0,0 +1,506 @@
...
+        for local_anchor_node_id in range(positive_labels.size(0)):
+            positive_mask = (
+                local_node_to_global_node.unsqueeze(1)
+                == positive_labels[local_anchor_node_id]
+            )  # shape [N, P], where N is the number of nodes and P is the number of positive labels for the current anchor node
+
+            # Gets the indexes of the items in local_node_to_global_node which match any of the positive labels for the current anchor node
+            output_positive_labels[local_anchor_node_id] = torch.nonzero(positive_mask)[
+                :, 0
+            ].to(self.to_device)
```
kmontemayor2-sc, line 475: add shape here?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -248,6 +253,103 @@ def _run_multiple_neighbor_loader(
...
+        context=context,
+        local_process_rank=0,
+        local_process_world_size=1,
+        supervision_edge_type=supervision_edge_type,
+    )
+    count = 0
+    for datum in loader:
+        assert isinstance(datum, HeteroData)
+        assert hasattr(datum, "y_positive")
+        assert isinstance(datum.y_positive, dict)
```
kmontemayor2-sc, line 283: this is SSL right? So no negative labels? 


Can we make other assertions about what's in `y_positive`, such as:

1. the keys are all part of `batch`.
2. the values are all part of  `node`
3. that there is no `y_negative`

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -248,6 +253,103 @@ def _run_multiple_neighbor_loader(
...
+        count += 1
+    assert count == dataset.train_node_ids[NodeType("paper")].size(0)
```
kmontemayor2-sc, line 285: btw I feel like we should be using `self.assertEqual` and friends as much as possible here.

We use `assert isinstance` for type narrowing reasons and imo that shoul dbe the only reason we do so.

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -248,6 +253,103 @@ def _run_multiple_neighbor_loader(
...
+    for local_anchor_node, local_positive_supervision_nodes in datum.y_positive.items():
+        global_anchor_node = datum[anchor_node_type].node[local_anchor_node]
+        global_positive_supervision_nodes = datum[supervision_node_type].node[
+            local_positive_supervision_nodes
+        ]
+        # Check that the current anchor node from y_positive is found in the batch
+        assert global_anchor_node in datum[anchor_node_type].batch
+        # Check that the current anchor node from y_positive is found in the expected anchor tensor
+        assert global_anchor_node.item() in expected_anchor_nodes
+        # Check that all positive supervision nodes from y_positive are found in the expected positive supervision tensor
```
kmontemayor2-sc, line 343: this doesn't check that all expected nodes are found right, like we could expect `8` but if it doesn't exist in data then we wouldn't fail?

python/tests/unit/distributed/distributed_neighborloader_test.py
mkolodner-sc: > the keys are all part of batch.
> the values are all part of node

The y_positive contains a local mapping now of `local_anchor_node_id: [local_positive_node_id]`, where local indicates that it provides the index in the data[node_type].x tensor to extract the embedding, rather than the original global node id. I can add checks that these local ids are in range of the `batch` and `node` tensor respectively though.

python/tests/unit/distributed/distributed_neighborloader_test.py
mkolodner-sc: I'll add this condition for `assert global_anchor_node in datum[anchor_node_type].batch`, for the subsequent line `assert global_anchor_node.item() in expected_anchor_nodes`, we cannot make this check since the `expected_anchor_nodes` contains all anchor nodes, including ones that may be included in the validation split. I'll update the naming to be `all_anchor_nodes` and `all_positive_supervision_nodes` to more accurately reflect this

PR #78, opened 2025-06-04
Have splitter return intersection, not union, of shared node ids

kmontemayor2-sc:
Something that came up is if we have:

```
positive_labels = [(1, 2), (1, 3), (2, 3)]
negative_labels = [(2, 5), (3, 1)]
```

Then the splitter would return the node ids `(1, 2, 3)` which means we will run into issues when we try to use `3` with positive labels and `1` with negative labels.

As such, let's change this to return the intersection of node ids for a given type.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,3 +5,7 @@
 .github/ @svij-sc @kmontemayor2-sc @nshah-sc
 requirements @svij-sc @kmontemayor2-sc @nshah-sc
 python/pyproject.toml @svij-sc @kmontemayor2-sc @nshah-sc
+
+# run_command_on_active_checkout.yaml gets updated when we build new images
+# As all devs may setup new images, we allow all of us to review it
+.github/cloud_builder/run_command_on_active_checkout.yaml @svij-sc @kmontemayor2-sc @nshah-sc @yliu2-sc @mkolodner-sc


python/gigl/utils/data_splitters.py
```diff
@@ -271,9 +290,27 @@ def __call__(
...
+
+            # We want to keep track of the node ids, by edge type.
+            # That way we can enforce that for a given anchor node type,
+            # the splits will contain node node ids that are present in all edge types.
+            # e.g:
+            # positive_labels = [(1, 2), (1, 3), (2, 3)]
+            # negative_labels = [(2, 5), (3, 1)]
+            # We should only use `2` as a node id.
+            node_count_by_edge: dict[EdgeType, torch.Tensor] = defaultdict(
+                lambda: torch.zeros(max_node_id, dtype=torch.int64)
```
mkolodner-sc, line 302: Does this need to be of type torch.int64 if we are doing counts? Can we use torch.int32 here instead, or potentially even less, like torch.uint8? Ditto elsewhere.

Since we just end up doing torch.nonzero() on this after, I wonder if we really need to keep track of the magnitude of the counts for each node, or if we can keep track of the counted nodes some other binary way that would only need `torch.uint8`, and thus be memory-wise cheaper.

python/gigl/utils/data_splitters.py
```diff
@@ -271,9 +290,27 @@ def __call__(
...
+            )
+            for anchor_nodes, edge_type in collected_anchor_nodes:
+                node_count_by_edge[edge_type].add_(
+                    torch.bincount(anchor_nodes, minlength=max_node_id)
+                )
+
+            node_id_count = torch.ones(max_node_id, dtype=torch.int64)
+            # Do the intersection calculation.
+            for anchor_counts in node_count_by_edge.values():
+                node_id_count = (node_id_count != 0) & (anchor_counts != 0)
```
mkolodner-sc, line 312: Can we also del and gc.collect() the other created tensors in `node_count_by_edge ` here?

mkolodner-sc (commented):
Thanks Kyle!

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -373,11 +373,11 @@ def test_node_based_link_splitter(
                     EdgeType(_NODE_C, _TO, _NODE_A),
                 ],
                 hash_function=lambda x: x,
-                val_num=0.1,
-                test_num=0.1,
+                val_num=1,
+                test_num=1,
```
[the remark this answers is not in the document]
kmontemayor2-sc: yes it means 1 in both.

python/gigl/utils/data_splitters.py
kmontemayor2-sc: good callout! thanks :)

PR #81, opened 2025-06-05
Allow for non-existant anchor node ids when creating labels

kmontemayor2-sc:
Instead of https://github.com/Snapchat/GiGL/pull/78 let's go with this approach.

Now, if there isa. positive node id with no corresponding hard negative labels, the nodes negative labels will be padded with our padding node, which is later stripped out.

### DIFF
--- a/python/gigl/utils/data_splitters.py
+++ b/python/gigl/utils/data_splitters.py
@@ -436,9 +436,13 @@ def _get_padded_labels(
         mask, torch.full_like(ranges, PADDING_NODE.item()), indices[ranges]
     )
     labels = torch.cat(
-        [labels, torch.ones(extra_nodes_to_pad, max_range) * PADDING_NODE], dim=0
+        [
+            labels,
+            torch.ones(extra_nodes_to_pad, max_range, dtype=torch.int64) * PADDING_NODE,
+        ],
+        dim=0,
     )
-    return labels.to(dtype=torch.int64)
+    return labels
 
 
 def _check_sampling_direction(sampling_direction: str):
--- a/python/gigl/utils/data_splitters.py
+++ b/python/gigl/utils/data_splitters.py
@@ -271,7 +271,7 @@ class HashedNodeAnchorLinkSplitter:
         splits: dict[NodeType, Tuple[torch.Tensor, torch.Tensor, torch.Tensor]] = {}
         for anchor_node_type, collected_anchor_nodes in node_ids_by_node_type.items():
             max_node_id = max_node_id_by_type[anchor_node_type]
-            node_id_count = torch.zeros(max_node_id, dtype=torch.int64)
+            node_id_count = torch.zeros(max_node_id, dtype=torch.uint8)
             for anchor_nodes in collected_anchor_nodes:
                 node_id_count.add_(torch.bincount(anchor_nodes, minlength=max_node_id))
             # This line takes us from a count of all node ids, e.g. `[0, 2, 0, 1]`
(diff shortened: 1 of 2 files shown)


python/gigl/utils/data_splitters.py
```diff
@@ -418,7 +435,10 @@ def _get_padded_labels(anchor_node_ids: torch.Tensor, topo: Topology) -> torch.T
     labels = torch.where(
         mask, torch.full_like(ranges, PADDING_NODE.item()), indices[ranges]
     )
-    return labels
+    labels = torch.cat(
+        [labels, torch.ones(extra_nodes_to_pad, max_range) * PADDING_NODE], dim=0
+    )
+    return labels.to(dtype=torch.int64)
```
mkolodner-sc, line 441: why do we need `.to(dtype=torch.int64)` here? What is the dtype prior to this being called?

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -614,16 +616,20 @@ def test_get_labels_for_anchor_nodes_heterogeneous_positive_and_negative(self):
...
             dataset=ds,
             node_ids=node_ids,
             positive_label_edge_type=a_to_b,
             negative_label_edge_type=a_to_c,
         )
 
-        expected_positive = torch.tensor([[10, 11], [12, 13]])
-        expected_negative = torch.tensor([[20, -1], [30, 50]])
+        expected_positive = torch.tensor(
+            [[10, 11], [12, 13], [10, -1]], dtype=torch.int64
```
mkolodner-sc, line 628: Would this be a [3x2] tensor, instead of the [2 x 3] tensor we should be expecting for edges?

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -614,16 +616,20 @@ def test_get_labels_for_anchor_nodes_heterogeneous_positive_and_negative(self):
...
+        )
+        expected_negative = torch.tensor(
```
mkolodner-sc, line 630: Can you provide some context for how can we have these anchor nodes for our negative edges if our node_ids is [10, 11, 13]?

python/tests/unit/utils/data_splitters_test.py
kmontemayor2-sc: no? we expect to return `[num anchor nodes x num labels]` and we have three anchor nodes in this case

python/tests/unit/utils/data_splitters_test.py
kmontemayor2-sc: these are the other side of the edges, so we have `10 -> {10, 11}` as a positive label, and then our first set of labels (for node `10`) is `[10, 11]`.

python/gigl/utils/data_splitters.py
kmontemayor2-sc: it's sometimes float32... though maybe it's just testing?

The """problem""" here is the cat may convert everything to float if one part of it is float. Fixed now.

```
>>> torch.cat([torch.tensor([1, 2], dtype=torch.float32), torch.tensor([3], dtype=torch.int64)])
tensor([1., 2., 3.])
>>> torch.cat([torch.tensor([1, 2], dtype=torch.int64), torch.tensor([3], dtype=torch.float32)])
tensor([1., 2., 3.])
>>> torch.cat([torch.tensor([1, 2], dtype=torch.int64), torch.tensor([])])
tensor([1., 2.])
>>> torch.cat([torch.tensor([1, 2], dtype=torch.int64), torch.tensor([])])
tensor([1., 2.])
```

python/tests/unit/utils/data_splitters_test.py
mkolodner-sc: Awesome, thanks!

mkolodner-sc (approved):
Thanks Kyle! This LGTM provided tests pass :)

PR #84, opened 2025-06-09
Skip edge_partition_book and edge_id generation for edge types with no features

mkolodner-sc:
**Scope of work done**

- Remove generation of edge partition book for edge type that have no features
- Remove use of edge_ids for edge types that have no features
- store the inferred edge_ids for tensor-based partitioning as `tuple[start_ind, end_ind]` which only get converted to the torch.arange(start_ind, end_ind) at the start of edge_feature partitioning. This way, we don't unnecessarily generate edge ids for edge types which don't need them.
- Increase range-based partitioning test coverage to include range-based partitioning without features
- Simplified return types of `partition_edge_index_and_features`

### DIFF
--- a/python/gigl/distributed/dist_link_prediction_dataset.py
+++ b/python/gigl/distributed/dist_link_prediction_dataset.py
@@ -479,11 +479,12 @@ class DistLinkPredictionDataset(DistDataset):
             isinstance(partition_output.partitioned_edge_features, abc.Mapping)
             and len(partition_output.partitioned_edge_features) > 0
         ):
+            assert isinstance(
+                partition_output.edge_partition_book, abc.Mapping
+            ), f"Found heterogeneous partitioned edge features, but no corresponding heterogeneous edge partition book. Got edge partition book of type {type(partition_output.edge_partition_book)}"
             assert (
-                isinstance(partition_output.edge_partition_book, abc.Mapping)
-                and len(partition_output.edge_partition_book) > 0
-            ), f"Found heterogeneous partitioned edge features, but no corresponding heterogeneous edge partition book with at least one edge type. \
-                Got edge partition book of type {type(partition_output.edge_partition_book)}"
+                len(partition_output.edge_partition_book) > 0
+            ), f"Expected at least one edge type in edge partition book, got {partition_output.edge_partition_book.keys()}."
             edge_type_to_partitioned_edge_features = {
                 edge_type: feature_partition_data.feats
                 for edge_type, feature_partition_data in partition_output.partitioned_edge_features.items()
--- a/python/gigl/distributed/dist_partitioner.py
+++ b/python/gigl/distributed/dist_partitioner.py
@@ -832,8 +832,8 @@ class DistPartitioner:
 
         Returns:
             GraphPartitionData: The graph data of the current partition.
-            FeaturePartitionData: The edge features on the current partition
-            Optional[PartitionBook]: The partition book of graph edges.
+            Optional[FeaturePartitionData]: The edge features on the current partition, will be None if there are no edge features for the current edge type
+            Optional[PartitionBook]: The partition book of graph edges, will be None if there are no edge features for the current edge type
         """
 
         assert (
--- a/python/gigl/distributed/dist_range_partitioner.py
+++ b/python/gigl/distributed/dist_range_partitioner.py
@@ -154,8 +154,8 @@ class DistRangePartitioner(DistPartitioner):
 
         Returns:
             GraphPartitionData: The graph data of the current partition.
-            Optional[FeaturePartitionData]: The edge features on the current partition
-            Optional[PartitionBook]: The partition book of graph edges.
+            Optional[FeaturePartitionData]: The edge features on the current partition, will be None if there are no edge features for the current edge type
+            Optional[PartitionBook]: The partition book of graph edges, will be None if there are no edge features for the current edge type
         """
 
         assert (


python/gigl/distributed/dist_partitioner.py
```diff
@@ -861,14 +871,9 @@ def _edge_pfn(_, chunk_range):
             rank_indices=edge_ids,
             partition_function=_edge_pfn,
             total_val_size=num_edges,
-            generate_pb=True,
+            generate_pb=False if should_skip_edge_feats else True,
```
kmontemayor2-sc, line 874: nit. 
```suggestion
            generate_pb=not should_skip_edge_feats
```

python/gigl/distributed/dist_link_prediction_dataset.py
```diff
@@ -474,37 +476,41 @@ def build(
...
-            del (
-                edge_type_to_partitioned_edge_features,
-                edge_type_to_partitioned_edge_feature_ids,
-                edge_type_to_id2idx,
-            )
+            if len(partition_output.partitioned_edge_features) > 0:
+                assert (
+                    isinstance(partition_output.edge_partition_book, abc.Mapping)
+                    and len(partition_output.edge_partition_book) > 0
+                )
```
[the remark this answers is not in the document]
kmontemayor2-sc: or error message in the assert would be better :)

python/gigl/distributed/dist_partitioner.py
```diff
@@ -829,7 +832,7 @@ def _partition_edge_index_and_edge_features(
         Returns:
             GraphPartitionData: The graph data of the current partition.
             FeaturePartitionData: The edge features on the current partition
-            PartitionBook: The partition book of graph edges.
```
kmontemayor2-sc, line 832: let's note when we return None here? e.g. if the edge type doesn't have features then both the second and third item of the tuple will be `None`?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -838,10 +841,17 @@ def _partition_edge_index_and_edge_features(
...
+        )
+
         # Partitioning Edge Indices
 
         edge_index = self._edge_index[edge_type]
-        edge_ids = self._edge_ids[edge_type]
+        edge_ids_start, edge_ids_end = self._edge_ids[edge_type]
+        edge_ids = convert_to_tensor(
+            torch.arange(edge_ids_start, edge_ids_end), dtype=torch.int64
+        )
```
kmontemayor2-sc, line 854: good catch! btw do you think it should be possible for us to never need the `torch.arange()` bit for range partitioner, and just generate the indexed tensors on-the-fly?

python/gigl/distributed/dist_range_partitioner.py
```diff
@@ -332,27 +339,27 @@ def partition_edge_index_and_edge_features(
...
-            partitioned_edge_features if partitioned_edge_features else None
-        )
         if self._is_input_homogeneous:
             return (
                 to_homogeneous(partitioned_edge_index),
-                to_homogeneous(return_edge_features),
+                to_homogeneous(partitioned_edge_features)
+                if len(partitioned_edge_features) > 0
+                else None,
                 to_homogeneous(edge_partition_book),
```
kmontemayor2-sc, line 358: why not also return `None` here if `len(edge_partition_book) == 0`?

python/gigl/types/graph.py
```diff
@@ -49,7 +49,7 @@ class GraphPartitionData:
     # edge index (rows, cols)
     edge_index: torch.Tensor
     # edge ids tensor corresponding to `edge_index`
-    edge_ids: torch.Tensor
+    edge_ids: Optional[torch.Tensor] = None
```
kmontemayor2-sc, line 52: nit. can we add a comment when this is `None` / why it defaults that way?

python/gigl/distributed/dist_partitioner.py
mkolodner-sc: Yup, exactly! This is why we override this functionality for range-based partitioning.

python/gigl/distributed/dist_range_partitioner.py
mkolodner-sc: Good catch, thanks!

python/gigl/types/graph.py
mkolodner-sc: Added comment, removed default here since upon reflection this should always be explicitly provided

python/gigl/distributed/dist_link_prediction_dataset.py
```diff
@@ -473,8 +475,15 @@ def build(
...
-            assert isinstance(partition_output.edge_partition_book, abc.Mapping)
+        elif (
+            isinstance(partition_output.partitioned_edge_features, abc.Mapping)
+            and len(partition_output.partitioned_edge_features) > 0
+        ):
+            assert (
+                isinstance(partition_output.edge_partition_book, abc.Mapping)
+                and len(partition_output.edge_partition_book) > 0
+            ), f"Found heterogeneous partitioned edge features, but no corresponding heterogeneous edge partition book with at least one edge type. \
+                Got edge partition book of type {type(partition_output.edge_partition_book)}"
```
kmontemayor2-sc, line 486: nit. let's add the keys of `partitioned_edge_features` too?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -829,7 +833,7 @@ def _partition_edge_index_and_edge_features(
         Returns:
             GraphPartitionData: The graph data of the current partition.
             FeaturePartitionData: The edge features on the current partition
-            PartitionBook: The partition book of graph edges.
+            Optional[PartitionBook]: The partition book of graph edges.
```
kmontemayor2-sc, line 836: Sorry to be clear, I think we should document in the returns block when we are retuning None.

Also it should be `Optional[FeaturePartitionData]` right?

python/gigl/distributed/dist_partitioner.py
kmontemayor2-sc: lmao! it's such a good idea we keep coming back to it...

python/gigl/distributed/dist_range_partitioner.py
```diff
@@ -137,21 +137,25 @@ def _partition_edge_index_and_edge_features(
...
         Args:
             node_partition_book (Dict[NodeType, PartitionBook]): The partition books of all graph nodes.
             edge_type (EdgeType): The edge type for input edges
 
         Returns:
             GraphPartitionData: The graph data of the current partition.
-            FeaturePartitionData: The edge features on the current partition
-            PartitionBook: The partition book of graph edges.
+            Optional[FeaturePartitionData]: The edge features on the current partition
+            Optional[PartitionBook]: The partition book of graph edges.
```
kmontemayor2-sc, line 158: ditto on when these may be None in the `Returns` block

python/tests/unit/distributed/distributed_partitioner_test.py
```diff
@@ -114,41 +115,47 @@ def _assert_graph_outputs(
...
             is_heterogeneous=is_heterogeneous,
             expected_entity_types=expected_edge_types,
         )
+
+        if len(expected_edge_feat_types) != 0:
+            self._assert_data_type_correctness(
+                output_data=output_edge_partition_book,
+                is_heterogeneous=is_heterogeneous,
+                expected_entity_types=expected_edge_feat_types,
+            )
```
kmontemayor2-sc, line 136: if it is empty, can we make other asserts? like that `output_edge_partition_book` is empty?

kmontemayor2-sc (approved):
Thanks Matt! Good work :)

python/tests/unit/distributed/distributed_partitioner_test.py
mkolodner-sc: we check this with the subsequent logic in this function, specifically the line `self.assertTrue(edge_partition_book is None)`

PR #87, opened 2025-06-10
ignore downloaded cs gzip and remove

kmontemayor2-sc:
We weren't removing the `cs` gzip, let's do so, and also ignore the file from git/docker.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -25,7 +25,6 @@ tmp
 gigl.egg-info/
 gigl_dataflow_setup.egg-info/
 miniconda/
-cs-x86_64-pc-linux.gz
 
 # ==================== IDE / Editor folders
 .idea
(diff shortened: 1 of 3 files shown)


.dockerignore
```diff
@@ -25,6 +25,7 @@ tmp
 gigl.egg-info/
 gigl_dataflow_setup.egg-info/
 miniconda/
+cs-x86_64-pc-linux.gz
```
svij-sc, line 28: should we just remove all zip files?
*.gz
?

kmontemayor2-sc: Synced offline, we can just download to `tools/`, which we ignore anyways. 

Tested this works offline.

svij-sc (approved):
Thanks for the work!

PR #89, opened 2025-06-11
Test correct node ids are set too

kmontemayor2-sc:
Originally, I wanted to test against a heterogeneous dataset too, but we don't have any one with labels setup, and we don't expose `_ssl_positive_label_percentage` here.

But I added these checks and might as well go for them :P


mkolodner-sc (approved):
Thanks!

PR #92, opened 2025-06-11
Update DistNeighborLoader if Dataset has labeled edge types

mkolodner-sc:
**Scope of work done**

### DIFF
--- a/python/gigl/distributed/dist_ablp_neighborloader.py
+++ b/python/gigl/distributed/dist_ablp_neighborloader.py
@@ -424,9 +424,11 @@ class DistABLPLoader(DistLoader):
         Returns:
             data (Data): Homogeneous graph with the labeled edge type removed
         """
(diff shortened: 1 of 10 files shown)


python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -506,3 +500,308 @@ def _collate_fn(self, msg: SampleMessage) -> Union[Data, HeteroData]:
             data = self._supervised_to_homogeneous(data)
         data = self._set_labels(data, positive_labels, negative_labels)
         return data
+
+
+class DistRandomLoader(DistLoader):
```
mkolodner-sc, line 505: This class is currently added to do random negative neighbor fanout on an ABLP dataset (with pos+neg labels registered as edge types). TODO (@kmontemayor2-sc) investigate whether this is needed, if not we can just use DistABLPLoader and DistNeighborLoader in this case.

PR #94, opened 2025-06-12
add util function for counting total number of ids from a parition book

kmontemayor2-sc:
IT's nice to be able to determine how many total nodes we have.

### DIFF
--- a/python/tests/unit/distributed/partition_book_test.py
+++ b/python/tests/unit/distributed/partition_book_test.py
@@ -78,7 +78,7 @@ class PartitionBookTest(unittest.TestCase):
         self.assertEqual(get_total_ids(partition_book), expected_num_nodes)
 
     def test_check_partition_book(self):
-        valid_partition_book = torch.Tensor([0, 1, 1, 0, 3, 3, 2, 0, 1, 1])
+        valid_partition_book = torch.tensor([0, 1, 1, 0, 3, 3, 2, 0, 1, 1])
         _check_partition_book(valid_partition_book)
 
     @parameterized.expand(


python/gigl/distributed/utils/partition_book.py
```diff
@@ -33,3 +33,24 @@ def get_ids_on_rank(
         return _get_ids_from_range_partition_book(
             range_partition_book=partition_book, rank=rank
         )
+
+
+def get_num_ids(partition_book: Union[torch.Tensor, PartitionBook]) -> int:
```
svij-sc, line 38: Are we planning on `Union[torch.Tensor, PartitionBook]` --> just PartitionBook sometime in the future?
What is the purpose/when is `torch.Tensor` used as compared to `PartitionBook` ?

svij-sc (approved):
LGTM, just one question

python/gigl/distributed/utils/partition_book.py
```diff
@@ -33,3 +33,24 @@ def get_ids_on_rank(
...
+    """
+    Returns the number of ids (e.g. the total number of nodes) from a partition book.
+    Args:
+        partition_book (Union[torch.Tensor, PartitionBook]): Tensor or range-based partition book
+    Returns:
+        int: Total number of ids in the partition book
+    """
+    if isinstance(partition_book, torch.Tensor):
```
mkolodner-sc, line 46: Looking back, this probably also should have been done for `get_ids_on_rank` as well, but do you think we should make any checks about the shape of the partition_book tensor here, since it is user-provided?

python/gigl/distributed/utils/partition_book.py
mkolodner-sc: Tensor-based partitioning uses a torch.Tensor to store its partition book, while the PartitionBook class is used for range-based Partitioning [[1](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/partition/base.py#L30)], [[2](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/partition/partition_book.py#L6)]

mkolodner-sc (approved):
Thanks Kyle! LGTM provided tests pass

python/gigl/distributed/utils/partition_book.py
kmontemayor2-sc: We could probably swap to only using ParitionBook and using `GLTPartitionBook` instead but I think that's very low priority and tbch may make things more confusing.

kmontemayor2-sc:
Thanks folks! Some changes since you last reviewed:

1. Rename `get_num_ids` to `get_total_ids` to avoid abbreviations.
2. add `_check_parition_book` to check tensor-based parition books
3. swap `torch.**T**ensor` -> `torch.**t**ensor` in tests.

python/gigl/distributed/utils/partition_book.py
svij-sc: gotcha.
It was just confusing when I was reading it; because `RangePartitionBook` is a type of `PartitionBook` but so is torch.Tensor ?

svij-sc (approved):

PR #96, opened 2025-06-14
dataset

svij-sc:
**Scope of work done**

### DIFF
--- a/python/gigl/distributed/dataset_factory.py
+++ b/python/gigl/distributed/dataset_factory.py
@@ -18,6 +18,7 @@ from graphlearn_torch.distributed import (
     shutdown_rpc,
 )
 
+import gigl.distributed.utils
 from gigl.common import Uri, UriFactory
 from gigl.common.data.dataloaders import TFRecordDataLoader
 from gigl.common.data.load_torch_tensors import (
@@ -207,8 +208,10 @@ def _build_dataset_process(
     process_number_on_current_machine: int,
     output_dict: MutableMapping[str, DistLinkPredictionDataset],
     serialized_graph_metadata: SerializedGraphMetadata,
-    distributed_context: DistributedContext,
-    dataset_building_port: int,
+    master_ip_address: str,
+    node_rank: int,
+    node_world_size: int,
+    master_dataset_building_port: int,
     sample_edge_direction: Literal["in", "out"],
     should_load_tensors_in_parallel: bool,
     partitioner_class: Optional[Type[DistPartitioner]],
@@ -240,7 +243,7 @@ def _build_dataset_process(
             will be written to for use by the parent process
         serialized_graph_metadata (SerializedGraphMetadata): Metadata about TFRecords that are serialized to disk
         distributed_context (DistributedContext): Distributed context containing information for master_ip_address, rank, and world size
-        dataset_building_port (int): RPC port to use to build the dataset
+        master_dataset_building_port (int): RPC port to use to build the dataset
         sample_edge_direction (Literal["in", "out"]): Whether edges in the graph are directed inward or outward
         should_load_tensors_in_parallel (bool): Whether tensors should be loaded from serialized information in parallel or in sequence across the [node, edge, pos_label, neg_label] entity types.
         partitioner_class (Optional[Type[DistPartitioner]]): Partitioner class to partition the graph inputs. If provided, this must be a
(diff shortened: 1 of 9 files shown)


kmontemayor2-sc (commented):
Thanks Shubham!

python/gigl/distributed/dataset_factory.py
```diff
@@ -354,33 +407,52 @@ def build_dataset(
...
             be read from inferenceArgs. Otherwise, arguments witll be read from trainerArgs.
         _tfrecord_uri_pattern (str): INTERNAL ONLY. Regex pattern for loading serialized tf records. Defaults to ".*-of-.*\.tfrecord(\.gz)?$".
     """
+
+    if distributed_context is not None:
+        logger.warning(
+            "The `distributed_context` argument is deprecated and will be removed in a future release. "
+            "Please setup the `torch.distributed.init_process_group` in the caller context and prevent "
+            "passing `distributed_context` argument."
+        )
```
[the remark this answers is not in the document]
svij-sc: because both of them take it is an argument.
Not the worst thing to log it twice, hopefully people read it then eh?

python/tests/test_assets/distributed/run_distributed_dataset.py
```diff
@@ -16,16 +19,41 @@
...
     world_size: int,
     mocked_dataset_info: MockedDatasetInfo,
     output_dict: MutableMapping[int, DistLinkPredictionDataset],
     should_load_tensors_in_parallel: bool,
-    master_ip_address: str,
-    master_port: int,
     partitioner_class: Optional[Type[DistPartitioner]] = None,
     splitter: Optional[NodeAnchorLinkSplitter] = None,
+    _use_process_group: bool = False,  # TODO: (svij) Marked for deprecation, use_process_group will default to be True in the future
+    _port: Optional[int] = None,  # TODO: (svij) Marked for deprecation
```
kmontemayor2-sc, line 56: Do we only have these for the tests in `distributed_dataset_test.py`? Specifically `test_dataset_correctness`?

Are we adding these so we can test both the new and old paths at the same time in our tests before we deprecate the old path?

python/gigl/distributed/utils/__init__.py
```diff
@@ -3,9 +3,11 @@
 """
 
 __all__ = [
+    "convert_pb_to_serialized_graph_metadata",
```
kmontemayor2-sc, line 6: more of a discussion point:

I feel like we should only be exposing "public" functions in __all__ (or __init__.py more generally).

I think we should be as explicit as possible in our own code (either from `gigl.distributed.utils networking import get_free_port`, or `from gigl.distributed.utils import networking; ... networking.get_free_port()`)

While __all__ and __init__ should be more like "what we want our users to use", and I'm not sure we want them to use convert_pb_to_serialized_graph_metadata, or even really be aware of it.

cc: @mkolodner-sc for the discussion. Though this is more something I wanted to bring up generally. If we're trying to simplify it should be for our users (and our examples). I think it's "fine" for us to go the longer route.

python/gigl/distributed/dataset_factory.py
```diff
@@ -286,21 +292,33 @@ def _build_dataset_process(
...
-    _dataset_building_port: int = DEFAULT_MASTER_DATA_BUILDING_PORT,
 ) -> DistLinkPredictionDataset:
     """
-    Launches a spawned process for building and returning a DistLinkPredictionDataset instance provided some SerializedGraphMetadata
+    Launches a spawned process for building and returning a DistLinkPredictionDataset instance provided some
+    SerializedGraphMetadata.
+
+    It is expected that there is only one `build_dataset` call per node (machine).
+    This requirement exists to ensure each machine only participates once in housing a parition of a dataset; otherwise
+    a machine may end up housing multiple partitions of the same dataset which may cause memory issues.
```
kmontemayor2-sc, line 310: What happens if multiple build_datasets are called concurrently on the same machine? won't they explode when they all try to establish the same process group [1]? 

[1]: https://github.com/Snapchat/GiGL/pull/96/files#diff-09c9908e0037ea4d00a203286f28f36662788d19e440b60e5c9cae662ccbc1c5R370

svij-sc: correct... this would be the case.
Are there foreseeable cases where we try to do this?

python/gigl/distributed/utils/__init__.py
svij-sc: This proposal makes sense to me.
I will make the change here, but we will have to revisit our existing imports and refactor them accordingly.

python/tests/test_assets/distributed/run_distributed_dataset.py
kmontemayor2-sc: > I was concerned that I'd get the opposite comment if I did it the other way 

Honestly maybe :P.

Though I am generally in favor of "legacy" paths being more KTLO vs the "modern" code paths being better tested but I could see how/why others would disagree.

PR #100, opened 2025-06-18
Tag E2E integration test pipelines with git branch

kmontemayor2-sc:
**Scope of work done**

Try to pipe in GIT_BRANCH to cloud build workers, so we can have our e2e pipelines named with the git branch.

e.g. https://console.cloud.google.com/vertex-ai/pipelines/locations/us-central1/runs/cora-glt-udl-test-on--04698?inv=1&invt=Ab0dkw&project=external-snap-ci-github-gigl will have the branch name in it.

Tested that the make file changes work locally. 

I cannot test the gh action changes easily locally so let's "test on main".

If the integration tests start failing let's roll this back immediately. 

### DIFF
--- a/.github/actions/run-cloud-run-command-on-active-checkout/action.yml
+++ b/.github/actions/run-cloud-run-command-on-active-checkout/action.yml
@@ -17,7 +17,7 @@ inputs:
   timeout:
     description: "Specified as duration i.e. '2h15m5s' is two hours, fifteen minutes, and five seconds"
     default: "3h"
-  
+
 
 runs:
   using: "composite"
@@ -30,5 +30,6 @@ runs:
           --service-account="projects/${{ inputs.project }}/serviceAccounts/${{ inputs.service_account }}" \
           --project="${{ inputs.project }}" \
           --machine-type="${{ inputs.machine_type }}" \
-          --timeout="${{ inputs.timeout }}"
+          --timeout="${{ inputs.timeout }}" \
+          --set-build-env-vars="GIT_HASH=$(git rev-parse HEAD),GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)"
       shell: bash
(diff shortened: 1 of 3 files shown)


.github/actions/run-cloud-run-command-on-active-checkout/action.yml
```diff
@@ -17,18 +17,20 @@ inputs:
...
           --config=.github/cloud_builder/run_command_on_active_checkout.yaml \
           --substitutions=_CMD="${{ inputs.cmd }}" \
           --service-account="projects/${{ inputs.project }}/serviceAccounts/${{ inputs.service_account }}" \
           --project="${{ inputs.project }}" \
           --machine-type="${{ inputs.machine_type }}" \
-          --timeout="${{ inputs.timeout }}"
-      shell: bash
\ No newline at end of file
+          --timeout="${{ inputs.timeout }}" \
+          --set-build-env-vars="GIT_HASH=$(git rev-parse HEAD),GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)"
```
svij-sc, line 35: btw, the GIT variables already should be available here: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

svij-sc (approved):
Thanks

.github/actions/run-cloud-run-command-on-active-checkout/action.yml
kmontemayor2-sc: I found https://stackoverflow.com/questions/58033366/how-to-get-the-current-branch-within-github-actions which implies sometimes I need to use `github.head_ref` and sometimes it should be `github.ref_name` depending on if it's an `push` or a `pull_request`, neither of which are applicable here.

I would assume invoking the git stuff manually is more reliable? If this works we can follow-up to try and use the more "native" ways here.

PR #104, opened 2025-06-19
Ensure node ids are always selected into the same split

kmontemayor2-sc:
Changes:
* update _fast_hash to run again to improve mixing - without this our N=20 node ids had splits like [18], [0], [2]
* enforce `val_num` and `test_num` as floats - since we can't guarantee counts easily anymore.
* Update splitting logic to ensure nodes are always selected into the same split [discussed more below].
* Added a (basic) test for the "distributed" case.
* Removed some tests that relied on integer counts.

The issue with our previous splitting logic, is that even though a given node id would always have the same hash on different machines, since we sorted the hashed, it's "position" may be different and so it may be selected into different splits.

For instance, let's assume the hash function is the identity, and `rank_0_nodes: [0, 1, 2, 3] rank_1_nodes: [3, 4, 5, 6]`

On rank 0, `3` would be selected into Test, as its hash value is the greatest sorted, while it would be in train on rank 1.

Now what we do is fine the globally largest/smallest hash and then normalize the hash values per machine.

We then select the nodes based on the normalized values, since the hashes are consistent, and they get normalized the same, the same node id will be selected into the same split always now.

### DIFF
--- a/python/gigl/distributed/dataset_factory.py
+++ b/python/gigl/distributed/dataset_factory.py
@@ -264,6 +264,16 @@ def _build_dataset_process(
         master_port=dataset_building_port,
         num_rpc_threads=16,
     )
+    # HashedNodeAnchorLinkSplitter requires rpc to be initialized, so we initialize it here.
+    should_teardown_process_group = False
+    if isinstance(splitter, HashedNodeAnchorLinkSplitter):
+        should_teardown_process_group = True
+        torch.distributed.init_process_group(
+            backend="gloo",
+            init_method=f"tcp://{distributed_context.main_worker_ip_address}:{dataset_building_port}",
+            world_size=distributed_context.global_world_size,
+            rank=distributed_context.global_rank,
+        )
 
     output_dataset: DistLinkPredictionDataset = _load_and_build_partitioned_dataset(
         serialized_graph_metadata=serialized_graph_metadata,
@@ -282,6 +292,8 @@ def _build_dataset_process(
     # machines may require rpc setup for partitioning, which will result in failure.
     barrier()
     shutdown_rpc()
+    if should_teardown_process_group:
+        torch.distributed.destroy_process_group()
 
 
 def build_dataset(
--- a/python/gigl/utils/data_splitters.py
+++ b/python/gigl/utils/data_splitters.py
@@ -96,7 +96,19 @@ def _fast_hash(x: torch.Tensor) -> torch.Tensor:
         x.bitwise_xor_(x >> 15)
         x.multiply_(0x846CA68B)
         x.bitwise_xor_(x >> 16)
+        # And again for better mixing ;)
+        x.bitwise_xor_(x >> 16)
+        x.multiply_(0x7FEB352D)
+        x.bitwise_xor_(x >> 15)
+        x.multiply_(0x846CA68B)
+        x.bitwise_xor_(x >> 16)
     elif x.dtype == torch.int64:
+        x.bitwise_xor_(x >> 30)
+        x.multiply_(0xBF58476D1CE4E5B9)
+        x.bitwise_xor_(x >> 27)
+        x.multiply_(0x94D049BB133111EB)
+        x.bitwise_xor_(x >> 31)
+        # And again for better mixing ;)
         x.bitwise_xor_(x >> 30)
         x.multiply_(0xBF58476D1CE4E5B9)
         x.bitwise_xor_(x >> 27)
@@ -129,8 +141,8 @@ class HashedNodeAnchorLinkSplitter:
     def __init__(
         self,
         sampling_direction: Union[Literal["in", "out"], str],
-        num_val: Union[float, int] = 0.1,
-        num_test: Union[float, int] = 0.1,
+        num_val: float = 0.1,
+        num_test: float = 0.1,
         hash_function: Callable[[torch.Tensor], torch.Tensor] = _fast_hash,
         supervision_edge_types: Optional[list[EdgeType]] = None,
         should_convert_labels_to_edges: bool = True,
@@ -139,10 +151,8 @@ class HashedNodeAnchorLinkSplitter:
 
         Args:
             sampling_direction (Union[Literal["in", "out"], str]): The direction to sample the nodes. Either "in" or "out".
-            num_val (Union[float, int]): The percentage of nodes to use for training. Defaults to 0.1 (10%).
-                                         If an integer is provided, than exactly that number of nodes will be in the validation split.
-            num_test (Union[float, int]): The percentage of nodes to use for validation. Defaults to 0.1 (10%).
-                                          If an integer is provided, than exactly that number of nodes will be in the test split.
+            num_val (float): The percentage of nodes to use for training. Defaults to 0.1 (10%).
+            num_test (float): The percentage of nodes to use for validation. Defaults to 0.1 (10%).
             hash_function (Callable[[torch.Tensor, torch.Tensor], torch.Tensor]): The hash function to use. Defaults to `_fast_hash`.
             supervision_edge_types (Optional[list[EdgeType]]): The supervision edge types we should use for splitting.
                 Must be provided if we are splitting a heterogeneous graph. If None, uses the default message passing edge type in the graph.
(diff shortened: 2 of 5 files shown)


python/gigl/utils/data_splitters.py
```diff
@@ -96,12 +96,24 @@ def _fast_hash(x: torch.Tensor) -> torch.Tensor:
         x.bitwise_xor_(x >> 15)
         x.multiply_(0x846CA68B)
         x.bitwise_xor_(x >> 16)
+        # And again for better mixing ;)
```
nshah-sc, line 99: is there any guidance on how these xors and multiplies are chosen?  assuming it doesn't materially affect the randomness of the outputs

python/gigl/utils/data_splitters.py
```diff
@@ -293,32 +303,55 @@ def __call__(
...
-            if isinstance(self._num_test, int):
-                num_test = self._num_test
+            hash_values = self._hash_function(nodes_to_select)  # 1 x M
+            # Now, we want to normalize the hash values to [0, 1) range so we can select them easily into splits.
+            # We want to do this *globally* e.g. across all processes,
+            # so that we can ensure that the same nodes are selected for the same split across all processes.
+            # If we don't do this, then if we have `[0, 1, 2, 3, 4]` on one process and `[4, 5, 6, 7]` on another,
+            # with the identity hash `4` may end up in Test in one rank and Train in another.
+            max_hash_value = hash_values.max().item()
+            min_hash_value = hash_values.min().item()
```
nshah-sc, line 313: i believe you can use [aminmax](https://docs.pytorch.org/docs/stable/generated/torch.aminmax.html) which is more efficient (single pass) to do this rather than making two passes (prob not a big deal but these tensors may be large)

python/gigl/utils/data_splitters.py
```diff
@@ -129,8 +141,8 @@ class HashedNodeAnchorLinkSplitter:
     def __init__(
         self,
         sampling_direction: Union[Literal["in", "out"], str],
-        num_val: Union[float, int] = 0.1,
-        num_test: Union[float, int] = 0.1,
+        num_val: float = 0.1,
+        num_test: float = 0.1,
```
nshah-sc, line 145: maybe worth renaming these to val_ratio, test_ratio etc.  num_ is a bit misleading

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -553,8 +552,15 @@ def test_cora_supervised(self):
             splitter=splitter,
         )
 
+        assert dataset.train_node_ids is not None
```
nshah-sc, line 555: nit: opportunity for a helpful error message

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -30,23 +34,73 @@
 _INVALID_TEST_EDGE_INDEX = torch.arange(0, _NUM_EDGES * 10).reshape((10, _NUM_EDGES))
 
 
+def rank_0(process_num: int):
+    """Test for "distributed" splitting in that the same nodes are selected into the same splits across ranks."""
+    del process_num  # Unused in this function
```
nshah-sc, line 39: this is a level of pettiness i aspire to

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -30,23 +34,73 @@
...
+            torch.arange(0, 40, 2, dtype=torch.int64),
+            torch.zeros(20, dtype=torch.int64),
+        ]
+    )
+    train, val, test = splitter(edges)
+    assert_tensor_equality(
+        train,
+        torch.tensor(
+            [0, 2, 4, 6, 12, 14, 18, 20, 22, 28, 30, 36, 38], dtype=torch.int64
+        ),
```
nshah-sc, line 56: how were these values chosen exactly?  is this based on the output from fast_hash that was precomputed?

i figured another way to test this more straightforwardly that could be resilient to the precise edge definition would be to have each rank be responsible for some segment, and then allgather the train/val/test lists and sanity check that the set intersections between rank_0's train to rank_1's val/test are null and so on.

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -426,6 +435,11 @@ def test_node_based_link_splitter_heterogenous(
         test_num,
         expected,
     ):
+        os.environ["MASTER_ADDR"] = "localhost"
+        os.environ["MASTER_PORT"] = str(get_free_port())
```
nshah-sc, line 439: just pointing out the maybe_teardowns and os.environ setting are reused a lot and could be moved to some small utilities at the top of this file for convenience but i understand if you are opting for compartmentalizatoin/readability of each test here

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -448,6 +462,29 @@ def test_node_based_link_splitter_heterogenous(
             assert_close(val, expected_val, rtol=0, atol=0)
             assert_close(test, expected_test, rtol=0, atol=0)
 
+    def test_node_based_link_splitter_parallized(self):
```
nshah-sc, line 465: parallelized*

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -481,22 +518,6 @@ def test_node_based_link_splitter_heterogenous_invalid(
             )
             splitter(edge_index=edges)
 
-    def test_node_based_link_splitter_no_train_nodes(self):
```
nshah-sc, line 484: is this covered by another testcase?  why drop?

python/gigl/utils/data_splitters.py
svij-sc: yea lets tick to this convention for now.
Maybe we can emphasize in docstring a quick blurb on why we dont support raw number asin in pyg.

python/gigl/utils/data_splitters.py
```diff
@@ -293,32 +305,58 @@ def __call__(
...
-            del hash_values
-            gc.collect()
-
-            if isinstance(self._num_val, int):
-                num_val = self._num_val
-            else:
-                num_val = int(nodes_to_select.numel() * self._num_val)
-            if isinstance(self._num_test, int):
-                num_test = self._num_test
+            hash_values = self._hash_function(nodes_to_select)  # 1 x M
```
svij-sc, line 308: opportunity to break this function into smaller composable pieces for better readability and maintainability.

python/gigl/utils/data_splitters.py
```diff
@@ -293,32 +305,58 @@ def __call__(
...
+            # Now, we want to normalize the hash values to [0, 1) range so we can select them easily into splits.
+            # We want to do this *globally* e.g. across all processes,
+            # so that we can ensure that the same nodes are selected for the same split across all processes.
+            # If we don't do this, then if we have `[0, 1, 2, 3, 4]` on one process and `[4, 5, 6, 7]` on another,
+            # with the identity hash `4` may end up in Test in one rank and Train in another.
+            min_hash_value, max_hash_value = map(
+                torch.Tensor.item, hash_values.aminmax()
+            )
+            if torch.distributed.is_initialized():
```
svij-sc, line 317: Won't the partition book already have information on min/max node_ids ? Could that be used?

python/gigl/utils/data_splitters.py
```diff
@@ -293,32 +305,58 @@ def __call__(
...
+                        global_min_hash_value, max_and_min[1].item()
+                    )
             else:
-                num_test = int(nodes_to_select.numel() * self._num_test)
-
-            num_train = nodes_to_select.numel() - num_val - num_test
-            if num_train <= 0:
-                raise ValueError(
-                    f"Invalid number of nodes to split. Expected more than 0. Originally had {nodes_to_select.numel()} nodes but due to having `num_test` = {self._num_test} and `num_val` = {self._num_val} got no training node.."
+                logger.warning(
```
svij-sc, line 336: Should we just error out if  process group is not initialized to gaurd against user errors ?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -553,8 +552,15 @@ def test_cora_supervised(self):
...
+
         mp.spawn(
-            fn=_run_cora_supervised, args=(dataset, self._context, expected_data_count)
+            fn=_run_cora_supervised,
+            args=(
+                dataset,
+                self._context,
+                to_homogeneous(dataset.train_node_ids).numel(),
```
svij-sc, line 562: comment on why we are converting `to_homogeneous`

python/gigl/utils/data_splitters.py
```diff
@@ -90,6 +89,9 @@ def _fast_hash(x: torch.Tensor) -> torch.Tensor:
         The hash values of the input tensor `x`. N x M
     """
     x = x.clone().detach()
+
+    # Add one so that _fast_hash(0) != 0
+    x.add_(1)
```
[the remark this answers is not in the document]
kmontemayor2-sc: the edge index is composed of non-negative ints.

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -77,30 +78,52 @@ def rank_1(process_num: int):
...
+    def setUp(self):
+        self._original_master_addr = os.environ.get(
+            "MASTER_ADDR", "SHOULD NEVER BE AN ADDR"
+        )
+        self._original_master_port = os.environ.get(
+            "MASTER_PORT", "SHOULD NEVER BE A PORT"
+        )
+        os.environ["MASTER_ADDR"] = "localhost"
+        os.environ["MASTER_PORT"] = str(get_free_port())
+        super().setUp()
```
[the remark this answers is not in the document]
kmontemayor2-sc: Not all of these tests spawn new processes. My understanding is we don't run each test suite in a new process [1] so we could potentially pollute the global process here. 

[1]: https://github.com/Snapchat/GiGL/blob/c221fcdd8f1d747a1974b38454af70164fec8a72/python/gigl/common/utils/test_utils.py#L82-L84

python/gigl/utils/data_splitters.py
kmontemayor2-sc: > i meant that they shouldn't impact the actual entropy of the distribution (should still be uniform, and arguably closer to uniform post the mixing)

I think they still should right? For example if we `xor` against `7` then most of the bits of the number wont be swapped right? Or if we multiple against say 4 then we'd be losing 2 bits of entropy every time?

Though tbf you may be right here, but I'd prefer to not mess with the numbers if we don't need to.

python/gigl/utils/data_splitters.py
kmontemayor2-sc: Yes, the partition book has information on all node ids, but the hash function isn't monotonic so we'd need to hash all node ids to find the min/max hash values there.

python/gigl/utils/data_splitters.py
kmontemayor2-sc: Hmmm - I don't think the splitter need be used only for distributed right?

I don't have super strong feelings either way here but I could see it being "more user friendly" to support "local" use cases - though maybe it's more error prone in our target use case?

python/gigl/utils/data_splitters.py
kmontemayor2-sc: I have a minor preference to leave it as is - this stuff in kind of complicated and the pieces all need to work together. For me, it's easier to read one longish function as opposed to several smaller ones that jump around to each other.

python/gigl/utils/data_splitters.py
kmontemayor2-sc: Actually - I thought about this more and it's probably best to error out here.

We can enable the local usecase in the future if we want :).

python/gigl/utils/data_splitters.py
svij-sc: got it.

python/gigl/utils/data_splitters.py
svij-sc: got it; I will defer to you as the author but reading this on first glace I thought if the code was restructured to be in smaller functions it would have been self documenting and easier to read.

python/gigl/utils/data_splitters.py
kmontemayor2-sc: that makes sense, it's more like, should we be maximally convenient for "pure local" users, since it is non-zero overhead to init a size 1 process group? But we can revisit in the future if/when we run into that.

python/gigl/utils/data_splitters.py
kmontemayor2-sc: ack, I may end up regret this in 2 months for for not seeing what the code is doing is easier for me to understand than reading some function named "parse_node_metadata" or whatever and then "run_node_splits" and if I need to debug there's more jumping around?

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -30,23 +34,96 @@
...
+        ),
+    )
+    assert_tensor_equality(val, torch.tensor([14, 24, 36], dtype=torch.int64))
+    assert_tensor_equality(test, torch.tensor([18], dtype=torch.int64))
+
+
+def rank_1(process_num: int):
+    """Test for "distributed" splitting in that the same nodes are selected into the same splits across ranks."""
+    del process_num
+    torch.distributed.init_process_group(rank=1, world_size=2)
```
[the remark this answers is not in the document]
svij-sc: you mean like having a utility for this?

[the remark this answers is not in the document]
kmontemayor2-sc: Hmmm made some changes, WDYT?

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -30,23 +34,96 @@
...
+    train, val, test = splitter(edges)
+    assert_tensor_equality(
+        train,
+        torch.tensor(
+            [0, 2, 4, 6, 8, 10, 12, 16, 20, 22, 26, 28, 30, 32, 34, 38],
+            dtype=torch.int64,
+        ),
+    )
+    assert_tensor_equality(val, torch.tensor([14, 24, 36], dtype=torch.int64))
+    assert_tensor_equality(test, torch.tensor([18], dtype=torch.int64))
```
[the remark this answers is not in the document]
svij-sc: Since we are not testing the correctness of the hash function here one could re-use it here.
i.e.  in this function we could call something like

```python
...
expected_train_nodes, expected_val_nodes, expected_test_nodes = HashedNodeAnchorLinkSplitter._split_train_test_val(hash_fn, ..., global_max_hash, global_min_hash, etc) # Assuming we already have test coverage for this
...
...

train, val, test = splitter(edges)
assert_tensor_equality(val, torch.tensor([14, 24, 36], dtype=torch.int64))
```

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -448,6 +452,61 @@ def test_node_based_link_splitter_heterogenous(
...
+            ),
+            (
+                # For process 1, all of the nodes would be selected into train for this example,
+                # As the identity hash does not mix, and on process 0 [0, 9] are all train too.
+                torch.arange(10, dtype=torch.int64),
+                torch.tensor([], dtype=torch.int64),
+                torch.tensor([], dtype=torch.int64),
+            ),
+            (
+                torch.arange(0, 16, 2, dtype=torch.int64),
```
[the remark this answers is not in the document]
kmontemayor2-sc: Hmmm I added a comment trying to explain this here - I think it'll be difficult to explain fully in a comment but hopefully the combination of identity hash and this comment helps make it more clear :)

python/gigl/distributed/dataset_factory.py
```diff
@@ -264,6 +264,16 @@ def _build_dataset_process(
         master_port=dataset_building_port,
         num_rpc_threads=16,
     )
+    # HashedNodeAnchorLinkSplitter requires rpc to be initialized, so we initialize it here.
```
kmontemayor2-sc, line 267: btw @svij-sc I added this since we now require a process group

PR #108, opened 2025-06-23
Patch SerializedGraphMetadata Translator

mkolodner-sc:
**Scope of work done**

PR which patches two items with the translator
- Fix a bug where we weren't loading positive labels if the label's `feature_dim` is set to 0
- Modify code so that an edge type exists which doesn't have a `mainEdgeInfo` field and only a `positiveEdgeInfo`, we still are able to load, partition, and register that label to the dataset

### DIFF
--- a/python/gigl/common/data/load_torch_tensors.py
+++ b/python/gigl/common/data/load_torch_tensors.py
@@ -46,11 +46,13 @@ class SerializedGraphMetadata:
     edge_entity_info: Union[
         SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]
     ]
-    # Positive Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. If data has no positive labels across all edge types, this value is None
+    # Positive Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases.
+    # If there are no positive labels for any edge type, this value is None
     positive_label_entity_info: Optional[
         Union[SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]]
     ] = None
-    # Negative Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. If input has no negative labels across all edge types, this value is None.
+    # Negative Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases.
+    # If there are no negative labels for any edge type, this value is None.
     negative_label_entity_info: Optional[
         Union[SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]]
     ] = None
--- a/python/gigl/common/data/load_torch_tensors.py
+++ b/python/gigl/common/data/load_torch_tensors.py
@@ -46,15 +46,13 @@ class SerializedGraphMetadata:
     edge_entity_info: Union[
         SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]
     ]
-    # Positive Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. May be None
-    # for specific edge types. If data has no positive labels across all edge types, this value is None
+    # Positive Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. If data has no positive labels across all edge types, this value is None
     positive_label_entity_info: Optional[
-        Union[SerializedTFRecordInfo, Dict[EdgeType, Optional[SerializedTFRecordInfo]]]
+        Union[SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]]
     ] = None
-    # Negative Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. May be None
-    # for specific edge types. If input has no negative labels across all edge types, this value is None.
+    # Negative Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. If input has no negative labels across all edge types, this value is None.
     negative_label_entity_info: Optional[
-        Union[SerializedTFRecordInfo, Dict[EdgeType, Optional[SerializedTFRecordInfo]]]
+        Union[SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]]
     ] = None
 
 
--- a/python/gigl/distributed/dist_partitioner.py
+++ b/python/gigl/distributed/dist_partitioner.py
@@ -1269,6 +1269,7 @@ class DistPartitioner:
         """
         Partitions positive or negative labels of a graph. If heterogeneous, partitions labels for all edge type.
         Must call `partition_node` first to get the node partition book as input.
+        Note that labels are always partitioned by the source node, since the edges are always assumed to be anchor_node -> to -> supervision node.
         Args:
             node_partition_book (Union[PartitionBook, Dict[NodeType, PartitionBook]]): The computed Node Partition Book
             is_positive (bool): Whether positive labels are currently being registered. If False, negative labels will be partitioned.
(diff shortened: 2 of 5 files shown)


python/gigl/common/data/load_torch_tensors.py
```diff
@@ -46,15 +46,13 @@ class SerializedGraphMetadata:
     edge_entity_info: Union[
         SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]
     ]
-    # Positive Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. May be None
-    # for specific edge types. If data has no positive labels across all edge types, this value is None
+    # Positive Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. If data has no positive labels across all edge types, this value is None
```
kmontemayor2-sc, line 49: "If data has no positive labels across all edge types, this value is None"

Can we clarify what that means?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -989,6 +989,10 @@ def _partition_label_edge_index(
         Returns:
             torch.Tensor: Edge index tensor of positive or negative labels, depending on is_positive flag
         """
+        if edge_type.src_node_type not in node_partition_book:
```
kmontemayor2-sc, line 992: Does this not depend on the edge_dir?

python/gigl/distributed/utils/serialized_graph_metadata_translator.py
kmontemayor2-sc, line 1: Can we add tests for these changes?

python/gigl/common/data/load_torch_tensors.py
```diff
@@ -46,15 +46,13 @@ class SerializedGraphMetadata:
...
     positive_label_entity_info: Optional[
-        Union[SerializedTFRecordInfo, Dict[EdgeType, Optional[SerializedTFRecordInfo]]]
+        Union[SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]]
     ] = None
-    # Negative Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. May be None
-    # for specific edge types. If input has no negative labels across all edge types, this value is None.
+    # Negative Label Entity Info, if present, a SerializedTFRecordInfo for homogeneous and Dict[EdgeType, SerializedTFRecordInfo] for heterogeneous cases. If input has no negative labels across all edge types, this value is None.
     negative_label_entity_info: Optional[
-        Union[SerializedTFRecordInfo, Dict[EdgeType, Optional[SerializedTFRecordInfo]]]
+        Union[SerializedTFRecordInfo, Dict[EdgeType, SerializedTFRecordInfo]]
```
kmontemayor2-sc, line 55: Why'd we stop making these `dict[EdgeType, Optional[SerializedTFRecordInfo]]`? 

IIRC it was like this because a negative label may not have edge features to load, right?

mkolodner-sc: We made this field optional previously since there could be specific edge types which don't have labels to load (but might have edges). This change opts for a cleaner approach by, rather than putting `None` in the dictionary, just not putting the edge type key in the dictionary at all if it is not present for that edge type.

kmontemayor2-sc: I see, should we update some usages [1], [2], [3] now since this can never be None?

[1]: https://github.com/Snapchat/GiGL/blob/c221fcdd8f1d747a1974b38454af70164fec8a72/python/tests/unit/distributed/dataset_input_metadata_translator_test.py#L245-L247
[2]: https://github.com/Snapchat/GiGL/blob/c221fcdd8f1d747a1974b38454af70164fec8a72/python/tests/unit/distributed/dataset_input_metadata_translator_test.py#L330-L332
[3]: https://github.com/Snapchat/GiGL/blob/c221fcdd8f1d747a1974b38454af70164fec8a72/python/gigl/common/data/load_torch_tensors.py#L49-L58

python/gigl/distributed/dist_partitioner.py
kmontemayor2-sc: Do we assert this anywhere in our code?

mkolodner-sc: I'm not sure what the assert would look like in this case, but I've added a comment to the `partition_labels` function repeating the above. Hopefully this is sufficient since we'll be deprecating this field in the future anyway.

python/tests/unit/distributed/distributed_dataset_test.py
```diff
@@ -401,6 +416,87 @@ def test_build_and_split_dataset_heterogeneous(
...
+        dataset = build_dataset(
+            serialized_graph_metadata=augmented_serialized_graph_metadata,
+            distributed_context=distributed_context,
+            sample_edge_direction="in",
+            should_load_tensors_in_parallel=True,
+            _dataset_building_port=master_port,
+        )
+
+        self.assertTrue(
+            isinstance(dataset.positive_edge_label, abc.Mapping)
```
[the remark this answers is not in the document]
mkolodner-sc: Resolved offline

python/gigl/distributed/dist_partitioner.py
kmontemayor2-sc: > deprecating this field

Which field? 

What I mean about asserts is I am curious if we ever enforce that `anchor to label` bit in our code anyways, what if someone sets `label from anchor` or `label to anchor`? In that case this check would not be correct right?

So I'd prefer if we could guarantee this somewhere.

python/tests/unit/distributed/dataset_input_metadata_translator_test.py
```diff
@@ -242,16 +244,17 @@ def test_translator_correctness(self, _, mocked_dataset_info: MockedDatasetInfo)
...
                 serialized_graph_metadata.positive_label_entity_info, abc.Mapping
             ):
                 serialized_positive_label_info_iterable = list(
                     serialized_graph_metadata.positive_label_entity_info.values()
                 )
-            else:
+            elif isinstance(
+                serialized_graph_metadata.positive_label_entity_info,
+                SerializedTFRecordInfo,
+            ):
```
kmontemayor2-sc, line 257: should we raise an error if this is also false?

python/tests/unit/distributed/distributed_dataset_test.py
```diff
@@ -401,6 +416,88 @@ def test_build_and_split_dataset_heterogeneous(
...
+            global_world_size=1,
+        )
+
+        dataset = build_dataset(
+            serialized_graph_metadata=augmented_serialized_graph_metadata,
+            distributed_context=distributed_context,
+            sample_edge_direction="in",
+            should_load_tensors_in_parallel=True,
+            _dataset_building_port=master_port,
+        )
```
kmontemayor2-sc, line 489: cc: @svij-sc depending on when his PR's get in with relation to this one :)

python/tests/unit/distributed/distributed_dataset_test.py
```diff
@@ -401,6 +416,88 @@ def test_build_and_split_dataset_heterogeneous(
...
+
+        assert isinstance(
+            dataset.positive_edge_label, abc.Mapping
+        ), f"Positive edge indices must be a dictionary, got {type(dataset.positive_edge_label)}"
+        self.assertTrue(labeled_edge_type in dataset.positive_edge_label)
+
+        assert isinstance(
+            dataset.negative_edge_label, abc.Mapping
+        ), f"Negative edge indices must be a dictionary, got {type(dataset.negative_edge_label)}"
+        self.assertTrue(labeled_edge_type in dataset.negative_edge_label)
```
kmontemayor2-sc, line 499: should we also assert the label edge type is not in dataset.edge_types or whatnot?

python/gigl/distributed/dist_partitioner.py
mkolodner-sc: > Which field?

The `positive_labels` and `negative_labels`, which `partition_labels` generates.

>What I mean about asserts is I am curious if we ever enforce that anchor to label bit in our code anyways, what if someone sets label from anchor or label to anchor? In that case this check would not be correct right?
So I'd prefer if we could guarantee this somewhere.

I understand your point, but given we are provided an edge index tensor with positive labels, as far as I know there's not a clear way we can assert that the edge direction here is `anchor_to_label` as opposed to `label_to_anchor`. I'm open to suggestions if you have any ideas -- although I really don't think this is in scope of this PR which is focused on patching the `SerializedGraphMetadata` translator. Additionally, as I mentioned, we will be deprecating the `positive_labels` and `negative_labels` in the future, and the `partition_labels` function as a result, so I don't think this should be very high priority right now.

kmontemayor2-sc (approved):

PR #109, opened 2025-06-24
Refactor homogeneous to not use dist context + leverage automatic infering ports

svij-sc:
**Scope of work done**

Leverages https://github.com/Snapchat/GiGL/pull/96 and https://github.com/Snapchat/GiGL/pull/98; reducing the need to call `connect_worker_pool` and passing around a `distributed_context` object. This also moves the logic of this inferencer to automatically infer ports, ip address from master node, et al using process groups that now the inferencer initializes.

Homogeneous inference is used by our e2e `cora_glt_udl_test`; similarly backwards compatibility is tested by `dblp_glt_test` which uses heterogeneous inference and has not been updated. Thus running e2e integration tests below.

### DIFF
--- a/examples/distributed/homogeneous_inference.py
+++ b/examples/distributed/homogeneous_inference.py
@@ -395,8 +395,9 @@ def _run_example_inference(
     inference_batch_size = gbml_config_pb_wrapper.inferencer_config.inference_batch_size
 
     local_world_size: int
-    if inferencer_args.get("local_world_size") is not None:
-        local_world_size = int(inferencer_args.get("local_world_size"))
+    arg_local_world_size = inferencer_args.get("local_world_size")
+    if arg_local_world_size is not None:
+        local_world_size = int(arg_local_world_size)
         logger.info(f"Using local_world_size from inferencer_args: {local_world_size}")
         if torch.cuda.is_available() and local_world_size != torch.cuda.device_count():
             logger.warning(


examples/distributed/homogeneous_inference.py
```diff
@@ -174,15 +175,31 @@ def _inference_process(
...
+        )
+        torch.cuda.set_device(device)
+    rank = node_rank * local_world_size + local_rank
+    world_size = node_world_size * local_world_size
+    logger.info(
+        f"Local rank {local_rank} in machine {node_rank} has rank {rank}/{world_size} and using device {device} for inference"
+    )
+    torch.distributed.init_process_group(
+        backend="gloo" if device.type == "cpu" else "nccl",
+        init_method=f"tcp://{master_ip_address}:{master_default_process_group_port}",
```
kmontemayor2-sc, line 193: for my own understanding, why do we need to specify init_method here? 

Is it just a preference to do so over setting os.environ?

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -146,7 +146,7 @@ def __init__(
                     f"Initializing process group with master ip address: {master_ip_address}, rank: {rank}, world size: {world_size}, local_rank: {local_rank}, local_world_size: {local_world_size}, device: {device}."
                 )
                 torch.distributed.init_process_group(
-                    backend="gloo", # We just default to gloo for this temporary process group
+                    backend="gloo",  # We just default to gloo for this temporary process group
                     init_method=f"tcp://{master_ip_address}:{DEFAULT_MASTER_INFERENCE_PORT}",
```
kmontemayor2-sc, line 150: sorry I'm not super sure how all these PRs stack together, but why aren't we using an inferred port here?

examples/distributed/homogeneous_inference.py
svij-sc: We could adopt the default init method `env://`  - which uses environ.
Or, set it up directly w/  `init_method = tcp://...`.

The second ^ option is more explicit and doesnt require use off side effects of setitng up OS environs, et al. Thus, it could be prefered in such a setup where we are explicitly computing ip_address, ports, etc.

`env://` based set is usually used by tools like `torchrun` or `vai` setup the context as those are external tooling (not in our code) creating seperate runtime environments for each process.

examples/distributed/homogeneous_inference.py
```diff
@@ -174,15 +175,31 @@ def _inference_process(
...
+        local_process_rank=local_rank,
     )  # The device is automatically inferred based off the local process rank and the available devices
+    if device.type == "cuda":
+        # If using GPU, we set the device to the local process rank's GPU
+        logger.info(
+            f"Using GPU {device} with index {device.index} on local rank: {local_rank} for inference"
+        )
+        torch.cuda.set_device(device)
+    rank = node_rank * local_world_size + local_rank
+    world_size = node_world_size * local_world_size
```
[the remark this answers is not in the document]
svij-sc: I tried this, but it seems like its probably better to leave these one liners as is.
Feels like we are just inflating the repo with not a lot of added value.

```suggestion

   rank = compute_rank(
        node_rank, 
        local_world_size,
        local_rank
    )
    world_size = compute_world_size(
        node_world_size=node_world_size,
        local_world_size=local_world_size
    )

```

examples/distributed/homogeneous_inference.py
```diff
@@ -328,70 +343,87 @@ def _run_example_inference(
...
+        logger.info(
+            f"Detected {local_world_size} GPUs. Thus, setting local_world_size to {local_world_size}"
+        )
+    else:
+        logger.info(
+            "No GPUs detected. Thus, setting local_world_size to `num_inference_processes_per_machine`, and running inference on CPU."
+        )
+        local_world_size = int(
+            inferencer_args.get("num_inference_processes_per_machine", "4")
+        )
```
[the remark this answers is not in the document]
kmontemayor2-sc: this is inference though right, where # processes can be > # gpus.

I agree for training we should throw (if users set something incorrectly, I don't think we should try to fix it for them - maybe they set a machine spec wrong, for instance).

> Should I rename / remove this var completely?
Are we talking about `local_world_size` here? If by that you mean always setting num procs to be num gpus then I guess that's fine but what do we do in the CPU case?

examples/distributed/homogeneous_inference.py
kmontemayor2-sc: ah sg! thanks for the explanation :)

examples/distributed/homogeneous_inference.py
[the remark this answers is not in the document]
kmontemayor2-sc: Yeah that makes sense.

WDYT about something like `get_rank_info() -> tuple[int, int, int, int]` where they represent machine rank, num machines, process num, num process, or something? 

Or maybe that's still too confusing...

Regardless this is fine for the moment )

kmontemayor2-sc (approved):

PR #110, opened 2025-06-24
Add util function to strip label edges

kmontemayor2-sc:
As we talked about in https://github.com/Snapchat/GiGL/pull/105 adding a util to strip out all label edge types from a PyG Data object.

Calling this in both neighbor loaders too.

### DIFF
--- a/python/gigl/distributed/distributed_neighborloader.py
+++ b/python/gigl/distributed/distributed_neighborloader.py
@@ -2,8 +2,10 @@ from collections import abc
 from typing import Dict, List, Optional, Tuple, Union
 
 import torch
+from graphlearn_torch.channel import SampleMessage
 from graphlearn_torch.distributed import DistLoader, MpDistSamplingWorkerOptions
 from graphlearn_torch.sampler import NodeSamplerInput, SamplingConfig, SamplingType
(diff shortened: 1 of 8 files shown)


python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -175,6 +176,25 @@ def __init__(
...
+                    num_neighbors = patch_fanout_for_sampling(
+                        dataset.get_edge_types(), num_neighbors
+                    )
+                else:
+                    raise ValueError(
+                        f"For heterogeneous datasets, input_nodes must be a tuple of (node_type, node_ids) OR if it is a labeled homogeneous dataset, input_nodes may be a torch.Tensor. Received node types: {dataset.node_ids.keys()}"
+                    )
+            else:
+                node_type = None
+
```
mkolodner-sc, line 197: Is this supposed to be here? I don't recall this from the previous `strip label edges` PR

mkolodner-sc (approved):

python/gigl/distributed/distributed_neighborloader.py
kmontemayor2-sc: oops snuck in through merge, good catch!

PR #112, opened 2025-06-24
Add InfiniteIterator Utility

mkolodner-sc:
**Scope of work done**

- Updates the import path for the infinite iterator utility, which allows us to iterate over a iterator indefinitely, and adds a test for this


python/gigl/utils/loader.py
```diff
@@ -0,0 +1,27 @@
+import collections
+
+from gigl.common.logger import Logger
+
+logger = Logger()
+
+
+class InfiniteIterator(collections.abc.Iterator):
```
nshah-sc, line 8: is this just https://github.com/Snapchat/GiGL/blob/62a2a7109743a464e626c4692d00ba0168167688/python/gigl/src/training/v1/lib/data_loaders/tf_records_iterable_dataset.py#L85?  IIUC this is the initial implementation for doing something like this

python/gigl/utils/loader.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __init__(self, loader: collections.abc.Iterator):
+        self.loader = loader
+        # We don't expect the iter to be called outside of the class
+        self._iter = iter(loader)
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
```
kmontemayor2-sc, line 21: ditto on type annotations for these too

python/gigl/utils/loader.py
```diff
@@ -0,0 +1,27 @@
...
+
+logger = Logger()
+
+
+class InfiniteIterator(collections.abc.Iterator):
+    """
+    A wrapper around iterators (objects with __iter__ and __next__ methods) that loop indefinitely over the data.
+    """
+
+    def __init__(self, loader: collections.abc.Iterator):
```
kmontemayor2-sc, line 13: nit. name this `iterator` or something, this isn't restricted to just dataloaders right?

python/gigl/utils/loader.py
kmontemayor2-sc, line 1: In the spirit of "not just limited to data loaders" should we move this to gig.utils or something? It can work with any iterator.

python/gigl/utils/loader.py
mkolodner-sc: Ah thanks for the pointer, I had thought this was hard-coded for the V1 GiGL use-case but seems like it could work in this case as well -- I'll opt to use that one then but keep the unit test we've added to make sure it works as expected.

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -399,6 +430,27 @@ def test_distributed_neighbor_loader(self):
             args=(dataset, self._context, expected_data_count),
         )
 
+    def test_infinite_distributed_neighbor_loader(self):
+        master_port = glt.utils.get_free_port(self._master_ip_address)
+        max_num_batches = 10000
```
kmontemayor2-sc, line 435: can we make this like, `num_node_ids * 2` or something?

python/gigl/utils/loader.py
mkolodner-sc: After trying to join the InfiniteIterator and LoopyIterableDataset implementations, it seems that there are some differences in expectation as Kyle mentioned, with the LoopyIterableDataset taking in a `torch.utils.data.IterableDataset` and the InfiniteIterator being more general. 

> I'd prefer if we could add the new class, and tag the existing thing as deprecated with a note to migrate to the new thing.

As a result, I've opted to go with this approach for now.

> Or do we know if there are perf differences between the two versions?

Since we aren't using itertools.cycle, I don't believe there are any performance differences in this case.

python/gigl/src/training/v1/lib/data_loaders/tf_records_iterable_dataset.py
```diff
@@ -86,6 +86,8 @@ class LoopyIterableDataset(torch.utils.data.IterableDataset, Generic[T]):
     """
     Takes as input an IterableDataset and makes it "loopy," so that the dataset
```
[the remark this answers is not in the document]
svij-sc: Btw, one practice we could adopt here is logging a warning inside the constructor and pointing them to the right direction.
This comment may not be read but the warning might get more visibility.

PR #115, opened 2025-06-26
wip to allow dict inputs

kmontemayor2-sc:
**Scope of work done**

### DIFF
--- a/python/gigl/distributed/dist_ablp_neighborloader.py
+++ b/python/gigl/distributed/dist_ablp_neighborloader.py
@@ -24,6 +24,8 @@ from gigl.distributed.dist_sampling_producer import DistSamplingProducer
 from gigl.distributed.distributed_neighborloader import DEFAULT_NUM_CPU_THREADS
 from gigl.distributed.sampler import ABLPNodeSamplerInput
 from gigl.distributed.utils.neighborloader import (
+    NodeSamplerInput,
+    infer_node_sampler_input_from_user_input,
     labeled_to_homogeneous,
     patch_fanout_for_sampling,
     shard_nodes_by_process,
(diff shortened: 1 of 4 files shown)


python/gigl/distributed/utils/neighborloader.py
```diff
@@ -118,3 +121,88 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
         del data.num_sampled_edges[edge_type]
 
     return data
+
+
+@dataclass(frozen=True)
+class _ResolvedNodeSamplerInput:
```
mkolodner-sc, line 127: I'd prefer not to add another`NodeSamplerInput` if we can avoid it for the sake of reducing the complexity of our codebase -- is the value here primarily the `is_labeled_homogeneous` field? Is there any way we can just have `resolve_node_sampler_input_from_user_input` return a Tuple[NodeSamplerInput, bool], since it seems like the only place this is used in in the `__init__` of the ABLPLoader?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -166,50 +161,41 @@ def __init__(
...
+            anchor_node_type is None
+            or anchor_node_type == DEFAULT_HOMOGENEOUS_NODE_TYPE
+        ):
             if supervision_edge_type is not None:
                 raise ValueError(
                     f"Expected supervision edge type to be None for homogeneous input nodes, got {supervision_edge_type}"
                 )
-            anchor_node_ids = input_nodes
-            anchor_node_type = DEFAULT_HOMOGENEOUS_NODE_TYPE
+            node_type = DEFAULT_HOMOGENEOUS_NODE_TYPE
```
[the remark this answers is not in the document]
kmontemayor2-sc: Should be gone now :)

python/gigl/distributed/utils/neighborloader.py
kmontemayor2-sc: Yeah I wanted to do this to avoid the three sized tuple. 

I understand the apprehension here about the new dataclass though - do you think renaming it could help? Something like `_ParsedInputs` or equivalent?

mkolodner-sc (approved):
Thanks! LGTM provided comments are addressed

PR #116, opened 2025-06-26
Add documentation for y_positive, y_negative

kmontemayor2-sc:
As preparation for our launches we should document the fields we're adding in our ABLP loader.

This change only updates the documentation.

I chose explicitly to not further document num_sampled_nodes and num_sampled_edges as they're GLT details and I'm not sure we should be tying ourself to them, and those fields aren't super relevant to users, whereas the label fields are super important.

### DIFF
--- a/python/gigl/distributed/dist_ablp_neighborloader.py
+++ b/python/gigl/distributed/dist_ablp_neighborloader.py
@@ -89,6 +89,11 @@ class DistABLPLoader(DistLoader):
             - (optional) `y_negative`: A dict[torch.Tensor] mapping from local anchor node id to a tensor of negative
                 label node ids. This will only be present if the supervision edge type has negative labels.
 
+        NOTE: for both y_positive, and y_negative, the values represented in both the key and value of the dicts are
+        the *local* node ids of the sampled nodes, not the global node ids.
+        In order to get the global node ids, you can use the `node` field of the Data/HeteroData object.
+        e.g. global_positive_node_id_labels = data.node[data.y_positive[local_anchor_node_id]].
+
         The underlying graph engine may also add the following fields to the output Data object:
             - num_sampled_nodes: If heterogeneous. a dictionary mapping from node type to the number of sampled nodes for that type, by hop.
             if homogeneous, a tensor the number of sampled nodes, by hop.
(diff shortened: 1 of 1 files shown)


python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -83,7 +83,34 @@ def __init__(
...
         Otherwise, the output will be a torch_geometric.data.HeteroData object.
 
-           Args:
+        The output of this loader will contain the additional following fields:
+            - `y_positive`: A dict[int, torch.Tensor] mapping from local anchor node id to a tensor of positive
+                label node ids.
+            - (optional) `y_negative`: A dict[torch.Tensor] mapping from local anchor node id to a tensor of negative
+                label node ids. This will only be present if the supervision edge type has negative labels.
+
+        The underlying graph engine may also add the following fields to the output Data object:
```
svij-sc, line 92: Nit: Suggestions:
```python
"""
        ....
        ...
        In **heterogeneous** graphs, `input_nodes` must be a `(node_type, node_ids)` tuple, and supervision edge types
        must be specified. In **homogeneous** graphs, `input_nodes` can be a `Tensor`, and supervision edge types will
        be inferred automatically from the dataset.


        Each sample returned from the loader is either a {py:class}`torch_geometric.data.Data` (homogeneous) or
       {py:class} `torch_geometric.data.HeteroData` (heterogeneous) object. The following fields may also be present:
        - `y_positive`: `Dict[int, torch.Tensor]`  ....
        - `y_negative`: (Optional) `Dict[int, torch.Tensor]`  ....
        - `num_sampled_nodes`: Per-hop sample counts. A tensor (homogeneous) or a `Dict[str, Tensor]` (heterogeneous) .....
"""
```

I love that we are improving this doc, small opportunity to add types and markdown formatting that sphinx will pick up and render on the website.

svij-sc (approved):
Thanks

python/gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Updated, thanks

PR #117, opened 2025-06-26
try to swap path to relative

kmontemayor2-sc:
**Scope of work done**

Where is the documentation for this feature?: N/A

Did you add automated tests or write a test plan?

***Updated Changelog.md?*** NO

***Ready for code review?:*** NO

PR #119, opened 2025-06-27
Add notebook for CORA pipeline

kmontemayor2-sc:
As part of our documentation work I figured we can add this now while we are working on the example training loop.
The actual notebook is *heavily* based on https://github.com/Snapchat/GiGL/blob/main/examples/MAG240M/mag240m.ipynb but we removed SGS and split gen.

Note the changes in `python/gigl/common/constants.py` and `python/gigl/src/mocking/dataset_asset_mocking_suite.py` are a part of https://github.com/Snapchat/GiGL/pull/117

cc: @mkolodner-sc who is working on the training loop.

This all runs locally :)


examples/distributed/configs/example_resource_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+  edge_preprocessor_config:
+    num_workers: 1
+    max_num_workers: 128
+    machine_type: "n2d-highmem-32"
+    disk_size_gb: 300
+  node_preprocessor_config:
+    num_workers: 1
+    max_num_workers: 128
+    machine_type: "n2d-highmem-64"
+    disk_size_gb: 300
```
svij-sc, line 39: i wonder if we should lessen # of machines here.
We probbaly dont need these many for cora esp for an example - seems a bit scary 🤣

python/gigl/common/constants.py
```diff
@@ -18,6 +18,10 @@
     GIGL_ROOT_DIR, "dep_vars.env"
 ).absolute()
 
+PYTHON_ROOT_DIR: Final[Path] = (
+    Path(__file__).resolve().parent.parent.parent
+)  # common -> gigl -> python
+
```
svij-sc, line 24: `PYTHON_ROOT_DIR` is already defined above.

examples/distributed/README.md
```diff
@@ -0,0 +1,16 @@
+# Examples for Training and Inference using live subgraph sampling
```
svij-sc, line 1: lets also add a link to this README in our docs: https://github.com/Snapchat/GiGL/blob/main/docs/user_guide/examples/index.md

examples/distributed/cora.ipynb
```diff
@@ -0,0 +1,307 @@
...
+    "\n",
+    "from gigl.common import LocalUri\n",
+    "from gigl.env.pipelines_config import get_resource_config\n",
+    "from gigl.src.common.types.pb_wrappers.gigl_resource_config import GiglResourceConfigWrapper\n",
+    "from gigl.src.common.types.pb_wrappers.gbml_config import GbmlConfigPbWrapper\n",
+    "\n",
+    "curr_datetime = datetime.datetime.now().strftime(\"%Y%m%d_%H%M%S\")\n",
+    "\n",
+    "# Firstly, let's give your job a name and ensure that the resource and task configs exist and can be loaded\n",
+    "JOB_NAME = f\"{getpass.getuser()}_cora_{curr_datetime}\"\n",
```
svij-sc, line 53: nit: job name includes `gigl_cora`, or `test_cora`, etc.

examples/distributed/README.md
svij-sc: Also, doesnt need to happen now, but maybe we can instead of calling the folder `distributed` 
just call it `in_memory_subgraphsampling`.
The other examples are also distributed i.e. MAG240M, etc. so its a big confusing.

It woould be nice for all the examples to be in their own dir.
i.e. homogeneous training/inference w/ cora is in one folder vs dblp hetrogeneous is in another folder

examples/distributed/cora.ipynb
```diff
@@ -0,0 +1,307 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e80e910c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "runner.run_trainer(pipeline_config=pipeline_config)"
```
svij-sc, line 239: maybe we could showcase that a model was built and uploaded to the URI specified in the frozen config?

svij-sc: We can probably add some guidance as to what parts of the pipeline config does the trainer read from and what the input/output locations and how they are determined.

examples/distributed/cora.ipynb
```diff
@@ -0,0 +1,307 @@
...
+   "id": "b29d52b7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Looking at inference results\n",
+    "from gigl.src.common.utils.bq import BqUtils\n",
+    "from gigl.src.inference.lib.assets import InferenceAssets\n",
+    "\n",
+    "\n",
+    "bq_emb_out_table = InferenceAssets.get_enumerated_embedding_table_path(gbml_config_pb_wrapper=frozen_config, node_type=\"paper\")\n",
```
svij-sc, line 275: enumerated table is an intermediary asset.
We should probably print
`sharedConfig.inferenceMetadata.embeddingsBqPath` right?
Or is this being populated in the postProcessor? 

Probably also adding a note somewhere on why we enumerate node_ids --> sequential_int_ids

svij-sc (commented):
Thanks for the work!
Excited for the docs to improve 😄

examples/distributed/cora.ipynb
kmontemayor2-sc: Hmm, added some stuff here, LMK what you think :)

examples/distributed/README.md
svij-sc: waiting for renaming sg.
cc @mkolodner-sc , since I left a similar comment re: renaming on his PRs too.

svij-sc: > Personally I kind of prefer having one folder with both heterogeneous and homogeneous live subgraph sampling but we can discuss that later.
ye sg to discuss later. We may have multiple examples in the future i.e. tasks: node classification, NABLP, Link Prediction; graph type: heterogeenous, homogeneous ; data size: cora, MAG240 ; distributed modality: single machine, DDP setup.

examples/distributed/cora.ipynb
svij-sc: Preprocessor outputs the map to BQ and stores it here in the preprocess metadata: https://github.com/Snapchat/GiGL/blob/924522caef86fd07d07e39c770d87376c7fdedbe/proto/snapchat/research/gbml/preprocessed_metadata.proto#L19

Subsequently, postprocessor does the unenumeration in the in-mememory SGS setting: https://github.com/Snapchat/GiGL/blob/924522caef86fd07d07e39c770d87376c7fdedbe/python/gigl/src/post_process/post_processor.py#L127-L134

Thus, maybe worth discussing this + showcasing the output results from after running the post processor?

examples/distributed/cora.ipynb
```diff
@@ -0,0 +1,441 @@
+{
```
svij-sc, line 1: actually, one more thing - this popped up in initial reviews w/ @yliu2-sc , and @nshah-sc .
Could you add a self link of this notebook at the top of the notebook . This is so when it gets rendered on the website, folks have an easy way to navigate directly to the notebook.

I need to do the same for the other notebooks as well.
"""
Latest version of this notebook can be found on [github](https://github.com/Snapchat/GiGL/blob/main/examples/distributed/cora.ipynb) 
"""

svij-sc: Example: https://github.com/Snapchat/GiGL/pull/134

examples/distributed/cora.ipynb
```diff
@@ -0,0 +1,441 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9b433f48",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from gigl.src.common.utils.file_loader import FileLoader\n",
+    "frozen_config_uri = runner.run_config_populator(pipeline_config=pipeline_config)\n",
```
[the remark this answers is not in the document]
kmontemayor2-sc: I intentionally didn't copy that over - I left some note at the top of this file to look at the toy graph example for some more details about the components, protos, etc.

if you really want this in this file, I can add it as a follow up, since `show_task_config_colored_unified_diff` should get moved to the `jupyter_magics` module.

examples/distributed/cora.ipynb
```diff
@@ -0,0 +1,441 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "13cb217c",
+   "metadata": {},
+   "source": [
+    "## Post Processor\n",
+    "\n",
+    "The inferencer outputs embeddings for *enumerated* node ids, e.g. the node ids that the preprocessor outputs. The preprocessor stores a mappings between unenumerated node ids and enumnerated node ids in `PreprocessedMetadata.condensed_node_type_to_preprocessed_metadata[node_type].enumerated_node_ids_bq_table` [1]\n",
+    "\n",
+    "The postprocessor unemuerates the embeddings and stores them in `PreprocessedMetadata.condensed_node_type_to_preprocessed_metadata[node_type].enumerated_node_ids_bq_table`\n",
```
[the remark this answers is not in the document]
kmontemayor2-sc: I'm not sure for this case "frozen_config" isn't the proto and I was hoping to actually make this the proto path.

We have the actual code later that I'd expect people to run/copy past whereas this is more just like "indicative"

examples/distributed/cora.ipynb
kmontemayor2-sc: Sure will do, will wait for https://github.com/Snapchat/GiGL/pull/133 to get merged in so I don't need to rename paths (again).

PR #120, opened 2025-06-27
Add Port as parameter to `build_dataset`

mkolodner-sc:
**Scope of work done**

- Adds back the port as a parameter to the `build_dataset` utility

### DIFF
--- a/python/gigl/distributed/dataset_factory.py
+++ b/python/gigl/distributed/dataset_factory.py
@@ -314,13 +314,15 @@ def build_dataset(
     serialized_graph_metadata: SerializedGraphMetadata,
     sample_edge_direction: Union[Literal["in", "out"], str],
     distributed_context: Optional[DistributedContext] = None,
-    dataset_building_port: Optional[int] = None,
     should_load_tensors_in_parallel: bool = True,
     partitioner_class: Optional[Type[DistPartitioner]] = None,
     node_tf_dataset_options: TFDatasetOptions = TFDatasetOptions(),
     edge_tf_dataset_options: TFDatasetOptions = TFDatasetOptions(),
     splitter: Optional[NodeAnchorLinkSplitter] = None,
     _ssl_positive_label_percentage: Optional[float] = None,
+    _dataset_building_port: Optional[
+        int
+    ] = None,  # WARNING: This field will be deprecated in the future
 ) -> DistLinkPredictionDataset:
     """
     Launches a spawned process for building and returning a DistLinkPredictionDataset instance provided some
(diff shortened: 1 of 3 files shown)


python/gigl/distributed/dataset_factory.py
```diff
@@ -246,7 +246,7 @@ def _build_dataset_process(
             will be written to for use by the parent process
         serialized_graph_metadata (SerializedGraphMetadata): Metadata about TFRecords that are serialized to disk
         master_ip_address (str): IP address of the master node
-        master_dataset_building_port (int): Free port on the master node to use to build the dataset
+        master_dataset_building_ports (Tuple[int, int]): Free ports on the master node to use to build the dataset, the first port is used for partitioning and the second is used for splitting
```
kmontemayor2-sc, line 249: this is just `Optional[int]` right?

python/gigl/distributed/dataset_factory.py
```diff
@@ -314,6 +314,7 @@ def build_dataset(
     serialized_graph_metadata: SerializedGraphMetadata,
     sample_edge_direction: Union[Literal["in", "out"], str],
     distributed_context: Optional[DistributedContext] = None,
+    dataset_building_port: Optional[int] = None,
```
svij-sc, line 317: can we add this as last argument and mark the field as deprecated 
```suggestion
    _dataset_building_ports: Tuple[int, int] = None, # WARNING: This field will be deprecated in the future
```

svij-sc: Also can we move this as last argument?

kmontemayor2-sc (approved):

svij-sc (commented):
Thanks for the quick work.
Can we add a test case where we specify the ports?

We can obv pass in a port that is inferred from our new get free port util.

python/gigl/distributed/dataset_factory.py
mkolodner-sc: For `_build_dataset_process`, it will be a Tuple[int, int]. This was incorrectly typed from the previous PR.

python/gigl/distributed/dataset_factory.py
```diff
@@ -314,6 +314,7 @@ def build_dataset(
     serialized_graph_metadata: SerializedGraphMetadata,
```
svij-sc, line 314: Is this the only function being used and not `build_dataset_from_task_config_uri` ?

mkolodner-sc: For the use cases that will want to specify the ports manually, only `build_dataset` is being used. `build_dataset_from_task_config_uri` can just use the automatic port-inferring utilities or use the default ports provided.

PR #122, opened 2025-06-27
Allow DistABLPLoader to auto-infer ports

kmontemayor2-sc:
Follow up of https://github.com/Snapchat/GiGL/pull/96.

This is a step towards removing DistContext alltogether, and also makes it so we can auto-infer ports which is safer and less error-prone.

I did not update DistABLPLoader in a backwards compatible way - as we don't have any users atm so we don't really need to support that path, though it's easy enough to add in the future if we have asks in that sense.

Swapped DistABLPLoader usages in tests to auto-infer ports.

### DIFF
--- a/python/gigl/distributed/dist_ablp_neighborloader.py
+++ b/python/gigl/distributed/dist_ablp_neighborloader.py
@@ -328,8 +328,8 @@ class DistABLPLoader(DistLoader):
 
         curr_process_nodes = shard_nodes_by_process(
             input_nodes=anchor_node_ids,
-            local_process_rank=local_process_rank,
-            local_process_world_size=local_process_world_size,
+            local_process_rank=local_rank,
+            local_process_world_size=local_world_size,
         )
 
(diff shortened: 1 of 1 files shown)


python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -132,6 +126,84 @@ def __init__(
...
+        local_rank: int
+        local_world_size: int
+
+        master_ip_address: str
+        should_cleanup_distributed_context: bool = False
+        device = (
+            pin_memory_device
+            if pin_memory_device
+            else gigl.distributed.utils.get_available_device(
+                local_process_rank=local_rank
```
svij-sc, line 142: just realized local_rank is not assigned here

svij-sc: A bug with `DistNeighborLoader` too :S 

But it looksl ike device assignemnt can just move post the `if/else` conditional. Maybe some log statement changes

kmontemayor2-sc: Good catch, fixed!

PR #124, opened 2025-06-27
Enable Homogeneous Training in E2E

mkolodner-sc:
**Scope of work done**


examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,888 @@
...
+            random_negative_data.batch_size
+        ).to(device)
+        positive_ids: torch.Tensor = torch.cat(list(main_data.y_positive.values())).to(
+            device
+        )
+
+        # We also extract a repeated query node id tensor which upsamples each query node based on the number of positives it has
+        repeated_query_node_ids = query_node_ids.repeat_interleave(
+            torch.tensor([len(v) for v in main_data.y_positive.values()]).to(device)
+        )
```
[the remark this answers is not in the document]
mkolodner-sc: Agreed -- once these examples are finished, I have plans to update how we return the `y_positive` and `y_negative` labels to be the single padded tensor instead of the dict[root_node_id, labels] we have currently.

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,888 @@
...
+            device=training_device,
+        ),
+        device_ids=[training_device],
+    )
+
+    learning_rate = float(trainer_args.get("learning_rate", "0.0005"))
+    weight_decay = float(trainer_args.get("weight_decay", "0.0005"))
+    num_max_train_batches = int(trainer_args.get("num_max_train_batches", "1000"))
+    num_val_batches = int(trainer_args.get("num_val_batches", "100"))
+    val_every_n_batch = int(trainer_args.get("val_every_n_batch", "50"))
```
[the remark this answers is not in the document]
mkolodner-sc: These arguments aren't used for `testing_process` (specific to `training_process`) and I'd prefer to keep the number of arguments to `training_process` shorter if possible.

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,888 @@
...
+        if num_batches and batch_idx >= num_batches:
+            # If num_batches is set, we stop the validation loop after processing num_batches batches. This is not expected to be used for _testing_process.
+            break
+        try:
+            main_data = next(main_loader)
+            random_data = next(random_negative_loader)
+        except StopIteration:
+            # If the test data loader is exhausted, we stop the test loop.
+            # Note that validation dataloaders are infinite, so this is not expected to happen during training.
+            break
```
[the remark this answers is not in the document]
mkolodner-sc: One concern I have about doing `for data, random in zip(loaders)` is that we won't be able to resume our progress for validation between multiple validation loops. For example, if I call validation on batches 0-99, then break, I'd want the next validation to be on 100-199. The above solution handles this scenario.

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,888 @@
...
+    for main_data, random_data in zip(train_main_loader, train_random_negative_loader):
+        if (
+            num_max_train_batches_per_process
+            and batch_idx >= num_max_train_batches_per_process
+        ):
+            logger.info(
+                f"num_max_train_batches_per_process={num_max_train_batches_per_process} reached, "
+                f"stopping training on machine {node_rank} local rank {local_rank}"
+            )
+            break
```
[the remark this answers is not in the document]
mkolodner-sc: We have a `barrier` after the loop in case there are any other conditions in which we want to break in the future

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,888 @@
...
+        model (DistributedDataParallel): DDP-wrapped torch model for training and testing
+        main_data (Union[Data, HeteroData]): The batch of data containing query nodes, positive nodes, and hard negative nodes
+        random_negative_data (Union[Data, HeteroData]): The batch of data containing random negative nodes
+        loss_cls (RetrievalLoss): Initialized class to use for loss calculation
+        use_automatic_mixed_precision (bool): Whether we should use automatic mixed precision with training
+        device (torch.device): Device for training or validation
+    Returns:
+        torch.Tensor: Final loss for the current batch on the current process
+    """
+    with torch.autocast("cuda", enabled=use_automatic_mixed_precision):
```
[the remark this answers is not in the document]
mkolodner-sc: Needs to be investigated in the follow-up, thanks!

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,886 @@
...
+
+trainerConfig:
+  trainerArgs:
+    # Example argument to trainer
+    log_every_n_batch: "50"
+  command: python -m examples.distributed.homogeneous_training
+featureFlags:
+  should_run_glt_backend: 'True'
+
+You can run this example in a full pipeline with `make run_cora_glt_udl_kfp_test` from GiGL root.
```
kmontemayor2-sc, line 16: btw I think it'd be useful to provide how to run this locally, e.g. `RANK=0 WORLD_SIZE=1 MASTER_ADDR=localhost MASTER_PORT=5435 python -m examples.distributed.homogeneous_training --task_config_uri=...` but maybe best to leave that for when we have CPU support :)

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,886 @@
...
+
+    # We add a barrier to wait for all processes to finish preparing the dataloader prior to the start of training
+    torch.distributed.barrier()
+
+    # Entering the training loop
+    training_start_time = time.time()
+    batch_idx = 0
+    avg_train_loss = 0.0
+    last_n_batch_avg_loss = []
+    last_n_batch_time = []
```
kmontemayor2-sc, line 355: nit. add types here?

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,886 @@
...
+    # We explicitly delete all the dataloaders to reduce their memory footprint. Otherwise, experimentally we have
+    # observed that not all memory may be cleaned up, leading to OOM.
+    del test_main_loader, test_random_negative_loader
+
+
+def _run_example_training(
+    task_config_uri: str,
+):
+    """
+    Runs an example training pipeline using GiGL Orchestration.
```
kmontemayor2-sc, line 712:
```suggestion
    Runs an example training loop using GiGL Orchestration.
```

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,886 @@
...
+    master_process_group_port_for_training = master_default_process_group_ports[0]
+    master_process_group_port_for_testing = master_default_process_group_ports[1]
+    # Destroying the process group as one will be re-initialized in the training process using above information
+    torch.distributed.destroy_process_group()
+
+    logger.info(f"--- Launching data loading process ---")
+    dataset = build_dataset_from_task_config_uri(
+        task_config_uri=task_config_uri,
+        is_inference=False,
+        _tfrecord_uri_pattern=".*tfrecord",
```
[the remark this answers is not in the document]
mkolodner-sc: good catch, thanks!

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
+"""
+This file contains an example for how to run homogeneous training using live subgraph sampling power by GraphLearn-for-PyTorch (GLT).
```
kmontemayor2-sc, line 2:
```suggestion
This file contains an example for how to run homogeneous training using live subgraph sampling powered by GraphLearn-for-PyTorch (GLT).
```

oops sorry.

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+            random_negative_loader, InfiniteIterator
+        ):
+            raise ValueError(
+                "Must set `num_batches` field when the provided data loaders are wrapped with InfiniteIterator"
+            )
+
+    model.eval()
+    batch_idx = 0
+    batch_losses: list[float] = []
+    last_n_batch_time = []
```
kmontemayor2-sc, line 496:
```suggestion
    last_n_batch_time: list[float] = []
```

nit.

examples/distributed/homogeneous_inference.py
```diff
@@ -273,9 +212,7 @@ def _inference_process(
 
         # These arguments to forward are specific to the GiGL LinkPredictionGNN model.
         # If just using a nn.Module, you can just use output = model(data)
-        output = model(
-            data=data, output_node_types=[inference_node_type], device=device
-        )[inference_node_type]
+        output = model(data=data, device=device)
```
kmontemayor2-sc, line 215: btw why can we get rid of the `[inference_node_type]` here?

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+        f"--- Data loading process finished, took {time.time() - start_time:.3f} seconds"
+    )
+    gbml_config_pb_wrapper = GbmlConfigPbWrapper.get_gbml_config_pb_wrapper_from_uri(
+        gbml_config_uri=UriFactory.create_uri(task_config_uri)
+    )
+
+    trainer_args = dict(gbml_config_pb_wrapper.trainer_config.trainer_args)
+
+    local_world_size = int(trainer_args.get("local_world_size", "1"))
+    if torch.cuda.is_available() and torch.cuda.device_count() > 0:
```
kmontemayor2-sc, line 751: btw what's the reason for the second check here? What if it *is* zero (for some reason), we'd probably want to throw then too right?

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+    parser.add_argument("--task_config_uri", type=str, help="Gbml config uri")
+
+    # We use parse_known_args instead of parse_args since we only need job_name and task_config_uri for distributed trainer
+    args, unused_args = parser.parse_known_args()
+    logger.info(f"Unused arguments: {unused_args}")
+
+    # We only need `task_config_uri` for running trainer
+    _run_example_training(
+        task_config_uri=args.task_config_uri,
+    )
```
kmontemayor2-sc, line 898:
```suggestion
    )

```

Does our linter not pick up these files? I can look into that...

kmontemayor2-sc (approved):
Thanks matt!

examples/distributed/homogeneous_training.py
mkolodner-sc: Good call, the `torch.cuda.is_available` seems to check if the device_count is larger than 0 as its source of truth, so this second check is redundant.

examples/distributed/configs/e2e_cora_udl_glt_task_config.yaml
```diff
@@ -20,13 +25,8 @@ inferencerConfig:
 sharedConfig:
   shouldSkipAutomaticTempAssetCleanup: false
   shouldSkipInference: false
-  shouldSkipTraining: true  # Live subgraph sampling pipelines do not have a training phase that is being tested right now
+  # Model Evaluation is only applicable for offline SGS pipelines -- we do not need it for live SGS pipelines
```
svij-sc, line 28: nit: I am guessing it is important but we just dont provide support for it right now?

examples/models.py
```diff
@@ -0,0 +1,55 @@
+from typing import Optional
```
svij-sc, line 1: For future:
Could we just co-locate this file with where its being used?
Id imagine this cant be used for anything but CORA so maybe we can collocate all the cora training and inference loops under `examples/cora_link_pred` instead of  `examples/distributed` - all examples are distributed.

examples/models.py
```diff
@@ -0,0 +1,55 @@
...
+    Args:
+        node_feature_dim (int): Input node feature dimension for the model
+        edge_feature_dim (int): Input edge feature dimension for the model
+        args (Dict[str, str]): Arguments for training or inference
+        device (Optional[torch.device]): Torch device of the model, if None defaults to CPU
+        state_dict (Optional[Dict[str, torch.Tensor]]): State dictionary for pretrained model
+    Returns:
+        LinkPredictionGNN: Link Prediction model for training or inference
+    """
+    encoder_model = GraphSAGE(
```
svij-sc, line 30: Can we add a comment on why we cant use PyG's default `GraphSage` implementation?
Actually, is this still true?

python/gigl/common/utils/torch_training.py
```diff
@@ -80,3 +80,18 @@ def is_distributed_available_and_initialized() -> bool:
...
+    Args:
+        metric (torch.Tensor): The metric, expressed as a torch Tensor, which should be synced across multiple processes
+    Returns:
+        float: The average of the provided metric across all training processes
+    """
+    assert is_distributed_available_and_initialized(), "DDP is not initialized"
+    # Make a copy of the local loss tensor
+    loss_tensor = metric.detach().clone()
+    torch.distributed.all_reduce(loss_tensor, op=torch.distributed.ReduceOp.SUM)
+    return loss_tensor.item() / torch.distributed.get_world_size()
```
svij-sc, line 97: nit: assuming practitioners have knowledge of DDP, this code did not add value for me as a practitioner - it actually added some confusion because I had to make sure this did all_recude + averaged it. Although a practitioner may implement something like this for themselves, I am not sure if this is a value add of gigl esp because its only being used by examples and not by gigl src.

Maybe remove?

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+    """
+    with torch.autocast("cuda", enabled=use_automatic_mixed_precision):
+        # Forward pass through encoder
+        main_embeddings = model(data=main_data, device=device)
+        random_negative_embeddings = model(data=random_negative_data, device=device)
+
+        # Extracting local query, random negative, positive, hard_negative, and random_negative ids.
+        # Local in this case refers to the local index in the batch, while global subsequently refers to the node's unique global ID across all nodes in the dataset.
+        # Global ids are stored in data.node, ex. `data.node = [50, 20, 10]` where the `0` is the local id for global id `50`
+        query_node_ids: torch.Tensor = torch.arange(main_data.batch_size).to(device)
```
svij-sc, line 81: can the same node be queries multiple times?
i.e. could we have `data.node = [50, 20, 20]`

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+        random_negative_ids: torch.Tensor = torch.arange(
+            random_negative_data.batch_size
+        ).to(device)
```
svij-sc, line 84: If I am understanding this correctnly, if we have `[50, 20, 10]` --> `[0, 1, 2]` in main then random negative `[20, 100, 30]` -> `[0, 1, 2]` ?
 
 Should the ids across random negatives and query nodes not be the same?

examples/distributed/homogeneous_training.py
mkolodner-sc: Nope, each is unique -- in a follow-up I'll add this detail, thanks!

examples/distributed/homogeneous_training.py
svij-sc: should these be "idx" then?
As in index vs. id ?

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+
+    train_main_loader = InfiniteIterator(train_main_loader)
+    train_random_negative_loader = InfiniteIterator(train_random_negative_loader)
+
+    logger.info(
+        f"---Machine {node_rank} local rank {local_rank} training data loaders initialized"
+    )
+
+    # We add a barrier to wait for all the training loaders to finish initializing before we initialize the val loaders to reduce the peak memory usage.
+    torch.distributed.barrier()
```
svij-sc, line 284: Do we not need a time.sleep here?

Any reason we have a `barrier` here but not in between `train` and `random` loaders?

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+        pin_memory_device=training_device,
+        worker_concurrency=sampling_workers_per_process,
+        channel_size=sampling_worker_shared_channel_size,
+        # Each val_main_loader will wait for `process_start_gap_seconds` * `local_process_rank` seconds before initializing to reduce peak memory usage.
+        process_start_gap_seconds=process_start_gap_seconds,
+    )
+
+    logger.info("Finished setting up val main loader.")
+
+    # The random_negative_loader will wait for `process_start_gap_seconds / 2` seconds after initializing the main_loader so that it doesn't interfere with other processes.
```
svij-sc, line 306: Same comment as above re: process_start_gap_seconds

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+        log_every_n_batch (int): The frequency we should log batch information when training and validating
+    """
+    # TODO (mkolodner-sc): Investigate work needed + add support for CPU training
+    if not torch.cuda.is_available():
+        raise NotImplementedError(
+            "Currently, only GPU training is supported with this example training loop"
+        )
+
+    test_process_world_size = node_world_size * local_world_size
+    test_process_global_rank = node_rank * local_world_size + local_rank
```
svij-sc, line 602: Maybe lets define `world_size` and `rank` once and re-use them instead of redefining it 
`train_process_world_size` and `test_process_world_size` .

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+        loss=torch.nn.CrossEntropyLoss(reduction="mean"),
+        temperature=0.07,
+        remove_accidental_hits=True,
+    )
+    logger.info(
+        f"Model initialized on machine {node_rank} training device {training_device}\n{model.module}"
+    )
+
+    # We add a barrier to wait for all processes to finish preparing the dataloader prior to the start of training
+    torch.distributed.barrier()
```
svij-sc, line 353: if the barrier is for dataloaders, should we move this above DDP setup?
IIRC, DDP itself need not have a barrier to work?

examples/distributed/homogeneous_training.py
yliu2-sc: and also use sth like global world size instead of node world size, since node has other meanings in this code case

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+
+    # We use one training device for each local process
+    training_device = torch.device(f"cuda:{local_rank % torch.cuda.device_count()}")
+    torch.cuda.set_device(training_device)
+    logger.info(
+        f"---Machine {node_rank} local rank {local_rank} training process set device {training_device}"
+    )
+
+    training_process_world_size = node_world_size * local_world_size
+    training_process_global_rank = node_rank * local_world_size + local_rank
```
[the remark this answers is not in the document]
yliu2-sc: and there shouldn't be a difference between train vs test world size & rank

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+    """
+    Runs an example training + testing loop using GiGL Orchestration.
+    Args:
+        task_config_uri (str): Path to YAML-serialized GbmlConfig proto.
+    """
+    start_time = time.time()
+    mp.set_start_method("spawn")
+    logger.info(f"Starting sub process method: {mp.get_start_method()}")
+
+    torch.distributed.init_process_group(backend="gloo")
```
svij-sc, line 723: would be nice to have a comment that this is only for getting `master_ip_address` and port information, et al to help with data loading.

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+    )
+    global_avg_val_loss = sync_metric_across_processes(
+        metric=torch.tensor(local_avg_loss, device=device)
+    )
+    logger.info(
+        f"process_rank={local_rank} got global validation loss {global_avg_val_loss=:.6f}"
+    )
+
+    # Put the model back in train mode
+    model.train()
```
yliu2-sc, line 546: this should be done in the training loop not inside val func

svij-sc (commented):
Thanks so much for the work here. Leaving some comments for further improvement.
cc @yliu2-sc if you could take a look at this to.

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+    return
+
+
+def _testing_process(
```
yliu2-sc, line 550: do we actually have to spin up separate processes for test? Can it just be run after the training loop, since most stuff here are the same to train except for the data loaders, and we use the val loop

mkolodner-sc: The value of having the `_testing_process` is so that if we are doing pipelines that should skip training, we can still run `_testing_process` to generate the offline metrics for monitoring purposes.

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+        )
+
+        # Compute the global candidate ids and concatentate into a single tensor
+
+        global_candidate_ids = torch.cat(
+            (
+                main_data.node[positive_ids],
+                main_data.node[hard_negative_ids],
+                random_negative_data.node[random_negative_ids],
+            )
```
svij-sc, line 131: Does this serve the same purpose as `n_id` in pyg?
https://pytorch-geometric.readthedocs.io/en/2.5.2/modules/loader.html#torch_geometric.loader.NeighborLoader

If so, it might be a source of confusion due to the name changing.

cc @yozenliu

examples/distributed/homogeneous_training.py
svij-sc: @yozenliu hmmm... thats a good point, alas we are using node_world_size elsewhere too.
Maybe we should  be using `machine_world_size`, and `machine_rank`?

examples/distributed/homogeneous_training.py
svij-sc: Reducing side effects that happen inside functions is always cleaner.
And, calling model.train() inside that function is a function side effect.
Like @yozenliu is saying,  the main training loop should do all  side effects i.e. `model.eval()` and `model.train()`.

examples/models.py
mkolodner-sc: Can add a comment. While this is possible to use now, there will be a large loss in modeling levers if we use the PyG default GraphSAGE implementation, which doesn't support edge features [[1](https://pytorch-geometric.readthedocs.io/en/2.4.0/_modules/torch_geometric/nn/models/basic_gnn.html#GraphSAGE)] and is missing utilities like `should_l2_normalize_embedding_layer_output`. In terms of what is most useful for customers as a result, IMO it would be better to use the GiGL implementation.

examples/distributed/homogeneous_training.py
mkolodner-sc: Sounds good to change the name to `machine_world_size` and `machine_rank`. 

One note is that we cannot define `rank` outside of the `_training_process` and `_testing_process` since they use `local_rank`, which is specific to each process on that machine and is the first argument after calling `mp.spawn`. This means we must at least pass in `machine_rank`, `local_world_size`, and `local_rank`. 

This means the only argument that can be computed outside of these processes is the `world_size = machine_world_size * local_world_size`. However, I feel like its a bit cleaner to pass in `local_rank, local_world_size, machine_rank, machine_world_size` than `local_rank, local_world_size, machine_rank, world_size` from a consistency perspective.

examples/distributed/homogeneous_training.py
yliu2-sc, line 723: why is it gloo here where everywhere else is nccl?

examples/distributed/homogeneous_training.py
```diff
@@ -0,0 +1,898 @@
...
+        "sampling_worker_shared_channel_size", "4GB"
+    )
+
+    process_start_gap_seconds = int(trainer_args.get("process_start_gap_seconds", "0"))
+    use_automatic_mixed_precision = bool(
+        strtobool(trainer_args.get("use_automatic_mixed_precision", "False"))
+    )
+    log_every_n_batch = int(trainer_args.get("log_every_n_batch", "25"))
+
+    learning_rate = float(trainer_args.get("learning_rate", "0.0005"))
```
yliu2-sc, line 798: nit: can we have all args parsing at the top of the function

examples/distributed/homogeneous_training.py
mkolodner-sc: Added a comment
```
We can use `gloo` here since these fields we are extracting don't require GPU capabilities provided by `nccl`.
```

examples/models.py
```diff
@@ -0,0 +1,55 @@
+from typing import Optional
+
+import torch
+
+from gigl.module.models import LinkPredictionGNN
+from gigl.src.common.models.pyg.homogeneous import GraphSAGE
+from gigl.src.common.models.pyg.link_prediction import LinkPredictionDecoder
+
+
+def init_example_gigl_homogeneous_cora_model(
```
yliu2-sc, line 10: nit: this is not cora specific

examples/models.py
mkolodner-sc: > Could we just co-locate this file with where its being used?

The reason this was separated was so this function could be shared across the `homogeneous_training.py` and `homogeneous_inference.py` files. I'd prefer to have this live here than for this model initialization function to be duplicated across both files.


> Id imagine this cant be used for anything but CORA


One note is that the `homogeneous_training.py` doesn't need to be cora-specific. After taking @yliu2-sc 's suggestion to remove the `cora` naming from the `init_example_gigl_homogeneous_model` function,  the only reference to cora here is a comment at the top of the file: `You can run this example in a full pipeline with make run_cora_glt_udl_kfp_test from GiGL root.` . Similarly, the model initialization is also not CORA specific. I believe this homogeneous training loop should work for any homogeneous dataset, although this needs to be more explicitly tested with MAG240M. Similarly, the next `heterogeneous_training.py` should work for any heterogeneous dataset.

> all examples are distributed.

I can rename this to `examples/link_prediction` in that case -- hopefully this can set us up in the future for when we  add SNC support.

PR #127, opened 2025-06-30
Add gigl.Module with RetrievalLoss and LinkPredictionGNN

mkolodner-sc:
**Scope of work done**

Currently, it is a bit awkward to use the existing LinkPredictionGNN and RetrievalLoss due to assumptions we made as part of offline SGS GiGL. For example, we need to provide a `output_node_types` to the forward pass of the model even if we are in the homogeneous setting, and index into a dictionary as a result as well when getting the output from the forward pass. Additionally, we need to populate an opinionated `BatchCombinedScores` dataclass in order to calculate RetrievalLoss. This is not PyG-conformant. 

This PR moves these two fields to their own files in a new `gigl.module` folder and removes these awkward assumptions from these classes. 

### DIFF
--- a/python/gigl/module/loss.py
+++ b/python/gigl/module/loss.py
@@ -1,0 +1,192 @@
+from typing import Optional
+
+import torch
+import torch.nn as nn
+
+
+class RetrievalLoss(nn.Module):
+    """
+    A loss layer built on top of the tensorflow_recommenders implementation.
+    https://www.tensorflow.org/recommenders/api_docs/python/tfrs/tasks/Retrieval
+
+    The loss function by default calculates the loss by:
+    ```
+    cross_entropy(torch.mm(query_embeddings, candidate_embeddings.T), positive_indices, reduction='sum'),
+    ```
+    where the candidate embeddings are `torch.cat((positive_embeddings, random_negative_embeddings))`. It encourages the model to generate query embeddings that yield the highest similarity score with their own first hop compared with others' first hops and random negatives. We also filter out the cases where, in some rows, the query could accidentally treat its own positives as negatives.
+
+    Args:
+        loss (Optional[nn.Module]): Custom loss function to be used. If `None`, the default is `nn.CrossEntropyLoss(reduction="sum")`.
+        temperature (Optional[float]): Temperature scaling applied to scores before computing cross-entropy loss. If not `None`, scores are divided by the temperature value.
+        remove_accidental_hits (bool): Whether to remove accidental hits where the query's positive items are also present in the negative samples.
+    """
+
+    def __init__(
+        self,
+        loss: Optional[nn.Module] = None,
+        temperature: Optional[float] = None,
+        remove_accidental_hits: bool = False,
+    ):
+        super(RetrievalLoss, self).__init__()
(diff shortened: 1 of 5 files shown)


python/gigl/module/loss.py
```diff
@@ -0,0 +1,192 @@
...
+            repeated_query_ids (torch.Tensor): Repeated query user IDs. Tensor shape: [num_positives]
+            candidate_sampling_probability (Optional[torch.Tensor]): Optional tensor of candidate sampling probabilities.
+                When given will be used to correct the logits to reflect the sampling probability of negative candidates.
+                Tensor shape: [num_positives + num_hard_negatives + num_random_negatives]
+        """
+        if repeated_query_ids.numel():
+            loss = self._calculate_batch_retrieval_loss(
+                scores=repeated_candidate_scores,
+                candidate_sampling_probability=candidate_sampling_probability,
+                query_ids=repeated_query_ids,
```
yliu2-sc, line 186: would the behavior be the same if you just feed `None` when there are no elements in `repeated_query_ids` and let `_calculate_batch_retrieval_loss` handle it?

yliu2-sc (approved):

python/gigl/module/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+from gigl.src.common.models.pyg.link_prediction import LinkPredictionDecoder
+from gigl.src.common.types.graph_data import NodeType
+
+
+class LinkPredictionGNN(nn.Module):
+    """
+    Link Prediction GNN model for both homogeneous and heterogeneous use cases
+    Args:
+        encoder (nn.Module): Either BasicGNN or Heterogeneous GNN for generating embeddings
```
svij-sc, line 15: If these (BasicGNN, Heterogeneous GNN) are GiGL classes, lets decorate them so sphinx doc can link them.
See example here: https://snapchat.github.io/GiGL/_modules/gigl/distributed/utils/networking.html#get_internal_ip_from_master_node

Using the following
```
    using :py:obj:`gigl.distributed.dataset_factory`
```

visualizes as:
https://snapchat.github.io/GiGL/docs/api/gigl/distributed/utils/networking/index.html#gigl.distributed.utils.networking.get_internal_ip_from_master_node

Screenshot 2025-07-01 at 11 04 27 PM

python/gigl/src/common/models/layers/loss.py
```diff
@@ -205,6 +208,10 @@ def __init__(
                 f"The temperature is expected to be greater than 1e-12, however you provided {self._temperature}"
             )
         self._remove_accidental_hits = remove_accidental_hits
+        logger.warning(
+            "Calculating retrieval loss with the class at gigl.src.common.models.layers.loss.RetrievalLoss is deprecated and will be removed in a future release. "
+            "Please use the `gigl.module.loss.RetrievalLoss` class instead."
```
svij-sc, line 213: Same comment as  above.

Lets, make sure these class / object paths get dynamically linked them using following syntax :py:obj:`gigl.module.loss.RetrievalLoss`. This will help us create better docs.

cc @kmonte , and @yliu2-sc as well for visibility to help improve our documentation

svij-sc (commented):
Just some nits for the future.

PR #129, opened 2025-07-01
Add tests for example notebooks

kmontemayor2-sc:
We should do this to ensure our examples are continually functional - nothing is worse than broken examples :(

There is some duplicity here in the sense that we also run most of the same business logic (but not """orchestration""" code) in our e2e tests - but I think it's worth it to ensure our examples are working.

Changes:

* Added the ability to stub-in resource configs to cora.ipynb
* Added python/tests/integration/examples/notebooks_test.py to run notebooks
* Added ipython/jupyter dependencies (which we should have in https://github.com/Snapchat/GiGL/pull/135 or https://github.com/Snapchat/GiGL/pull/68)

As I mentioned in the tests - we launch all the notebooks in parallel so the test can be faster.

Another thing of note is that the toy example is currently broken - I'd rather get this in first and the follow up to fix that one too :)


python/tests/integration/examples/notebooks_test.py
```diff
@@ -0,0 +1,106 @@
...
+        # if not, set it to some default value.
+        resource_config_uri = os.environ.get(
+            "GIGL_TEST_DEFAULT_RESOURCE_CONFIG",
+            str(GIGL_ROOT_DIR / "deployment/configs/e2e_glt_resource_config.yaml"),
+        )
+        logger.info(f"Using resource config URI: {resource_config_uri}")
+        # Copy over resource config to GCS so all machines can access it.
+        # If we don't do this, then since different machines run this test case,
+        # And run some of the buisness logic, they will not have access to the resource config.
+        gcs_uri = GcsUri.join(
```
svij-sc, line 56:
```suggestion
        resource_config_uri = GcsUri.join(
```

svij-sc: Should we clean this up after job finishes?

svij-sc (approved):
LGTM, given test passes when it should and fails when it should - thanks!

python/tests/integration/examples/notebooks_test.py
kmontemayor2-sc: hmmm I kind of like this to stay around (in the temp bucket) so that we can more easily debug if needed.

PR #131, opened 2025-07-02
Enable Heterogeneous Training in E2E

mkolodner-sc:
**Scope of work done**

- Enables Heterogeneous E2E Training on the DBLP dataset using GLT
- Updates the DBLP E2E config
- Small fixes to enable heterogeneous training

### DIFF
--- a/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
+++ b/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
@@ -32,6 +32,7 @@ trainerConfig:
   trainerArgs:
     # Example argument to trainer
     log_every_n_batch: "50"
+    ssl_positive_label_percentage: "0.05"
   command: python -m examples.distributed.heterogeneous_training
 inferencerConfig:
   inferencerArgs:
--- a/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
+++ b/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
@@ -32,6 +32,8 @@ trainerConfig:
   trainerArgs:
     # Example argument to trainer
     log_every_n_batch: "50"
+    # The DBLP Dataset does not have specified labeled edges, so we provide this field to indicate what percentage of edges we should
+    # select as self-supervised labeled edges.
     ssl_positive_label_percentage: "0.05"
   command: python -m examples.distributed.heterogeneous_training
 inferencerConfig:
--- a/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
+++ b/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
@@ -28,6 +28,11 @@ datasetConfig:
     dataPreprocessorArgs:
       # This argument is specific for the `PassthroughPreprocessorConfigForMockedAssets` preprocessor to indicate which dataset we should be using
       mocked_dataset_name: 'dblp_node_anchor_edge_features_lp'
+trainerConfig:
+  trainerArgs:
+    # Example argument to trainer
+    log_every_n_batch: "50"
+  command: python -m examples.distributed.heterogeneous_training
 inferencerConfig:
   inferencerArgs:
     # Example argument to inferencer
@@ -37,12 +42,7 @@ inferencerConfig:
 sharedConfig:
   shouldSkipAutomaticTempAssetCleanup: false
   shouldSkipInference: false
-  shouldSkipTraining: true  # Live subgraph sampling pipelines do not have a training phase that is being tested right now
+  # Model Evaluation is only applicable for offline SGS pipelines -- we do not need it for live SGS pipelines
   shouldSkipModelEvaluation: true
-  trainedModelMetadata:
-    # This is an example model trained on the DBLP dataset. To train a custom DBLP model, please refer to the
-    # `gigl/src/mocking/configs/dblp_node_anchor_based_link_prediction_template_gbml_config.yaml` config template.
-    # You can run a pipeline for training on DBLP using `make run_dblp_nalp_e2e_kfp_test`.
-    trainedModelUri: gs://public-gigl/mocked_assets/2024-08-01--17-11-27-UTC/dblp_node_anchor_edge_features_lp/trainer/models/model.pt
 featureFlags:
   should_run_glt_backend: 'True'
--- a/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
+++ b/examples/distributed/configs/e2e_dblp_glt_task_config.yaml
@@ -32,8 +32,10 @@ trainerConfig:
   trainerArgs:
     # Example argument to trainer
     log_every_n_batch: "50"
-    # The DBLP Dataset does not have specified labeled edges, so we provide this field to indicate what percentage of edges we should
-    # select as self-supervised labeled edges.
+    # The DBLP Dataset does not have specified labeled edges so we provide this field to indicate what
+    # percentage of edges we should select as self-supervised labeled edges. Doing this randomly sets 5% as "labels".
+    # Note that the current GiGL implementation does not remove these selected edges from the global set of edges, which may
+    # have a slight negative impact on training specifically with self-supervised learning. This will improved on in the future.
     ssl_positive_label_percentage: "0.05"
   command: python -m examples.distributed.heterogeneous_training
 inferencerConfig:
--- a/examples/distributed/heterogeneous_inference.py
+++ b/examples/distributed/heterogeneous_inference.py
@@ -95,13 +95,15 @@ def _inference_process(
     num_neighbors: List[int] = [fanout_per_hop, fanout_per_hop]
 
     # While the ideal value for `sampling_workers_per_inference_process` has been identified to be between `2` and `4`, this may need some tuning depending on the
-    # production pipeline. We default this value to `4` here for simplicity.
+    # pipeline. We default this value to `4` here for simplicity. A `sampling_workers_per_process` which is too small may not have enough parallelization for
+    # sampling, which would slow down inference, while a value which is too large may slow down each sampling process due to competing resources, which would also
+    # then slow down inference.
     sampling_workers_per_inference_process: int = int(
         inferencer_args.get("sampling_workers_per_inference_process", "4")
     )
 
     # This value represents the the shared-memory buffer size (bytes) allocated for the channel during sampling, and
-    # is the place to store pre-fetched data, so if it is too small then prefetching is limited. This parameter is a string
+    # is the place to store pre-fetched data, so if it is too small then prefetching is limited, causing sampling slowdown. This parameter is a string
     # with `{numeric_value}{storage_size}`, where storage size could be `MB`, `GB`, etc. We default this value to 4GB,
     # but in production may need some tuning.
     sampling_worker_shared_channel_size: str = inferencer_args.get(
--- a/examples/distributed/heterogeneous_inference.py
+++ b/examples/distributed/heterogeneous_inference.py
@@ -148,7 +148,6 @@ def _inference_process(
     model: LinkPredictionGNN = init_example_gigl_heterogeneous_dblp_model(
         node_type_to_feature_dim=node_type_to_feature_dim,
         edge_type_to_feature_dim=edge_type_to_feature_dim,
-        args=inferencer_args,
         device=device,
         state_dict=model_state_dict,
     )
--- a/examples/distributed/heterogeneous_inference.py
+++ b/examples/distributed/heterogeneous_inference.py
@@ -42,10 +42,6 @@ from gigl.distributed import (
     build_dataset_from_task_config_uri,
 )
 from gigl.src.common.models.pyg.heterogeneous import HGT
-from gigl.src.common.models.pyg.link_prediction import (
-    LinkPredictionDecoder,
-    LinkPredictionGNN,
-)
 from gigl.src.common.types import AppliedTaskIdentifier
 from gigl.src.common.types.graph_data import EdgeType, NodeType
 from gigl.src.common.types.pb_wrappers.gbml_config import GbmlConfigPbWrapper
--- a/examples/distributed/heterogeneous_inference.py
+++ b/examples/distributed/heterogeneous_inference.py
@@ -26,7 +26,7 @@ from typing import Dict, List, Optional, Union
 
 import torch
 import torch.multiprocessing as mp
-import torch.nn as nn
+from examples.models import init_example_gigl_heterogeneous_dblp_model
 from graphlearn_torch.distributed import barrier, shutdown_rpc
 
 import gigl.distributed
@@ -41,7 +41,7 @@ from gigl.distributed import (
     DistributedContext,
     build_dataset_from_task_config_uri,
 )
-from gigl.src.common.models.pyg.heterogeneous import HGT
+from gigl.module.models import LinkPredictionGNN
 from gigl.src.common.types import AppliedTaskIdentifier
 from gigl.src.common.types.graph_data import EdgeType, NodeType
 from gigl.src.common.types.pb_wrappers.gbml_config import GbmlConfigPbWrapper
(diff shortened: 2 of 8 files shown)


examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,976 @@
...
+        gbml_config_pb_wrapper.gbml_config_pb.shared_config.trained_model_metadata.trained_model_uri
+    )
+
+    supervision_edge_types = (
+        gbml_config_pb_wrapper.task_metadata_pb_wrapper.get_supervision_edge_types()
+    )
+    if len(supervision_edge_types) != 1:
+        raise NotImplementedError(
+            "GiGL Training currently only supports 1 supervision edge type."
+        )
```
kmontemayor2-sc, line 843: is this true? we do support multiple data loaders for different supervision edge types right?

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,976 @@
...
+    supervision_edge_type = supervision_edge_types[0]
+
+    # Training Hyperparameters shared among the training and test processes
+
+    fanout_per_hop = int(trainer_args.get("fanout_per_hop", "10"))
+    subgraph_fanout: list[int] = [fanout_per_hop, fanout_per_hop]
```
kmontemayor2-sc, line 849: should we demonstrate that the fanout can be like {edgeA: [10, 10], edgeB: [5, 15]}`?

examples/distributed/heterogeneous_inference.py
```diff
@@ -26,7 +26,7 @@
 
 import torch
 import torch.multiprocessing as mp
-import torch.nn as nn
+from examples.models import init_example_gigl_heterogeneous_dblp_model
```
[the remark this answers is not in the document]
mkolodner-sc: Yes, this is from running `make_format`

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,976 @@
...
+        trainer_args: dict[str, str]: Arguments for training
+        model_uri (Uri): URI Path to save the model to
+        subgraph_fanout: list[int]: Fanout for subgraph sampling, where the ith item corresponds to the number of items to sample for the ith hop
+        sampling_workers_per_process (int): Number of sampling workers per training process
+        main_batch_size (int): Batch size for main dataloader with query and labeled nodes
+        random_batch_size (int): Batch size for random negative dataloader
+        sampling_worker_shared_channel_size (str): Shared-memory buffer size (bytes) allocated for the channel during sampling
+        process_start_gap_seconds (int): The amount of time to sleep for initializing each dataloader. For large-scale settings, consider setting this
+            field to 30-60 seconds to ensure dataloaders don't compete for memory during initialization, causing OOM.
+        use_automatic_mixed_precision (bool): Whether we should train with mixed precision
```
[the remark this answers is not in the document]
mkolodner-sc: @svij-sc had a suggestion to remove this for the examples since it may not be necessary for a simple example -- I'll instead aim to remove this in this case.

examples/distributed/heterogeneous_training.py
mkolodner-sc: I don't think this is the way we should suggest training multiple supervision edge types with the same anchor node. Ideally, if we have `A` -> 'to' -> `B` as a supervision edge and `A` -> `to`  -> `1` as a supervision edge for a different supervision edge type, we should have the model compute loss on each supervision edge type at the same time. However, with multiple data loaders for each supervision edge type, its highly unlikely that `A` appears in both dataloaders in the same batch. Ideally, we instead support this by having multiple supervision edges with the same anchor node type in the same data loader. Since we don't have this right now, I've marked this as unsupported as so.

examples/distributed/heterogeneous_training.py
mkolodner-sc: Added a comment on this, thanks!

examples/distributed/configs/e2e_dblp_glt_task_config.yaml
```diff
@@ -28,6 +28,14 @@ datasetConfig:
     dataPreprocessorArgs:
       # This argument is specific for the `PassthroughPreprocessorConfigForMockedAssets` preprocessor to indicate which dataset we should be using
       mocked_dataset_name: 'dblp_node_anchor_edge_features_lp'
+trainerConfig:
+  trainerArgs:
+    # Example argument to trainer
+    log_every_n_batch: "50"
+    # The DBLP Dataset does not have specified labeled edges, so we provide this field to indicate what percentage of edges we should
+    # select as self-supervised labeled edges.
```
kmontemayor2-sc, line 36:
```suggestion
    # The DBLP Dataset does not have specified labeled edges
    # so we provide this field to indicate what percentage of edges we should select as self-supervised labeled edges.
    # Doing this randomly sets 5% [NOTE: this should not be named "percent" in that case] as "labels" as removes those edges otherwise from the graph.
```

examples/distributed/heterogeneous_inference.py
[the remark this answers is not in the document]
kmontemayor2-sc: odd I wonder why it doesn't think `examples` is a "project module".

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,943 @@
...
+
+    # Extracting local query, random negative, positive, hard_negative, and random_negative ids.
+    # Local in this case refers to the local index in the batch, while global subsequently refers to the node's unique global ID across all nodes in the dataset.
+    # Global ids are stored in data.node, ex. `data.node = [50, 20, 10]` where the `0` is the local id for global id `50`
+    query_node_ids: torch.Tensor = torch.arange(
+        main_data[query_node_type].batch_size
+    ).to(device)
+    random_negative_ids: torch.Tensor = torch.arange(
+        random_negative_data[labeled_node_type].batch_size
+    ).to(device)
```
kmontemayor2-sc, line 101: btw for this why not just do:

```
    random_negative_embeddings = random_negative_embeddings[labeled_node_type][
        :random_negative_data[labeled_node_type].batch_size
    ]
```

on line 126, that way we save the tensor allocation.

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,943 @@
...
+    process_start_gap_seconds: int,
+    log_every_n_batch: int,
+    learning_rate: float,
+    weight_decay: float,
+    num_max_train_batches: int,
+    num_val_batches: int,
+    val_every_n_batch: int,
+) -> None:
+    """
+    This function is spawned by each machine for training a GNN model given some loaded distributed dataset.
```
kmontemayor2-sc, line 194:
```suggestion
   Each machine spawns process(es) running this function for training a GNN model given some loaded distributed dataset.
```

WDYT?

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,943 @@
...
+        raise NotImplementedError(
+            "Currently, only GPU training is supported with this example training loop"
+        )
+
+    logger.info(
+        f"---Machine {node_rank} local rank {local_rank} training process started"
+    )
+
+    # We use one training device for each local process
+    training_device = torch.device(f"cuda:{local_rank % torch.cuda.device_count()}")
```
kmontemayor2-sc, line 232: nit. since local_rank must be less than torch.cuda.device_count() we don't need the modulus here.

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -246,6 +246,12 @@ def __init__(
                 node_type = None
         else:
             node_type, node_ids = input_nodes
+            assert isinstance(
+                dataset.node_ids, abc.Mapping
+            ), "Dataset must be heterogeneous if provided input nodes are a tuple."
+            num_neighbors = patch_fanout_for_sampling(
+                dataset.get_edge_types(), num_neighbors
+            )
```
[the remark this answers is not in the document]
kmontemayor2-sc: hmmm - I'm surprised our tests didn't catch this... is it because the DBLP tests are disabled?

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,943 @@
...
+    num_batches: Optional[int] = None,
+) -> None:
+    """
+    Runs validation using the provided models and dataloaders.
+    This function is shared for both validation while training and testing after training has completed.
+    Args:
+        local_rank (int): Process number on the current machine
+        model (DistributedDataParallel): DDP-wrapped torch model for training and testing
+        main_loader (Iterator): Dataloader for loading main batch data with query and labeled nodes
+        random_negative_loader (Iterator): Dataloader for loading random negative data
```
kmontemayor2-sc, line 499: nit. types here?

examples/distributed/heterogeneous_training.py
kmontemayor2-sc: why not actually do different fanouts for both? This seems like a "common" usecase for hetereogeneous datasets right?

I don't think that saying "we're being opinionated here" is a good argument. Why are we being opinionated? The point of the examples is IMO to:

1. Demonstrate what GiGL is capable of.
2. Give people a starting point for writing their own loops.

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,943 @@
...
+    # Training Hyperparameters shared among the training and test processes
+    fanout_per_hop = int(trainer_args.get("fanout_per_hop", "10"))
+
+    # If `subgraph_fanout` is provided as a list, it will use that fanout for each edge type. Otherwise, subgraph fanout should be provided as a
+    # dict[EdgeType, list[int]], indicating the fanout behavior for each edge type. Different edge types can have different fanout strategies. For simplicity,
+    # we make an opinionated decision to keep the fanouts for all edge types the same, specifying the `subraph_fanout` with a `list[int]`.
+
+    subgraph_fanout: list[int] = [fanout_per_hop, fanout_per_hop]
+
+    # While the ideal value for `sampling_workers_per_process` has been identified to be between `2` and `4`, this may need some tuning depending on the
```
kmontemayor2-sc, line 826: "While ..."?

Also should we say how/why people should tune this? What does it meant o increase this, or decrease this, and what performance metrics are impacted by doing so?

kmontemayor2-sc: The comment over `sampling_worker_shared_channel_size` is better - though imo it should also say what happens if it's too large - e.g. OOMS?

examples/distributed/homogeneous_training.py
```diff
@@ -65,82 +63,79 @@ def _compute_loss(
...
-        main_embeddings = model(data=main_data, device=device)
-        random_negative_embeddings = model(data=random_negative_data, device=device)
-
-        # Extracting local query, random negative, positive, hard_negative, and random_negative ids.
-        # Local in this case refers to the local index in the batch, while global subsequently refers to the node's unique global ID across all nodes in the dataset.
-        # Global ids are stored in data.node, ex. `data.node = [50, 20, 10]` where the `0` is the local id for global id `50`
-        query_node_ids: torch.Tensor = torch.arange(main_data.batch_size).to(device)
-        random_negative_ids: torch.Tensor = torch.arange(
-            random_negative_data.batch_size
+    # Forward pass through encoder
```
kmontemayor2-sc, line 70: Should all of these changes also be in this PR?

examples/distributed/heterogeneous_training.py
mkolodner-sc: We still need to subsequently do `random_negative_data[labeled_node_type].node[random_negative_ids]` for populating the `global_candidate_ids` which is important for calculating RetrievalLoss, so it makes more sense IMO to just use the `random_negative_ids` in this case.

kmontemayor2-sc: we could do the same index strategy in the latter case too, right?

examples/distributed/configs/e2e_dblp_glt_task_config.yaml
mkolodner-sc: The secondary part of this `as removes those edges otherwise from the graph ` is not currently the case -- we don't remove those labeled edges from the graph (this is a TODO for the future) -- I'll not this down as well.

examples/distributed/configs/e2e_dblp_glt_task_config.yaml
```diff
@@ -37,12 +47,7 @@ inferencerConfig:
 sharedConfig:
   shouldSkipAutomaticTempAssetCleanup: false
   shouldSkipInference: false
-  shouldSkipTraining: true  # Live subgraph sampling pipelines do not have a training phase that is being tested right now
+  # Model Evaluation is only applicable for offline SGS pipelines -- we do not need it for live SGS pipelines
```
kmontemayor2-sc, line 50: not blocking. is this a thing we want to enable down the line?

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,965 @@
...
+percentage of edges we should select as self-supervised labeled edges. Doing this randomly sets 5% as "labels". Note that the current GiGL implementation
+does not remove these selected edges from the global set of edges, which may have a slight negative impact on training specifically with self-supervised learning.
+This will improved on in the future.
+"""
+
+import argparse
+import statistics
+import time
+from collections import abc
+from collections.abc import Iterator
```
kmontemayor2-sc, line 29: nit. do we need both? let's just use one, either `from collections import abc; abc.Iterator` or `from collections.abc import Iterator`.

examples/distributed/heterogeneous_training.py
kmontemayor2-sc: any reason we didn't go with this change? I wanted to highlight that there may be multiple of these processes, per machine.

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,965 @@
...
+        worker_concurrency=sampling_workers_per_process,
+        channel_size=sampling_worker_shared_channel_size,
+        process_start_gap_seconds=0,
+        shuffle=True,
+    )
+
+    logger.info("Finished setting up train random negative loader")
+
+    train_main_loader = InfiniteIterator(train_main_loader)
+    train_random_negative_loader = InfiniteIterator(train_random_negative_loader)
```
kmontemayor2-sc, line 312: IIRC calling `iter(loader)` starts sampling in the backgroud, if we are spacing out creating the loaders for resource allocation reasons, should we not start sampling in the background too?

this is a nit tho - just something to keep in mind for the future

examples/distributed/heterogeneous_training.py
```diff
@@ -0,0 +1,965 @@
...
+    supervision_edge_type = supervision_edge_types[0]
+
+    # Training Hyperparameters shared among the training and test processes
+    fanout_per_hop = int(trainer_args.get("fanout_per_hop", "10"))
+
+    # If `subgraph_fanout` is provided as a list, it will use that fanout for each edge type. Otherwise, subgraph fanout should be provided as a
+    # dict[EdgeType, list[int]], indicating the fanout behavior for each edge type. Different edge types can have different fanout strategies. For simplicity,
+    # we make an opinionated decision to keep the fanouts for all edge types the same, specifying the `subraph_fanout` with a `list[int]`.
+
+    # TODO (mkolodner-sc): Make the heterogeneous example use a dict[EdgeType, list[int]] instead of list[int] for specifying fanout
```
kmontemayor2-sc, line 842:
```suggestion
    # TODO (mkolodner-sc): Make the heterogeneous example use a dict[EdgeType, list[int]] instead of list[int] for specifying fanout
    # For example:
    # subgraph_fanout: dict[EdgeType, list[int]] = {
    #    EdgeType(NodeType("author"), Relation("to"), NodeType("paper")): [10, 10].
    #    EdgeType(NodeType("term"), Relation("to"), NodeType("paper")): [10, 10].
    # }
```

As an explicit example?

kmontemayor2-sc (approved):
Thanks for the iterations, Matt!

examples/distributed/configs/e2e_dblp_glt_task_config.yaml
mkolodner-sc: Yup, we can enable it in the future -- for now modelEvaluation uses GiGL V1 constructs.

examples/distributed/heterogeneous_training.py
mkolodner-sc: Good call -- I moved the `iter` and `InfiniteIterator` wrapping for each to happen with their corresponding data loader (i.e. we initialize train_main, then wrap that with infinite_iterator, then sleep, then initialize train_random_negative, then wrap that with infinite_iterator, etc.)

PR #136, opened 2025-07-07
Patch Example Test Dataloader Initialization

mkolodner-sc:
**Scope of work done**

-  Updates our example training loops so that we initialize the test dataloaders only after training and properly shutdown the train and val dataloaders


python/gigl/utils/iterator.py
```diff
@@ -16,14 +18,27 @@ class InfiniteIterator(Iterator[_T]):
...
         try:
             return next(self._iter)
         except StopIteration:
             logger.info("InfiniteIterator restarting the internal iterator")
             self._iter = iter(self._iterable)
             return next(self._iter)
+
+    def shutdown(self):
+        if self._shutdowned:
+            return
```
mkolodner-sc, line 40: This mirrors the functionality of the DistLoader shutdown call [[1](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/distributed/dist_loader.py#L270-L271)]

examples/link_prediction/heterogeneous_training.py
```diff
@@ -505,16 +495,18 @@ def _training_process(
...
-
         logger.info(f"---Rank {rank} finished training")
 
-        # We explicitly delete all the dataloaders to reduce their memory footprint. Otherwise, experimentally we have
-        # observed that not all memory may be cleaned up, leading to OOM.
-        del train_main_loader, train_random_negative_loader
-        del val_main_loader, val_random_negative_loader
+        torch.cuda.empty_cache()
+        torch.cuda.synchronize()
+        torch.distributed.barrier()
```
mkolodner-sc, line 502: These lines reflect how GLT documentation cleans up their iterations through a dataloader in their examples: https://github.com/alibaba/graphlearn-for-pytorch/blob/main/examples/distributed/dist_train_sage_supervised.py#L146-L148

kmontemayor2-sc: shouldn't `shutdown` get called automatically? [1]

Do we have any suspicions that it doesn't? If so what are they?

[1]: https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_loader.py#L264-L267

python/gigl/utils/iterator.py
kmontemayor2-sc: oh is the bug here when we `del` the `InfiniteIterator`, the internal inter doesn't get cleaned up additionally?

Why not overwrite `__del__` here in that case, to also `del self._iterable, self._iter`?

mkolodner-sc: I don't believe so. This is because when doing testing and just removing the original `del` call without doing any `shutdown()`, all of the code is still able to function as expected. This makes me believe that cleaning up the internal dataloader is not a strict requirement for initializing the next dataloader (although it does create a higher memory usage). Additionally, when adding `del self._iterable, self._iter` to the `shutdown()` the bug once again appears. [[earlier commit which still had this bug](https://github.com/Snapchat/GiGL/pull/136/commits/42e9d1e98f0a83cd770143903367bc7a01ed42bd)]


Rather, this bug seems to surface iff we explicitly call `del` on the original dataloader. Rather, it appears that the safe way to "clean up" a dataloader is to explicitly invoke its `shutdown` method.

examples/link_prediction/heterogeneous_training.py
mkolodner-sc: It gets called automatically whenever the dataloader goes out of scope. However, we want to call shutdown explicitly after train/val has finished but before testing has started, at which point it has not been automatically called yet. As mentioned earlier, `del` is appears to be unsafe to use with the GLT dataloaders, which is why we need the explicit `shutdown` call instead.

python/gigl/utils/iterator.py
mkolodner-sc: > Though that begs the question of "should" we do the cleanup in our examples?

I think we should -- otherwise some of the more large-scale use cases that copy this example over may run into OOM when doing testing. Remembering that each of the dataloaders do prefetching for some amount of batches based on the `channel_size` we pass into each dataloader, this is a lot of memory overhead that we should be clearing by the time we start the test dataloader.

> That commit doesn't do the torch.cuda.empty_cache() etc which is the important thing here I think.

I was also very hopeful that this would fix the issue, but this was added in an [earlier commit](https://github.com/Snapchat/GiGL/pull/136/commits/bba839f6f824be2a44c8caca6e5bc2a0569aad30) in isolation and still observed the same bug 

> There shouldn't be any difference between shutdown() and del on the dataloaders - if we want to swap from del to shutdown we can but we use del elsewhere and tbch I kind of prefer doing it the del way.

My observations based on my experiments I've ran so far is that there **is** a difference in this case specifically when we are trying to initialize another dataloader (test) after looping through the initial ones (train, val). Based on my findings, it seems unsafe to use `del` and we should be using `shutdown` instead to safely clear the pre-fetched memory [[1](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/distributed/dist_sampling_producer.py#L229-L230)]

kmontemayor2-sc: Hmm - I  think it's odd that `del` is unsafe here but if it's what we've found then going with `shutdown` sg!

python/tests/unit/utils/iterator_test.py
```diff
@@ -1,14 +1,16 @@
 import unittest
+from typing import Final
 
 from gigl.utils.iterator import InfiniteIterator
 
+_ITERABLE_LIST: Final[list[int]] = [3, 4, 5, 6]
```
kmontemayor2-sc, line 6: nit. but for such small tests I don't really like having module level constants, I think they make things harder to read.

python/tests/unit/utils/iterator_test.py
```diff
@@ -17,6 +19,15 @@ def test_infinite_iterator(self):
...
         self.assertEqual(actual_list, expected_list)
 
+    def test_infinite_iterator_shutdown(self):
+        infinite_iterator = InfiniteIterator(iterable=_ITERABLE_LIST)
+        for _ in infinite_iterator:
+            break
+        infinite_iterator.shutdown()
+        with self.assertRaises(ValueError):
+            for _ in infinite_iterator:
+                break
```
kmontemayor2-sc, line 29: nit.
```suggestion
            next(infinite_iterator)
```

python/gigl/utils/iterator.py
kmontemayor2-sc: I do think we should add tests for this case though - specifically, "can we shutdown a dataloader, and then *in the same process*, start a new dataloader.

Maybe it's not possible because we don't have GPUs on CICD atm but maybe modifying [1] with this case would be a good idea?

[1]: https://github.com/Snapchat/GiGL/blob/f750c1bde0b56c5729fd4624ccb23bfdc3083209/python/tests/unit/distributed/distributed_neighborloader_test.py#L273-L314

examples/link_prediction/homogeneous_training.py
```diff
@@ -461,16 +451,18 @@ def _training_process(
...
         logger.info(f"---Rank {rank} finished training")
 
-        # We explicitly delete all the dataloaders to reduce their memory footprint. Otherwise, experimentally we have
-        # observed that not all memory may be cleaned up, leading to OOM.
-        del train_main_loader, train_random_negative_loader
-        del val_main_loader, val_random_negative_loader
-        gc.collect()
+        torch.cuda.empty_cache()
+        torch.cuda.synchronize()
+        torch.distributed.barrier()
```
kmontemayor2-sc, line 458: let's add a comment here for why we do this?

examples/link_prediction/homogeneous_training.py
```diff
@@ -371,6 +357,10 @@ def _training_process(
             sampling_worker_shared_channel_size=sampling_worker_shared_channel_size,
             process_start_gap_seconds=process_start_gap_seconds,
         )
+        assert isinstance(train_main_loader, InfiniteIterator)
+        assert isinstance(train_random_negative_loader, InfiniteIterator)
+        assert isinstance(val_main_loader, InfiniteIterator)
+        assert isinstance(val_random_negative_loader, InfiniteIterator)
```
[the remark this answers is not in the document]
mkolodner-sc: In the test case though it returns `iter(dataloader)` instead of `InfiniteIterator(dataloader)`

python/gigl/utils/iterator.py
xgao4-sc: Code pointer to the above comment: https://github.com/alibaba/graphlearn-for-pytorch/blob/v0.2.5/graphlearn_torch/python/distributed/dist_loader.py#L263-L266

zfan3-sc: > Yes, calling del on the DistLoader will call its shutdown protocol. 

I did a little bit of research and found out this may not be true. `del my_object` only decreased the refcount to `my_object` and it's the interpreters job to destruct it. So `__del__` isn't necessarily called. [stackoverflow ref](https://stackoverflow.com/questions/1481488/what-is-the-del-method-and-how-do-i-call-it)

> However, from experimentation, my findings is that when you call del with a subsequent step that attempts to initialize a new test dataloader in the same process, this will produce error:  

Thanks for the context! With this behavior, my guess is that there is still some refcount hidden within the code so the member iterator's `shutdown` isn't being called. So explicitly call it makes sense.

examples/link_prediction/homogeneous_training.py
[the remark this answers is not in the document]
kmontemayor2-sc: I see, I'm not super happy about all these type narrowing asserts in our code tbch...

one alternative here is do overwrite `__del__` on InfiniteIterator which as Zihao mentioned could call `shutdown` internally, though I'm not sure if you think that approach is "cleaner"?

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -299,6 +299,9 @@ def _run_multiple_neighbor_loader(
     # https://paperswithcode.com/dataset/cora
     assert count == expected_data_count
 
+    loader_one.shutdown()
+    loader_two.shutdown()
```
kmontemayor2-sc, line 303: thanks!

examples/link_prediction/homogeneous_training.py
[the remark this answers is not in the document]
mkolodner-sc: Discussed offline, I've moved the `InfiniteIterator`/`iter` casting to be outside of the function call. Hopefully that can reduce the type narrow asserts we have here, thanks!

examples/link_prediction/heterogeneous_training.py
```diff
@@ -371,21 +358,12 @@ def _training_process(
...
-        device=device,
-        sampling_worker_shared_channel_size=sampling_worker_shared_channel_size,
-        process_start_gap_seconds=process_start_gap_seconds,
-    )
-
     if not should_skip_training:
+        train_main_loader: Iterator[HeteroData]
+        train_random_negative_loader: Iterator[HeteroData]
+        val_main_loader: Iterator[HeteroData]
+        val_random_negative_loader: Iterator[HeteroData]
```
kmontemayor2-sc, line 365: qq do we need these defs?

why not `train_main_loader, train_random_loader = _setup_dataloaders(); train_iter, train_random_inter = InfiniteIter(),...`?

kmontemayor2-sc: More generally - I have some preference against adding type annotations *outside* signatures, I think that they make the code messier and these specifically are error prone.

mkolodner-sc: Done, let me know how this looks, thanks!

examples/link_prediction/heterogeneous_training.py
```diff
@@ -505,16 +486,19 @@ def _training_process(
...
+        torch.cuda.synchronize()  # Ensures all CUDA operations have finished
+        torch.distributed.barrier()  # Waits for all processes to reach the current point
 
-        gc.collect()
+        # We explicitly shutdown all the dataloaders to reduce their memory footprint. Otherwise, experimentally we have
+        # observed that not all memory may be cleaned up, leading to OOM.
+        train_main_loader_iter.shutdown()
+        train_random_negative_loader_iter.shutdown()
+        val_main_loader_iter.shutdown()
+        val_random_negative_loader_iter.shutdown()
```
kmontemayor2-sc, line 501: now we can just do `train_main_loader.shutdown()` and remove that bit from InfiniteIterator (which is imo a bit confusing) right?

examples/link_prediction/heterogeneous_training.py
```diff
@@ -399,6 +372,11 @@ def _training_process(
             process_start_gap_seconds=process_start_gap_seconds,
         )
 
+        train_main_loader_iter = InfiniteIterator(train_main_loader)
```
kmontemayor2-sc, line 375:
```suggestion
        # We keep track of both the dataloader and the InfiniteIterator for it
        # so we can clean up resources from the dataloader later.
        train_main_loader_iter = InfiniteIterator(train_main_loader)
```

And add this elsewhere?

kmontemayor2-sc (approved):
Thanks Matt!

PR #141, opened 2025-07-09
[In Progress] Bump + Release KFP + Docker

svij-sc:
**Scope of work done**

Where is the documentation for this feature?: N/A

Did you add automated tests or write a test plan?

***Updated Changelog.md?*** NO

***Ready for code review?:*** NO

PR #142, opened 2025-07-09
Add ability to push to a new branch in `commit-and-push` github action

svij-sc:
**Scope of work done**

Needed for: https://github.com/Snapchat/GiGL/pull/141

Providing ability for the `commit-and-push` action to have a new param `to_new_branch_named` specific.
If specific, it will push the changes to this new branch.

This was tested on the following branch: https://github.com/Snapchat/GiGL/pull/141
w/ resulting branch push here: https://github.com/Snapchat/GiGL/pull/150

### DIFF
--- a/.github/actions/commit-and-push/action.yml
+++ b/.github/actions/commit-and-push/action.yml
@@ -1,5 +1,12 @@
 name: "Commit and Push Changes"
-description: "Commits and pushes changes to the current or specified branch with a customizable commit message, including changes the workflow has made"
+description: |
+  Commits and pushes changes to the current or specified branch with a customizable commit message.
+  Includes changes the workflow has made.
+  Only commits changes to files that have been modified or deleted, without adding new files.
+  If a new branch is specified, the action will create a new branch and push the changes to it.
+  If no new branch is specified, the action will push the changes to the current branch.
+
+
 inputs:
   commit_message:
     description: "Commit message for the changes"
@@ -29,7 +36,7 @@ runs:
           git push --set-upstream origin $TO_NEW_BRANCH_NAMED
         fi
 
-        git add --all
+        git add --update
 
         # If no changes, exit gracefully. If not done, git commit will fail.
         if git diff --cached --exit-code; then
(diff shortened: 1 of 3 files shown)


.github/actions/commit-and-push/action.yml
```diff
@@ -8,24 +8,36 @@ inputs:
   github_token:
     description: "GitHub token to authenticate and push the changes"
     required: true
+  to_new_branch_named:
+    description: "Create a new branch and push the changes to it"
+    required: false
+    default: "false"
```
kmontemayor2-sc, line 14: why not make this empty string? having the default branch name be called "false" is a bit odd here?

.github/actions/commit-and-push/action.yml
```diff
@@ -8,24 +8,36 @@ inputs:
...
-        
         # Check if there are any changes
-        if git diff --cached --exit-code; then
-          echo "No changes to commit."
-          exit 0
+        if [ "$TO_NEW_BRANCH_NAMED" == "false" ]; then
+          if git diff --cached --exit-code; then
+            echo "No changes to commit."
+            exit 0
+          fi
```
kmontemayor2-sc, line 38: I'm a little confused - why do we need this? it's saying "if we don't want to push to a new branch, and there are no diffs, then don't push?

What would happen if we didn't have this check?

.github/actions/commit-and-push/action.yml
```diff
@@ -8,24 +8,36 @@ inputs:
   github_token:
     description: "GitHub token to authenticate and push the changes"
     required: true
+  to_new_branch_named:
+    description: "Create a new branch and push the changes to it"
```
kmontemayor2-sc, line 12: nit. but what are "the" changes here? All changes in your branch (compared to main)? The changes the gh actions have made?

kmontemayor2-sc (approved):

.github/actions/commit-and-push/action.yml
svij-sc: Actually, `git commit -m "$COMMIT_MESSAGE"` with nothing to commit returns error code, thus we need to have this check

PR #146, opened 2025-07-09
Fix toy example + add to notebook tests

kmontemayor2-sc:
Fixed toy example notebook (details below) and add toy example and dblp to notebook tests

The toy example notebook was not functional (on my machine) for a few reasons:

1. I didn't migrate some of the visualization tools to the common file (whoops - tests should catch that now though :))
2. We didn't build the scala jars locally prior - so SGS and split gen couldn't run, now we do and copy the files over to the expected locations
3. I believe we need to reload the frozen config pb wrapper after pre processor - as otherwise the preprocessed metadata wrapper will be unpopulated. 

Note: we should probably run `notebooks_test` as a part of the e2e tests but I can do that as a follow up :)


examples/toy_visual_example/toy_example_walkthrough.ipynb
```diff
@@ -333,6 +334,69 @@
...
+    "        component=GiGLComponents.SplitGenerator, use_spark35=True\n",
+    "    ).uri,\n",
+    "]\n",
+    "for directory in dirs_to_delete:\n",
+    "    shutil.rmtree(directory, ignore_errors=True)\n",
+    "\n",
+    "packager = ScalaPackager()\n",
+    "sgs_spark31 = packager.package_subgraph_sampler()\n",
+    "sgs_spark35 = packager.package_subgraph_sampler(use_spark35=True)\n",
+    "split_gen = packager.package_split_generator(use_spark35=True)\n",
```
svij-sc, line 381: Maybe we just get these functions to delete the dirs so the instructions are not conflated?

We could introduce a ScalaPacker().package_all() that does everything here and just have it inside the docker cell.

examples/toy_visual_example/toy_example_walkthrough.ipynb
```diff
@@ -333,6 +334,69 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Copy over the jars to the appropriate directories so that we can use them in the next cells.\n",
+    "!cp $sgs_spark31.uri python/gigl/deps/scala/subgraph_sampler/jars/toy_sgs.jar\n",
```
svij-sc, line 395: As I understand it, the cell above is just a copy off what is in `if __name__ == "__main__":` off `scala_packager.py`. This is called by `compile_jars` which is usually called as a prerequisite to `push_new_docker_images`. Thus, In the Makefile flows, it doesnt seem like we do any `cp` operations

do we actually need to do these `cp` operations here?

examples/toy_visual_example/toy_example_walkthrough.ipynb
```diff
@@ -520,6 +584,9 @@
     }
    ],
    "source": [
+    "# Reload frozen config as config populator has modified the file.\n",
```
svij-sc, line 587: Do you mean preprocessor?
Also,  I don't think this should be the case / needed.

See: https://github.com/Snapchat/GiGL/blob/bb68d3d113e48265a864806ced4f478ac53e8765/python/gigl/src/common/types/pb_wrappers/gbml_config.py#L353-L368

examples/toy_visual_example/toy_example_walkthrough.ipynb
kmontemayor2-sc: sg, done :)

examples/toy_visual_example/toy_example_walkthrough.ipynb
svij-sc: hu weird....
why dont we do `cp` operations in any of our make workflows then?
i.e. if I look at 

https://github.com/Snapchat/GiGL/blob/d38c6fd8de6c864772c7f98ced096479a3586f13/Makefile#L409


We do the steps in following order:
1. compile_jars
https://github.com/Snapchat/GiGL/blob/d38c6fd8de6c864772c7f98ced096479a3586f13/Makefile#L223-L225
The scala packager itself doesnt seem to be doing any `cp`
https://github.com/Snapchat/GiGL/blob/33afd3844eeb39393e01e77bdc3d286dc2f37212/scripts/scala_packager.py#L136-L159



2. push_new_docker_images
https://github.com/Snapchat/GiGL/blob/d38c6fd8de6c864772c7f98ced096479a3586f13/Makefile#L232-L241
Which is just calling `build_and_push` targets
https://github.com/Snapchat/GiGL/blob/91170a13116a8a5fee802ef8b20a2ae230442465/scripts/build_and_push_docker_image.py#L139-L172

svij-sc: I wonder if its because its being executed from a diff directory inside the jupyter notebook?
But this would be weird...

python/gigl/common/utils/test_utils.py
```diff
@@ -49,6 +49,11 @@ def parse_args() -> TestArgs:
 
 
 def _run_individual_test(test: unittest.TestCase) -> Tuple[bool, int]:
+    if test.countTestCases() == 0:
```
svij-sc, line 52: can we add a comment what this is for

svij-sc (approved):
Thanks for the iterations

python/gigl/common/utils/test_utils.py
kmontemayor2-sc: Done.

PR #147, opened 2025-07-10
Add fossa integration

svij-sc:
**Scope of work done**

See: [gigl-issue/1292](https://go/gigl-issue/1292) for context.


.github/workflows/fossa-analyze.yml
```diff
@@ -0,0 +1,61 @@
...
+
+permissions:
+  contents: read
+  id-token: write
+
+# Concurrency configuration to prevent multiple Fossa analyses from running simultaneously
+# This ensures only one analysis runs at a time to avoid conflicts
+concurrency:
+  group: "fossa"
+  cancel-in-progress: false
```
kmontemayor2-sc, line 23: hmmm why cancel in progress ones? 

If we merge PR A, which triggers workflow a, then later merge PR B, while a is still running, wouldn't we want to run the workflow on B instead?

.github/workflows/fossa-analyze.yml
```diff
@@ -0,0 +1,61 @@
...
+            SHA256=0c59fed2b76ddd6a9dbd0b2a937796a162b8ccc730c63a3a33528b30291dc5cb
+            wget "https://github.com/fossas/fossa-cli/releases/download/v3.9.42/fossa_3.9.42_linux_amd64.zip" \
+                && echo "$SHA256 fossa_3.9.42_linux_amd64.zip" | sha256sum -c - \
+                && unzip fossa_3.9.42_linux_amd64.zip
+            ./fossa --version
+      - name: Run Fossa
+        env:
+          FOSSA_API_KEY: ${{ secrets.FOSSA_API_KEY }}
+        run: |
+          # Suppress stdout, capture stderr
```
kmontemayor2-sc, line 55: any reason we suppress stdout?

kmontemayor2-sc (approved):
The changes as-is seem fine - do we have a way to ensure this works before we submit? I'd be curious to see the workflow :)

Otherwise this LGTM.

.github/workflows/fossa-analyze.yml
svij-sc: workflow B would still run.
This just means it would let workflow A to finish first and enforce sequentiality so we only ever publish the latest report.

.github/workflows/fossa-analyze.yml
svij-sc: See [gigl-issue/1292](https://go/gigl-issue/1292)

.github/workflows/fossa-analyze.yml
kmontemayor2-sc: I think in this case there's a time where workflow A finishes, but PR B is merged and workflow B is running, so we are sort of "out of date" then right?

Like, if we know workflow A is going to be outdated immediately why finish it?

svij-sc: We could be in the middle of an upload - the behavior of cancelling a report upload half way through is undefined atm. 
This just saves the hassle to figure it out by enforcing sequentiality to circumvent the problem.

svij-sc: We also get nice green checks after every commit :)

PR #154, opened 2025-07-10
Fix Fanout in Dataloaders

mkolodner-sc:
**Scope of work done**

- In the case where num_neighbors is a dict, we were previously checking if the edge types of the dataset is equal to the edge types of num_neighbors **before** patching it with the positive label edge types, resulting in failure. This order should be switched, which is patched in this PR
- Additionally, we don't call patching if dataset.get_edge_types() is None, indicating a homogeneous dataset.
- Also restructures the ABLPNeighborLoader and DistNeighborLoader to be more similar in the order where they patch fanout.
- Update `patch_subgraph_from_fanout` to not infer fanout for non-label edge types if provided fanout is a dictionary
- Update tests for `patch_subgraph_from_fanout` utility

### DIFF
--- a/python/gigl/distributed/dist_ablp_neighborloader.py
+++ b/python/gigl/distributed/dist_ablp_neighborloader.py
@@ -311,20 +311,9 @@ class DistABLPLoader(DistLoader):
             )
         )
 
-        if dataset.get_edge_types() is not None:
-            num_neighbors = patch_fanout_for_sampling(
-                dataset.get_edge_types(), num_neighbors
-            )
-            hops = len(next(iter(num_neighbors.values())))
-            if not all(len(fanout) == hops for fanout in num_neighbors.values()):
-                raise ValueError(
-                    f"num_neighbors must be a dict of edge types with the same number of hops. Received: {num_neighbors}"
-                )
-        else:
-            if isinstance(num_neighbors, abc.Mapping):
-                raise ValueError(
-                    "When dataset is homogeneous, the num_neighbors field cannot be a dictionary."
-                )
+        num_neighbors = patch_fanout_for_sampling(
+            dataset.get_edge_types(), num_neighbors
+        )
 
         curr_process_nodes = shard_nodes_by_process(
             input_nodes=anchor_node_ids,
--- a/python/gigl/distributed/dist_ablp_neighborloader.py
+++ b/python/gigl/distributed/dist_ablp_neighborloader.py
@@ -311,20 +311,20 @@ class DistABLPLoader(DistLoader):
             )
         )
 
-        # TODO(kmonte): stop setting fanout for positive/negative once GLT sampling is fixed.
-        num_neighbors = patch_fanout_for_sampling(
-            dataset.get_edge_types(), num_neighbors
-        )
-
-        if num_neighbors.keys() != dataset.graph.keys():
-            raise ValueError(
-                f"num_neighbors must have all edge types in the graph, received: {num_neighbors.keys()} with for graph with edge types {dataset.graph.keys()}"
+        if dataset.get_edge_types() is not None:
+            num_neighbors = patch_fanout_for_sampling(
+                dataset.get_edge_types(), num_neighbors
             )
-        hops = len(next(iter(num_neighbors.values())))
-        if not all(len(fanout) == hops for fanout in num_neighbors.values()):
-            raise ValueError(
-                f"num_neighbors must be a dict of edge types with the same number of hops. Received: {num_neighbors}"
-            )
+            hops = len(next(iter(num_neighbors.values())))
+            if not all(len(fanout) == hops for fanout in num_neighbors.values()):
+                raise ValueError(
+                    f"num_neighbors must be a dict of edge types with the same number of hops. Received: {num_neighbors}"
+                )
+        else:
+            if isinstance(num_neighbors, abc.Mapping):
+                raise ValueError(
+                    "When dataset is homogeneous, the num_neighbors field cannot be a dictionary."
+                )
 
         curr_process_nodes = shard_nodes_by_process(
             input_nodes=anchor_node_ids,
--- a/python/gigl/distributed/distributed_neighborloader.py
+++ b/python/gigl/distributed/distributed_neighborloader.py
@@ -202,25 +202,6 @@ class DistNeighborLoader(DistLoader):
                 )
             input_nodes = dataset.node_ids
 
-        if isinstance(num_neighbors, abc.Mapping):
-            # TODO(kmonte): We should enable this. We have two blockers:
-            # 1. We need to treat `EdgeType` as a proper tuple, not the GiGL`EdgeType`.
-            # 2. There are (likely) some GLT bugs around https://github.com/alibaba/graphlearn-for-pytorch/blob/26fe3d4e050b081bc51a79dc9547f244f5d314da/graphlearn_torch/python/distributed/dist_neighbor_sampler.py#L317-L318
-            # Where if num_neighbors is a dict then we index into it improperly.
-            if not isinstance(dataset.graph, abc.Mapping):
-                raise ValueError(
-                    "When num_neighbors is a dict, the dataset must be heterogeneous."
-                )
-            if num_neighbors.keys() != dataset.graph.keys():
-                raise ValueError(
-                    f"num_neighbors must have all edge types in the graph, received: {num_neighbors.keys()} with for graph with edge types {dataset.graph.keys()}"
-                )
-            hops = len(next(iter(num_neighbors.values())))
-            if not all(len(fanout) == hops for fanout in num_neighbors.values()):
-                raise ValueError(
-                    f"num_neighbors must be a dict of edge types with the same number of hops. Received: {num_neighbors}"
-                )
-
         # Determines if the node ids passed in are heterogeneous or homogeneous.
         self._is_labeled_heterogeneous = False
         if isinstance(input_nodes, torch.Tensor):
@@ -235,9 +216,6 @@ class DistNeighborLoader(DistLoader):
                 ):
                     node_type = DEFAULT_HOMOGENEOUS_NODE_TYPE
                     self._is_labeled_heterogeneous = True
-                    num_neighbors = patch_fanout_for_sampling(
-                        dataset.get_edge_types(), num_neighbors
-                    )
                 else:
                     raise ValueError(
                         f"For heterogeneous datasets, input_nodes must be a tuple of (node_type, node_ids) OR if it is a labeled homogeneous dataset, input_nodes may be a torch.Tensor. Received node types: {dataset.node_ids.keys()}"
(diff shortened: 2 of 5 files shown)


python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -311,20 +311,20 @@ def __init__(
...
 
-        # TODO(kmonte): stop setting fanout for positive/negative once GLT sampling is fixed.
-        num_neighbors = patch_fanout_for_sampling(
-            dataset.get_edge_types(), num_neighbors
-        )
-
-        if num_neighbors.keys() != dataset.graph.keys():
-            raise ValueError(
-                f"num_neighbors must have all edge types in the graph, received: {num_neighbors.keys()} with for graph with edge types {dataset.graph.keys()}"
-            )
```
mkolodner-sc, line 322: This check is no longer needed since it will always be equal to each other after we patch the fanout

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -202,25 +202,6 @@ def __init__(
...
-            if num_neighbors.keys() != dataset.graph.keys():
-                raise ValueError(
-                    f"num_neighbors must have all edge types in the graph, received: {num_neighbors.keys()} with for graph with edge types {dataset.graph.keys()}"
-                )
-            hops = len(next(iter(num_neighbors.values())))
-            if not all(len(fanout) == hops for fanout in num_neighbors.values()):
-                raise ValueError(
-                    f"num_neighbors must be a dict of edge types with the same number of hops. Received: {num_neighbors}"
-                )
-
```
kmontemayor2-sc, line 223: btw should we consolidate this logic as much as possible between the two loaders? WDYT about bringing over this function? https://github.com/Snapchat/GiGL/pull/115/files#diff-1f13197ec18bb37bbbbd8146057e17430c427831dfb2825978bdb86d8fb65b49R141

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -749,11 +748,38 @@ def test_dblp_supervised(self):
...
+                partitioner_class=DistPartitioner,
+                fanout=[2, 2],
+            ),
+            param(
+                "Range-based partitioning, list fanout",
+                partitioner_class=DistRangePartitioner,
+                fanout=[2, 2],
+            ),
+            param(
+                "Range-based partitioning, list fanout",
```
kmontemayor2-sc, line 762:
```suggestion
                "Range-based partitioning, dict fanout",
```

python/tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -749,11 +748,38 @@ def test_dblp_supervised(self):
...
+                partitioner_class=DistPartitioner,
```
kmontemayor2-sc, line 763: Also should this be `DistRangePartitioner`?

python/gigl/distributed/dist_ablp_neighborloader.py
mkolodner-sc: Sure, we can keep this around in that case

python/gigl/distributed/distributed_neighborloader.py
mkolodner-sc: Hmm, I'm fine with reducing duplicity, but the linked function seems to be working around the `NodeSamplerInput`, and the change in this PR is focusing on the `num_neighbors` field, so it seems a bit out of scope. Maybe we can have the DistABLPLoader subclass from the DistNeighborLoader have some private function there which reduces this duplicated code? WDYT?

kmontemayor2-sc: oh you're right! I guess my point was we could add some `validate_num_neighbors` or something to do the checks for us in a shared manner :)

I generally prefer to use functions to share functionality over inheritance, as inheritance can get *very* confusing very quickly when different classes and different layers override some methods and not others, where as with functions it's always *right there*.

mkolodner-sc: Sounds good, I've moved some of the logic into the `patch_subgraph_fanout` function and updated the tests accordingly -- let me know how this look.

I was also looking into that function and noticed we have a behavior where if we have a `num_neighbors = (User->To->Item: [10, 10])` and `edge_types = (User->to->Story, Story->to->User)` that we set the `Story->To->User` edge type fanout to `[10, 10]` as well. I'm not sure we should be doing this -- if num_neighbors is missing a edge type that a user should be aware of and pass in, I feel like we should error out in this case. I've also updated the logic in `patch_subgraph_fanout` -- lmk if this looks ok here. Thanks!

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -27,26 +28,49 @@ def patch_fanout_for_sampling(
...
+            raise ValueError(
+                "When dataset is homogeneous, the num_neighbors field cannot be a dictionary."
+            )
+        return num_neighbors
     if isinstance(num_neighbors, list):
         original_fanout = num_neighbors
+        should_broadcast_fanout = True
         num_neighbors = {}
     else:
+        for edge_type in num_neighbors.keys():
```
kmontemayor2-sc, line 48:
```suggestion
        for edge_type in num_neighbors:
```
nit.

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -27,26 +28,49 @@ def patch_fanout_for_sampling(
...
         if is_label_edge_type(edge_type):
             num_neighbors[edge_type] = zero_samples
-        elif edge_type not in num_neighbors:
+        elif should_broadcast_fanout and edge_type not in num_neighbors:
             num_neighbors[edge_type] = original_fanout
-
+        elif not should_broadcast_fanout and edge_type not in num_neighbors:
+            raise ValueError(
+                f"Found non-labeled edge type in dataset {edge_type} which is not in the provided fanout {num_neighbors.keys()}"
+            )
```
kmontemayor2-sc, line 67: can we add the notes I put before where we should reduce these restrictions once we we fix some of the glt sampling logic?

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -27,26 +28,49 @@ def patch_fanout_for_sampling(
...
     for edge_type in edge_types:
         if is_label_edge_type(edge_type):
             num_neighbors[edge_type] = zero_samples
-        elif edge_type not in num_neighbors:
+        elif should_broadcast_fanout and edge_type not in num_neighbors:
             num_neighbors[edge_type] = original_fanout
-
+        elif not should_broadcast_fanout and edge_type not in num_neighbors:
+            raise ValueError(
+                f"Found non-labeled edge type in dataset {edge_type} which is not in the provided fanout {num_neighbors.keys()}"
```
kmontemayor2-sc, line 66:
```suggestion
                f"Found non-labeled edge type in dataset {edge_type} which is not in the provided fanout {num_neighbors.keys()}. If fanout is provided as a dict, all edges must be present."
```

Let's say the expectation here?

Also maybe in the doc comment?

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -53,7 +54,7 @@ def test_shard_nodes_by_process(
     @parameterized.expand(
         [
             param(
-                "Test set_labeled_edge_type on num_neighbors dict with labeled edge type",
+                "Test patch_fanout_for_sampling on num_neighbors dict with labeled edge type",
```
kmontemayor2-sc, line 57: hmmm, what if a user provides a label edge type in the fanout?

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -84,18 +76,63 @@ def test_shard_nodes_by_process(
...
+        num_neighbors: Union[list[int], dict[EdgeType, list[int]]],
+        expected_num_neighbors: Union[list[int], dict[EdgeType, list[int]]],
     ):
         num_neighbors = patch_fanout_for_sampling(edge_types, num_neighbors)
         self.assertEqual(num_neighbors, expected_num_neighbors)
 
+    @parameterized.expand(
+        [
+            param(
+                "Test failure when homogeneous dataset has heterogeneous num_neighbors",
```
kmontemayor2-sc, line 100: nit. we don't need `"test failure" when "failure" is in the test name.

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -84,18 +76,63 @@ def test_shard_nodes_by_process(
...
-        num_neighbors: dict[EdgeType, list[int]],
-        expected_num_neighbors: dict[EdgeType, list[int]],
+        edge_types: Optional[list[EdgeType]],
+        num_neighbors: Union[list[int], dict[EdgeType, list[int]]],
+        expected_num_neighbors: Union[list[int], dict[EdgeType, list[int]]],
     ):
         num_neighbors = patch_fanout_for_sampling(edge_types, num_neighbors)
         self.assertEqual(num_neighbors, expected_num_neighbors)
 
+    @parameterized.expand(
```
kmontemayor2-sc, line 97: nice tests!

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -13,40 +14,67 @@
...
+) -> Union[list[int], dict[EdgeType, list[int]]]:
     """
-    Setups an approprirate fanout for sampling.
+    Sets up an approprirate fanout for sampling.
 
     Does the following:
     - For all label edge types, sets the fanout to be zero.
     - For all other edge types, if the fanout is not specified, uses the original fanout.
 
+    Note that if fanout is provided as a dict, all edges must be present.
```
kmontemayor2-sc, line 27:
```suggestion
    Note that if fanout is provided as a dict, the keys (edges) in the fanout must be in `edge_types`.
```

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -13,40 +14,67 @@
...
+                f"Found extra edge types {extra_edge_types} in fanout which is not in dataset edge types {edge_types}."
+            )
         original_fanout = next(iter(num_neighbors.values()))
+        should_broadcast_fanout = False
         num_neighbors = deepcopy(num_neighbors)
 
     num_hop = len(original_fanout)
     zero_samples = [0 for _ in range(num_hop)]
     for edge_type in edge_types:
+        # TODO(kmonte): stop setting fanout for positive/negative edges once GLT sampling correctly ignores those edges during fanout.
```
kmontemayor2-sc, line 62: minor preference to copy over the old comment with more details :P

https://github.com/Snapchat/GiGL/pull/154/files#diff-e751a9cd787e206721cd05c941effd36bae4e56559ac269889b7e9fdb27eb240L206-L209

kmontemayor2-sc (approved):
Thanks Matt!

python/gigl/distributed/utils/neighborloader.py
mkolodner-sc: Hmm, it seems like that comment may be a bit outdated from when we used to raise a NotImplementedError when `num_neighbors` was an abc.Mapping. We have provided support for and tested this behavior now.

PR #157, opened 2025-07-11
add gcsfuse to docker img

nshah-sc:
**Scope of work done**

### DIFF
--- a/containers/Dockerfile.cuda.base
+++ b/containers/Dockerfile.cuda.base
@@ -8,9 +8,19 @@ SHELL ["/bin/bash", "-c"]
 # Install basic dependencies
 # TODO(mkolodner-sc): iputils-ping temporarily needed to setup inter-job VAI communication for GLT Inference.
 # Once VAI natively supports this communication, we can remove this requirement.
-RUN apt-get update \
-    && apt-get upgrade -y \
-    && apt-get install -y build-essential git wget cmake iputils-ping \
+RUN apt-get update && apt-get upgrade -y && apt-get install -y \
+    build-essential \
+    git \
+    wget \
+    cmake \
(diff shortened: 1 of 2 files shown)


containers/Dockerfile.cuda.base
```diff
@@ -5,12 +5,24 @@ FROM nvidia/cuda:12.1.0-cudnn8-devel-ubuntu22.04
...
 
-# Install basic dependencies
+# We install some basic dependencies and gcsfuse for GCS access.
+# We can use gcsfuse to mount GCS buckets in the container.
 # TODO(mkolodner-sc): iputils-ping temporarily needed to setup inter-job VAI communication for GLT Inference.
 # Once VAI natively supports this communication, we can remove this requirement.
-RUN apt-get update \
-    && apt-get upgrade -y \
-    && apt-get install -y build-essential git wget cmake iputils-ping \
+RUN apt-get update && apt-get upgrade -y && apt-get install -y \
```
svij-sc, line 12: For future reader, maybe we link where we are pulling the GCSFUSE installation instructions from:
https://cloud.google.com/storage/docs/cloud-storage-fuse/install#configure-package-manager

Also seems installing from source is more straightforward; curious if you tried it and if its too slow / possible:

https://cloud.google.com/storage/docs/cloud-storage-fuse/install#install-source-code
```bash
RUN go install github.com/googlecloudplatform/gcsfuse/v3@master
```

nshah-sc: I'll add a note -- I didn't try installing from source but anyway in order to do this we'd need to add go 1.24.0 to the images as well just to build gcsfuse and that also seems like a bit more work just to do this (and is likely to be slower).  hence, opting against it

PR #158, opened 2025-07-11
Make setting up DDP for LinkPredictionGNN easier

kmontemayor2-sc:
**Scope of work done**
* Add `LinkPredictionGNN.to_dpp` method and implement in examples

Since DDP.forward() calls `_pre_forward()` [1] and `self.reducer.prepare_for_forward() ` [2], calling `forward` directly on modules DDP wraps is not expected or supported.

As such, we do this so that we can "safely" use the LinkPredictionGNN in a distributed context. 

Add some unit tests :)

[1]: https://github.com/pytorch/pytorch/blob/26807dcf277feb2d99ab88d7b6da526488baea93/torch/nn/parallel/distributed.py#L1657
[2]: https://github.com/pytorch/pytorch/blob/26807dcf277feb2d99ab88d7b6da526488baea93/torch/nn/parallel/distributed.py#L1529


examples/link_prediction/heterogeneous_training.py
```diff
@@ -396,23 +396,22 @@ def _training_process(
...
+            node_type_to_feature_dim=node_type_to_feature_dim,
+            edge_type_to_feature_dim=edge_type_to_feature_dim,
+            device=device,
+            init_for_ddp=True,
+            dummy_data=next(
+                train_main_loader_iter
+            ),  # We need to pass a dummy data for DDP initialization
         )
-
+        model.encoder.find_unused_parameters = True
```
zfan3-sc, line 408: Ideally we shouldn't set it to `True` unless the training task requires it. IIUC in the example use case it's because of one node type not participating in the model forward and backend, can we skip init params for it when we init the model?

examples/link_prediction/heterogeneous_training.py
```diff
@@ -396,23 +396,22 @@ def _training_process(
...
+        model.decoder.find_unused_parameters = True
```
zfan3-sc, line 409: similarly, would prefer leave this as False for decoder as well

examples/link_prediction/models.py
```diff
@@ -49,16 +51,26 @@ def init_example_gigl_homogeneous_model(
...
-        device = torch.device("cpu")
-    model.to(device)
+    if for_ddp:
+        # Initialize the model for DDP training.
+        model: LinkPredictionGNN = LinkPredictionGNN.for_ddp(
+            encoder=encoder_model,
+            decoder=decoder_model,
+            device=device,
+            dummy_data=dummy_data,
+            find_unused_parameters=True,
```
zfan3-sc, line 61: same here

examples/link_prediction/models.py
```diff
@@ -105,16 +118,28 @@ def init_example_gigl_heterogeneous_model(
...
-    model.to(device)
+    if init_for_ddp:
+        # Initialize the model for DDP training.
+        model: LinkPredictionGNN = LinkPredictionGNN.for_ddp(
+            encoder=encoder_model,
+            decoder=decoder_model,
+            device=device,
+            dummy_data=dummy_data,
+            output_node_types=list(node_type_to_feature_dim.keys()),
+            find_unused_parameters=True,
```
zfan3-sc, line 129: same here

examples/link_prediction/heterogeneous_training.py
kmontemayor2-sc: That's true - Let's continue the discussion in https://github.com/Snapchat/GiGL/pull/158#discussion_r2201915811 :)

examples/link_prediction/models.py
mkolodner-sc: This is correct, we should just make sure to add comments explaining why we set this to True, and note that graph definitions which don't have this constraint with unused parameters do not need to set this field to True. 

One note is that I do not believe the decoder needs to receive `find_unused_parameters=True`

zfan3-sc: I'm fine with leaving a note here explaining why we need it. And ditto for the decoder we don't need it, would be nice to have separate control for encoder and decoder.

examples/link_prediction/heterogeneous_training.py
```diff
@@ -396,23 +396,20 @@ def _training_process(
...
-            # We should set `find_unused_parameters` to True since not all of the model parameters may be used in backward pass in the heterogeneous setting
-            find_unused_parameters=True,
+        model = init_example_gigl_heterogeneous_model(
+            node_type_to_feature_dim=node_type_to_feature_dim,
+            edge_type_to_feature_dim=edge_type_to_feature_dim,
+            device=device,
+            init_for_ddp=True,
+            dummy_data=next(
+                train_main_loader_iter
+            ),  # We need to pass a dummy data for DDP initialization
```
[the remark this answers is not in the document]
kmontemayor2-sc: good call out - somehow I remember needing to do this but we shouldn't need to.

Removed - runinng e2e tests

examples/link_prediction/models.py
```diff
@@ -105,16 +122,35 @@ def init_example_gigl_heterogeneous_model(
...
+            raise ValueError(
+                "Device must be specified when initializing for DDP training"
+            )
+
+        # Initialize the model for DDP training.
+        # Since the HGT encoder has params for *all* node and edge types [1]
+        # We need to allow DDP to find unused parameters.
+        # As not all node types may be present in the training task.
+        # 1. https://github.com/Snapchat/GiGL/blob/766fae5dc313e1224998ed5618cf70cf0fb4da30/python/gigl/src/common/models/pyg/heterogeneous.py#L47-L51
+        model = LinkPredictionGNN.for_ddp(
```
mkolodner-sc, line 136: Generally, the way that DDP is used with PyG is we initialize some base model and, when we want to use DDP, we subsequently wrap that initialized model with DDP. In our case, I feel like we are not following this usage pattern by having two `constructor` methods for `LinkPredictionGNN`, which feels confusing and doesn't follow usual PyG patterns. I'd prefer it if we could have one constructor for initializing a LinkPredictionGNN and a subsequent method we could call on it which wraps the encoder/decoder with DDP

examples/link_prediction/models.py
```diff
@@ -105,16 +122,35 @@ def init_example_gigl_heterogeneous_model(
...
+            encoder=encoder_model,
+            decoder=decoder_model,
+            device=device,
+            find_unused_encoder_parameters=True,
```
mkolodner-sc, line 140: Can we expose this parameter at a higher level, rather than just defaulting this to True? While this is a model that lives in `examples/link_prediction/models`, I imagine when folks copy over this file they will still be using this function, which technically would still work regardless of what dataset they are using. In that case, it would be helpful if we could have this flag exposed so that users can disable this if their dataset doesn't have the same constraint as our DBLP dataset. 

I'd rather have our example also be more explicit that DDP is using `find_unused_params=True` for the encoder, rather than keeping that omitted from users who are trying to run it. Moving this parameter and above comment to the training loops would be preferred in this case.

mkolodner-sc: Also, if we have a `find_unused_encoder_parameters` field, should we also expose a `find_unused_decoder_parameters` field?

mkolodner-sc (commented):
Thanks Kyle! Left some initial comments

examples/link_prediction/models.py
```diff
@@ -105,16 +130,30 @@ def init_example_gigl_heterogeneous_model(
...
         encoder=encoder_model,
         decoder=decoder_model,
     )
-
-    # Push the model to the specified device.
-    if device is None:
-        device = torch.device("cpu")
-    model.to(device)
+    if wrap_with_ddp:
+        if device is None:
```
mkolodner-sc, line 138: If `device` is None, can we just have this get passed to `to_ddp` instead`? This is more conformant to how wrapping a model with DDP directly would look. [[1](https://github.com/pytorch/pytorch/blob/v2.7.0/torch/nn/parallel/distributed.py#L529-L530)]

mkolodner-sc (commented):
Thanks Kyle! Generally this LGTM provided tests pass, will leave it to @zfan3-sc to stamp this since he raised this issue

zfan3-sc (approved):

PR #159, opened 2025-07-11
Enable CPU training in examples

mkolodner-sc:
**Scope of work done**

- Enables CPU training in homogeneous and heterogeneous examples

### DIFF
--- a/examples/link_prediction/homogeneous_training.py
+++ b/examples/link_prediction/homogeneous_training.py
@@ -40,6 +40,7 @@ from gigl.distributed import (
     build_dataset_from_task_config_uri,
 )
 from gigl.distributed.distributed_neighborloader import DistNeighborLoader
+from gigl.distributed.utils import get_available_device
 from gigl.module.loss import RetrievalLoss
 from gigl.src.common.types.pb_wrappers.gbml_config import GbmlConfigPbWrapper
 from gigl.src.common.utils.model import load_state_dict_from_uri, save_state_dict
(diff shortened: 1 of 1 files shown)


examples/link_prediction/heterogeneous_training.py
```diff
@@ -349,8 +345,8 @@ def _training_process(
     logger.info(f"---Rank {rank} training process group initialized")
 
     # We use one training device for each local process
-    device = torch.device(f"cuda:{local_rank}")
-    torch.cuda.set_device(device)
```
mkolodner-sc, line 353: `torch.cuda.set_device(device)` automatically gets called in `init_neighbor_loader_worker`, so we don't need to call it explicitly here

kmontemayor2-sc: hmmm I remember this throwing at some point if I didn't add this line... this still passes e2e tests right?

kmontemayor2-sc (commented):
Thanks Matt! Shall we add instructions for how one can run these files locally too?

examples/link_prediction/heterogeneous_training.py
mkolodner-sc: Yup, still passes e2e

examples/link_prediction/heterogeneous_training.py
```diff
@@ -840,6 +840,10 @@ def _run_example_training(
     logger.info(f"--- Training finished, took {time.time() - start_time} seconds")
 
 
+"""
+Given a frozen task config with some already populated data preprocessor output, the following training script can be run locally with CPU training using:
+WORLD_SIZE=1 RANK=0 MASTER_ADDR="localhost" MASTER_PORT=20000 python -m examples.link_prediction.heterogeneous_training --task_config_uri=<task_config>
+"""
```
kmontemayor2-sc, line 846: can we put these at the top of the file?

Also provide instructions for how to generate a task config uri?

kmontemayor2-sc: Also this isn't restricted to local CPU - it could also do GPU, depending on the machine setup.

kmontemayor2-sc (approved):
Thanks Matt

PR #170, opened 2025-07-15
Stop running notebooks tests as part of "integration tests"

kmontemayor2-sc:
They are big tests and if we have our "e2e" tests separate then these should be separate too, as they (essentially) run the same code.

* created new `/notebook_test` command to run these tests on a PR
* created new "on-per-merge" `ci-integration-example-notebooks-test` workflow(?) to run the tests when merging into main,

Tested in https://github.com/Snapchat/GiGL/actions/runs/16351820323 and https://github.com/Snapchat/GiGL/actions/runs/16379413218/job/46287263462?pr=170


.github/workflows/on-pr-comment.yml
```diff
@@ -75,6 +75,27 @@ jobs:
         command: |
           make run_all_e2e_tests
 
+ notebooks-e2e-test:
```
svij-sc, line 78: have you had a chance to test this?

You can do this by adding
```
on:
  issue_comment:
    types: [created]
  + push:
```

.github/workflows/on-pr-comment.yml
```diff
@@ -75,6 +75,27 @@ jobs:
...
+    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/notebook_tests') }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 155 # Tests as of 2025-05-16 are taking ~110 mins to complete, 40% buffer
```
svij-sc, line 81: does this need to be updated?

.github/workflows/on-pr-comment.yml
svij-sc: btw, the tab on `notebooks-e2e-test:` might not be correct.

svij-sc (approved):
LGTM, caveat on us testing this before we merge this.

Makefile
```diff
@@ -193,6 +193,9 @@ integration_test:
 	--test_file_pattern=$(PY_TEST_FILES) \
 	)
 
+notebooks_test:
+	RESOURCE_CONFIG_PATH=${GIGL_TEST_DEFAULT_RESOURCE_CONFIG} python -m e2e_tests.notebooks_test
```
svij-sc, line 197: We also have `shared/tests/dep_vars_check.py` ... wondering if we should move these together into some generic folder?

svij-sc: like a top level testing folder?
```
testing/notebooks_test.py
testing/dep_vars_check.py
```

Sure, if ^

thanks!

PR #171, opened 2025-07-15
remove typing.List in favor of list

kmontemayor2-sc:
It's actually pretty easy to do this with:

1. replace `List[` -> `list[`
2. autoflake/isort to do the rest

Can send out follow up PRs for other collections as well.


svij-sc (approved):
Thanks

Issue #169, opened 2025-07-15
Deprecate typing imports off Dict, List, etc.


svij-sc:
From Python 3.9 onwards, you can directly use the built-in collection types for type annotations:
i.e.

```python
def process_items(items: list[str]) -> dict[str, int]:
```

is now possible,
vs before we standardized around

```python
from typing import List, Dict
def process_items(items: List[str]) -> Dict[str, int]:
```

We should shift away from use of typing module

PR #183, opened 2025-07-18
Replace typing.Dict with dict

kmontemayor2-sc:
For https://github.com/Snapchat/GiGL/issues/169

Test plan is for our cicd to check it all :P (`make type_check`) passes locally tho


svij-sc (approved):
TY

PR #184, opened 2025-07-18
Revert "Remove unnecessary distributed calls in Dataloader setup (#140)"

kmontemayor2-sc:
This reverts commit 460a82910b4c913a0698ea02198c117df6202638.

This breaks some internal pipelines which still rely on `glt.distributed.barrier`

Let's rollback and forward later with more fixes :)

cc: @mkolodner-sc


svij-sc (approved):
Thanks

PR #189, opened 2025-07-21
Add `hid_dim` and `out_dim` for examples

mkolodner-sc:
**Scope of work done**

We should add `hid_dim` and `out_dim` to our examples so that when folks copy over our code they can more easily start playing around with these levers. This also allows our MAG240M example to use these training/inference loops directly and still converge.

- Adds `hid_dim` and `out_dim` as trainer/inferencer args for our examples
- Some small comment improvements

### DIFF
--- a/examples/link_prediction/homogeneous_inference.py
+++ b/examples/link_prediction/homogeneous_inference.py
@@ -325,6 +325,9 @@ def _run_example_inference(
     inferencer_args = dict(gbml_config_pb_wrapper.inferencer_config.inferencer_args)
     inference_batch_size = gbml_config_pb_wrapper.inferencer_config.inference_batch_size
 
+    hid_dim = int(inferencer_args.get("hid_dim", "16"))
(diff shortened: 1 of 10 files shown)


examples/link_prediction/homogeneous_inference.py
```diff
@@ -367,19 +370,21 @@ def _run_example_inference(
...
-            master_default_process_group_port,
-            embedding_output_gcs_folder,
-            model_uri,
-            inference_batch_size,
-            dataset,
-            inferencer_args,
-            graph_metadata.homogeneous_node_type,
-            node_feature_dim,
-            edge_feature_dim,
+            local_world_size,  # local_world_size
```
mkolodner-sc, line 373: Looks like we forgot to add the corresponding var names for this file -- added it here

kmontemayor2-sc: Thanks Matt!

PR #191, opened 2025-07-21
Update HGT model to handle differently ordered edge types

mkolodner-sc:
**Scope of work done**

This was a tricky one. When we initialize a HGTConv layer, we provide some edge types, which it uses to create an offset mapping for each edge type. When we forward some `data.edge_index_dict` through this layer, we require that the edge types there have the same order as the edge types in the constructor, otherwise the offsets will be off. For large graphs, this will lead to indexing errors during segmented matrix multiplication [[1](https://pyg-lib.readthedocs.io/en/stable/_modules/pyg_lib/ops.html#segment_matmul)] as such
```
split_with_sizes expects split_sizes have only non-negative entries, but got split_sizes=[66491, -41288, 41288, 66491, 0, 0]
```

which is caused by the `ptr` tensor being in a non-ascending order` 

For smaller graphs, segmented matrix multiplication is not used (based on some heuristic in PyG, see torch_geometric.backend.use_segment_matmul in [[2](https://pytorch-geometric.readthedocs.io/en/2.5.1/_modules/torch_geometric/nn/dense/linear.html#HeteroLinear)]) and we don't observe the above error when the indexing is incorrect, which is why we haven't observed this error prior and only instead when we were using MAG240M with larger batch sizes. Even in the small graph case though, the indices are still wrong and likely lead to incorrect forward passes, hurting the model performance.

[[3](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/hgt_conv.html#HGTConv)]

This fix was validated on a MAG240M run which was previously failing due to the above issue. 

### DIFF
--- a/python/gigl/src/common/models/pyg/heterogeneous.py
+++ b/python/gigl/src/common/models/pyg/heterogeneous.py
@@ -95,9 +95,9 @@ class HGT(nn.Module):
         # The HGTConv layer requires that the edge index dictionary during forward is in the same order
         # as the edge types that the were provided to the model during initialization. Otherwise, we may run into
         # indexing errors with the HeteroLinear layer inside of the HGTConv.
-        if sorted(self._edge_types) != sorted(data.edge_index_dict):
+        if sorted(self._edge_types) != sorted(data.edge_index_dic.keys()):
             raise ValueError(
-                f"Found mismatching edge types between HGTConv initialized edge types {self._edge_types} and HeteroData edge types {sorted(data.edge_index_dict)}"
+                f"Found mismatching edge types between HGTConv initialized edge types {self._edge_types} and HeteroData edge types {sorted(data.edge_index_dict)}. These must be the same."
             )
         edge_index_dict = {
             edge_type: data.edge_index_dict[edge_type] for edge_type in self._edge_types
--- a/python/gigl/src/common/models/pyg/heterogeneous.py
+++ b/python/gigl/src/common/models/pyg/heterogeneous.py
@@ -92,10 +92,14 @@ class HGT(nn.Module):
                 for node_type, x in node_type_to_features_dict.items()
             }
 
-        # The HGTConv layer requires that the edge index dictionary during forward is in the same order
-        # as the edge types that the were provided to the model during initialization. Otherwise, we may run into
-        # indexing errors with the HeteroLinear layer inside of the HGTConv.
-        if sorted(self._edge_types) != sorted(data.edge_index_dic.keys()):
+        # When we initialize a HGTConv layer, we provide some edge types, which it uses to create an offset mapping for
+        # each edge type. When we forward some data.edge_index_dict through this layer, we require that the edge types
+        # there have the same order as the edge types in the constructor, otherwise the offsets will be off. For large graphs,
+        # this will lead to indexing errors during segmented matrix multiplication. For smaller graphs, segmented matrix
+        # multiplication is not used (based on some heuristic in PyG) and we don't observe this error. However, the indices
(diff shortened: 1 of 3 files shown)


python/tests/unit/module/models_test.py
kmontemayor2-sc, line 1: can we put these changes in `python/tests/unit/src/common/models/pyg/heterogeneous_test.py`

Let's be consistent about where we place the tests for different things?

python/gigl/src/common/models/pyg/heterogeneous.py
```diff
@@ -92,14 +92,25 @@ def forward(
                 for node_type, x in node_type_to_features_dict.items()
             }
 
+        # The HGTConv layer requires that the edge index dictionary during forward is in the same order
+        # as the edge types that the were provided to the model during initialization. Otherwise, we may run into
+        # indexing errors with the HeteroLinear layer inside of the HGTConv.
```
kmontemayor2-sc, line 97: meta comment: I like the description and detail you put into the pr description, I think it helps say *why* we make these checks, etc, can we move some of that over?

python/tests/unit/module/models_test.py
```diff
@@ -136,5 +138,72 @@ def test_unwrap_from_ddp(self):
         self.assertIs(unwrapped.decoder, decoder)
 
 
+class TestHGT(unittest.TestCase):
+    def setUp(self):
+        # Set this field to True so that it will be able to raise error if the HGTConv forward pass does not behave as expected
+        torch_geometric.backend.use_segment_matmul = True
+
+    def test_hgt_forward(self):
```
kmontemayor2-sc, line 146: should we also test for the error case? https://github.com/Snapchat/GiGL/pull/191/files#diff-40e5bd9670955a06ee468ef4c430fb7937acde3dc40191f4b74f25fb56436b5eR98-R100

python/tests/unit/module/models_test.py
```diff
@@ -136,5 +138,72 @@ def test_unwrap_from_ddp(self):
         self.assertIs(unwrapped.decoder, decoder)
 
 
+class TestHGT(unittest.TestCase):
+    def setUp(self):
+        # Set this field to True so that it will be able to raise error if the HGTConv forward pass does not behave as expected
+        torch_geometric.backend.use_segment_matmul = True
```
kmontemayor2-sc, line 144: can we store the original value for this and then reset in tearDown().

Also we should call `super().setUp()` as best practice.

kmontemayor2-sc (approved):
LGTM - though we should create the new test file - it can get hairy to maintain if we're not consistent about the business logic <=> test logic path mapping.

python/tests/unit/src/common/models/pyg/heterogeneous_test.py
```diff
@@ -0,0 +1,93 @@
...
+        with self.assertRaises(ValueError):
+            data.edge_index_dict = {
+                location_to_item_edge_type: torch.tensor([[0, 1, 2, 3], [0, 1, 2, 0]]),
+                user_to_user_edge_type: torch.tensor([[0, 1], [1, 0]]),
+            }
+            encoder(
+                data=data,
+                output_node_types=[user_node_type, item_node_type, location_node_type],
+                device=torch.device("cpu"),
+            )
```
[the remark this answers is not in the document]
mkolodner-sc: I initially did it this way, but ended up avoiding doing this in this case to reduce duplicity with the other fields which are redundantly initialized between the two tests, such as the HGT model -- is this blocking the PR merge?

PR #192, opened 2025-07-21
Force conda forge channel only

svij-sc:
**Scope of work done**

Noticed behaviour:
Terms of Service implicitly accepted for conda default channel - even though we don't want to use it.
This forces us not to use it this case. 

Will follow up w/ task where we can just do this at a global level i.e. reliance on `actions/setup-python@v4` which has a pre-installation of conda which uses default channel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ get_ver_hash:
 	$(eval GIT_COMMIT=$(shell git log -1 --pretty=format:"%H"))
 
 initialize_environment:
-	conda create -y -c conda-forge --name ${CONDA_ENV_NAME} python=${PYTHON_VERSION} pip=${PIP_VERSION} pip-tools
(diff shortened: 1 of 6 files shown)


kmontemayor2-sc (commented):
Thanks Shubham! 

Should we also update the other places we run `conda`? [1] 

ex, `Dockerfile.builder` [2], `Makefile` [3]

[1]: https://github.com/search?q=repo%3ASnapchat%2FGiGL%20conda&type=code
[2]: https://github.com/Snapchat/GiGL/blob/4f81daf8993be47aea10d3c11b81c651f6afb768/containers/Dockerfile.builder#L44
[3]: https://github.com/Snapchat/GiGL/blob/4f81daf8993be47aea10d3c11b81c651f6afb768/Makefile#L47

kmontemayor2-sc (approved):
Minor preference for us to test by rebuilding our base images but LGTM.

Thanks for the quick work!

svij-sc:
> Minor preference for us to test by rebuilding our base images but LGTM.
I have 3 other prs which require rebuilding base images - will do it as part of that 🤣 

But this should work - I did validate locally atleast.

PR #193, opened 2025-07-21
Handle Missing Features after Fanout

mkolodner-sc:
**Scope of work done**

### DIFF
--- a/python/gigl/distributed/dist_link_prediction_dataset.py
+++ b/python/gigl/distributed/dist_link_prediction_dataset.py
@@ -61,6 +61,8 @@ class DistLinkPredictionDataset(DistDataset):
         num_train: Optional[Union[int, Dict[NodeType, int]]] = None,
         num_val: Optional[Union[int, Dict[NodeType, int]]] = None,
         num_test: Optional[Union[int, Dict[NodeType, int]]] = None,
+        node_feature_dim: Optional[Union[int, dict[NodeType, int]]] = None,
+        edge_feature_dim: Optional[Union[int, dict[EdgeType, int]]] = None,
     ) -> None:
         """
         Initializes the fields of the DistLinkPredictionDataset class. This function is called upon each serialization of the DistLinkPredictionDataset instance.
@@ -77,9 +79,13 @@ class DistLinkPredictionDataset(DistDataset):
             positive_edge_label (Optional[Union[torch.Tensor, Dict[EdgeType, torch.Tensor]]]): Positive Edge Label Tensor
             negative_edge_label (Optional[Union[torch.Tensor, Dict[EdgeType, torch.Tensor]]]): Negative Edge Label Tensor
             node_ids (Optional[Union[torch.Tensor, Dict[NodeType, torch.Tensor]]]): Node IDs on the current machine
-            num_train: (Optional[Mapping[NodeType, int]]): Number of training nodes on the current machine. Will be a dict if heterogeneous.
-            num_val: (Optional[Mapping[NodeType, int]]): Number of validation nodes on the current machine. Will be a dict if heterogeneous.
-            num_test: (Optional[Mapping[NodeType, int]]): Number of test nodes on the current machine. Will be a dict if heterogeneous.
+            num_train:  Optional[Union[int, Dict[NodeType, int]]]): Number of training nodes on the current machine. Will be a dict if heterogeneous.
+            num_val: (Optional[Union[int, Dict[NodeType, int]]]): Number of validation nodes on the current machine. Will be a dict if heterogeneous.
+            num_test: (Optional[Union[int, Dict[NodeType, int]]]): Number of test nodes on the current machine. Will be a dict if heterogeneous.
+            node_feature_dim: Optional[Union[int, dict[NodeType, int]]]: Dimension of node features, will be a dict if heterogeneous. Note this will be None in the
+                homogeneous case if the data has no node features, or will only contain node types with node features in the heterogeneous case.
+            edge_feature_dim: Optional[Union[int, dict[EdgeType, int]]]: Dimension of edge features, will be a dict if heterogeneous. Note this will be None in the
+                homogeneous case if the data has no edge features, or will only contain edge types with edge features in the heterogeneous case.
         """
         self._rank: int = rank
         self._world_size: int = world_size
@@ -109,6 +115,8 @@ class DistLinkPredictionDataset(DistDataset):
         self._num_train = num_train
         self._num_val = num_val
         self._num_test = num_test
+        self._node_feature_dim = node_feature_dim
+        self._edge_feature_dim = edge_feature_dim
 
     # TODO (mkolodner-sc): Modify so that we don't need to rely on GLT's base variable naming (i.e. partition_idx, num_partitions) in favor of more clear
     # naming (i.e. rank, world_size).
@@ -214,8 +222,16 @@ class DistLinkPredictionDataset(DistDataset):
         return self._negative_edge_label
 
     @property
-    def node_ids(self) -> Optional[Union[torch.Tensor, Dict[NodeType, torch.Tensor]]]:
+    def node_ids(self) -> Optional[Union[torch.Tensor, Dict[NodeType, int]]]:
         return self._node_ids
+
+    @property
+    def node_feature_dim(self) -> Optional[Union[int, dict[NodeType, int]]]:
+        return self._node_feature_dim
+
+    @property
+    def edge_feature_dim(self) -> Optional[Union[int, dict[EdgeType, int]]]:
+        return self._edge_feature_dim
 
     @property
     def train_node_ids(
@@ -421,6 +437,7 @@ class DistLinkPredictionDataset(DistDataset):
                 id2idx=node_id2idx,
                 with_gpu=False,
             )
+            self._node_feature_dim = partitioned_node_features.size(1)
             del partitioned_node_features, partitioned_node_feature_ids, node_id2idx
         elif isinstance(partition_output.partitioned_node_features, abc.Mapping):
             assert isinstance(partition_output.node_partition_book, abc.Mapping)
@@ -448,6 +465,10 @@ class DistLinkPredictionDataset(DistDataset):
                 id2idx=node_type_to_id2idx,
                 with_gpu=False,
             )
+            self._node_feature_dim = {
+                node_type: feature_partition_data.feats.size(1)
+                for node_type, feature_partition_data in partition_output.partitioned_node_features.items()
+            }
             del (
                 node_type_to_partitioned_node_features,
                 node_type_to_partitioned_node_feature_ids,
@@ -474,6 +495,7 @@ class DistLinkPredictionDataset(DistDataset):
                 id2idx=edge_id2idx,
                 with_gpu=False,
             )
+            self._edge_feature_dim = partitioned_edge_features.size(1)
             del partitioned_edge_features, partition_edge_feat_ids, edge_id2idx
         elif (
             isinstance(partition_output.partitioned_edge_features, abc.Mapping)
@@ -510,6 +532,10 @@ class DistLinkPredictionDataset(DistDataset):
                 id2idx=edge_type_to_id2idx,
                 with_gpu=False,
             )
+            self._edge_feature_dim = {
+                edge_type: feature_partition_data.feats.size(1)
+                for edge_type, feature_partition_data in partition_output.partitioned_edge_features.items()
+            }
             del (
                 edge_type_to_partitioned_edge_features,
                 edge_type_to_partitioned_edge_feature_ids,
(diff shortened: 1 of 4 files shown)


python/gigl/distributed/dist_link_prediction_dataset.py
```diff
@@ -626,14 +652,16 @@ def share_ipc(
         Optional[Union[Graph, Dict[EdgeType, Graph]]],
         Optional[Union[Feature, Dict[NodeType, Feature]]],
         Optional[Union[Feature, Dict[EdgeType, Feature]]],
-        Optional[Union[torch.Tensor, Dict[NodeType, torch.Tensor]]],
-        Optional[Union[torch.Tensor, Dict[EdgeType, torch.Tensor]]],
+        Optional[Union[PartitionBook, Dict[NodeType, PartitionBook]]],
```
mkolodner-sc, line 655: Some of the typing and docstrings were not up to date -- I've updated them here

python/gigl/distributed/dist_link_prediction_dataset.py
```diff
@@ -217,11 +225,19 @@ def negative_edge_label(
...
+    @property
+    def edge_feature_dim(self) -> Optional[Union[int, dict[EdgeType, int]]]:
+        return self._edge_feature_dim
+
     @property
     def train_node_ids(
         self,
     ) -> Optional[Union[torch.Tensor, abc.Mapping[NodeType, torch.Tensor]]]:
-        if self._num_train is None:
+        if not self._num_train:
```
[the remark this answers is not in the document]
kmontemayor2-sc: empty dict and None are not the same, 

None implies we *never* split, empty dict implies we split but didn't split for any nodes types.

I am not sure we should be conflating the two here.

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -153,3 +155,76 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
...
+def set_missing_features(
+    data: _GraphType,
+    node_feature_dim: Optional[Union[int, dict[NodeType, int]]],
+    edge_feature_dim: Optional[Union[int, dict[EdgeType, int]]],
+    device: torch.device,
+    dtype: torch.dtype = torch.float32,
+) -> _GraphType:
+    """
+    If a feature is missing from a produced Data or HeteroData object due to not fanning out to it, populates it with an empty tensor with the appropriate feature dim.
+    Note that PyG natively does this with their DistNeighborLoader for missing edge features + edge indices and missing node features.
```
kmontemayor2-sc, line 169: Can we link to where they do it here?

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -153,3 +155,76 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
...
+    However, native Graphlearn-for-PyTorch only does this for edge indices -- we should do this for node/edge features as well.
```
kmontemayor2-sc, line 170: Once we get our own fork of GLT up and running we should do it there right? Let's add a TODO about that.

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -153,3 +155,76 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
         del data.num_sampled_edges[edge_type]
 
     return data
+
+
+def set_missing_features(
```
kmontemayor2-sc, line 160: Add a note that this modifies the input inplace?

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -153,3 +155,76 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
...
+    data: _GraphType,
+    node_feature_dim: Optional[Union[int, dict[NodeType, int]]],
+    edge_feature_dim: Optional[Union[int, dict[EdgeType, int]]],
```
kmontemayor2-sc, line 163: qq why would these be None?

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -153,3 +155,76 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
...
+    if isinstance(data, Data):
+        if isinstance(node_feature_dim, dict):
+            raise ValueError(
+                f"Expected node feature dimension to be an int or None for homogeneous data, got {node_feature_dim} of type {type(node_feature_dim)}"
+            )
+        if isinstance(edge_feature_dim, dict):
+            raise ValueError(
+                f"Expected edge feature dimension to be an int or None for homogeneous data, got {edge_feature_dim} of type {type(edge_feature_dim)}"
+            )
+        if node_feature_dim and (not hasattr(data, "x") or data.x is None):
```
kmontemayor2-sc, line 193: Again does `data.x = None` and `data.x` not being present have different semantic meanings?

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -211,3 +218,135 @@ def test_strip_label_edges(self):
...
+                num_node_features=20,
+                num_edge_features=0,
+            ),
+            param(
+                "Node features, edge features",
+                num_node_features=20,
+                num_edge_features=40,
+            ),
+        ]
+    )
```
kmontemayor2-sc, line 240: shouldn't  we be testing for the `None` case as well here?

python/gigl/distributed/dist_link_prediction_dataset.py
```diff
@@ -217,11 +225,19 @@ def negative_edge_label(
...
         return self._node_ids
 
+    @property
+    def node_feature_dim(self) -> Optional[Union[int, dict[NodeType, int]]]:
+        return self._node_feature_dim
+
+    @property
+    def edge_feature_dim(self) -> Optional[Union[int, dict[EdgeType, int]]]:
+        return self._edge_feature_dim
+
```
kmontemayor2-sc, line 235: btw why don't we just derive these on the fly from however we store the node/edge features?

We could cache the result but I think adding new fields to our constructor/etc is a bit of an overhead and maybe we should avoid it?

python/gigl/distributed/dist_link_prediction_dataset.py
kmontemayor2-sc: Thanks!

python/gigl/distributed/distributed_neighborloader.py
```diff
@@ -325,9 +329,14 @@ def __init__(
...
+            data=data,
+            node_feature_dim=self._node_feature_dim,
+            edge_feature_dim=self._edge_feature_dim,
+            device=self.to_device,
+        )
         if isinstance(data, HeteroData):
             data = strip_label_edges(data)
-
         if self._is_labeled_heterogeneous:
             data = labeled_to_homogeneous(DEFAULT_HOMOGENEOUS_EDGE_TYPE, data)
```
kmontemayor2-sc, line 341: btw this stuff is exactly why I wanted to expose `transforms` in the first place but I don't think we need to change it at this point.

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -153,3 +155,76 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
...
 
     return data
+
+
+def set_missing_features(
+    data: _GraphType,
+    node_feature_dim: Optional[Union[int, dict[NodeType, int]]],
+    edge_feature_dim: Optional[Union[int, dict[EdgeType, int]]],
+    device: torch.device,
+    dtype: torch.dtype = torch.float32,
```
kmontemayor2-sc, line 165: any reason we expose this and then never use it?

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -211,3 +218,135 @@ def test_strip_label_edges(self):
...
+            node_feature_dim=2,
+            edge_feature_dim=4,
+            device=self._device,
+            dtype=self._feature_dtype,
+        )
+        assert_tensor_equality(
+            data.x,
+            torch.ones(
+                (num_node_features, 2), device=self._device, dtype=self._feature_dtype
+            ),
```
kmontemayor2-sc, line 262: why is this `torch.ones`? Shouldn't it be zeros?

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -211,3 +218,135 @@ def test_strip_label_edges(self):
...
+                num_edge_features=0,
+            ),
+            param(
+                "Node features, edge features",
+                num_node_features=20,
+                num_edge_features=40,
+            ),
+        ]
+    )
+    def test_homogeneous_set_missing_features(
```
kmontemayor2-sc, line 241: we should also probably add a test that we don't overwrite features that are already existing in the data, btw.

kmontemayor2-sc (commented):
Thanks Matt!

One thing I'm a bit confused about is why we're treating `not set`/`None`/`empty dict` as the same case now, they seem different to me.

python/gigl/distributed/dist_link_prediction_dataset.py
mkolodner-sc: From the PR description:

Note that we don't extract the features directly from the dataset, but rather store the node/edge_feature_dim as a property of the dataset class during dataset building. This is done since the features are lazily initialized whenever they are re-serialized, and we want to avoid unnecessary memory spikes when possible which would be caused by directly accessing the feature fields and causing lazy initialization.

python/tests/unit/distributed/utils/neighborloader_test.py
mkolodner-sc: This is tested whenever one of the `num_node_features` here is non-zero, since we assert the feature shape is equal to that size after we call the `missing_edge_features` function

python/tests/unit/distributed/utils/neighborloader_test.py
mkolodner-sc: I made an arbitrary choice to initialize all the non-empty features in this file with ones values, but I'm fine doing it with zeros if that's the preference here.

python/gigl/distributed/utils/neighborloader.py
mkolodner-sc: This is a good question -- I looked into it and it actually seems like its different for the `homogeneous` and `heterogeneous` cases. 

For homogeneous, it seems that hasattr(data, "x") always returns True and we should be checking if this field is None to see if it has been set.

For heterogeneous, it seems that checking if this field is None for a given node type will always produce `AttributeError: 'NodeStorage' object has no attribute 'x'` and we should, in that case, use the `hasattr(hetero_data[node_type], "x")` to see if the x field is present.

python/gigl/distributed/utils/neighborloader.py
mkolodner-sc: From the docstring, if we have no node features or edge features in the graph, they will be None. See [[1](https://github.com/Snapchat/GiGL/blob/5ae5a286097a9097e07ca7749690884dfda0131a/python/gigl/distributed/dist_link_prediction_dataset.py#L483-L543)], where we only populate the `self._edge_feature_dim` field if it goes through any of these conditionals. If there are no edge features (meaning `partition_output.partitioned_edge_features = None`, then it will never change from its default value of None.

python/gigl/distributed/dist_link_prediction_dataset.py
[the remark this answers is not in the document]
mkolodner-sc: Makes sense, will update the logic here to instead have the heterogeneous case set `self._num_train, etc.` to None if a splitter is not provided instead of an empty dict

python/gigl/distributed/utils/neighborloader.py
```diff
@@ -153,3 +155,81 @@ def strip_label_edges(data: HeteroData) -> HeteroData:
...
+            raise ValueError(
+                f"Expected node feature dimension to be an int or None for homogeneous data, got {node_feature_dim} of type {type(node_feature_dim)}"
+            )
+        if isinstance(edge_feature_dim, dict):
+            raise ValueError(
+                f"Expected edge feature dimension to be an int or None for homogeneous data, got {edge_feature_dim} of type {type(edge_feature_dim)}"
+            )
+        # For homogeneous case, the Data object will always have the x or edge_attr fields -- we should check if it is None to see if it set
+        if node_feature_dim and data.x is None:
+            data.x = torch.empty((0, node_feature_dim), dtype=torch.float32).to(device)
```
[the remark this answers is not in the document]
mkolodner-sc: Makes sense, as discussed offline I've added dtype as another item that gets computed and stored it in a `FeatureInfo` dataclass along with the feature_dim field

python/tests/unit/distributed/utils/neighborloader_test.py
mkolodner-sc: Actually, coming back to this, it seems like the unit tests started failing when I switched to use `torch.empty()` -- I inspected the values a bit more closely, and it seems like its a bit unsafe to call torch.empty on a non-empty tensor, there are some elements which end up transiently obtaining some very small values (i.e. 10^-43 small) which causes failures with the `assert_equal`. Going to switch back to `torch.zeros` in this case.

python/tests/unit/distributed/utils/neighborloader_test.py
mkolodner-sc: Sure, added a test for explicit values

python/tests/unit/distributed/utils/neighborloader_test.py
kmontemayor2-sc: > See test_set_missing_features_no_feats

Gotcha thanks!

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -211,3 +217,192 @@ def test_strip_label_edges(self):
...
+                "Node features, no edge features",
+                num_node_features=20,
+                num_edge_features=0,
+            ),
+            param(
+                "Node features, edge features",
+                num_node_features=20,
+                num_edge_features=40,
+            ),
+        ]
```
kmontemayor2-sc, line 238: nit. probably best to test against `dtype` too.

python/tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -211,3 +217,192 @@ def test_strip_label_edges(self):
...
+            )
+        with self.assertRaises(ValueError):
+            set_missing_features(
+                data=HeteroData(),
+                node_feature_info=FeatureInfo(dim=2, dtype=torch.float32),
+                edge_feature_info=FeatureInfo(dim=2, dtype=torch.float32),
+                device=self._device,
+            )
+
+    def test_set_custom_features(self):
```
kmontemayor2-sc, line 383: hmm for maximum coverage we should also have an equivalent of this for hetero data.

python/gigl/distributed/dist_link_prediction_dataset.py
```diff
@@ -217,6 +236,18 @@ def negative_edge_label(
...
+    def node_feature_info(
+        self,
+    ) -> Optional[Union[FeatureInfo, dict[NodeType, FeatureInfo]]]:
+        return self._node_feature_info
+
+    @property
+    def edge_feature_info(
+        self,
+    ) -> Optional[Union[FeatureInfo, dict[EdgeType, FeatureInfo]]]:
+        return self._edge_feature_info
```
kmontemayor2-sc, line 249: nit. We should add doc comments here.

kmontemayor2-sc (commented):
mostly lgtm, I can stamp tomorrow but I'd like to review with fresh eyes (there's been a lot of iteration here, I want to make sure we don't miss anything).

Thanks for. the interations, Matt!

kmontemayor2-sc (approved):
LGTM, I'd prefer if you could address my comments tho

PR #196, opened 2025-07-22
Create Release: Bump + Release KFP + Docker 

svij-sc:
**Scope of work done**

- Allows for us to create + publish all release assets (except for wheels - which are final artifact and [currently a separate workflow](https://github.com/Snapchat/GiGL/blob/f81175116f8b6eb5f5470b0ce830df7ff86e0fbb/.github/workflows/release.yml#L1). We still need to think how/when wheels should be released and should they be manual?)
- In the future, we will add a [cron schedule](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule) to release nightly - so it can be used for e2e canary testing. cc @xgao4-sc  for vis.
- We are also tracking all release artifact paths in `dep_vars.env` now so that future orchestration apis like  `gigl run_dev_gnn_kubeflow_pipeline`, et al can use these directly.  Our components like trainer are now using these as backfall too.

Ran a create_release workflow, the respective PR was created here: https://github.com/Snapchat/GiGL/pull/203


yliu2-sc (approved):
There are quite a bit of TODOs still in this PR

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,187 @@
...
+        gcp_project_id: ${{ secrets.GCP_PROJECT_ID }}
+        workload_identity_provider: ${{ secrets.WORKLOAD_IDENTITY_PROVIDER }}
+        gcp_service_account_email: ${{ secrets.GCP_SERVICE_ACCOUNT_EMAIL }}
+
+    # Capture current version before bumping
+    - name: Get current version
+      id: get_version
+      run: |
+        CURRENT_VERSION=$(python -m scripts.bump_version --get_current_version)
+        echo "current_version=$CURRENT_VERSION" >> $GITHUB_OUTPUT
```
kmontemayor2-sc, line 53: QQ where does the `$GITHUB_OUTPUT` var get defined?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,187 @@
...
+
+    - name: Release GiGL KFP pipeline
+      uses: snapchat/gigl/.github/actions/run-cloud-run-command-on-active-checkout@main
+      with:
+        cmd: |
+          source dep_vars.env && make \
+            DOCKER_IMAGE_DATAFLOW_RUNTIME_NAME_WITH_TAG=\$DEFAULT_GIGL_RELEASE_SRC_IMAGE_CUDA \
+            DOCKER_IMAGE_MAIN_CUDA_NAME_WITH_TAG=\$DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU \
+            DOCKER_IMAGE_MAIN_CPU_NAME_WITH_TAG=\$DEFAULT_GIGL_RELEASE_SRC_IMAGE_DATAFLOW_CPU \
+            compiled_pipeline_path=\$DEFAULT_GIGL_RELEASE_KFP_PIPELINE_PATH \
```
kmontemayor2-sc, line 124: Hmmm any reason this isn't versioned?

dep_vars.env
```diff
@@ -3,6 +3,12 @@ DOCKER_LATEST_BASE_CUDA_IMAGE_NAME_WITH_TAG=us-central1-docker.pkg.dev/external-
 DOCKER_LATEST_BASE_CPU_IMAGE_NAME_WITH_TAG=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/gigl-cpu-base:59e780eeb5831d4ce75747262b8e57b46f3dc17a.24.1
 DOCKER_LATEST_BASE_DATAFLOW_IMAGE_NAME_WITH_TAG=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/gigl-dataflow-base:59e780eeb5831d4ce75747262b8e57b46f3dc17a.24.1
 
+DEFAULT_GIGL_RELEASE_SRC_IMAGE_CUDA=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/src-cuda:latest
+DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/src-cpu:latest
+DEFAULT_GIGL_RELEASE_SRC_IMAGE_DATAFLOW_CPU=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/src-cpu-dataflow:latest
+DEFAULT_GIGL_RELEASE_DEV_WORKBENCH_IMAGE=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/gigl-dev-workbench:latest
+DEFAULT_GIGL_RELEASE_KFP_PIPELINE_PATH=gs://public-gigl/releases/gigl-pipeline-1.yaml
```
kmontemayor2-sc, line 10: why `gigl-pipeline-1.yaml`?

scripts/build_and_push_docker_image.py
```diff
@@ -126,14 +126,24 @@ def build_and_push_image(
...
 
     # Push image if it's not a multi-arch build (multi-arch images are pushed in the build step)
     if not multi_arch:
         push_command = ["docker", "push", image_name]
-        subprocess.run(push_command, check=True)
+        result_push = subprocess.run(
+            push_command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
+        if result_push.returncode != 0:
+            print(result_push.stdout.decode())
```
kmontemayor2-sc, line 143: use logger here?

scripts/build_and_push_docker_image.py
```diff
@@ -126,14 +126,24 @@ def build_and_push_image(
     build_command.append(root_dir.as_posix())
 
     logger.info(f"Running command: {' '.join(build_command)}")
-    subprocess.run(
-        build_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True
+    result = subprocess.run(
+        build_command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
     )
+    if result.returncode != 0:
+        print(result.stdout.decode())
```
kmontemayor2-sc, line 133: ditto on logger.

scripts/bump_version.py
```diff
@@ -1,47 +1,89 @@
+"""
```
kmontemayor2-sc, line 1: Should we note that this script doesn't actually update the docker images or pipeline yaml, just their "labels"?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,187 @@
...
+      run: |
+        git tag "v${{ needs.bump_version.outputs.new_version }}"
+        git push origin "v${{ needs.bump_version.outputs.new_version }}"
+
+    # Create PR to merge release branch back to main
+    - name: Create a new PR
+      shell: bash
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
```
kmontemayor2-sc, line 159: Does this mean in the future we will need to approve the nightly bump every day?

dep_vars.env
```diff
@@ -3,6 +3,12 @@ DOCKER_LATEST_BASE_CUDA_IMAGE_NAME_WITH_TAG=us-central1-docker.pkg.dev/external-
 DOCKER_LATEST_BASE_CPU_IMAGE_NAME_WITH_TAG=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/gigl-cpu-base:59e780eeb5831d4ce75747262b8e57b46f3dc17a.24.1
 DOCKER_LATEST_BASE_DATAFLOW_IMAGE_NAME_WITH_TAG=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/gigl-dataflow-base:59e780eeb5831d4ce75747262b8e57b46f3dc17a.24.1
 
+DEFAULT_GIGL_RELEASE_SRC_IMAGE_CUDA=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/src-cuda:latest
+DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU=us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/src-cpu:latest
```
kmontemayor2-sc, line 7: unrelated to this PR but I'm not sure I like having these "defaults" it seems "unsafe" to have potentially bad code to be run, if we forget to specify a docker container.

I Ideally think we should require specifying docker containers at all times, and these just shouldn't exist?

Though this change is out of the scope of this pr.

scripts/bump_version.py
```diff
@@ -1,47 +1,89 @@
...
+def update_dep_vars_env(
+    cuda_image_name: str,
+    cpu_image_name: str,
+    dataflow_image_name: str,
+    dev_workbench_image_name: str,
+    kfp_pipeline_path: str,
+) -> None:
+    print(
+        f"Updating dep_vars.env with {cuda_image_name}, {cpu_image_name}, {dataflow_image_name}"
+    )
```
kmontemayor2-sc, line 54: should we also put the workbench image and kfp pipeline path here?

And specify which vars each will be signed to?

scripts/bump_version.py
```diff
@@ -54,49 +96,62 @@ def update_pyproject(version: str) -> None:
...
-        base_image=DOCKER_LATEST_BASE_CPU_IMAGE_NAME_WITH_TAG,
-        image_name=dataflow_image,
-        dockerfile_name="Dockerfile.dataflow.src",
-        multi_arch=True,
+    base_image_registry = f"us-central1-docker.pkg.dev/{project}/public-gigl"
+    cuda_image_name = f"{base_image_registry}/src-cuda:{new_version}"
+    cpu_image_name = f"{base_image_registry}/src-cpu:{new_version}"
+    dataflow_image_name = f"{base_image_registry}/src-cpu-dataflow:{new_version}"
+    dev_workbench_image_name = f"{base_image_registry}/gigl-dev-workbench:{new_version}"
+    kfp_pipeline_path = f"gs://{GIGL_PUBLIC_BUCKET_NAME}/releases/gigl-pipeline-{new_version}.yaml"
```
kmontemayor2-sc, line 145: nit. Maybe `releases/pipelines/gigl-pipeline-{version}.yaml`? Maybe we put other things in `releases` too in the future?

scripts/bump_version.py
```diff
@@ -1,7 +1,10 @@
 """
 This script can be used to:
 - Get the current version of GiGL
-- Bump the version of GiGL, and subsequently release the src images and KFP pipeline
+- Bump the version labels of GiGL
+Note: This does not release the src images and KFP pipeline, it only updates the version labels.
+The usage of bump_version.py, and the overall release process of the src images and KFP pipeline
+is handled by the create_release.yml workflow.
```
kmontemayor2-sc, line 7:
```suggestion
is handled by the .github/workflows/create_release.yml workflow.
```

nit.

.github/workflows/create_release.yml
kmontemayor2-sc: no that's what I meant, thanks for the work!

btw 0.0.2.dev20250723232244 is a bit long, do you think it'd be sufficient to have `yyyymmddhh`?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,196 @@
...
+        echo "new_version=$NEW_VERSION" >> $GITHUB_OUTPUT
+        echo "release_branch_name=$RELEASE_BRANCH_NAME" >> $GITHUB_OUTPUT
+        echo "release_version_labels=$RELEASE_VERSION_LABELS" >> $GITHUB_OUTPUT
+        echo "New version: $NEW_VERSION"
+        echo "Release branch name: $RELEASE_BRANCH_NAME"
+        echo "Release version labels: $RELEASE_VERSION_LABELS"
+
+    # Commit version changes to new release branch
+    - name: Commit and push version bump
+      uses: ./.github/actions/commit-and-push
```
kmontemayor2-sc, line 91: should this be the thing pointing to main?

dep_vars.env
kmontemayor2-sc: hmmm, does this mean our "default" will primarily be the nightly release, unless have had a major/minor/patch bump that day?

or are we not going to have the nightly branch merge into main, in which case, shouldn't this be for `0.0.3` or whatever?

scripts/build_and_push_docker_image.py
```diff
@@ -126,14 +126,24 @@ def build_and_push_image(
...
 
     logger.info(f"Running command: {' '.join(build_command)}")
-    subprocess.run(
-        build_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True
+    result = subprocess.run(
+        build_command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
     )
+    if result.returncode != 0:
+        logger.info(result.stdout.decode())
+        raise RuntimeError(f"Docker build failed with exit code {result.returncode}")
```
kmontemayor2-sc, line 134: nit. add the command that failed here (I know there's duplication but imo it's useful for debugging).

ditto later?

kmontemayor2-sc (commented):
Mostly lgtm - the only thing blocking my stamp is this thread now. https://github.com/Snapchat/GiGL/pull/196/files#r2226839931

dep_vars.env
svij-sc: nightly will not merge into main.
Yea, I will do a proper release as a follow up once I am able to get this PR and the follow up and any issues addressed. Which should overwrite these to 0.0.3, etc.

scripts/bump_version.py
kmontemayor2-sc: wait is there a reason we didn't att `releases/`**pipelines**/gigl-pipeline.yaml`?

dep_vars.env
svij-sc: To clarify:

In follow ups:
1. There will be a cron to do a nightly release - this will not merge to main. A branch will be created, release will be made, and branch will be deleted. No PRs will be created. The reason I don't have this right now is strictly to test and make sure this workflow works as it should
2. We can workflow dispatch, from actions tab, an actual release i.e. patch, minor, major upgrade - this will create a branch that will merge to main.

.github/workflows/create_release.yml
svij-sc: could work, until two people try releasing at the same time.
Is this a nit? Id pref just keeping the name as no one is technically going to use nightlys except for our automated testing suites or for beta testing.

scripts/build_and_push_docker_image.py
svij-sc: sure, done

.github/workflows/create_release.yml
svij-sc: thanks, updated

PR #202, opened 2025-07-23
Fix notebook integration tests

kmontemayor2-sc:
Updated examples/toy_visual_example/toy_example_walkthrough.ipynb so that it runs on the `notebook_tests`, see [this run](https://github.com/Snapchat/GiGL/actions/runs/16528919878).

Changes made:
1. If the resource config uri was already on GCS, don't try to reupload it. `file_loader.load` doesn't work with `(GCS, GCS)` uris for some reason, we should probably fix that at some point (I can do so as a follow up)
2. Expose some more vars as env vars so the `%% capture` cells can use them, without them for whatever reason those cells could not read those variables and would just pass in an empty string
3. Swap to using ProcessPoolExecutor for notebook tests so that the environment overrides are isolated. 

Again now passes the tests :)

### DIFF
--- a/examples/toy_visual_example/toy_example_walkthrough.ipynb
+++ b/examples/toy_visual_example/toy_example_walkthrough.ipynb
@@ -1790,22 +1790,22 @@
    "outputs": [],
    "source": [
     "# Copy over resource config to GCS so that training and inference can access it.\n",
-    "# If we are already using a GCS URI, we will just use it as is.\n",
+    "# If we are already using a non-local URI, we will just use it as is.\n",
     "from gigl.common import GcsUri, LocalUri\n",
     "from gigl.src.common.utils.file_loader import FileLoader\n",
     "\n",
     "if isinstance(RESOURCE_CONFIG_PATH, LocalUri):\n",
-    "    GCS_RESOURCE_CONFIG_PATH = GcsUri.join(resource_config.temp_assets_regional_bucket_path, JOB_NAME, \"resource_config.yaml\")\n",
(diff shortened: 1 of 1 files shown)


examples/toy_visual_example/toy_example_walkthrough.ipynb
```diff
@@ -1790,16 +1790,23 @@
...
-    ")\n",
-    "print(f\"GCS_RESOURCE_CONFIG_PATH: {GCS_RESOURCE_CONFIG_PATH}\")"
+    "if isinstance(RESOURCE_CONFIG_PATH, LocalUri):\n",
+    "    GCS_RESOURCE_CONFIG_PATH = GcsUri.join(resource_config.temp_assets_regional_bucket_path, JOB_NAME, \"resource_config.yaml\")\n",
+    "    loader = FileLoader()\n",
+    "    loader.load_file(\n",
+    "        file_uri_src=RESOURCE_CONFIG_PATH,\n",
+    "        file_uri_dst=GCS_RESOURCE_CONFIG_PATH\n",
+    "    )\n",
+    "else:\n",
```
svij-sc, line 1804: this might be an HTTP URI too.

Maybe restructure conditionals so :
```
if GCS_URI_PROVIDED:
   assignment
else:
   file_loader: upload_to_gcs
```

svij-sc: I guess im not sure if all our components works w/ HTTPURI as of now, thus asked to have the conditionals where the `else clause` uploads to GCP

PR #220, opened 2025-07-26
Improving KFP runner CLI

svij-sc:
**Scope of work done**

* You need not specify docker images in KFP runner, the default latest version release are used if not provided
* You also need not compile your own custom src code images now, for simple images where only source code is different you can specify `extra_source_dir` as an argument.
* We now also do a nightly release of GiGL, upon which the API tests run.
    * The nightly release will be subsequently used to run internal benchmark tests


.github/workflows/create_release.yml
```diff
@@ -214,9 +227,10 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        PR_URL=$(gh pr view ${{ needs.bump_version.outputs.release_branch_name }} --json url --jq '.url')
-        echo "::notice title=Release PR URL::$PR_URL"
-        echo "::notice title=Release Branch::${{ needs.bump_version.outputs.release_branch_name }}"
+        if [ "${{ inputs.bump_type }}" != "nightly" ]; then
```
kmontemayor2-sc, line 230: Any reason we check twice here?

.github/workflows/nightly_release_&_test.yml
```diff
@@ -0,0 +1,43 @@
...
+permissions:
+  contents: write
+  pull-requests: write
+  id-token: write
+
+
+# Calls create_release w/ nightly and then run the api tests
+jobs:
+  create-release:
+    uses: ./.github/workflows/create_release.yml
```
kmontemayor2-sc, line 18: revert?

.github/workflows/nightly_release_&_test.yml
kmontemayor2-sc, line 1: Hmm do you think putting an & into the filename like this might cause issues later?

python/gigl/orchestration/Dockerfile.customer_src
```diff
@@ -0,0 +1,13 @@
...
+
+ARG BASE_IMAGE
+FROM $BASE_IMAGE
+
+WORKDIR /gigl
+
+COPY . .
+
+# Find out if there is 'setup.py' or 'pyproject.toml' in the current directory, if so install it
+RUN if [ -f "setup.py" ] || [ -f "pyproject.toml" ]; then pip install -e .; fi
```
kmontemayor2-sc, line 13: Why do we add this? If we want to do this for our code we should do so, customers may or may not want this and should decide on their own if this is needed.

python/gigl/orchestration/img_builder.py
kmontemayor2-sc, line 1: Should we merge this with https://github.com/Snapchat/GiGL/blob/main/scripts/build_and_push_docker_image.py?

python/gigl/orchestration/img_builder.py
```diff
@@ -0,0 +1,146 @@
...
+        dockerfile_path=CUSTOMER_SRC_DOCKERFILE_PATH,
+        context_path=context_path,
+    )
+    logger.info(f"Building and pushing dataflow image to {export_dataflow_image_name}")
+    build_and_push_image(
+        base_image=base_image_dataflow,
+        image_name=export_dataflow_image_name,
+        dockerfile_path=CUSTOMER_SRC_DOCKERFILE_PATH,
+        context_path=context_path,
+    )
```
kmontemayor2-sc, line 75: mild. preference to do this in a multithreaded way, e.g. `with ThreadPoolExecutor() as pool: ...` so that we can speed up builds.

AFAICT doing so is entirely safe.

python/gigl/orchestration/img_builder.py
```diff
@@ -0,0 +1,146 @@
...
+        logger.info(result.stdout.decode())
+        logger.error(f"Command failed: {' '.join(build_command)}")
+        raise RuntimeError(f"Docker build failed with exit code {result.returncode}")
+
+    # Push image if it's not a multi-arch build (multi-arch images are pushed in the build step)
+    if not multi_arch:
+        push_command = ["docker", "push", image_name]
+        result_push = subprocess.run(
+            push_command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
```
kmontemayor2-sc, line 140: we have some util function for this right? `from gigl.common.utils.os_utils import run_command_and_stream_stdout`? Let's use it?

python/gigl/orchestration/kubeflow/runner.py
```diff
@@ -113,12 +117,41 @@ def from_string(s: str) -> Action:
...
-        "container_image_cuda",
-        "container_image_cpu",
-        "container_image_dataflow",
+        "compiled_pipeline_path",
     ]
 )
 
+
+def _assert_required_flags(args: argparse.Namespace) -> None:
+    required_flags: frozenset[str]
```
kmontemayor2-sc, line 126: btw I have a strong preference against pre-declaring variables like this, since it means we can use them without them actually existing later.

Why not remove this have a final else after line 132 to throw, that way `required_flags` should always be defined later.

python/gigl/orchestration/kubeflow/runner.py
```diff
@@ -160,23 +193,32 @@ def _parse_additional_job_args(
...
         help="The docker image name and tag to use for cuda pipeline components ",
     )
     parser.add_argument(
         "--container_image_cpu",
+        default=DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU,
         help="The docker image name and tag to use for cpu pipeline components ",
     )
     parser.add_argument(
         "--container_image_dataflow",
+        default=DEFAULT_GIGL_RELEASE_SRC_IMAGE_DATAFLOW_CPU,
```
kmontemayor2-sc, line 219: Preference against having defaults here - I think this will lead to confusion down the road as people will forget to supply some image and be confused when their code doesn't work.

Why did we swap to having defaults?

python/gigl/orchestration/kubeflow/runner.py
```diff
@@ -235,35 +283,33 @@ def _parse_additional_job_args(
...
-    missing_values: list[str] = []
-    for flag in required_flags:
-        if not hasattr(args, flag):
-            missing_flags.append(flag)
-        elif len(getattr(args, flag)) == 0:
-            missing_values.append(flag)
+    if args.compiled_pipeline_path:
+        compiled_pipeline_path = UriFactory.create_uri(args.compiled_pipeline_path)
+    else:
+        compiled_pipeline_path = DEFAULT_KFP_COMPILED_PIPELINE_DEST_PATH
```
kmontemayor2-sc, line 291: this actually could be a default arg tho right?

python/gigl/orchestration/img_builder.py
kmontemayor2-sc: Oh I see we did, can we just delete that script all together?

python/gigl/orchestration/img_builder.py
```diff
@@ -0,0 +1,146 @@
...
+    DOCKER_LATEST_BASE_DATAFLOW_IMAGE_NAME_WITH_TAG,
+)
+from gigl.common.logger import Logger
+
+logger = Logger()
+
+
+CUSTOMER_SRC_DOCKERFILE_PATH = (
+    Path(__file__).parent.resolve() / "Dockerfile.customer_src"
+).as_posix()
```
kmontemayor2-sc, line 18: I feel like this should also be an input argument, I'd imagine users would want to customize the docker file easily.

python/gigl/orchestration/kubeflow/runner.py
```diff
@@ -211,17 +248,28 @@ def _parse_additional_job_args(
...
     parser.add_argument(
         "--compiled_pipeline_path",
         help="A custom URI that points to where you want the compiled pipeline is to be saved to."
         + "In the case you want to run an existing pipeline that you are not compiling, this is the path to the compiled pipeline.",
-        default=DEFAULT_KFP_COMPILED_PIPELINE_DEST_PATH.uri,
+    )
+    parser.add_argument(
+        "--export_docker_artifact_registry",
+        help="The docker artifact registry to push the customer src images to.",
+        default=None,
```
kmontemayor2-sc, line 265: should we give an example of what we expect this to look like, e.g. `us-central1-docker.pkg.dev/external-snap-ci-github-gigl/` vs `us-central1-docker.pkg.dev/external-snap-ci-github-gigl/public-gigl/`

Makefile
```diff
@@ -136,6 +136,10 @@ precondition_tests:
 	python testing/dep_vars_check.py
 
 
+run_api_test:
+	cd testing/api_test && make run_api_test
```
kmontemayor2-sc, line 140: are we not concerned about the "subshell makefile" stuff here?

.github/workflows/create_release.yml
```diff
@@ -18,6 +18,18 @@ on:
           - patch
           - nightly
 
+  workflow_call:
+    inputs:
+      bump_type:
+        description: 'Type of version bump'
+        required: true
+        type: string
```
kmontemayor2-sc, line 26: is it possible for us to add the allowed options here?

python/pyproject.toml
```diff
@@ -173,6 +173,7 @@ Homepage = "https://github.com/snapchat/gigl"
 
 [project.scripts]
 gigl-post-install = "gigl.scripts.post_install:main"
+gigl = "gigl.scripts.gigl_cli:main"
```
kmontemayor2-sc, line 176: this file doesn't exist in this pr?

python/gigl/orchestration/img_builder.py
svij-sc: Agreed, we can enable that option in the future too.

Makefile
svij-sc: not here, its because we are running a seperate make project.
Different then running a make command in the same project.

.github/workflows/create_release.yml
svij-sc: not allowed for `workflow_call`, I can add a comment.

.github/workflows/create_release.yml
svij-sc: this if statement is inside the step, so the step still gets executed except for whats inside the if, vs one on line 203 skips the entire step since its on the step level.

.github/workflows/nightly_release_&_test.yml
svij-sc: I can change to `and`

python/gigl/orchestration/kubeflow/runner.py
kmontemayor2-sc: > doing something like on line 126 is just for type checking and doesnt actually define the var. example:
I know what it's for - the issue is that it *doesn't* define the var, and can lead to those `foo is not defined` errors which are annoying and that we can catch statically (but not with the pre-declaration!)

Anyways I don't think that this line is required here for mypy, or is it? 

it's possible to get mypy to complain about this case tho, see an example with [pre-declaration](https://mypy-play.net/?mypy=latest&python=3.12&flags=strict&enable-error-code=possibly-undefined&gist=b7e2adc0b75bd87a39f859212a2643f2) vs, [without](https://mypy-play.net/?mypy=latest&python=3.12&flags=strict&enable-error-code=possibly-undefined&gist=354fe8ea86251aaab2301837660cf36d).

> Pretty sure I have seen this pattern in official python docs.

Maybe, but sometimes python makes weird decisions.

> line 136 should fail, if u want i can explicity throw another exception to be "clean"

Yes that'd be nice - because it'd give an error message for what failed (unknown action).

python/gigl/orchestration/kubeflow/runner.py
kmontemayor2-sc: hmmm, I was originally thinking you just reference the make file vars when you launch but I guess they're not  available right now in `testing/api_test/Makefile`

Again I do think that exposing these defaults will cause us headaches down the road, and would prefer if we didn't have them if possible.

python/gigl/orchestration/kubeflow/runner.py
kmontemayor2-sc: >     # for compile action this is a required flag so we cannot provide it a default value.

Does the code fail if we put a default here? 

```
# args.py
import argparse


parser = argparse.ArgumentParser(description="Process some integers.")
parser.add_argument('--i', type=int, default=1, help='an integer for the accumulator')

args = parser.parse_args()

print(f"{args.i=} is the value of i")
print(f"{hasattr(args, 'i')=} is the value of hasattr(args, 'i')")
print(f"{getattr(args, 'i')=} is the value of i")


python args.py 
args.i=1 is the value of i
hasattr(args, 'i')=True is the value of hasattr(args, 'i')
getattr(args, 'i')=1 is the value of i
````

File: python/gigl/orchestration/img_builder.py [hunk 14]
```diff
@@ -0,0 +1,139 @@
...
+    DOCKER_LATEST_BASE_DATAFLOW_IMAGE_NAME_WITH_TAG,
+)
+from gigl.common.logger import Logger
+from gigl.common.utils.os_utils import run_command_and_stream_stdout
+
+logger = Logger()
+
+
+CUSTOMER_SRC_DOCKERFILE_PATH = (
+    Path(__file__).parent.resolve() / "Dockerfile.customer_src"
```
kmontemayor2-sc, line 17: qq for my own learning, why do we something do `Path.parent.resolve()` and `Path.resolve().parent`?

.github/workflows/create_release.yml
```diff
@@ -5,6 +5,7 @@ name: Create Release
 # 2. TODO: (svij)Cron trigger: Nightly build
```
kmontemayor2-sc, line 5: remove?
```suggestion
```

kmontemayor2-sc (approved):
prefer if you address my comments but this should be fine (provided we revert https://github.com/Snapchat/GiGL/pull/220/files#diff-148b43903866b0ecfa81aa1732edc570eef44a30b9f67a33d0dd505da1a6b5d7R17)

python/gigl/orchestration/kubeflow/runner.py
svij-sc: > Again I do think that exposing these defaults will cause us headaches down the road, and would prefer if we didn't have them if possible.

I want to understand more offline. We need some way to package these with the GiGL releases - I dont see a way around this. Maybe there could be a discussion on how we expose them and where they are exposed?

Right now as a user i.e. `api_test` reads this from the GiGL version installed by the host env calling this test.

python/gigl/orchestration/img_builder.py
svij-sc: OOM

python/gigl/orchestration/kubeflow/runner.py
kmontemayor2-sc: > I want to understand more offline. We need some way to package these with the GiGL releases - I dont see a way around this. Maybe there could be a discussion on how we expose them and where they are exposed?

Why do we need to package them with GiGL releases?

Though I am ok with packaging them (if needed). What I do not want is to have the cli flags default to them. e.g. your suggestion like

```
The alternative being, if I do pip install GiGL, then I have to do following to kick off a job?

CONTAINER_IMAGE_DATAFLOW=$(python -c "from gigl.common.constants import DEFAULT_GIGL_RELEASE_SRC_IMAGE_DATAFLOW_CPU; 

...etc

print(DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU)")
python -m gigl.orchestration.kubeflow.runner --container_image_dataflow=${CONTAINER_IMAGE_DATAFLOW}

... etc
```

would be fine imo.

python/gigl/orchestration/img_builder.py
svij-sc: the wheel itself is not going to have a python dir or a top level dir.
The wheel only packages `gigl` and `snapchat` dirs.
```
[tool.setuptools.packages.find]
where = ["."]  # list of folders that contain the packages (["."] by default)
include = ["gigl*", "snapchat*"]  # Include only packages that match the gigl* pattern
```

Also it doesnt make sense for our gigl python APIs to be calling something that lives outside of the python codebase i.e. Dockerfiles in top level container dir - also lets not loose the convenience of having those functions in `build_and_push_docker_image` vs. trying to pass around the dockerfile names/paths everywhere we need to compile these images.

python/gigl/orchestration/kubeflow/runner.py
svij-sc: I don't understand ^.
We need to enforce that the user provide a value for this in some cases and in other cases we can default to some value.

You could say then the argument becomes subjective?:
1. assert `some_arg != DEFAULT_KFP_COMPILED_PIPELINE_DEST_PATH` for case A, B
2.  for case A, B assert some_arg has some value; assign default value for some_arg if not provided

I pref 2.

python/gigl/orchestration/kubeflow/runner.py
svij-sc: From an api usability perspective, if someone is using a released version of GiGL, this is unideal:
```
CONTAINER_IMAGE_DATAFLOW=$(python -c "from gigl.common.constants import DEFAULT_GIGL_RELEASE_SRC_IMAGE_DATAFLOW_CPU; 
CONTAINER_IMAGE_CUDA=$(python -c "from gigl.common.constants import DEFAULT_GIGL_RELEASE_SRC_IMAGE_CUDA; 
CONTAINER_IMAGE_CPU=$(python -c "from gigl.common.constants import DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU; 


python -m gigl.orchestration.kubeflow.runner \
    --container_image_dataflow=${CONTAINER_IMAGE_DATAFLOW} \
    --container_image_cuda=${CONTAINER_IMAGE_CUDA} \
    --container_image_cpu=${CONTAINER_IMAGE_CPU} \
    --job_name='$${job_name}' \
    --start_at='config_populator' \
    --task_config_uri='$${task_config_uris[$$i]}' \
    --resource_config_uri='$${resource_config_uris[$$i]}'
```

Even if it was only passing one of these arguments i.e. `COMPILED_KFP_PIPELINE_PATH`, its still unideal.
We have released a version of GiGL, the package should know what the release paths are and be able to use them directly. 

Even though like you mentioned, we might be preventing some accidental issues by forcing them to pass in variables; but on the other we are making our apis more complicated. Id opt for making our apis less complicated most times here in the tradeoff otherwise it will turn away folks.

svij-sc: Lets chat more offline, I will resolve this for now as it seems non blocking based on approval.

python/gigl/orchestration/kubeflow/runner.py
svij-sc: Will resolve this for now to get the PR merged since this seems non-blocking.
I am happy to address post if there are still concerns here after

PR #233, opened 2025-07-29
Add some 'shared' KGE utils

nshah171:
These utils are used in KGE experimentation and are likely to be generically useful for other graph applications (outside KGE).


python/gigl/experimental/knowledge_graph_embedding/common/torchrec/utils.py
```diff
@@ -0,0 +1,170 @@
...
+    Args:
+        model: The model to be wrapped.
+        device: The device to use for the model.
+        sharding_plan: An optional ShardingPlan to use for the DistributedModelParallel.
+    Returns:
+        The model wrapped in DistributedModelParallel if in a distributed environment,
+        otherwise the model itself.
+    """
+
+    if torch.distributed.is_initialized():
```
svij-sc, line 48: im assuming this also only works on nccl backend?
i.e. both cuda needs to be available and nccl backend enabled?

python/gigl/experimental/knowledge_graph_embedding/common/torchrec/utils.py
```diff
@@ -0,0 +1,170 @@
...
+        # Build a sharding plan
+        logger.info("***** Wrapping in DistributedModelParallel *****")
+        logger.info(f"Model before wrapping: {model}")
+        model = DistributedModelParallel(
```
svij-sc, line 52: QQ: I see there are a lot more configurable params here.
Curious why we only chose to parameterize `sharding_plan` ?

python/gigl/experimental/knowledge_graph_embedding/common/torchrec/utils.py
```diff
@@ -0,0 +1,170 @@
...
+            module=model,
+            device=device,
+            plan=sharding_plan,
+        )
+        logger.info(f"Model after wrapping: {model}")
+
+    return model
+
+
+def get_sharding_plan(
```
svij-sc, line 62: I am suprised `DistributedModelParallel` doesnt do this for us already?
Is there something here that I am not seeing that would be model/business logic specific?

svij-sc (approved):
Still working through this, but approving ot unblock

python/gigl/experimental/knowledge_graph_embedding/common/torchrec/utils.py
nshah-sc: no, this also works without nccl and cuda.  e.g. this could also be a gloo backend and device="gpu".  hence, just need to check for distributed existance

yliu2-sc (approved):

PR #234, opened 2025-07-29
Update Tabularized Notebook to be more aligned with In-Memory notebook (1/2)

mkolodner-sc:
**Scope of work done**

- Base PR which contained originally contained discussion for removing outputs from tabularized notebook, after some discussion it was decided we would not take this approach.


yliu2-sc (approved):
Can we remove the %%capture as well. And just add notes at the very top, you can ctrl/cmd + A, right click, and `enable scrolling for outputs`

PR #235, opened 2025-07-29
Add experimental parsing libs to parse KGE configs

nshah171:
Add class structure which is used to parse the KGE experiment configs.  These interface w/ hydra + omegaconf to create a structured configuration dataclass which is used in the modeling/training code.


svij-sc (commented):
Can we add doc strings to @dataclass that are missing it?

python/gigl/experimental/knowledge_graph_embedding/lib/config/evaluation.py
```diff
@@ -0,0 +1,26 @@
...
+from applied_tasks.knowledge_graph_embedding.lib.config.training import SamplingConfig
+
+
+@dataclass
+class EvaluationPhaseConfig:
+    dataloader: DataloaderConfig = field(default_factory=DataloaderConfig)
+    step_frequency: Optional[int] = None
+    num_batches: Optional[int] = None
+    hit_rates_at_k: List[int] = field(default_factory=lambda: [1, 10, 100])
+    sampling: SamplingConfig = field(default_factory=SamplingConfig)
```
svij-sc, line 16: would be nice to explain what each of these arguments do w/ docstrings.
Same below and above.

python/gigl/experimental/knowledge_graph_embedding/lib/config/model.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@dataclass
+class ModelConfig:
+    node_embedding_dim: int = 128
+    embedding_similarity_type: SimilarityType = SimilarityType.COSINE
+    src_operator: OperatorType = OperatorType.IDENTITY
+    dst_operator: OperatorType = OperatorType.IDENTITY
+
+    # Below fields are populated at runtime.
```
svij-sc, line 19: does this mean that user does not provide these?

python/gigl/experimental/knowledge_graph_embedding/lib/config/model.py
```diff
@@ -0,0 +1,24 @@
...
+    training_sampling: Optional[SamplingConfig] = None
+    validation_sampling: Optional[SamplingConfig] = None
+    testing_sampling: Optional[SamplingConfig] = None
```
svij-sc, line 22: btw, why do we have different sampling configs for train/val/test?

python/gigl/experimental/knowledge_graph_embedding/lib/config/training.py
```diff
@@ -0,0 +1,63 @@
...
+        default_factory=lambda: OptimizerParamsConfig(lr=0.01, weight_decay=0.001)
+    )
+
+
+@dataclass
+class DistributedConfig:
+    num_processes_per_machine: int = (
+        torch.cuda.device_count() if torch.cuda.is_available() else 1
+    )
+    storage_reservation_percentage: float = 0.1
```
svij-sc, line 33: could we also attach an example in this PR of what a saturated yaml might look like for these configs?

yliu2-sc (approved):

PR #239, opened 2025-07-30
Add KGE constants

nshah171:
Add constants filepaths for experimental KGE work.

### DIFF
--- a/python/gigl/experimental/knowledge_graph_embedding/lib/constants/gcs.py
+++ b/python/gigl/experimental/knowledge_graph_embedding/lib/constants/gcs.py
@@ -1,0 +1,134 @@
+import gigl.src.common.constants.gcs as gcs_constants
+from gigl.common import GcsUri
+from gigl.src.common.types import AppliedTaskIdentifier
+from gigl.src.common.types.graph_data import EdgeType
+
+
+def get_applied_task_staging_path(
+    applied_task_identifier: AppliedTaskIdentifier,
+) -> GcsUri:
+    """
+    Returns the GCS URI for the staging/temp path of the applied task.
+
+    Args:
+        applied_task_identifeir: The identifier of the applied task.
+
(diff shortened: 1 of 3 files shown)


python/gigl/experimental/knowledge_graph_embedding/lib/constants/local.py
```diff
@@ -0,0 +1,5 @@
+from pathlib import Path
+
+PROJECT_ROOT_DIR = Path(__file__).resolve().parent.parent.parent
```
svij-sc, line 3: comment on what the dir is so reader does not need to do the work to resolve it.

python/gigl/experimental/knowledge_graph_embedding/lib/constants/gcs.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Returns the GCS URI for the staging/temp path of the applied task.
+
+    Args:
+        applied_task_identifeir: The identifier of the applied task.
+
+    Returns:
+        GcsUri: The GCS URI for the staging path of the applied task.
+    """
+    return gcs_constants.get_applied_task_temp_regional_gcs_path(
```
svij-sc, line 19: nit: should we prefix w/  `kge` or something similar?

svij-sc (approved):

python/gigl/experimental/knowledge_graph_embedding/lib/constants/gcs.py
nshah-sc: i wanted to avoid naming all these new constructs all with kge and rather resort on using the directory structure to make this somewhat implicit (and in usage it can be something like `import gigl.experimental.kge.lib.constants.local as kge_local_constants` to avoid padding all the names.   happy to revisit if it gets confusing later on.

PR #241, opened 2025-07-30
make README prettier

kmontemayor2-sc:
It's quite inconvenient that you can't get directly to the docs by clicking from the README...

Docs self-linking is ok, so I exposed the link and now people can click it easily :) And it makes the README (which is the first thing people see when they lookup GiGL) look nicer too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,10 @@
 
 GiGL is an open-source library for training and inference of Graph Neural Networks at very large (billion) scale.
 
-```{only}
-::not html
 
 For best experience in reading GiGL documentation, visit our website:
-https://snapchat.github.io/GiGL/
+<a href="https://snapchat.github.io/GiGL"/>https://snapchat.github.io/GiGL/</a>
 
-```
 
 ## Key Features 🌟
 
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,8 +1,0 @@
-.. only:: not html
-
-   This is a placeholder for GiGL's API documentation:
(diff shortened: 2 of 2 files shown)


docs/api/index.rst
```diff
@@ -1,8 +0,0 @@
-.. only:: not html
```
svij-sc, line 1: lets put this file back in.
We had this as 
1. pointer that this folder exists
2. our links on github pointing to api still function

README.md
```diff
@@ -2,13 +2,10 @@
 
 GiGL is an open-source library for training and inference of Graph Neural Networks at very large (billion) scale.
 
-```{only}
-::not html
 
 For best experience in reading GiGL documentation, visit our website:
-https://snapchat.github.io/GiGL/
+<a href="https://snapchat.github.io/GiGL"/>https://snapchat.github.io/GiGL/</a>
```
svij-sc, line 7:
```suggestion
For best experience in reading GiGL documentation, visit our [website](https://snapchat.github.io/GiGL/)
```

docs/api/index.rst
kmontemayor2-sc: `make build_docs` deleted this - my local index.html seems to be working without this file.

I can leave it in as an empty file if you'd like but I don't see much use, I don't see the point of #2 here, what sort of link are you referring to?

kmontemayor2-sc: I played around a bit locally - I didn't see any side effects, worst case this is easy to revert :P

PR #243, opened 2025-07-30
Enforce exported config is to GCS.

kmontemayor2-sc:
As disscussed offline, require GCS exports for the generated resource config.

Also update notebooks to run the scripts on-the-fly. 

We will need to revert some of these changes in the future


examples/tutorial/KDD_2025/lab_instructions.md
```diff
@@ -56,8 +56,8 @@ python -m scripts.bootstrap_resource_config \
   --template_resource_config_uri="examples/tutorial/KDD_2025/resource_config.yaml"
 ```
```
svij-sc, line 57: cc @yozenliu 
to make this usable outside off QWIK_LABS we can directly use the following env variable for project: `$GOOGLE_CLOUD_PROJECT` - this is available on all GCP VMs

Also, QWIK_LABS_USER can be replaced with $GOOGLE_CLOUD_PROJECT for creating the buckets; this simplifies things and again, makes it usable for all GCP setups.

kmontemayor2-sc: $GOOGLE_CLOUD_PROJECT seems to no exist on my GCP vm, saddly.

> Also, QWIK_LABS_USER can be replaced with $GOOGLE_CLOUD_PROJECT for creating the buckets; this simplifies things and again, makes it usable for all GCP setups.

Agreed but we'd need to modify the post-startup script so let's leave it as-is for now?.

Auto-inferred the project tho.

scripts/bootstrap_resource_config.py
```diff
@@ -280,6 +280,12 @@ def assert_gcs_bucket_exists(bucket_name: str):
         or default_resource_config_dest_path
     )
 
+    file_uri_dest = UriFactory.create_uri(uri=destination_file_path)
+    if not isinstance(file_uri_dest, GcsUri):
+        raise ValueError(
+            f"Destination file path must be a GCS URI starting with 'gs://'. Please provide a valid GCS path. Recived URI: {file_uri_dest.uri}"
```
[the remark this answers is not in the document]
kmontemayor2-sc: Oops I could have sworn I had that originally...

Updated.

kmontemayor2-sc:
>Lets also update the notebook cells ?

I'd prefer not to do this - we do require the resource config to be on GCS,. and `GIGL_TEST_DEFAULT_RESOURCE_CONFIG` *may* point to a local uri (even if it wont in practice for the tutorial run through.

Does it hurt to have the additional few lines of code here?

svij-sc (approved):
> Does it hurt to have the additional few lines of code here?

Readability yes, otherwise no - the user can always change cells to what they want if they dont conform to what we have. (non-blocking)


Have we at least validated that we can  the two notebooks after this change?

kmontemayor2-sc:
> Have we at least validated that we can the two notebooks after this change?

I'll do some testing (on qwiklabs ideally...) with this change and without the notebooks updated.

If the notebooks break when updated I will not update them.

yliu2-sc (approved):

svij-sc:
FYI, I am doing a release: https://github.com/Snapchat/GiGL/pull/253
I can re-release after this is in

nshah171 (approved):

PR #255, opened 2025-08-01
Include script folder in the workbench image

svij-sc:
**Scope of work done**


svij-sc:
> This is fine but I don't think it will resolve the root issue here
It will get us closer ... docker not being available to come.

yliu2-sc (approved):

PR #267, opened 2025-08-08
Add `/help` command.

kmontemayor2-sc:
Added /help command so we can see what we can run. 

See https://github.com/Snapchat/GiGL/pull/267#issuecomment-3176785175 for example output.

This parses `.github/workflows/on-pr-comment.yml` and outputs some (hopefully useful!) help text.

### DIFF
--- a/.github/scripts/get_help_text.py
+++ b/.github/scripts/get_help_text.py
@@ -83,7 +83,6 @@ def main():
         with open(github_output, "a") as f:
             # Use EOF delimiter for multiline output
             f.write(f"help_message<<EOF\n{help_text}\nEOF\n")
-            f.write(f"some_value=20\n")
     else:
         # Fallback: print to stdout for local testing
         print(help_text)
--- a/.github/scripts/get_help_text.py
+++ b/.github/scripts/get_help_text.py
@@ -12,10 +12,10 @@ import yaml
 
 def get_help_text():
     """Generate help text by parsing the workflow file."""
-    workflow_path = Path(".github/workflows/")
+    workflow_path = Path(".github/workflows")
     commands = []
 
-    for workflow_file in workflow_path.rglob("*.yml"):
+    for workflow_file in workflow_path.glob("**/*.yml"):
         try:
             # Read and parse the YAML file
             with open(workflow_file, "r") as f:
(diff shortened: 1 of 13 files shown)


.github/workflows/on-pr-comment.yml
```diff
@@ -13,6 +13,80 @@ permissions:
...
 
 jobs:
+  help:
+    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/help') }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
```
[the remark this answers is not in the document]
svij-sc: Id imagine you have tried doing something like following and it doesnt work?


```python
# python code ....
# help_message = ....

github_output = os.getenv("GITHUB_OUTPUT")
if github_output:
    with open(github_output, "a") as f:
        f.write(f"help_message={help_message}\n")
```

Subsequently, the action can change

```yaml
    - name: Checkout code
       uses: actions/checkout@v4
    - name: Set up Python
        uses: actions/setup-python@v4
        with:
          python-version: '3.9.13'
     - name: Parse available commands
       id: parse_commands
       run: |
             python .github/scripts/parse.py
      ...
```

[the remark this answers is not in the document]
kmontemayor2-sc: yeah we can use the `GITHUB_OUTPUT`.

I was more posting about using some new wrapper library for the GH api.

.github/scripts/get_help_text.py
```diff
@@ -0,0 +1,93 @@
...
+import os
+import re
+from pathlib import Path
+
+import yaml
+
+
+def get_help_text():
+    """Generate help text by parsing the workflow file."""
+    workflow_path = Path(".github/workflows/on-pr-comment.yml")
```
svij-sc, line 15: should this discover all `.ymls`  and parse them?
Less prone to breaking in the future i.e. renaming and/or we add new workflows that have comment triggers.

.github/scripts/get_help_text.py
```diff
@@ -0,0 +1,93 @@
...
+def main():
+    """Main function to generate help text and set GitHub output."""
+    help_text = get_help_text()
+
+    # Set the output for GitHub Actions
+    github_output = os.environ.get("GITHUB_OUTPUT")
+    if github_output:
+        with open(github_output, "a") as f:
+            # Use EOF delimiter for multiline output
+            f.write(f"help_message<<EOF\n{help_text}\nEOF\n")
```
svij-sc, line 86: could we double check if we append in this action another varaible to this file that it can be read as well by next steps? 

i.e. line 87:
```
f.write("some_var_ensure_can_be_read=20"
```

kmontemayor2-sc: Tested see https://github.com/Snapchat/GiGL/pull/267#issuecomment-3177605156

.github/scripts/get_help_text.py
kmontemayor2-sc: Sure, done.

kmontemayor2-sc:
> Looks like the bot is commenting twice
> 
> Screenshot 2025-08-12 at 11 51 23 AM

yeah I did that intentionally so I could test that future outputs would still work https://github.com/Snapchat/GiGL/pull/267#discussion_r2268515186

The changes that caused that e.g. [1], [2] are no longer present

[1]: https://github.com/Snapchat/GiGL/pull/267/commits/262ba94761cb20518a983ba76b270db81d8ff376#diff-36870ed5102c99f0dc4389f7d11f519e3d74852bfb4a8530468ca294b9b9e712R86
[2]: https://github.com/Snapchat/GiGL/pull/267/commits/262ba94761cb20518a983ba76b270db81d8ff376#diff-062547b8745e455f2dd42a830138617b653d9b486eb29085e4549d1c919a38baR147-R151

kmontemayor2-sc:
> /help

this wont work until merge, I was testing with the on-push stuff

PR #273, opened 2025-08-14
Add HashedNodeSplitter

mkolodner-sc:
**Scope of work done**

- Adds the NodeAnchorSplitter/HashedNodeAnchorSplitter classes which aims to split node ids instead of edge indices, which the HashedNodeAnchorLinkSplitter is responsible for.
- Re-uses constructs of the HashedNodeLinkSplitter where possible
- Adds relevant tests

This implementation provides additional value over the [GLT splitting implementation](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/data/dataset.py#L504-L515) by making less rigorous assumptions about the node ids that are provided to the splitter, such as not requiring every node to be split upon (by requiring as an argument the total number of nodes N and splitting the ids from 0 to N). This may be useful in cases where we may only want to provide a subset of nodes to the splitter when we have an imbalanced dataset. Additionally, if we want to include a node id multiple times, it ensures that it will be hashed to the same split each time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support `URI / URI_LIKE`, similar to `pathlib.Path`.
 
 ### Added
+- Added support for NodeAnchorSplitter
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support `URI / URI_LIKE`, similar to `pathlib.Path`.
 
 ### Added
+
 - Added support for NodeAnchorSplitter
 
 ### Changed
--- a/python/gigl/utils/data_splitters.py
+++ b/python/gigl/utils/data_splitters.py
@@ -11,6 +11,7 @@ from typing import (
     Tuple,
     Union,
     overload,
+    runtime_checkable,
 )
 
 import torch
@@ -32,6 +33,76 @@ logger = Logger()
 PADDING_NODE: Final[torch.Tensor] = torch.tensor(-1, dtype=torch.int64)
 
 
+def _create_distributed_splits_from_hash(
+    nodes_to_select: torch.Tensor,
+    hash_values: torch.Tensor,
+    num_val: float,
+    num_test: float,
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """
+    Creates train, val, test splits from hash values using distributed coordination.
+
+    This function performs the complete splitting workflow:
+    1. Validates that a distributed process group is initialized
+    2. Normalizes hash values globally across all processes
+    3. Creates train/val/test splits from the normalized hash values
+
+    Args:
+        nodes_to_select (torch.Tensor): The nodes to split.
+        hash_values (torch.Tensor): Raw hash values for the nodes.
+        num_val (float): Percentage of nodes for validation.
+        num_test (float): Percentage of nodes for testing.
+
+    Returns:
+        Tuple of (train_nodes, val_nodes, test_nodes).
+
+    Raises:
+        RuntimeError: If no distributed process group is found.
+    """
+    # Validate distributed process group
+    if not torch.distributed.is_initialized():
+        raise RuntimeError(
+            f"Splitter requires a Torch distributed process group, but none was found. "
+            "Please initialize a process group (`torch.distributed.init_process_group`) before using this splitter."
+        )
+
+    # Normalize hash values globally across all processes
+    min_hash_value, max_hash_value = map(torch.Tensor.item, hash_values.aminmax())
+
+    all_max_and_mins = [
+        torch.zeros(2, dtype=torch.int64)
+        for _ in range(torch.distributed.get_world_size())
+    ]
+    torch.distributed.all_gather(
+        all_max_and_mins,
+        torch.tensor([max_hash_value, min_hash_value], dtype=torch.int64),
+    )
+    global_max_hash_value = max_hash_value
+    global_min_hash_value = min_hash_value
+    for max_and_min in all_max_and_mins:
+        global_max_hash_value = max(
+            global_max_hash_value, max_and_min[0].item()
+        )
+        global_min_hash_value = min(
+            global_min_hash_value, max_and_min[1].item()
+        )
+
+    normalized_hash_values = (hash_values - global_min_hash_value) / global_max_hash_value
+
+    # Create splits from normalized hash values
+    test_inds = normalized_hash_values >= 1 - num_test # 1 x M
+    val_inds = (normalized_hash_values >= 1 - num_test - num_val) & ~test_inds # 1 x M
+    train_inds = ~test_inds & ~val_inds # 1 x M
+
+    # Apply masks to select nodes
+    train = nodes_to_select[train_inds] # 1 x num_train_nodes
+    val = nodes_to_select[val_inds] # 1 x num_val_nodes
+    test = nodes_to_select[test_inds] # 1 x num_test_nodes
+
+    return train, val, test
+
+
+@runtime_checkable
 class NodeAnchorLinkSplitter(Protocol):
     """Protocol that should be satisfied for anything that is used to split on edges.
 
@@ -68,6 +139,39 @@ class NodeAnchorLinkSplitter(Protocol):
 
     @property
     def should_convert_labels_to_edges(self):
+        ...
+
+
+@runtime_checkable
+class NodeAnchorSplitter(Protocol):
+    """Protocol that should be satisfied for anything that is used to split on nodes directly.
+
+    Args:
+        node_ids: The node IDs to split on. 1D tensor for homogeneous or mapping for heterogeneous
+    Returns:
+        The train (1 x X), val (1 x Y), test (1 x Z) nodes. X + Y + Z = N
+    """
+
+    @overload
+    def __call__(
+        self,
+        node_ids: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        ...
+
+    @overload
+    def __call__(
+        self,
+        node_ids: Mapping[NodeType, torch.Tensor],
+    ) -> Mapping[NodeType, Tuple[torch.Tensor, torch.Tensor, torch.Tensor]]:
+        ...
+
+    def __call__(
+        self, *args, **kwargs
+    ) -> Union[
+        Tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+        Mapping[NodeType, Tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
+    ]:
         ...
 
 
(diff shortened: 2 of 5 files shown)


python/gigl/utils/data_splitters.py
```diff
@@ -11,6 +11,7 @@
     Tuple,
     Union,
     overload,
+    runtime_checkable,
```
mkolodner-sc, line 14: We add the `runtime_checkable` argument so that we are able to check `isinstance()` with a protocol field. Otherwise, we may run into issue: `Only @runtime_checkable protocols can be used with instance and class checks `

kmontemayor2-sc (commented):
Thanks Matt!

python/gigl/utils/data_splitters.py
```diff
@@ -379,6 +439,116 @@ def should_convert_labels_to_edges(self):
...
+        ...
+
+    def __call__(
+        self,
+        node_ids: Union[torch.Tensor, Mapping[NodeType, torch.Tensor]],
+    ) -> Union[
+        Tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+        Mapping[NodeType, Tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
+    ]:
+        node_ids_dict: Mapping[NodeType, torch.Tensor]
```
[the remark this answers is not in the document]
mkolodner-sc: Was able to fix by switching `Mapping` -> `dict` in this file

python/gigl/utils/data_splitters.py
```diff
@@ -546,6 +716,18 @@ def _check_edge_index(edge_index: torch.Tensor):
         raise ValueError("Expected a dense tensor. Received a sparse tensor.")
 
 
+def _check_node_ids(node_ids: torch.Tensor):
```
yliu2-sc, line 719: nit: more explicit function naming

yliu2-sc (approved):
LGTM

python/gigl/utils/data_splitters.py
```diff
@@ -32,6 +32,73 @@
...
+    ]
+    torch.distributed.all_gather(
+        all_max_and_mins,
+        torch.tensor([max_hash_value, min_hash_value], dtype=torch.int64),
+    )
+    global_max_hash_value = max_hash_value
+    global_min_hash_value = min_hash_value
+    for max_and_min in all_max_and_mins:
+        global_max_hash_value = max(global_max_hash_value, max_and_min[0].item())
+        global_min_hash_value = min(global_min_hash_value, max_and_min[1].item())
```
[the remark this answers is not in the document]
mkolodner-sc: I think the current implementation is more memory performant, since the `.aminmax()` solution would require converting `all_max_and_mins` from a `list[tensor]` to a `tensor` which would require a `torch.cat` or `torch.stack` call. Since the world size is small, I'd like to leave this as is.

python/gigl/utils/data_splitters.py
```diff
@@ -546,6 +723,18 @@ def _check_edge_index(edge_index: torch.Tensor):
         raise ValueError("Expected a dense tensor. Received a sparse tensor.")
 
 
+def _check_node_ids_are_valid(node_ids: torch.Tensor):
```
kmontemayor2-sc, line 726: again I am not a fan of the `_valid` things since I feel like the `check` implies testing for validity, and just putting "valid" in the name says nothing about what "valid" means.

Like, what does adding `_valid` give us here? Do `check_foo` and `check_foo_valid` really mean different things?

python/gigl/utils/data_splitters.py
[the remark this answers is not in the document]
mkolodner-sc: Actually looks like order swapping works in this case, thanks!

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -728,6 +779,216 @@ def test_get_padded_labels(self, _, node_ids, topo, expected):
...
+                node_ids=torch.arange(20, dtype=torch.int64),
+                hash_function=_fast_hash,
+                val_num=0.1,
+                test_num=0.1,
+                expected_train=torch.tensor(
+                    [0, 1, 2, 3, 4, 5, 6, 8, 10, 11, 12, 13, 16, 17, 19],
+                    dtype=torch.int64,
+                ),
+                expected_val=torch.tensor([9, 14], dtype=torch.int64),
+                expected_test=torch.tensor([7, 15, 18], dtype=torch.int64),
```
svij-sc, line 805: I brought a similar thread up in one of @kmontemayor2-sc 's earlie PR's as well.

We should generally refrain from knowing magic numbers induced by knowing internals of a function that we are not testing. i.e. in this case we are coupling the test with intimiate knowledge of how _fast_hash works - meaning any time _fast_hash changes, not only will we have to update the _fast_hash test but these tests too.

Other way to test this functionality without coupling w/ knowing the internals of the hash function:
```python
assert len(set(train.tolist()) & set(val.tolist())) == 0
assert len(set(train.tolist()) & set(test.tolist())) == 0
assert len(set(val.tolist()) & set(test.tolist())) == 0
# All IDs covered
and set(train.tolist()) | set(val.tolist()) | set(test.tolist()) == set(original.tolist())


# Correct sizes (within 1 due to rounding)
assert abs(len(val) - round(len(original) * 0.1)) <= 1
assert abs(len(test) - round(len(original) * 0.1)) <= 1
```

svij-sc: FWIW, I think the tests w/ using identity hashes is fine since the implementation of it is simple and 1 line; albeit, documentation could help as to why we expect what we expect when using identity hash.

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -728,6 +779,216 @@ def test_get_padded_labels(self, _, node_ids, topo, expected):
...
+        hash_function,
+        val_num,
+        test_num,
+        expected_train,
+        expected_val,
+        expected_test,
+    ):
+        torch.distributed.init_process_group(
+            rank=0, world_size=1, init_method=get_process_group_init_method()
+        )
```
svij-sc, line 849: we need init_process_group to be able to initialize nodesplitter?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support `URI / URI_LIKE`, similar to `pathlib.Path`.
 
 ### Added
+- Added support for NodeAnchorSplitter
```
svij-sc, line 12: thanks.

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -504,8 +510,50 @@ def test_node_based_link_splitter_parallelized(self):
...
+            ),
+            (
+                torch.arange(0, 16, 2, dtype=torch.int64),
+                torch.tensor([16], dtype=torch.int64),
+                torch.tensor([18], dtype=torch.int64),
+            ),
+        ]
+        # Run the splitter in parallel
+        mp.spawn(
+            _run_splitter_distributed,
```
svij-sc, line 550: nit: a more readable design would include passing the hash function as an argument to `_run_splitter_distributed` so the author does not need to read the `_run_splitter_distributed` to understand what is the hashing function used

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -504,8 +510,50 @@ def test_node_based_link_splitter_parallelized(self):
...
+            torch.arange(0, 20, dtype=torch.int64),
+            torch.arange(0, 10, dtype=torch.int64),
+            torch.arange(0, 20, 2, dtype=torch.int64),
+        ]
+        # We need to guarantee that a given node id is always selected into the same split, on every rank.
+        expected_splits = [
+            (
+                torch.arange(16, dtype=torch.int64),
+                torch.tensor([16, 17], dtype=torch.int64),
+                torch.tensor([18, 19], dtype=torch.int64),
```
svij-sc, line 533: comments would help reader.
```suggestion
                torch.arange(16, dtype=torch.int64),  # train_split on process_1: using identity hash, we expect 80% of data  .... etc.
                torch.tensor([16, 17], dtype=torch.int64), # test_split on process_1; 
                torch.tensor([18, 19], dtype=torch.int64),
```

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -39,27 +40,32 @@ def _run_splitter_distributed(
     process_num: int,
     world_size: int,
     init_method: str,
-    edges: list[torch.Tensor],
+    tensors: list[torch.Tensor],
     expected: list[tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
+    is_node_split: bool,
```
svij-sc, line 45: More readable design + there may be more splitters in the future.
```suggestion
    splitter: Union[HashedNodeSplitter, HashedNodeAnchorLinkSplitter]
```

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -728,6 +779,216 @@ def test_get_padded_labels(self, _, node_ids, topo, expected):
...
+                "Node splitting with non-contiguous IDs",
+                node_ids=torch.arange(2, 22, 2, dtype=torch.int64),
+                hash_function=lambda x: x,
+                val_num=0.1,
+                test_num=0.1,
+                expected_train=torch.tensor(
+                    [2, 4, 6, 8, 10, 12, 14, 16], dtype=torch.int64
+                ),
+                expected_val=torch.tensor([18], dtype=torch.int64),
+                expected_test=torch.tensor([20], dtype=torch.int64),
```
svij-sc, line 832: what is this testing that the prior tests did not?
Could we not just have one of these i.e. you could craft something complicated
`node_ids=[1, 20, 2, 5, 200]`,

python/tests/unit/utils/data_splitters_test.py
```diff
@@ -728,6 +779,216 @@ def test_get_padded_labels(self, _, node_ids, topo, expected):
...
+                    ),
+                    _NODE_B: (
+                        torch.arange(0, 72, dtype=torch.int64),
+                        torch.arange(72, 96, dtype=torch.int64),
+                        torch.arange(96, 120, dtype=torch.int64),
+                    ),
+                    _NODE_C: (
+                        torch.arange(6, dtype=torch.int64),
+                        torch.tensor([6, 7], dtype=torch.int64),
+                        torch.tensor([8, 9], dtype=torch.int64),
```
svij-sc, line 924: generally same comments as the homogeneous tests.

python/gigl/utils/data_splitters.py
```diff
@@ -238,6 +269,12 @@ def __call__(
         Tuple[torch.Tensor, torch.Tensor, torch.Tensor],
         Mapping[NodeType, Tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
     ]:
+        # Validate distributed process group
+        if not torch.distributed.is_initialized():
+            raise RuntimeError(
+                f"Splitter requires a Torch distributed process group, but none was found. "
+                "Please initialize a process group (`torch.distributed.init_process_group`) before using this splitter."
+            )
```
svij-sc, line 277: why is this a requirement?
do we need communication between processes during splitting?

If it does should we be calling these classes `Dist...` ?

python/gigl/utils/data_splitters.py
```diff
@@ -379,6 +439,116 @@ def should_convert_labels_to_edges(self):
...
+            max_node_id = int(nodes_to_split.max().item() + 1)
+            # Set device explicitly here so we don't default to CPU.
+            node_id_count = torch.zeros(
+                max_node_id, dtype=torch.uint8, device=nodes_to_split.device
+            )
+            node_id_count.add_(torch.bincount(nodes_to_split, minlength=max_node_id))
+            # This line takes us from a count of all node ids, e.g. `[0, 2, 0, 1]`
+            # To a tensor of the non-zero counts, e.g. `[[1], [3]]`
+            # and the `squeeze` converts that to a 1d tensor (`[1, 3]`).
+            unique_nodes = torch.nonzero(node_id_count).squeeze()
```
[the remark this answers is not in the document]
svij-sc: can node_if_count ever be a size 1 tensor?

```python
>>> torch.nonzero(torch.tensor([0, 2, 0, 1])).squeeze()
tensor([1, 3])
>>> torch.nonzero(torch.tensor([1])).squeeze()
tensor(0)
```

The second case is a scaler - unsure if this will break stuff.

python/gigl/utils/data_splitters.py
```diff
@@ -379,6 +377,185 @@ def should_convert_labels_to_edges(self):
...
+        for node_type, nodes_to_split in node_ids_dict.items():
+            _check_node_ids(nodes_to_split)
+
+            # Use efficient bincount approach to find unique nodes instead of torch.unique()
+            max_node_id = int(nodes_to_split.max().item() + 1)
+            # Set device explicitly here so we don't default to CPU.
+            node_id_count = torch.zeros(
+                max_node_id, dtype=torch.uint8, device=nodes_to_split.device
+            )
+            node_id_count.add_(torch.bincount(nodes_to_split, minlength=max_node_id))
```
svij-sc, line 463: can we do direct assignment here?
is setting everything to zeros, then doing `add_` necessary?

python/gigl/utils/data_splitters.py
```diff
@@ -379,6 +377,185 @@ def should_convert_labels_to_edges(self):
...
+        hash_function: Callable[[torch.Tensor], torch.Tensor] = _fast_hash,
+    ):
+        """Initializes the HashedNodeSplitter.
+
+        Args:
+            num_val (float): The percentage of nodes to use for validation. Defaults to 0.1 (10%).
+            num_test (float): The percentage of nodes to use for testing. Defaults to 0.1 (10%).
+            hash_function (Callable[[torch.Tensor], torch.Tensor]): The hash function to use. Defaults to `_fast_hash`.
+        """
+        _check_val_test_percentage(num_val, num_test)
```
svij-sc, line 412: since this function throws, better naming convention is utilizing `assert`  
```suggestion
        _assert_valid_split_ratios(num_val, num_test)
```

python/gigl/utils/data_splitters.py
```diff
@@ -379,6 +377,185 @@ def should_convert_labels_to_edges(self):
...
+
+        for node_type, nodes_to_split in node_ids_dict.items():
+            _check_node_ids(nodes_to_split)
+
+            # Use efficient bincount approach to find unique nodes instead of torch.unique()
+            max_node_id = int(nodes_to_split.max().item() + 1)
+            # Set device explicitly here so we don't default to CPU.
+            node_id_count = torch.zeros(
+                max_node_id, dtype=torch.uint8, device=nodes_to_split.device
+            )
```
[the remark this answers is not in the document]
svij-sc: I guess generating potentially a very tensor w/ bincount is not a problem from a memory perspective?

If a node_type has cardinality of 1 billion nodes, then we are potentially allocating close to 1 billion*8 bits*n procs = 1 Gb * procs?


Would be good to add some comment around what is expected size here and if this is being used elsewhere to see if it could be extracted as a utility.

python/gigl/utils/data_splitters.py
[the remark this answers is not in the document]
kmontemayor2-sc: sg! thans for the explanation :)

python/gigl/utils/data_splitters.py
kmontemayor2-sc: We need this otherwise the same node id may be in different splits on different ranks. We encountered this before and fixed it in https://github.com/Snapchat/GiGL/pull/106

Re: naming it `Dist..` we could, no strong preference there.

python/tests/unit/utils/data_splitters_test.py
svij-sc: Looks like you were able to figure out an alternative.
Thanks.

python/tests/unit/utils/data_splitters_test.py
svij-sc: Thanks.

My general comment would be to develop tests with the same rigor as rest of code.
If tests are complex to follow - better to cut duplicity between them.

svij-sc (approved):
Thanks for the changes.
Just one blocking comment re: use of `.clamp_` - rest are nits, but would also appreciate addressing.

python/tests/unit/utils/data_splitters_test.py
svij-sc: maybe this is a nit: as I do see the comment on the top but its better if inline comments are used here imo.

svij-sc (commented):
Its probably also worth

PR #274, opened 2025-08-15
Revert "Replace typing.Dict with dict (#183)"

kmontemayor2-sc:
This reverts commit f4e3bc8b463039fd123c0945963835c71377efda.

Apparently dataflow / beam cares about the semantic differences here so we saw some failures that reverting this fixed


mkolodner-sc (approved):
Thanks for looking into this!

PR #275, opened 2025-08-15
Use `typing.Dict` for Beam type annotations.

kmontemayor2-sc:
We've seen some failures at very large datasizes when using `dict` vs `typing.Dict` here. I am not sure what the root cause here is, but I suspect it may be related to our version of Beam only parsing `typing` annotations [1] and dicts having some special Coder [2].
If/when we upgrade beam versions, we could revisit this.
[1]: https://github.com/apache/beam/blob/96d2a281cd69698463d5455861335b4d7d09e795/sdks/python/apache_beam/typehints/native_type_compatibility.py#L277-L279
[2]: https://github.com/apache/beam/blob/96d2a281cd69698463d5455861335b4d7d09e795/sdks/python/apache_beam/coders/typecoders.py#L96

### DIFF
--- a/python/gigl/src/data_preprocessor/lib/types.py
+++ b/python/gigl/src/data_preprocessor/lib/types.py
@@ -11,6 +11,14 @@ from gigl.common import Uri
 
 # TODO (mkolodner-sc): Move these variables to a more general location, as they are used even outside of context of data preprocessor
 
+# Note: we use `typing.Dict` here instead of `dict` as the version of Beam we are on (2.56.0)
+# will see some failures when running with very large data and `dict`.
+# I am not sure what the root cause here is,
+# but I suspect it may be related to our version of Beam only parsing `typing` annotations [1]
+# and dicts having some special Coder [2].
+# If/when we upgrade beam versions, we could revisit this.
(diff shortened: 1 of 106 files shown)


python/gigl/src/data_preprocessor/lib/types.py
```diff
@@ -11,12 +11,20 @@
...
 
-InstanceDict = dict[str, Any]
-TFTensorDict = dict[str, common_types.TensorType]
-FeatureSpecDict = dict[str, common_types.FeatureSpecType]
-FeatureIndexDict = dict[str, Tuple[int, int]]  # feature_name -> (start, end) index
-FeatureSchemaDict = dict[str, Feature]
-FeatureVocabDict = dict[str, list[str]]
+# Note: we use `typing.Dict` here instead of `dict` as the version of Beam we are on (2.56.0)
+# will see some failures when running with very large data and `dict`.
+# I am not sure what the root cause here is,
```
svij-sc, line 16: nit: Add ur @tag, or refrain from using first person.

svij-sc (approved):
Thanks

mkolodner-sc (approved):

PR #276, opened 2025-08-15
Add ability to load, partition, and separate node labels with features

mkolodner-sc:
**Scope of work done**

- Adds ability for TFRecordDataloader to load labels as part of the features and corresponding tests
- Adds function for separating labels from features and corresponding tests
- Updates dataset factory to call this function to separate node labels if they are present

Explanation of Node Classification Labels
1. As input to our node classification pipeline, we should have prepared bigquery node feature tables which have columns corresponding to the node classification label.

Taken from `external-snap-ci-github-gigl.public_gigl.cora_homogeneous_supervised_node_classification_edge_features_paper_nodes_2024-07-15--21-30-07-UTC`
Screenshot 2025-08-18 at 11 59 26 AM

2. In our Data Preprocessor config, labels are specified [[1](https://github.com/Snapchat/GiGL/blob/main/python/gigl/src/mocking/mocking_assets/passthrough_preprocessor_config_for_mocked_assets.py#L79-L107)] in the `get_nodes_preprocessing_spec` function.
3. After running data preprocessor, we have generated TFRecords with the node label as a feature. We can see the preprocessed metadata output at ` gs://public-gigl/mocked_assets/2024-07-15--21-30-07-UTC/cora_homogeneous_supervised_node_classification/data_preprocess/preprocessed_metadata.yaml`:
```
...
    - f1430
    - f1431
    - f1432
    labelKeys:
    - node_label
    nodeIdKey: node_id
    schemaUri: gs://public-gigl/mocked_assets/2024-07-15--21-30-07-UTC/cora_homogeneous_supervised_node_classification/data_preprocess/node_features_dir/paper/schema.pbtxt
    tfrecordUriPrefix: gs://public-gigl/mocked_assets/2024-07-15--21-30-07-UTC/cora_homogeneous_supervised_node_classification/data_preprocess/node_features_dir/paper/features/
```

4. We load the preprocessed metadata into a `PreprocessedMetadataPbWrapper`. This creates and isolates fields we'll need for loading the TFRecords like the `FeatureSchemaDict`, `FeatureDim`, `LabelKeys`, etc. As of this PR, the `FeatureSchemaDict` includes the feature schemas for both the feature keys and label keys.
5. We pass this information to the SerializedTFRecordInfo.
6. We load the node features and labels, with the labels being appended as an additional feature to the normal feature tensor.
7. We call our partitioner and partition the "node features" across all machines. The benefit of having labels as part of this process is that 1) there are less distributed calls and 2) there are guarantees that the mapping of global_id to local_id (`id2idx`) is correct for both node features and node labels.
8. We extract the labels from the partitioned node features with `_get_labels_from_features`.
9. We register those labels to the dataset in `dataset.build()` (TODO in the next PR)


python/gigl/distributed/dataset_factory.py
```diff
@@ -192,6 +194,44 @@ def _load_and_build_partitioned_dataset(
...
+        label_dim = len(serialized_graph_metadata.node_entity_info.label_keys)
+        if label_dim > 0:
+            node_features, node_labels = get_labels_from_features(
+                partition_output.partitioned_node_features.feats, label_dim=label_dim
+            )
+            partition_output.partitioned_node_features.feats = node_features
+
+    gc.collect()
+
+    # TODO (mkolodner-sc): Add node labels to the dataset
```
mkolodner-sc, line 233: This will be done in the next PR

python/gigl/common/data/dataloaders.py
```diff
@@ -79,23 +81,31 @@ class TFDatasetOptions:
...
 
     Args:
         feature_key_to_tf_tensor (dict[str, tf.Tensor]): A dictionary mapping feature names to their corresponding tf tensors.
-        feature_keys (list[str]): A list of feature names specifying the order in which tensors should be concatenated.
+        feature_keys (Sequence[str]): A list of feature names specifying the order in which tensors should be concatenated.
+        label_keys (Sequence[str]): Name of the label columns for the current entity.
 
     Returns:
-        tf.Tensor: A concatenated tensor of the features in the specified order.
+        tf.Tensor: A concatenated tensor of the features in the specified order, with the label being concatenated at the end if it exists
```
kmontemayor2-sc, line 98:
```suggestion
        tf.Tensor: A concatenated tensor of the features in the specified order, with the labels being concatenated at the end if it exists
```

nit.

python/gigl/common/data/dataloaders.py
```diff
@@ -298,11 +308,17 @@ def load_as_torch_tensors(
...
 
         # We make a deep copy of the feature spec dict so that future modifications don't redirect to the input
 
         feature_spec_dict = deepcopy(serialized_tf_record_info.feature_spec)
 
+        for label_key in label_keys:
+            feature_spec_dict[label_key] = tf.io.FixedLenFeature(
+                shape=[], dtype=tf.int64
+            )
+
```
kmontemayor2-sc, line 321: qq why is the shape here `[]`? Does this restrict us to having scalar labels?

python/gigl/common/data/dataloaders.py
```diff
@@ -356,11 +372,14 @@ def load_as_torch_tensors(
...
-                else None
-            )
+            if label_keys:
+                empty_feature = torch.empty(
+                    0, serialized_tf_record_info.feature_dim + len(label_keys)
+                )
+            elif feature_keys:
+                empty_feature = torch.empty(0, serialized_tf_record_info.feature_dim)
+            else:
+                empty_feature = None
```
kmontemayor2-sc, line 382: hmmm, if we have both features and labels, when we iterate wont all "empty" things be "empty_labels"?

I may not be communicating here well - my understanding is the code here would say "if there are no features, return None, otherwise return `(0, feature_dim)`.

Now, we should return `0` if it's a label, `(0, feature_dim)` if it's a feature, and None if there are no features or labels, right?

Does this code handle the "there are features and labels, and there was a feature that is empty" case appropriately?

python/gigl/distributed/dataset_factory.py
```diff
@@ -192,6 +194,44 @@ def _load_and_build_partitioned_dataset(
 
     partition_output = partitioner.partition()
 
+    node_labels: Optional[Union[torch.Tensor, dict[NodeType, torch.Tensor]]] = None
+    if isinstance(partition_output.partitioned_node_features, abc.Mapping):
+        node_labels = {}
```
kmontemayor2-sc, line 199: nit. add type.

python/gigl/distributed/dataset_factory.py
```diff
@@ -192,6 +194,44 @@ def _load_and_build_partitioned_dataset(
...
+                partition_output.partitioned_node_features[
+                    node_type
+                ].feats = node_features
+    elif isinstance(partition_output.partitioned_node_features, FeaturePartitionData):
+        if not isinstance(
+            serialized_graph_metadata.node_entity_info, SerializedTFRecordInfo
+        ):
+            raise ValueError(
+                f"Expected partitioned node features to be type SerializedTFRecordInfo, got {type(partition_output.partitioned_node_features)}"
+            )
```
kmontemayor2-sc, line 223: the other case here is if `serialized_graph_metadata.node_entity_info` is None, right? Should that be an error?

python/gigl/utils/node_labels.py
```diff
@@ -0,0 +1,24 @@
...
+    Given a combined tensor of features and labels, returns the features and labels separately.
+    Args:
+        feature_and_label_tensor (torch.Tensor): Tensor of features and labels
+        label_dim (int): Dimension of the labels
+    Returns:
+        feature_tensor (torch.Tensor): Tensor of features
+        label_tensor (torch.Tensor): Tensor of labels
+    """
+
+    _, feature_and_label_dim = feature_and_label_tensor.shape
```
kmontemayor2-sc, line 17: should we enforce that `len(feature_and_label_tensor.shape) = 2` here?

python/tests/unit/common/data/dataloaders_test.py
```diff
@@ -124,6 +135,134 @@ def tearDown(self):
...
+                        * 10,  # feature_0
+                        torch.tensor(
+                            [i % 2 for i in range(100)], dtype=torch.float32
+                        ).reshape(
+                            100, 1
+                        ),  # label_0
+                    ),
+                    dim=1,
+                ),
+            ),
```
kmontemayor2-sc, line 179: with this - we're also requiring that all nodes have some label right? e.g. in DP users should put -1 or something for unlabeled nodes?

python/tests/unit/utils/node_labels_test.py
```diff
@@ -0,0 +1,76 @@
...
+                expected_features=torch.empty((2, 0)),
+                expected_labels=torch.tensor([[3.0], [6.0]]),
+            ),
+            param(
+                "Test with minimal tensor (1x2)",
+                feature_and_label_tensor=torch.tensor([[1.0, 2.0]]),
+                label_dim=1,
+                expected_features=torch.tensor([[1.0]]),
+                expected_labels=torch.tensor([[2.0]]),
+            ),
```
kmontemayor2-sc, line 59: there's also a (weird!) case of "all '''features''' are labels" right? Should we test for that too?

python/gigl/distributed/dataset_factory.py
```diff
@@ -192,6 +194,44 @@ def _load_and_build_partitioned_dataset(
 
     partition_output = partitioner.partition()
```
svij-sc, line 196: Generally I feel this function, and the dataset factory is more and more tied to our specific* data preprocessor output schema. i.e. a person cannot use these unless they use GiGL's preprocessor.
This is fine, but maybe we need to think about how to rename the api to signify that more.

We will have future use cases where I may want to load from a BQ table directly i.e. I may want to `build_dataset` with `BqTableInfo` instead of  `SerializedGraphMetadata` - which is very coupled to tfrecords and our specific output schematics of tfrecords from gigl's preprocessor.

The naming atleast for public methods is probably better worked on now when there are only a few uses, before we have a lot more use cases and then we are locked in. cc @kmontemayor2-sc

python/gigl/distributed/dataset_factory.py
```diff
@@ -192,6 +194,44 @@ def _load_and_build_partitioned_dataset(
...
+        ):
+            raise ValueError(
+                f"Expected partitioned node features to be type SerializedTFRecordInfo, got {type(partition_output.partitioned_node_features)}"
+            )
+        label_dim = len(serialized_graph_metadata.node_entity_info.label_keys)
+        if label_dim > 0:
+            node_features, node_labels = get_labels_from_features(
+                partition_output.partitioned_node_features.feats, label_dim=label_dim
+            )
+            partition_output.partitioned_node_features.feats = node_features
```
svij-sc, line 229: instead of doing assignment like this and losing the benefits of immutable objects, we could just do something like:

```python
paritioned_output.partitioned_node_features = FeaturePartitionData(feats = node_features)
```

python/gigl/types/graph.py
```diff
@@ -32,7 +32,7 @@
 # TODO(kmonte, mkolodner): Move SerializedGraphMetadata and maybe convert_pb_to_serialized_graph_metadata here.
 
 
-@dataclass(frozen=True)
```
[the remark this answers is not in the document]
svij-sc: See comment: https://github.com/Snapchat/GiGL/pull/276/files#r2280048946

python/tests/unit/common/data/dataloaders_test.py
```diff
@@ -124,6 +135,134 @@ def tearDown(self):
                     dim=1,
                 ),
             ),
+            param(
+                "Two features with labels",
+                feature_spec=_FEATURE_SPEC_WITH_ENTITY_KEY,
```
svij-sc, line 140: A little confused, I thought `_FEATURE_SPEC_WITH_ENTITY_KEY` should have the label key too?

A small request to add some explaination in the PR description on what is expected as output from preprocessor in the case of labels, vs not, to help with the context to review this PR.

python/gigl/distributed/dataset_factory.py
mkolodner-sc: The type is defined in line 197 `node_labels: Optional[Union[torch.Tensor, dict[NodeType, torch.Tensor]]] = None` :P

python/gigl/distributed/dataset_factory.py
```diff
@@ -52,6 +53,39 @@
 logger = Logger()
 
 
+def _get_labels_from_features(
```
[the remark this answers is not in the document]
mkolodner-sc: Makes sense, thanks for the discussion here.

Just for full visibility as I've mentioned this in the PR description but I'll provide more detail to this point here -- there are some advantages in terms of complexity and performance that also result with how we group the labels with the node features, much like our preprocessor does.

One alternative in this case is to shuffle the labels separately from the features. The tricky part is how we should partition the labels. Let's say on one machine we've initially loaded:
- Node 0: Features `[0, 0]`, label: `[0]`
- Node 1: Features `[1, 1]`, label: `[1]`

We now need to shuffle these across machines. Let's say both of these nodes are assigned to the same machine post-shuffle. In the alternative setting where we are partitioning the features and labels separately, firstly we need to make 2x the number of distributed calls to partition the node labels (assuming each node contains 1 label). Additionally, we have no control over which order the labels and features are partitioned to each machine. We very much could end up with a new machine that sees:
- Features: `[[1, 1], [0, 0]]`
 - Labels: `[0, 1]`

The only way we can guarantee both the `[0, 0]` feature and the `[0]` label are both associated with the same node is for us to maintain **two** sets of ids per node during shuffling and **two** id2idx tensors which map {global id 0 -> local index 1} for features and {global id 0 -> local index 0}. Besides the additional memory this uses up, there is also additional complexity with having to manage these two separately and increased room for error in the shuffling step. 

In the current setting, since the features and labels are managed together, we only have guarantees that the local index of the features and the local index of the labels are the same on each machine. Additionally, we have fewer distributed calls that we need to make since these are considered together, and our code is generally much simpler since we only have to manage one global id -> local id mapping. 

> Thereafter, partitioning, etc. should be able to handle the tensors in a more generic nature - rather than having access to everything and being coupled to preprocessor logic.

Going back to this point, our partitioner doesn't actually see any of the preprocessor-coupled logic -- what it instead observes is that it has some features it needs to partition and it shuffles them accordingly so that they are redistributed and accessible across machines. There is no coupling with labels or our data preprocessor here, as it doesn't know or care about labels being part of the feature set.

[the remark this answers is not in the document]
mkolodner-sc: > But, can we add a tech debt task to clean this up - and for now refrain from making it even more.

However, I do think there is opportunity for further cleanup here. "Grouping" the features and labels can still be achieved without necessarily concatenating them together -- we just need to ensure that we shuffle them together, which can be achieved without concatenation. Right now we take the concatenation approach since it aligns with our data preprocessor and requires relatively few changes to our existing code base. That being said, I can add "non-concatenated grouping" approach as a tech debt item for the future.

python/gigl/distributed/dataset_factory.py
```diff
@@ -52,6 +53,39 @@
...
+    feature_dim = feature_and_label_dim - label_dim
+
+    return (
+        feature_and_label_tensor[:, :feature_dim],
+        feature_and_label_tensor[:, feature_dim:],
+    )
+
+
 @tf_on_cpu
 def _load_and_build_partitioned_dataset(
```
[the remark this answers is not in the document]
mkolodner-sc: I can look into extracting some sub-methods here in this PR -- if it ends up increasing the scope too much then I will opt to move it to the next one, thanks!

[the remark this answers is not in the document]
mkolodner-sc: Just drafted what this would look like [[1](https://github.com/Snapchat/GiGL/pull/276/commits/d7490e65625dc251c873e6060974990b60b3527e)] -- breaking down the sub-methods here involves a lot of movement and would be best done as a separate PR -- I'll add a task for this and raise a follow-up with this logic shortly.

python/gigl/common/data/dataloaders.py
```diff
@@ -356,11 +368,16 @@ def load_as_torch_tensors(
...
+            if label_keys and feature_keys:
+                empty_feature = torch.empty(
+                    0, serialized_tf_record_info.feature_dim + len(label_keys)
+                )
+            elif label_keys and not feature_keys:
+                empty_feature = torch.empty(0, len(label_keys))
+            elif not label_keys and feature_keys:
+                empty_feature = torch.empty(0, serialized_tf_record_info.feature_dim)
+            else:
+                empty_feature = None
```
kmontemayor2-sc, line 380: oops sorry to go back again but WDYT about 
```suggestion
            if feature_iterable:
                empty_feature = torch.empty(
                    0, len(feature_iterable)                )
            else:
                empty_feature = None
```

python/gigl/distributed/dataset_factory.py
```diff
@@ -192,6 +226,52 @@ def _load_and_build_partitioned_dataset(
...
+        for (
+            node_type,
+            node_feature,
+        ) in partition_output.partitioned_node_features.items():
+            if isinstance(serialized_graph_metadata.node_entity_info, Mapping):
+                label_dim = len(
+                    serialized_graph_metadata.node_entity_info[node_type].label_keys
+                )
+            else:
+                label_dim = len(serialized_graph_metadata.node_entity_info.label_keys)
```
kmontemayor2-sc, line 241: btw why do we allow map/scalar here but we enforce scalar on line 253?

python/gigl/src/common/types/pb_wrappers/preprocessed_metadata.py
```diff
@@ -68,13 +68,14 @@ def __post_init__(self):
             condensed_node_type_to_feature_dim_map[
                 CondensedNodeType(condensed_node_type)
             ] = node_metadata.feature_dim
-            node_feature_keys = list(node_metadata.feature_keys)
+            node_feature_keys = sorted(list(node_metadata.feature_keys))
+            label_keys = sorted(list(node_metadata.label_keys))
```
kmontemayor2-sc, line 72: why do we make this change here? this seems like a bigger change that shouldn't be included as part of this pr.

kmontemayor2-sc: is this just for testing? if so let's do the changes all on the test side.

mkolodner-sc: This is not just for testing -- this is done so that our FeatureSpecDict includes the label keys as well. This makes it so we don't need to be assuming what the feature spec looks like for the label keys [from [this comment](https://github.com/Snapchat/GiGL/pull/276#discussion_r2280020488)]. This logic is tested in the added `test_load_labels_from_pb` [function](https://github.com/Snapchat/GiGL/pull/276/files#diff-02132213959de2fbd619bfcab31747e5d3883ff8605d7a0542f45a3221a8769cR443).

python/gigl/common/data/dataloaders.py
mkolodner-sc: This would not necessarily work -- `len(feature_iterable) = len(feature_keys) + len(label_keys)`.  `len(feature_keys)` and `serialized_tf_record_info.feature_dim` are not necessarily the same, since we could have one feature key with dim > 1

PR #284, opened 2025-08-20
Fail early if invalid Node IDs are found prior partitioning

mkolodner-sc:
**Scope of work done**

There may be some scenarios where invalid node ids may be loaded as a result of unexpected input data, such as `NULL` values for node ids. This leads to behaviors where we may have the repeated or non-contiguous node ids in our input data. However, GLT requires that the node ids across all machines be enumerated from 0 to total_num_nodes - 1. If this condition is not met, there are failures late in the GLT neighbor-loading step, and its somewhat tricky to debug and isolate this issue. 

This PR adds a check much earlier than the above failure point when we register node ids to the partitioner. We add it here since this is the first `distributed` step in the dataset building work. We also add corresponding tests to ensure the error is raised as expected.

### DIFF
--- a/python/gigl/distributed/dist_partitioner.py
+++ b/python/gigl/distributed/dist_partitioner.py
@@ -396,19 +396,17 @@ class DistPartitioner:
             str, Tuple[int, dict[NodeType, Tuple[int, int]]]
         ] = glt_rpc.all_gather((self._rank, node_type_to_num_nodes_and_max_node))
 
-        # Looping through each of the registered node types in the graph
-        for node_type in self._node_types:
-            # Computing total number of nodes across all ranks of type `node_type`
-            for (
-                _,
-                gathered_node_type_to_num_nodes_and_max_node,
-            ) in gathered_node_info.values():
-                self._num_nodes[
-                    node_type
-                ] += gathered_node_type_to_num_nodes_and_max_node[node_type][0]
+        for (
+            _,
+            gathered_node_type_to_num_nodes_and_max_node,
+        ) in gathered_node_info.values():
+            for node_type, (
+                num_nodes,
+                max_id,
+            ) in gathered_node_type_to_num_nodes_and_max_node.items():
+                self._num_nodes[node_type] += num_nodes
                 self._max_node_ids[node_type] = max(
-                    self._max_node_ids[node_type],
-                    gathered_node_type_to_num_nodes_and_max_node[node_type][1],
+                    self._max_node_ids[node_type], max_id
                 )
 
     def register_edge_index(


python/gigl/distributed/dist_partitioner.py
```diff
@@ -371,28 +371,54 @@ def register_node_ids(
 
         self._node_ids = convert_to_tensor(input_node_ids, dtype=torch.int64)
 
-        # This tuple here represents a (rank, num_nodes_on_rank) pair on a given partition, specified by the str key of the dictionary of format `distributed_random_partitoner_{rank}`.
-        # num_nodes_on_rank is a dict[NodeType, int].
-        # Gathered_num_nodes is then used to identify the number of nodes on each rank, allowing us to access the total number of nodes across all ranks
-        gathered_node_info: dict[str, Tuple[int, dict[NodeType, int]]]
+        # This tuple here represents a (rank, (num_nodes_on_rank, max_node_id_on_rank)) pair on a given partition, specified by the str key of the dictionary of format `distributed_random_partitoner_{rank}`.
```
kmontemayor2-sc, line 374: btw why don't we just have the key be an int (the rank)?

kmontemayor2-sc: or is this something that `glt_rpc.all_gather` does?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -371,28 +371,54 @@ def register_node_ids(
...
+        # num_nodes_on_rank and max_node_id_on_rank are ints.
+        # Gathered_num_nodes is then used to identify the number of nodes and max node id on each rank,
+        # allowing us to access the total number of nodes and max node id across all ranks
+        gathered_node_info: dict[str, Tuple[int, dict[NodeType, Tuple[int, int]]]]
         self._num_nodes = defaultdict(int)
 
-        node_type_to_num_nodes: dict[NodeType, int] = {
-            node_type: input_node_ids[node_type].size(0)
+        node_type_to_node_info: dict[NodeType, Tuple[int, int]] = {
```
kmontemayor2-sc, line 381: nit. can we name this to `node_type_to_num_nodes_and_max_node` (or something clearer) so it's easy to read later.

Or, it may be easier to just have two dicts. so we don't have to keep track of which part of the tuple is what.

python/gigl/distributed/dist_partitioner.py
```diff
@@ -371,28 +371,54 @@ def register_node_ids(
...
+                )
+
+        for node_type in self._node_types:
+            total_num_nodes = self._num_nodes[node_type]
+            max_id_on_all_ranks = max_ids[node_type]
+            logger.info(
+                f"Registered {total_num_nodes} nodes with max node id {max_id_on_all_ranks} for node type {node_type} across all machines"
+            )
+            if total_num_nodes > 0 and max_id_on_all_ranks + 1 != total_num_nodes:
+                # Node IDS provided must be unique and contiguous from 0 to total_num_nodes - 1.
```
kmontemayor2-sc, line 417: This line is good, should we make it part of the error?

python/tests/unit/distributed/distributed_partitioner_test.py
```diff
@@ -992,6 +992,53 @@ def test_partitioning_failure(self) -> None:
...
+    @parameterized.expand(
+        [
+            param(
+                "Missing Node ID -- homogeneous",
+                node_ids=torch.tensor([0, 2, 3]),
+            ),
+            param(
+                "Duplicate Node ID -- homogeneous",
+                node_ids=torch.tensor([0, 0, 1]),
+            ),
```
kmontemayor2-sc, line 1004: btw does this catch `[0, 0, 3]`?

kmontemayor2-sc: er, `[0, 0, 2]`

python/gigl/distributed/dist_partitioner.py
mkolodner-sc: One dict has less calls to `all_gather`, which is why I've adopted that approach. Sg to update the name here.

python/tests/unit/distributed/distributed_partitioner_test.py
kmontemayor2-sc: Hmmm, we can use our favorite function `torch.bincount` here couldn't we? It seems pretty fast too.


```py
import torch
import time

def count(x: torch.Tensor):
    start = time.time()
    counts = torch.bincount(x)
    if torch.all(counts == 1):
        print("All counts are 1")
    else:
        print("Not all counts are 1")
    end = time.time()
    print(f"Time taken: {end - start} seconds")
    return counts

counts = count(torch.arange(1_000_000_000)) # 3.14 second
```

python/tests/unit/distributed/distributed_partitioner_test.py
```diff
@@ -992,6 +992,87 @@ def test_partitioning_failure(self) -> None:
...
+            node_features = {
+                node_type: torch.ones(node_ids[node_type].shape[0], 2)
+                for node_type in node_ids.keys()
+            }
+
+        with self.assertRaises(ValueError):
+            partitioner.register_node_ids(node_ids=node_ids)
+            partitioner.register_node_features(node_features=node_features)
+            node_pb = partitioner.partition_node()
+            partitioner.partition_node_features(node_pb)
```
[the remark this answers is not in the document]
kmontemayor2-sc: then let's have multiple tests :)

if you are worried about duplicating these lines [1] then we can break them out into some helper function

[1]: https://github.com/Snapchat/GiGL/pull/284/files/d61c3727a8a02bac080b5f3a336dc7b19155f0f3#diff-b17a1745879d8c7405b5ebc80a3102ab5cd64b37d0a3740e618763b8c4e30007R1047-R1059

python/tests/unit/distributed/distributed_partitioner_test.py
mkolodner-sc: This is a scenario where I'd have a strong preference for optimizing for the minimal memory footprint over the time it would take in this scenario. The partitioner (especially feature partitioning where the .unique() is currently called) is on the of the memory bottlenecks of the trainer/inferencer proceses, and between 
1) creating `num_machines` tensors of potential size (`num_total_ids`)
2) waiting 6 extra seconds
I'd have a strong preference for the latter here.

kmontemayor2-sc: > 2. waiting 6 extra seconds

Using `bincount` is 1 second, not 6, but either way the time difference here is neglible and so we shouldn't care about that. Memory is more important I and understand wanting to be cautious here.

If we *really* want to be careful about memory, then let's at least do all the checks in one place? Doing some checks in `register_node_ids` and some in `register_node_features` is IMO confusing. Especially if there are some consecutive nodes but no features for them.

Can we do both somewhere in `_partition_node` to be simpler?

Also, the `bincount` approach seems faster if we do care about wall time? I'm a bit surprised that `unique()` has lower memory usage across the board (since we are allocating tensors of size N in both cases, right?) and I'm able to get the memory usage to be the same across the board so I suspect there's specifically some torch optimization with the duplicate tensors that's coming into play here...

If we don't use the same process for each then we can see the two have ~equal  memory across the board and should we go with `bincount` as it's faster then?


```py
import torch
import time
from memory_profiler import profile
import argparse

@profile
def bincount(x: torch.Tensor):
    start = time.time()
    if torch.all(torch.bincount(x, minlength=x.max() + 1) == 1):
        print("All counts are 1")
    else:
        print("Not all counts are 1")
    end = time.time()
    print(f"Time taken: {end - start} seconds")
    return x

@profile
def unique(x: torch.Tensor):
    start = time.time()
    counts = torch.unique(x)
    if counts.numel() == x.numel():
        print("All unique")
    else:
        print("Not all unique")
    end = time.time()
    print(f"Time taken: {end - start} seconds")
    return x


def main(use_bincount: bool):
    if use_bincount:
        y = bincount(torch.arange(1_000_000_000)) # 1.07s
    else:
        y = unique(torch.arange(1_000_000_000)) # 6.98s

    print(y.shape)

if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--use_bincount", action="store_true")
    args = parser.parse_args()
    main(args.use_bincount)

```

```
python count.py                     
All unique
Time taken: 5.965315818786621 seconds
Filename: count.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    17   7790.6 MiB   7790.6 MiB           1   @profile
    18                                         def unique(x: torch.Tensor):
    19   7790.6 MiB      0.0 MiB           1       start = time.time()
    20   9323.6 MiB   1533.0 MiB           1       counts = torch.unique(x)
    21   9323.8 MiB      0.2 MiB           1       if counts.numel() == x.numel():
    22   9324.1 MiB      0.3 MiB           1           print("All unique")
    23                                             else:
    24                                                 print("Not all unique")
    25   9324.1 MiB      0.0 MiB           1       end = time.time()
    26   9324.1 MiB      0.0 MiB           1       print(f"Time taken: {end - start} seconds")
    27   9324.1 MiB      0.0 MiB           1       return x


torch.Size([1000000000])

###

python count.py --use_bincount
All counts are 1
Time taken: 1.2766740322113037 seconds
Filename: count.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     6   7789.0 MiB   7789.0 MiB           1   @profile
     7                                         def bincount(x: torch.Tensor):
     8   7789.0 MiB      0.0 MiB           1       start = time.time()
     9   7790.5 MiB      1.5 MiB           1       if torch.all(torch.bincount(x, minlength=x.max() + 1) == 1):
    10   7790.5 MiB      0.0 MiB           1           print("All counts are 1")
    11                                             else:
    12                                                 print("Not all counts are 1")
    13   7790.5 MiB      0.0 MiB           1       end = time.time()
    14   7790.5 MiB      0.0 MiB           1       print(f"Time taken: {end - start} seconds")
    15   7790.5 MiB      0.0 MiB           1       return x
```

kmontemayor2-sc: TLDR: 

1. Let's consolidate all of the checks to one place
2. I suspect bincount is much faster here in practice (especially if the data is shuffled) but there's definitely some torch optimizations going on too.

python/gigl/distributed/dist_partitioner.py
```diff
@@ -371,28 +373,59 @@ def register_node_ids(
...
-        # Gathered_num_nodes is then used to identify the number of nodes on each rank, allowing us to access the total number of nodes across all ranks
-        gathered_node_info: dict[str, Tuple[int, dict[NodeType, int]]]
         self._num_nodes = defaultdict(int)
 
-        node_type_to_num_nodes: dict[NodeType, int] = {
-            node_type: input_node_ids[node_type].size(0)
-            for node_type in sorted(input_node_ids.keys())
-        }
+        node_type_to_num_nodes_and_max_node: dict[NodeType, Tuple[int, int]] = {}
+        for node_type in sorted(input_node_ids.keys()):
```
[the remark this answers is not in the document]
mkolodner-sc: We've run into issues in the past when we have dictionaries passed into distributed calls where the keys are in different orders across machines, so the sort guarantees they are the same across all machines. I'll go ahead and `self._node_types` here which alrady does this operation on the input node id keys.

python/gigl/distributed/dist_partitioner.py
mkolodner-sc: Sg, since there are descript comments and variable naming here to explain what is happening, I will opt to leave the code as is to optimize for performance and minimal `all_gather` calls, thanks!

python/tests/unit/distributed/distributed_partitioner_test.py
kmontemayor2-sc: Thanks for the investigations!

I think in practice we would not need to call `nonzero` on the output on `bincount`, right? What's the case where `torch.all(torch.bincount(x) == 1)` would fail? The `nonzero` bit is to de-dupe iirc. Or am I misunderstanding the nature of the input data here? My impression is that these are dense tensors of node-ids, where if there is a dupe we should also throw?

>  I have some suspicion that with the memory profiler it calls gc.collect() after each line to get more precise diffs of the memory

Maybe! Looking at [the source] (https://github.com/pythonprofilers/memory_profiler/blob/master/memory_profiler.py#L1103-L1104) I think they only use `gc.collect()` for notebooks, but I could be misreading that and also I think the general point here that we shouldn't take `memory_profiler` as being an absolute truth and more pointing us in the general direction. (specifically, I have seen memory spikes in general with using torch that aren't captured by this, the sorting part of `unique()` causes some memory spike, for instance.

Overall I think it's fine to keep using `unique(), but I'd appreciate if you could clarify my first point if I'm incorrect as to what we need to do here.

kmontemayor2-sc: Oh I see, the issue here is we don't want to ever have all node ids on one machine (for the torch.all... approach to work), but we *also* don't want to have any duplicate node ids. and `unique()` does satisfy this for the "sparse" node ids.

Thanks again for the explanations and testing :)

python/tests/unit/distributed/distributed_partitioner_test.py
```diff
@@ -992,6 +992,89 @@ def test_partitioning_failure(self) -> None:
             ):
                 partitioner.partition()
 
+    @parameterized.expand(
+        [
+            param(
+                "Missing Node ID -- homogeneous",
+                node_ids=torch.tensor([0, 2, 3]),
+            ),
```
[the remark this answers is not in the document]
kmontemayor2-sc: if we can't, that's fine. Let's just leave a note about us having some best-effort checks for the validating of the input graph but stuff may still break later on.

PR #290, opened 2025-08-22
Fix cost association for glt/trainer

kmontemayor2-sc:
We must have missed this before - let's associate the cost labels correctly :)


mkolodner-sc (approved):
Nice catch, thanks!

PR #293, opened 2025-08-25
[KGE] Add some modeling code

nshah-sc:
Adding some files which support:

- the core heterogeneous graph embedding model which enables various combinations of in-batch vs. random sampling, src/dst operators, negative sampling corruptions, scoring functions etc.
- some negative sampling utilities
- tests for these utilities
- all code is written to be [torch.fx-traceable](https://docs.pytorch.org/docs/stable/fx.html)  -- this requires following certain patterns that avoid [well-known pitfalls](https://docs.pytorch.org/docs/stable/fx.html#limitations-of-symbolic-tracing), including dynamic control flow, and using non-torch functions and following miscellanea carefully.  

### DIFF
--- a/python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
+++ b/python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
@@ -1,0 +1,427 @@
+from typing import Callable, Tuple, Union
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torchrec
+from applied_tasks.knowledge_graph_embedding.common.torchrec.large_embedding_lookup import (
+    LargeEmbeddingLookup,
+)
+from applied_tasks.knowledge_graph_embedding.lib.config import ModelConfig
+from applied_tasks.knowledge_graph_embedding.lib.config.sampling import SamplingConfig
+from applied_tasks.knowledge_graph_embedding.lib.data.edge_batch import EdgeBatch
+from applied_tasks.knowledge_graph_embedding.lib.data.node_batch import NodeBatch
+from applied_tasks.knowledge_graph_embedding.lib.model.negative_sampling import (
+    against_batch_relationwise_contrastive_similarity,
+    in_batch_relationwise_contrastive_similarity,
+)
+from applied_tasks.knowledge_graph_embedding.lib.model.types import (
+    ModelPhase,
+    SimilarityType,
+)
+
+from gigl.common.logger import Logger
+
+logger = Logger()
+
+
+class HeterogeneousGraphSparseEmbeddingModel(nn.Module):
+    """
+    A backbone model to support sparse embedding of (possibly multi-relational) graphs.
+    Can also be used to implement matrix factorization and variants.
+
+    Args:
+        model_config (ModelConfig): Configuration object containing model parameters.
+    """
+
+    def __init__(self, model_config: ModelConfig):
+        """Initialize the model with the given embedding configurations."""
+        super().__init__()
+
+        self.num_edge_types = model_config.num_edge_types
+        self.node_emb_dim = model_config.node_embedding_dim
+        self.training_sampling_config = model_config.training_sampling
+        self.validation_sampling_config = model_config.validation_sampling
+        self.testing_sampling_config = model_config.testing_sampling
+        self.similarity_type = model_config.embedding_similarity_type
+        self._phase: ModelPhase = ModelPhase.TRAIN
+
+        # Validate the sampling configurations.
+        for sampling_config in (
+            self.training_sampling_config,
+            self.validation_sampling_config,
+            self.testing_sampling_config,
+        ):
+            assert sampling_config is not None, "Sampling config must be provided."
+            assert (
+                sampling_config.positive_edge_batch_size > 0
+            ), "Positive edge batch size must be greater than 0."
+            assert (
+                sampling_config.num_inbatch_negatives_per_edge
+                + sampling_config.num_random_negatives_per_edge
+                > 0
+            ), "At least one type of negative sampling must be specified."
+
+        # Define the embedding layers.
+        self.large_embeddings = LargeEmbeddingLookup(
+            embeddings_config=model_config.embeddings_config
+        )
+
+        # Define the operators applied to src and dst node types respectively
+        self.src_operator = model_config.src_operator.get_corresponding_module()(
+            num_edge_types=self.num_edge_types,
+            node_emb_dim=self.node_emb_dim,
+        )
+        self.dst_operator = model_config.dst_operator.get_corresponding_module()(
+            num_edge_types=self.num_edge_types,
+            node_emb_dim=self.node_emb_dim,
+        )
+
+        logger.info(f"Initialized model with: {self.__dict__}")
+
+    @property
+    def active_sampling_config(self) -> SamplingConfig:
+        if self.phase == ModelPhase.TRAIN:
+            return self.training_sampling_config
+        elif self.phase == ModelPhase.VAL:
+            return self.validation_sampling_config
+        elif self.phase == ModelPhase.TEST:
+            return self.testing_sampling_config
+        elif (
+            self.phase == ModelPhase.INFERENCE_SRC
+            or self.phase == ModelPhase.INFERENCE_DST
+        ):
+            raise ValueError(
+                "Active sampling config is not defined for inference phase. "
+            )
+        else:
+            raise ValueError(
+                f"Unknown model phase: {self.phase}. Cannot determine active sampling config."
+            )
+
+    def set_phase(self, phase: ModelPhase):
+        """
+        Set the phase of the model. This is used to determine which sampling
+        configuration to use during training, validation, or testing.
+
+        Note that this affects
+        (i) how data that is passed into the model is interpreted (e.g. #s of positives, negatives)
+        (ii) whether inbatch negatives are used to compute logits and labels
+
+        Args:
+            phase (ModelPhase): The current phase of the model (TRAIN, VALIDATION, TEST).
+        """
+        old_phase = self._phase
+        self._phase = phase
+        logger.info(f"Changed model phase from {old_phase} to {phase}")
+
+    @property
+    def phase(self) -> ModelPhase:
+        return self._phase
+
+    def fetch_src_and_dst_embeddings(
+        self, edge_batch: EdgeBatch
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        num_edges = edge_batch.batch_size
+        node_embeddings_kt: torchrec.KeyedTensor = self.large_embeddings(
+            edge_batch.src_dst_pairs
+        )
+        logger.debug(f"node embeddings kt: {node_embeddings_kt}")
+        node_embeddings = (
+            node_embeddings_kt.values()
+        )  # [2 * num_edges, num_node_types * node_dim]
+        logger.debug(
+            f"node embeddings kt values: {node_embeddings, node_embeddings.shape}"
+        )
+        node_embeddings = node_embeddings.reshape(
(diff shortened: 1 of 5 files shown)


python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
```diff
@@ -0,0 +1,427 @@
...
+
+    Args:
+        model_config (ModelConfig): Configuration object containing model parameters.
+    """
+
+    def __init__(self, model_config: ModelConfig):
+        """Initialize the model with the given embedding configurations."""
+        super().__init__()
+
+        self.num_edge_types = model_config.num_edge_types
```
svij-sc, line 41: I guess we are trying to learn a unique mapping per edge type?
Is this common ?

python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
```diff
@@ -0,0 +1,427 @@
...
+        old_phase = self._phase
+        self._phase = phase
+        logger.info(f"Changed model phase from {old_phase} to {phase}")
+
+    @property
+    def phase(self) -> ModelPhase:
+        return self._phase
+
+    def fetch_src_and_dst_embeddings(
+        self, edge_batch: EdgeBatch
```
svij-sc, line 123: where is EdgeBatch defined?

python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
nshah171: Pretty common -- e.g. see https://arxiv.org/pdf/2309.12501 Tables 2/3.  These typically all involve learning some parameters per each relation/edge-type r.  the idea is the function of similarity between  src / dst (or head / tail) should be relation-specific.

python/gigl/experimental/knowledge_graph_embedding/lib/model/negative_sampling.py
```diff
@@ -0,0 +1,248 @@
...
+
+    B, D = src_embeddings.shape
+    N = batch_src_embeddings.shape[0]
+
+    # Precompute similarity matrix between [B] queries and [N] candidates
+    if scoring_function == SimilarityType.DOT:
+        sim_fn = lambda x, y: x @ y.T
+    elif scoring_function == SimilarityType.COSINE:
+        sim_fn = lambda x, y: F.normalize(x, dim=1) @ F.normalize(y, dim=1).T
+    elif scoring_function == SimilarityType.EUCLIDEAN:
```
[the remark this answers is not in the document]
nshah-sc: its complete i think -- this if/else block is just assigning the value of sim_fn which is used to compute similarities below.

also what's wrong with ValueError for this as i used below?  its meant to be used when we receive an argument with inappropriate/invalid value IIUC

python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
```diff
@@ -0,0 +1,427 @@
...
+    ModelPhase,
+    SimilarityType,
+)
+
+from gigl.common.logger import Logger
+
+logger = Logger()
+
+
+class HeterogeneousGraphSparseEmbeddingModel(nn.Module):
```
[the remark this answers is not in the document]
nshah-sc: added the link and another survey paper resource.

re API: i agree the API you shared is probably cleaner.  The main reason I put everything in "one giant model" is because there is work required to make whatever model is used compatible / working with torch.fx.  This requires some of the implementation choices made in the giant model currently and it is easier to build/test that one giant model for being fx-able rather than test a variety of them. 

in practice I think we could actually do something like what you suggested if we just break up the API appropriately, and have these models inherit / reuse fx-able implementations from the ones built now.  

i won't scope for this PR but adding a comment

nshah-sc:
> Also, curious if we had thought to closely reflect the pyg KGE models API here instead of roling out our custom versions?
https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#kge-models

Left a comment  which contains part of the answer: https://github.com/Snapchat/GiGL/pull/293#discussion_r2335258397

What we _could_ do: We could very likely mimic the structure where we have one module per KGE method, and define its `forward` by using some torchrec modules but otherwise ~copying other operations. 

What we _can't_ do:  Even if we do the above, there will be some material changes in the dataloading/negative sampling logic between the two worlds.  in PyG, all the models like `TransR`, `TransE` etc inherit from a `KGEModel` class, which looks [like this](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/kge/base.html#KGEModel).  This class has inbuilt functionalities for loading, loss computation, metric computation, negative sampling, etc. which work on small-scale and are opinionated, but in our case due to added flexibility desiderata and constraints of scale (torchrec, fx-ability) we would likely want to organize them differently (though this may not affect the user-facing API much)

python/gigl/experimental/knowledge_graph_embedding/lib/model/negative_sampling.py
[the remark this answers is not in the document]
svij-sc: ahhh... man that was confusing lol.
Can we define `def sim_fn(x, y):` somewhere else as `euclidiean` scoring and assign it here?

Alternatively, `scoring_function.impl`.

python/gigl/experimental/knowledge_graph_embedding/lib/model/negative_sampling.py
```diff
@@ -0,0 +1,248 @@
...
+    )  # [B, 1]
+
+    # Negative similarity matrix (B x N), depends on corruption side
+    if corrupt_side == NegativeSamplingCorruptionType.SRC:
+        neg_sim_matrix = sim_fn(batch_src_embeddings, dst_embeddings)  # [N, B]
+        neg_sim_matrix = neg_sim_matrix.T  # [B, N]
+    elif corrupt_side == NegativeSamplingCorruptionType.DST:
+        neg_sim_matrix = sim_fn(src_embeddings, batch_dst_embeddings)  # [B, N]
+    elif corrupt_side == NegativeSamplingCorruptionType.BOTH:
+        is_src_corruption = torch.rand_like(
```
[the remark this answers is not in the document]
svij-sc: but this isnt deterministic right?
Is there a way to confirm this?

svij-sc (approved):
Thanks for addressing the comments.

python/gigl/experimental/knowledge_graph_embedding/lib/data/edge_batch.py
```diff
@@ -0,0 +1,362 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from functools import partial
+from typing import Dict, List, Tuple
```
kmontemayor2-sc, line 5: fyi I'm trying to migrate to just using `dict, list, tuple`, etc. 

I think our cursor rules should auto generate them.

python/gigl/experimental/knowledge_graph_embedding/tests/test_against_batch_contrastive_similarity_relationwise.py
kmontemayor2-sc, line 1: Do these tests run? We should at minimum need an `__init__.py` here right?

If we moved this to `gigl/tests/unit/experimental` or equivalent they should be automatically discovered :)

python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
```diff
@@ -0,0 +1,439 @@
...
+
+    def __init__(self, model_config: ModelConfig):
+        """Initialize the model with the given embedding configurations."""
+        super().__init__()
+
+        self.num_edge_types = model_config.num_edge_types
+        self.node_emb_dim = model_config.node_embedding_dim
+        self.training_sampling_config = model_config.training_sampling
+        self.validation_sampling_config = model_config.validation_sampling
+        self.testing_sampling_config = model_config.testing_sampling
```
kmontemayor2-sc, line 52: hmmm do all of these need to be accessed later? 

I think it could be simpler to assign `active_sampling_config` directly in init, would reduce indirection here.

python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
```diff
@@ -0,0 +1,439 @@
...
+                self.active_sampling_config.positive_edge_batch_size :
+            ],
+        )
+        pos_labels, batch_neg_labels = (
+            labels[: self.active_sampling_config.positive_edge_batch_size],
+            labels[self.active_sampling_config.positive_edge_batch_size :],
+        )
+
+        pos_logits: torch.Tensor
+        pos_labels: torch.Tensor
```
kmontemayor2-sc, line 295: btw we should really avoid doing these pre-declarations.

With them, the below will [pass mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=b4e86636cc9e2fc1cef5f71334f2ce8a)

```
a: str
a.startswith("foo")
```

but without them, [mypy will fail](https://mypy-play.net/?mypy=latest&python=3.12&gist=387e36121d4b779af947b42d8c6c7bf9) (this is good, referencing undeclared variables is an error, we should catch it as early as possible.

```
a.startswith("foo")
```

python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
```diff
@@ -0,0 +1,439 @@
...
+        neg_logits = list()
+        neg_labels = list()
```
kmontemayor2-sc, line 297: nit, can we add types here? 
```suggestion
        neg_logits: list[torch.Tensor] = list()
        neg_labels: list[torch.Tensor] = list()
```

python/gigl/experimental/knowledge_graph_embedding/lib/data/edge_batch.py
```diff
@@ -0,0 +1,362 @@
...
+
+    # Determine the number of edges
+    num_edges = len(inputs)
+
+    # Preallocate torch tensors
+    edges = torch.empty((num_edges, 2), dtype=torch.int32)
+    condensed_edge_types = torch.empty(num_edges, dtype=torch.int32)
+
+    # Fill the preallocated torch tensors using direct indexing
+    for i, row in enumerate(inputs):
```
kmontemayor2-sc, line 287: how big is `inputs` in practice? And does this code get called in a hot loop? I'm concerned this could end up being slow.

But I guess that's a bridge we can cross when we get to.

kmontemayor2-sc (commented):
I think you have the necessary approvals already, left some mostly nits.

Would prefer if we move the tests folder to somewhere they get run though :)

python/gigl/experimental/knowledge_graph_embedding/lib/data/edge_batch.py
nshah-sc: this does run in a hot loop although this has not proven to be a major performance bottleneck in my observation.  there may be future optimizations which are warranted here but not high-prio at the time

python/gigl/experimental/knowledge_graph_embedding/lib/model/heterogeneous_graph_model.py
nshah-sc: yeah, they do all need to be accessed later.   

i'll leave this here for now as this class may be worth a bigger refactor /decomposition later

python/gigl/experimental/knowledge_graph_embedding/lib/data/edge_batch.py
nshah-sc: done

PR #294, opened 2025-08-26
Support GCP region override for vertex ai jobs

kmontemayor2-sc:
**Scope of work done**


svij-sc (commented):
The tests feel like overkill for a 2 line change, and prone to constant refactoring.

An alternative here is to do this in the wrapper or use omegaconf resolutions so we dont need to add conditionals in our trainer/inferencer code.

python/tests/unit/src/common/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,118 @@
...
+        resource_config_wrapper = GiglResourceConfigWrapper(resource_config)
+        with self.assertRaises(ValueError):
+            resource_config_wrapper.vertex_ai_inferencer_region
+
+    def test_vertex_ai_inferencer_region_error_dataflow_inferencer_config(self):
+        resource_config = GiglResourceConfig()
+        resource_config.shared_resource_config.common_compute_config.region = (
+            "us-central1"
+        )
+        resource_config.inferencer_resource_config.dataflow_inferencer_config.SetInParent()
```
svij-sc, line 111: do we need a test for every possible conditional scenario?
Seems too verbose 🍭  IMO

python/gigl/src/training/v2/glt_trainer.py
```diff
@@ -120,9 +120,10 @@ def __execute_VAI_training(
             if trainer_resource_config.timeout
             else None,
         )
+        region = resource_config.vertex_ai_trainer_region
         vertex_ai_service = VertexAIService(
             project=resource_config.project,
-            location=resource_config.region,
+            location=region,
```
svij-sc, line 126: any reason we cant do inline similar to how we do project, sa, bucket, etc?

Same comment for inferencer.

svij-sc (approved):
two minor comments, otherwise LGTM - thanks for the iteration

PR #301, opened 2025-08-27
e2e tests on python

svij-sc:
**Scope of work done**

Carrying over work @kmonte did prior internally w/ modifications.


python/gigl/orchestration/kubeflow/kfp_orchestrator.py
```diff
@@ -166,7 +167,28 @@ def run(
...
+    @overload
+    def wait_for_completion(self, run: str):
+        ...
+
+    def wait_for_completion(
+        self,
+        run: Union[
+            aiplatform.PipelineJob, str, list[aiplatform.PipelineJob], list[str]
+        ],
+    ):
```
kmontemayor2-sc, line 191: I think `str | PipelineJob` makes sense here, but having `str | list[str]` may be too flexible? Any reason we don't want to have some `wait_for_runs_completion` or something instead?

python/gigl/orchestration/kubeflow/kfp_orchestrator.py
```diff
@@ -176,6 +198,32 @@ def wait_for_completion(self, run: Union[aiplatform.PipelineJob, str]):
         Returns:
             None
         """
-        resource_name = run if isinstance(run, str) else run.resource_name
-        VertexAIService.wait_for_run_completion(resource_name)
-        logger.info(f"Pipeline run {resource_name} completed successfully.")
+        resource_names: list[str]
```
kmontemayor2-sc, line 201:
```suggestion
        resource_names: list[str] = []
```

Why not just do this? If we really want the pre-declaration, why not this way? 

But again, I don't think this really makes the code more readable.

python/gigl/orchestration/kubeflow/kfp_orchestrator.py
```diff
@@ -176,6 +198,32 @@ def wait_for_completion(self, run: Union[aiplatform.PipelineJob, str]):
...
+            return resource_name  # Convenience to return the run name for logging
+
+        with ThreadPoolExecutor(max_workers=len(resource_names)) as executor:
+            futures = [
+                executor.submit(wait_for_run_completion, resource_name=resource_name)
+                for resource_name in resource_names
+            ]
+            for future in as_completed(futures):
+                resource_name = future.result()
+                logger.info(f"Pipeline run {resource_name} completed successfully.")
```
kmontemayor2-sc, line 227: does this throw if the run fails? Should we add a test for this? It's easy enough to make a pipeline that fails in https://github.com/Snapchat/GiGL/blob/main/python/tests/integration/common/services/vertex_ai_test.py

python/tests/e2e_tests/e2e_test.py
kmontemayor2-sc, line 1: Should we put this is `GiGL/testing` if we want to make that folder for testing the "gigl system"?

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+
+logger = Logger()
+
+
+@dataclass
+class E2ETest:
+    """Configuration for a single E2E test specification."""
+
+    task_config_uri: str
+    resource_config_uri: str
```
kmontemayor2-sc, line 76: not blocking, future work. at some point it'd be nice to have a way to resolve directly to `URI`, do you think that's possible?

scripts/bootstrap_resource_config.py
kmontemayor2-sc, line 1: why is this here?

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+        help="The compiled pipeline definition to use.",
+    )
+    parser.add_argument(
+        "--test_names",
+        type=str,
+        default=[],
+        action="append",
+        help="""Optional names of specific tests to run. If not provided, all tests in the spec files will be run.
+        Example: --test_names=cora_glt_udl_test --test_names=cora_nalp_test""",
+    )
```
kmontemayor2-sc, line 163: hmmm, I am fine to keep the one unified `test_defs.yaml` file, but I feel like if we swapped to one file/test, then we wouldn't have this "test_names" bit which is imo a little awkward, as by default it runs everything?

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+
+        pipeline_job: PipelineJob = orchestrator.run(
+            applied_task_identifier=applied_task_identifier,
+            task_config_uri=task_config_uri,
+            resource_config_uri=resource_config_uri,
+            start_at=job.start_at,
+            stop_after=job.stop_after,
+            compiled_pipeline_path=compiled_pipeline_path,
+        )
+        if job.wait_for_completion == True:
```
kmontemayor2-sc, line 135:
```suggestion
        if job.wait_for_completion:
```

nit.

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+    resource_config_uri: str
+    name_suffix: str = (
+        "_on_${now:}"  # Makes use of gigl.common.omegaconf_resolvers#now_resolver()
+    )
+    uses_in_memory_sampling: bool = True
+    start_at: str = "config_populator"
+    stop_after: str = "post_processor"
+    wait_for_completion: bool = True
+    pipeline_tag: str = ""
+    run_labels: List[str] = field(default_factory=list)
```
kmontemayor2-sc, line 85:
```suggestion
    run_labels: list[str] = field(default_factory=list)
```

ditto generally

python/gigl/orchestration/kubeflow/runner.py
```diff
@@ -167,7 +167,7 @@ def _assert_required_flags(args: argparse.Namespace) -> None:
 logger = Logger()
 
 
-def _parse_additional_job_args(
+def parse_additional_job_args(
```
kmontemayor2-sc, line 170: any reason for this change?

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
+from __future__ import annotations
```
kmontemayor2-sc, line 1: what does this import do in this file?

Makefile
```diff
@@ -29,6 +30,8 @@ PY_TEST_FILES?="*_test.py"
 # You can override GIGL_TEST_DEFAULT_RESOURCE_CONFIG by setting it in your environment i.e.
 # adding `export GIGL_TEST_DEFAULT_RESOURCE_CONFIG=your_resource_config` to your shell config (~/.bashrc, ~/.zshrc, etc.)
 GIGL_TEST_DEFAULT_RESOURCE_CONFIG?=${PWD}/deployment/configs/unittest_resource_config.yaml
+# Default path for compiled KFP pipeline
+GIGL_E2E_TEST_COMPILED_PIPELINE_PATH:=/tmp/gigl/pipeline_${DATE}_${GIT_HASH}.yaml
```
kmontemayor2-sc, line 34: is there ever a time we won't want to compile the pipeline when we run the tests? Even for daily integration tests we'd check out a specific commit and then could compile from there right?

I guess I'm asking if we can remove this bit.

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+            start_at: "config_populator" # Optional, default is config_populator.
+            stop_after: "post_processor" # Optional, default is post_processor.
+            wait_for_completion: "True" # Optional, default is True.
+            pipeline_tag: "my_pipeline_tag" # Optional, default is empty.
+            run_labels:
+                - "gigl_commit=$(GIT_HASH)" # Optional, default is empty.
+                - "gigl_version=$(GIGL_VERSION)" # Optional, default is empty.
+        ```
+    The following arguments are optional:
+        --test_names: [Optional] List of test names to run form the test spec file. The value is repeated for each test.
```
mkolodner-sc, line 32: nit: form -> from

Also an optional rewording:
`--test_names: [Optional] Test name to run from the test spec file. The value can be repeated for running multiple tests`

mkolodner-sc: I know it's listed below, but can we also add a note here as well about what the behavior is if test_names is not provided?

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+            task_config_uri: "path/to/task_config.yaml"
+            resource_config_uri: "path/to/resource_config.yaml"
+            uses_in_memory_sampling: "True"
+            start_at: "config_populator" # Optional, default is config_populator.
+            stop_after: "post_processor" # Optional, default is post_processor.
+            wait_for_completion: "True" # Optional, default is True.
+            pipeline_tag: "my_pipeline_tag" # Optional, default is empty.
+            run_labels:
+                - "gigl_commit=$(GIT_HASH)" # Optional, default is empty.
+                - "gigl_version=$(GIGL_VERSION)" # Optional, default is empty.
```
mkolodner-sc, line 29: What do the defaults here entail if they are empty (i.e. what is an empty GiGL version?)

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+        --test_spec_uri: URI to a YAML file containing test specifications.
+        Supports yaml format:
+        ```yaml
+        tests:
+        - "name_of_the_test":
+            name_suffix: "_on_$(now:)"
+            task_config_uri: "path/to/task_config.yaml"
+            resource_config_uri: "path/to/resource_config.yaml"
+            uses_in_memory_sampling: "True"
+            start_at: "config_populator" # Optional, default is config_populator.
```
mkolodner-sc, line 23: nit: This reads a bit strange since this field is optional but it's listed in a section with the note "the following arguments are required"

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+    test_spec_uri = UriFactory.create_uri(args.test_spec_uri)
+    logger.info(f"Will load test spec from: {test_spec_uri}")
+    logger.info(f"Will load compiled pipeline from: {os.getcwd()}")
+    test_spec: E2ETestsSpec = load_resolved_yaml(test_spec_uri, E2ETestsSpec)
+    filtered_tests: Dict[str, E2ETest]
+    if args.test_names:
+        filtered_tests = {
+            test_name: test_spec.tests[test_name]
+            for test_name in args.test_names
+            if test_name in args.test_names
```
mkolodner-sc, line 177: wouldn't this conditional always be True? Is this supposed to be `if test in test_spec.tests[test_name]`? 

If so, it'd be nice if there was a log we could provide to say that the test argument was not found in the test spec, rather than silently not run that test.

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+        --compiled_pipeline_path: The KFP pipeline definition to use.
+        --test_spec_uri: URI to a YAML file containing test specifications.
+        Supports yaml format:
+        ```yaml
+        tests:
+        - "name_of_the_test":
+            name_suffix: "_on_$(now:)"
+            task_config_uri: "path/to/task_config.yaml"
+            resource_config_uri: "path/to/resource_config.yaml"
+            uses_in_memory_sampling: "True"
```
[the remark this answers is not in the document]
mkolodner-sc: Also, where is this field being used? It doesn't seem like its needed as input to the runner [[1](https://github.com/Snapchat/GiGL/blob/47bd3f1ea9fb59271a0f1ca3fc541cdbcbe04873/python/gigl/orchestration/kubeflow/runner.py#L220)]

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+                --compiled_pipeline_path={compiled_pipeline_path} \
+                --job_name={full_job_name} \
+                --task_config_uri={job.task_config_uri} \
+                --resource_config_uri={job.resource_config_uri} \
+                --start_at={job.start_at} \
+                --stop_after={job.stop_after} \
+                --pipeline_tag={job.pipeline_tag} \
+                --uses_in_memory_sampling={job.uses_in_memory_sampling} \
+                --run_labels={job.run_labels} \
+                --wait_for_completion={job.wait_for_completion} \
```
mkolodner-sc, line 117: It seems the arg used here is `--wait` [[1](https://github.com/Snapchat/GiGL/blob/47bd3f1ea9fb59271a0f1ca3fc541cdbcbe04873/python/gigl/orchestration/kubeflow/runner.py#L20)]

python/tests/e2e_tests/e2e_test.py
```diff
@@ -0,0 +1,186 @@
...
+        Supports yaml format:
+        ```yaml
+        tests:
+        - "name_of_the_test":
+            name_suffix: "_on_$(now:)"
+            task_config_uri: "path/to/task_config.yaml"
+            resource_config_uri: "path/to/resource_config.yaml"
+            uses_in_memory_sampling: "True"
+            start_at: "config_populator" # Optional, default is config_populator.
+            stop_after: "post_processor" # Optional, default is post_processor.
```
mkolodner-sc, line 24: Should the default here be None? This aligns more closely with the default value of the runner [[1](https://github.com/Snapchat/GiGL/blob/47bd3f1ea9fb59271a0f1ca3fc541cdbcbe04873/python/gigl/orchestration/kubeflow/runner.py#L263)]

mkolodner-sc (commented):
Thanks Shubham, did an initial pass!

python/gigl/orchestration/kubeflow/kfp_orchestrator.py
svij-sc: done

python/tests/e2e_tests/e2e_test.py
svij-sc: sure

python/tests/e2e_tests/e2e_test.py
svij-sc: i changed this a bit.

python/tests/e2e_tests/e2e_test.py
svij-sc: fixed

Makefile
svij-sc: I don't think we can remove this from here.
This is essentially here so we have a common path to write/read from for our compile --> running e2e test step.

Id say the way to remove this is to increase complexity of e2e_test and adding compile step in there - but then we have two seperate compile entry points i.e. this make file and e2e_test file - we can fix this but this is a much bigger refactor.

PR #303, opened 2025-09-02
[DRAFT] Node label tech debt

mkolodner-sc:
**Scope of work done**

### DIFF
--- a/python/gigl/distributed/dist_partitioner.py
+++ b/python/gigl/distributed/dist_partitioner.py
@@ -192,6 +192,8 @@ class DistPartitioner:
         self._max_node_ids: Optional[dict[NodeType, int]] = None
         self._node_feat: Optional[dict[NodeType, torch.Tensor]] = None
         self._node_feat_dim: Optional[dict[NodeType, int]] = None
+        self._node_labels: Optional[dict[NodeType, torch.Tensor]] = None
+        self._node_labels_dim: Optional[dict[NodeType, int]] = None
 
         self._edge_index: Optional[dict[EdgeType, torch.Tensor]] = None
         self._edge_ids: Optional[dict[EdgeType, tuple[int, int]]] = None
@@ -510,6 +512,41 @@ class DistPartitioner:
         for node_type in input_node_features:
             self._node_feat_dim[node_type] = input_node_features[node_type].shape[1]
 
+    def register_node_labels(
+        self, node_labels: Union[torch.Tensor, dict[NodeType, torch.Tensor]]
+    ) -> None:
+        """
+        Registers the node labels to the partitioner.
+
+        This function assumes that node label at index N on the current machine corresponds to the node id at index N
+        on the current machine, which should be registered through `register_node_ids`.
+
+        For optimal memory management, it is recommended that the reference to node_labels tensor be deleted
+        after calling this function using del <tensor>, as maintaining both original and intermediate tensors can cause OOM concerns.
+        We do not need to perform `all_gather` calls here since register_node_ids is responsible for determining total number of nodes
+        across all ranks.
+        Args:
+            node_labels (Union[torch.Tensor, dict[NodeType, torch.Tensor]]): Input node labels which is either a torch.Tensor if homogeneous or a Dict if heterogeneous
+        """
+
+        self._assert_and_get_rpc_setup()
+
+        logger.info("Registering Node Labels ...")
+
+        input_node_labels = self._convert_node_entity_to_heterogeneous_format(
+            input_node_entity=node_labels
+        )
+
+        assert (
+            input_node_labels
+        ), "Node labels is an empty dictionary. Please provide node labels to register."
+
+        self._node_labels = convert_to_tensor(input_node_labels, dtype=torch.float32)
+
+        self._node_labels_dim = {}
+        for node_type in input_node_labels:
+            self._node_labels_dim[node_type] = input_node_labels[node_type].shape[1]
+
     def register_edge_features(
         self, edge_features: Union[torch.Tensor, dict[EdgeType, torch.Tensor]]
     ) -> None:
@@ -582,7 +619,7 @@ class DistPartitioner:
 
     def _partition_single_chunk_data(
         self,
-        input_data: Optional[Tuple[torch.Tensor, ...]],
+        input_data: Optional[Tuple[Optional[torch.Tensor], ...]],
         rank_indices: torch.Tensor,
         partition_function: Callable[[torch.Tensor, Tuple[int, int]], torch.Tensor],
         chunk_start_pos: int,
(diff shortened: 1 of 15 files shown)


python/gigl/distributed/dist_partitioner.py
```diff
@@ -510,6 +517,41 @@ def register_node_features(
...
+        )
+
+        assert (
+            input_node_labels
+        ), "Node labels is an empty dictionary. Please provide node labels to register."
+
+        self._node_labels = convert_to_tensor(input_node_labels, dtype=torch.int64)
+
+        self._node_labels_dim = {}
+        for node_type in input_node_labels:
```
kmontemayor2-sc, line 552: maybe this was already here but this means when people re-register that the existing labels (or whatever else, I assume this pattern is used everywhere) are deleted.

This means that:
1. We don't explicitly GC the removed data, this seems counter to our general philosophy. here
2. It may be surprising to users that "register" works this way.

e.g. if I have some:

```py

foo = Foo()
foo.register(a)
foo.register(b)

# I'd assume a and b are both registered now
```

If this is the case, I think we should throw if someone re-registers (I think it's fine to allow only one register/type/instance, but we should make it explicit).

This can be a follow-up though (maybe for Sam as a starter?

python/gigl/distributed/dist_partitioner.py
```diff
@@ -580,9 +622,30 @@ def register_labels(
...
+        Args:
+            field_name (str): The name of the instance attribute (e.g., '_node_feat')
+            entity_key (Union[NodeType, EdgeType]): The key to remove from the dictionary
+        Raises:
+            AttributeError: If the field_name does not exist in the class
+        """
+        field = getattr(self, field_name)
+        assert isinstance(
+            field, dict
+        ), f"Field {field_name} is not a dict, got {type(field)}"
```
kmontemayor2-sc, line 640: it's valid for `field` to be `None` right? 

Do our tests pass? I'd hope they fail here.

python/gigl/distributed/dist_partitioner.py
```diff
@@ -802,76 +865,125 @@ def _partition_node_features(
...
-            feature_partition_data.ids is not None
-            and feature_partition_data.ids.numel()
-            != torch.unique(feature_partition_data.ids).numel()
-        ):
-            raise ValueError(
-                f"Node ids are not unique for node type {node_type}. Please ensure that node ids are unique and contiguous across all machines from 0 to total_num_nodes - 1."
-            )
+            else:
+                node_label_partition_data = None
+        else:
```
kmontemayor2-sc, line 961: (style-nit.) which "if" does this "else" belong to? It may be useful to add "else: # if ..." to be explicit here.

kmontemayor2-sc (commented):
Do we need to update internal usages once this goes in?

python/gigl/distributed/dist_partitioner.py
mkolodner-sc: Oops you're right this is being caught by our tests currently, will fix this, thanks!

`AssertionError: Field _node_labels is not a dict, got <class 'NoneType'>`

python/gigl/distributed/dist_partitioner.py
```diff
@@ -510,6 +517,41 @@ def register_node_features(
         for node_type in input_node_features:
             self._node_feat_dim[node_type] = input_node_features[node_type].shape[1]
 
+    def register_node_labels(
```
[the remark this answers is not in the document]
mkolodner-sc: Yes, the renaming will happen on a larger scale in a separate PR [[current draft for this](https://github.com/Snapchat/GiGL/pull/305)]

python/gigl/distributed/dist_partitioner.py
```diff
@@ -628,7 +692,7 @@ def _partition_single_chunk_data(
 
     def _partition_by_chunk(
         self,
-        input_data: Optional[Tuple[torch.Tensor, ...]],
+        input_data: Optional[Tuple[Optional[torch.Tensor], ...]],
```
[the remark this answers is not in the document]
mkolodner-sc: I updated the logic a bit here so we don't need to have places where we are partitioning a tuple with some elements are None, thanks!

python/gigl/distributed/dist_partitioner.py
```diff
@@ -802,76 +871,134 @@ def _partition_node_features(
...
+            input_data = (node_features, node_ids)
+        elif node_labels is not None:
+            input_data = (node_labels, node_ids)
+        else:
+            raise ValueError(
+                f"Found no node features or node labels to partition for node type {node_type}"
+            )
+
+        has_node_features = node_features is not None
+        has_node_labels = node_labels is not None
```
mkolodner-sc, line 913: For some reason, doing 
```
input_data: Tuple[torch.Tensor, ...]
        if has_node_features and has_node_labels:
            input_data = (node_features, node_labels, node_ids)
        elif has_node_features:
            input_data = (node_features, node_ids)
        elif has_node_labels:
            input_data = (node_labels, node_ids)
        else:
            raise ValueError(
                f"Found no node features or node labels to partition for node type {node_type}"
            )
```
leads to mypy failures, not recognizing that the `node_features` and `node_labels` tensors are of type `torch.Tensor` and not `Optional[torch.Tensor]`

python/gigl/distributed/dist_partitioner.py
```diff
@@ -950,8 +963,9 @@ def _node_feature_partition_fn(node_feature_ids, _):
 
             # Partitioned node labels are stored at the 1st index in each tuple of the partitioned results.
             if has_node_labels:
+                node_label_ind = 1 if has_node_features else 0
                 node_label_partition_data = FeaturePartitionData(
-                    feats=torch.cat([r[1] for r in partitioned_results]),
+                    feats=torch.cat([r[node_label_ind] for r in partitioned_results]),
```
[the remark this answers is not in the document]
mkolodner-sc: Different partition functions may want to partition different items. See an example [here](https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/dist_range_partitioner.py#L206-L217), which contains edge indices and edge ids that are partitioned or here, which partitions [edge labels](https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/dist_partitioner.py#L1082-L1083), both of which are different from the current use case which partitions node ids, node features, and node labels.

I prefer the current solution to a giant dataclass which contains multiple unused fields, which for partitioning node ids and features, for example, may look something like
```
partittioned_dataclass(
   node_ids: torch.Tensor
   node_features: torch.Tensor
   node_labels: torch.Tensor
   edge_ids: None
   edge_src: None
   edge_dst: None
   edge_feat: None
)
```
Ideally, we should avoid needing to send RPC calls with None if possible and keep the size of the `input_data` as minimal as possible. It may be small, but it could add up when we are making hundreds of millions of RPC calls.

PR #307, opened 2025-09-03
Add ability to resolve any yaml --> dataclass using utility + introduce ${git_hash:} resolver

svij-sc:
**Scope of work done**

- Added suport to load arbritary yamls given a dataclass definition
- Added custom Omegaconf resolver for `${git_hash:}`

Both of these will be used here: https://github.com/Snapchat/GiGL/pull/301

### DIFF
--- a/docs/user_guide/config_guides/task_config_guide.md
+++ b/docs/user_guide/config_guides/task_config_guide.md
@@ -126,6 +126,7 @@ GiGL makes use of custom [Omegaconf resolvers](https://omegaconf.readthedocs.io/
 allow us to expose macros that are resolved at runtime instead of being hardcoded. Our resolvers are defined in
 [omegaconf_resolvers.py](python/gigl/common/omegaconf_resolvers.py)
 
+
 ```{note}
 In tabularized GiGL: Subgraph Sampler, and Split Generator do not have support for custom resolvers. In most cases this should not be a problem as the Config Populator should populate any needed custom resolution to your frozen configs.
 ```
@@ -166,4 +167,22 @@ yesterday: "2023-12-14"
 tomorrow_plus_5_hours_30_min_15_sec: "2023-12-16 20:00:37"
 next_week: "2023-12-22"
 multiple_args: "20231130:20231214"
+```
+
+### Git Hash Resolver
+
+
+Examples:
+```yaml
+experiment:
+    commit: "${git_hash:}"
+    model_version: "model_${git_hash:}"
+```
+
+Assuming you are scheduling workflows from an active git repo with the current commit hash: `ABC1235`, this would resolve:
+
+```yaml
+experiment:
+    commit: "ABC1235"
+    model_version: "model_ABC1235"
 ```
--- a/docs/user_guide/config_guides/task_config_guide.md
+++ b/docs/user_guide/config_guides/task_config_guide.md
@@ -126,7 +126,6 @@ GiGL makes use of custom [Omegaconf resolvers](https://omegaconf.readthedocs.io/
 allow us to expose macros that are resolved at runtime instead of being hardcoded. Our resolvers are defined in
 [omegaconf_resolvers.py](python/gigl/common/omegaconf_resolvers.py)
 
-
 ```{note}
 In tabularized GiGL: Subgraph Sampler, and Split Generator do not have support for custom resolvers. In most cases this should not be a problem as the Config Populator should populate any needed custom resolution to your frozen configs.
 ```
@@ -171,18 +170,26 @@ multiple_args: "20231130:20231214"
 
 ### Git Hash Resolver
 
+This resolver returns the current git hash if one is available. Takes no arguments and returns the git hash as a string.
+Specifically this returns the SHA that is returned when the following is run in the active working directory:
+
+```bash
+git rev-parse HEAD
+```
 
 Examples:
+
 ```yaml
 experiment:
     commit: "${git_hash:}"
     model_version: "model_${git_hash:}"
 ```
 
-Assuming you are scheduling workflows from an active git repo with the current commit hash: `ABC1235`, this would resolve:
+Assuming you are scheduling workflows from an active git repo with the current commit hash:
+`9d42b423b65961692ffc650a0714a63a1b695b12`, this would resolve:
 
 ```yaml
 experiment:
-    commit: "ABC1235"
-    model_version: "model_ABC1235"
+    commit: "9d42b423b65961692ffc650a0714a63a1b695b12"
+    model_version: "9d42b423b65961692ffc650a0714a63a1b695b12"
 ```
--- a/docs/user_guide/config_guides/task_config_guide.md
+++ b/docs/user_guide/config_guides/task_config_guide.md
@@ -177,6 +177,8 @@ Specifically this returns the SHA that is returned when the following is run in 
 git rev-parse HEAD
 ```
 
+If no git repository is found, or there is an error, will return empty string.
+
 Examples:
 
 ```yaml
@@ -191,5 +193,5 @@ Assuming you are scheduling workflows from an active git repo with the current c
 ```yaml
 experiment:
     commit: "9d42b423b65961692ffc650a0714a63a1b695b12"
-    model_version: "9d42b423b65961692ffc650a0714a63a1b695b12"
+    model_version: "model_9d42b423b65961692ffc650a0714a63a1b695b12"
 ```
--- a/python/gigl/common/omegaconf_resolvers.py
+++ b/python/gigl/common/omegaconf_resolvers.py
@@ -150,10 +150,11 @@ def git_hash_resolver() -> str:
         subprocess.CalledProcessError,
         subprocess.TimeoutExpired,
         FileNotFoundError,
-    ) as e:
-        raise RuntimeError(
-            f"Could not retrieve git hash - git command failed or not in a git repository: {e}"
+    ):
+        logger.info(
+            "Could not retrieve git hash - git command failed or not in a git repository"
         )
+        return ""
 
 
 def register_resolvers() -> None:
--- a/python/gigl/common/omegaconf_resolvers.py
+++ b/python/gigl/common/omegaconf_resolvers.py
@@ -150,11 +150,10 @@ def git_hash_resolver() -> str:
         subprocess.CalledProcessError,
         subprocess.TimeoutExpired,
         FileNotFoundError,
-    ):
-        logger.warning(
-            "Could not retrieve git hash - git command failed or not in a git repository"
+    ) as e:
+        raise RuntimeError(
+            f"Could not retrieve git hash - git command failed or not in a git repository: {e}"
         )
-        return ""
 
 
 def register_resolvers() -> None:
--- a/python/gigl/common/omegaconf_resolvers.py
+++ b/python/gigl/common/omegaconf_resolvers.py
@@ -4,6 +4,7 @@ This module contains custom OmegaConf resolvers that can be used in YAML configu
 files to provide dynamic values during configuration loading.
 """
 
+import subprocess
 from datetime import datetime, timedelta
 
 from omegaconf import OmegaConf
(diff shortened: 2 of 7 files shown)


docs/user_guide/config_guides/task_config_guide.md
```diff
@@ -167,3 +168,21 @@ tomorrow_plus_5_hours_30_min_15_sec: "2023-12-16 20:00:37"
...
 ```
+
+### Git Hash Resolver
+
+
+Examples:
+```yaml
+experiment:
+    commit: "${git_hash:}"
+    model_version: "model_${git_hash:}"
```
kmontemayor2-sc, line 179: Should we specify what the git hash is?

I think this is a useful feature, but if we ever want to move to a `pip install gigl` world then this "git hash" may be a bit confusing.

python/gigl/common/omegaconf_resolvers.py
```diff
@@ -120,16 +121,60 @@ def now_resolver(*args: str) -> str:
...
+        )
+        return result.stdout.strip()
+    except (
+        subprocess.CalledProcessError,
+        subprocess.TimeoutExpired,
+        FileNotFoundError,
+    ):
+        logger.warning(
+            "Could not retrieve git hash - git command failed or not in a git repository"
+        )
```
kmontemayor2-sc, line 156: related to my previous comment, I feel like we should throw if we can't resolve. 

I understand there's some issue here with this and our current e2e tests, but maybe we should just resolve the e2e test issue and use a datetime as a tag instead?

python/tests/unit/common/utils/yaml_loader_test.py
```diff
@@ -0,0 +1,87 @@
...
+        """Set up test fixtures."""
+        super().setUp()
+        register_resolvers()
+        self.temp_file = tempfile.NamedTemporaryFile(
+            mode="w", suffix=".yaml", delete=False
+        )
+
+    def tearDown(self):
+        self.temp_file.close()
+        os.remove(self.temp_file.name)
```
kmontemayor2-sc, line 39: nit.
```suggestion
        os.remove(self.temp_file.name)
        super().tearDown()
```

python/tests/unit/common/utils/yaml_loader_test.py
```diff
@@ -0,0 +1,87 @@
...
+        expected_result = _Complex_TestConfig(
+            basic_config=_SubConfig(
+                name=f"experiment_{patch_datetime.strftime('%Y%m%d')}",
+                value=42,
+                enabled=True,
+                tags=[f"tag_{patch_commit_hash}", "42"],
+            ),
+            description="This is a test description",
+        )
+        result: _Complex_TestConfig
```
kmontemayor2-sc, line 69: again, I am *really* not a fan of these, we are blowing a hole through our static checks everytime we do this.

I don't think this makes the code more readable, we annotate the type directly later anyways (which I also think is redundant but that's a separate issue...)

python/tests/unit/common/utils/yaml_loader_test.py
```diff
@@ -0,0 +1,87 @@
...
+from gigl.common.omegaconf_resolvers import register_resolvers
+from gigl.common.utils.yaml_loader import load_resolved_yaml
+
+
+@dataclass
+class _SubConfig:
+    name: str
+    value: int
+    enabled: bool = True
+    tags: List[str] = field(default_factory=list)
```
kmontemayor2-sc, line 19: should we add a test where we set this?

python/tests/unit/common/utils/yaml_loader_test.py
```diff
@@ -0,0 +1,87 @@
...
+        contents = """
+basic_config:
+    name: "experiment_${now:%Y%m%d}"
+    value: 42
+    enabled: true
+    tags:
+        - "tag_${git_hash:}"
+        - "${basic_config.value}" # resolves to 42
+description: "This is a test description"
+"""
```
kmontemayor2-sc, line 53: nit. textwrap.dedent here?

python/tests/unit/common/utils/yaml_loader_test.py
```diff
@@ -0,0 +1,87 @@
...
+                tags=[f"tag_{patch_commit_hash}", "42"],
+            ),
+            description="This is a test description",
+        )
+        result: _Complex_TestConfig
+        with patch(
+            "gigl.common.omegaconf_resolvers.subprocess.run"
+        ) as mock_subprocess_run, patch(
+            "gigl.common.omegaconf_resolvers.datetime"
+        ) as mock_datetime:
```
kmontemayor2-sc, line 74: meta-comment, not blocking, I don't think we should need to feel the need to test our omegaconf resolvers here, I feel it's a bit out of the scope of the unit test. 

And the yaml_loader doesn't use the resolvers by default anyways, we have to register them in this test case, so I think it's a bit misleading to test behavior that isn't supported out of the box.

python/gigl/common/utils/yaml_loader.py
```diff
@@ -0,0 +1,27 @@
...
+def load_resolved_yaml(uri: Uri, type_of_object: Type[T]) -> T:
+    with FileLoader().load_to_temp_file(uri) as tf:
+        test_spec_data = OmegaConf.load(tf.name)
+
+    # Merge OmegaConf structured config with loaded data for validation
+    merged_config = OmegaConf.merge(
+        OmegaConf.structured(type_of_object), test_spec_data
+    )
+
+    # Convert to strongly typed E2ETestsSpec object
```
nshah-sc, line 26: shouldn't say E2ETestsSpec but rather T?

python/gigl/common/omegaconf_resolvers.py
```diff
@@ -120,16 +121,60 @@ def now_resolver(*args: str) -> str:
...
+
 def register_resolvers() -> None:
     """Register all custom OmegaConf resolvers.
 
     This function should be called at application startup to register
     all custom resolvers with OmegaConf.
     """
-    logger.info("Registering OmegaConf resolvers")
     if not OmegaConf.has_resolver("now"):
+        logger.info("Registering OmegaConf resolver 'now'")
```
nshah-sc, line 167: what is the "now" resolver?  / any doc on it?

docs/user_guide/config_guides/task_config_guide.md
```diff
@@ -126,6 +126,7 @@ GiGL makes use of custom [Omegaconf resolvers](https://omegaconf.readthedocs.io/
 allow us to expose macros that are resolved at runtime instead of being hardcoded. Our resolvers are defined in
 [omegaconf_resolvers.py](python/gigl/common/omegaconf_resolvers.py)
 
+
```
nshah-sc, line 129: should this doc be in the task config guide?  

do we support resolvers in all the non-tabularized components?

python/gigl/common/utils/yaml_loader.py
```diff
@@ -0,0 +1,27 @@
...
+    return OmegaConf.to_object(merged_config)
```
nshah-sc, line 27: i believe this should work for nested dataclasses as well, e.g. when type_of_object is say a dataclass called "Config" and has sub-configs like AppConfig, ModelConfig, etc. which are also dataclasses.  have you tried?

nshah-sc (approved):

python/gigl/common/omegaconf_resolvers.py
```diff
@@ -120,16 +121,59 @@ def now_resolver(*args: str) -> str:
...
             "OmegaConf resolver 'now' already registered, skipping registration"
         )
+
+    if not OmegaConf.has_resolver("git_hash"):
+        logger.info("Registering OmegaConf resolver 'git_hash'")
+        OmegaConf.register_new_resolver("git_hash", git_hash_resolver)
+    else:
+        logger.info(
+            "OmegaConf resolver 'git_hash' already registered, skipping registration"
+        )
```
kmontemayor2-sc, line 179: unrelated but WDYT about demoting these to logging.debug? This thing may get called many times and I think as we add more resolvers it'll create a lot of logspam.

python/gigl/common/omegaconf_resolvers.py
```diff
@@ -120,16 +121,59 @@ def now_resolver(*args: str) -> str:
...
+        return result.stdout.strip()
+    except (
+        subprocess.CalledProcessError,
+        subprocess.TimeoutExpired,
+        FileNotFoundError,
+    ) as e:
+        raise RuntimeError(
+            f"Could not retrieve git hash - git command failed or not in a git repository: {e}"
+        )
```
kmontemayor2-sc, line 156: do our current e2e tests pass with it like this?

python/gigl/common/utils/yaml_loader.py
```diff
@@ -0,0 +1,27 @@
...
+from gigl.common import Uri
+from gigl.common.logger import Logger
+from gigl.common.omegaconf_resolvers import register_resolvers
+from gigl.src.common.utils.file_loader import FileLoader
+
+logger = Logger()
+
+T = TypeVar("T")
+
+register_resolvers()
```
kmontemayor2-sc, line 14: should we be calling `register_resolvers()` at a module level like this? I think generally we should avoid executing code as a part of imports (and `import yaml_loader` will execute this, right). We had some issues with this in the past e.g what changed inhttps://github.com/Snapchat/GiGL/commit/91170a13116a8a5fee802ef8b20a2ae230442465#diff-ff4187f18de1ebbfe8919528758efebb70d1767ecb08cf468342b46ededd297b

python/tests/unit/common/utils/yaml_loader_test.py
```diff
@@ -0,0 +1,91 @@
...
+class _Complex_TestConfig:
+    basic_config: _SubConfig
+    description: str
+
+
+class YamlLoaderTest(unittest.TestCase):
+    def setUp(self):
+        """Set up test fixtures."""
+        super().setUp()
+        register_resolvers()
```
kmontemayor2-sc, line 33: Hmm, maybe part of my concern here https://github.com/Snapchat/GiGL/pull/307/files#r2319352196 was this bit - it shouldn't be needed right? Let's remove it?

python/tests/unit/common/utils/yaml_loader_test.py
kmontemayor2-sc: Synced offline, my core issue here is that doing this allows the following code to go through [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=3c0f17812aa82bc2ff04b4af5b9850ce).

```py
a: str
a.startswith("foo")
```

Whereas no declaration [throws](https://mypy-play.net/?mypy=latest&python=3.12&gist=dd1c8da7b7c4d49d32848f6af5846a81) (which we want!)

```py
a.startswith("foo")
```

which is imo a huge issue and introduces an entirely new class of bugs. 

AIUI @svij-sc agrees we shouldn't be introducing these things, right?

docs/user_guide/config_guides/task_config_guide.md
```diff
@@ -167,3 +167,29 @@ tomorrow_plus_5_hours_30_min_15_sec: "2023-12-16 20:00:37"
...
+    model_version: "model_${git_hash:}"
+```
+
+Assuming you are scheduling workflows from an active git repo with the current commit hash:
+`9d42b423b65961692ffc650a0714a63a1b695b12`, this would resolve:
+
+```yaml
+experiment:
+    commit: "9d42b423b65961692ffc650a0714a63a1b695b12"
+    model_version: "9d42b423b65961692ffc650a0714a63a1b695b12"
```
kmontemayor2-sc, line 194:
```suggestion
    model_version: "model_9d42b423b65961692ffc650a0714a63a1b695b12"
```

nit.

python/gigl/common/omegaconf_resolvers.py
svij-sc: sure.

python/gigl/common/utils/yaml_loader.py
svij-sc: if not here we would have to  either 1. manually register it for every entry point, or call it repeatedly any time yaml is resolved.
Given that this is idempotent maybe its less concerning?

python/gigl/common/omegaconf_resolvers.py
svij-sc: Actually, this is changing logic from what existed before and is both: Failing what didnt fail before, increasing the scope of this PR if we want to discuss merits of our prior decision before, etc.

python/gigl/common/omegaconf_resolvers.py
svij-sc: See: https://github.com/Snapchat/GiGL/pull/301#issuecomment-3264075970

python/gigl/common/utils/yaml_loader.py
```diff
@@ -0,0 +1,27 @@
...
+    with FileLoader().load_to_temp_file(uri) as tf:
+        test_spec_data = OmegaConf.load(tf.name)
+
+    # Merge OmegaConf structured config with loaded data for validation
+    merged_config = OmegaConf.merge(
+        OmegaConf.structured(type_of_object), test_spec_data
+    )
+
+    # Convert to strongly typed T object
+    return cast(T, OmegaConf.to_object(merged_config))
```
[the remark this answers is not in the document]
svij-sc: it wasnt getting the right type here so an explicit cast is needed.
Test added to ensure this is correct runtime behaviour too.

python/tests/unit/common/utils/yaml_loader_test.py
```diff
@@ -0,0 +1,89 @@
...
+            "gigl.common.omegaconf_resolvers.datetime"
+        ) as mock_datetime:
+            mock_result = MagicMock()
+            mock_result.stdout = patch_commit_hash
+            mock_subprocess_run.return_value = mock_result
+            mock_datetime.now.return_value = patch_datetime
+
+            uri = LocalUri(self.temp_file.name)
+            result: _Complex_TestConfig = load_resolved_yaml(uri, _Complex_TestConfig)
+            self.assertTrue(isinstance(result, _Complex_TestConfig))
```
kmontemayor2-sc, line 83: nit. self.assertIsInstance

PR #309, opened 2025-09-04
Migrate DistDataset to it's own file

mkolodner-sc:
**Scope of work done**


python/gigl/distributed/dist_dataset.py
```diff
@@ -0,0 +1,1014 @@
+# Originally taken from https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/distributed/dist_dataset.py
```
mkolodner-sc, line 1: There is no logic change, this is just being moved from the previous file

kmontemayor2-sc (commented):
Should we also update https://github.com/Snapchat/GiGL/blob/main/python/gigl/distributed/__init__.py? 

Also the test files?

PR #315, opened 2025-09-08
Simplify Dataset `build` function

mkolodner-sc:
**Scope of work done**

Our dataset.build() function has greatly grown in complexity and is now several hundreds of lines long. We should aim to shorten this so that our code is cleaner and more readable

### DIFF
--- a/python/gigl/distributed/dist_dataset.py
+++ b/python/gigl/distributed/dist_dataset.py
@@ -1038,7 +1038,8 @@ def _prepare_feature_data(
     elif isinstance(partitioned_data, Mapping):
         assert (
             len(partitioned_data) > 0
-        ), f"Expected at least one entity type in partitioned data, but got {partitioned_data.keys()}."
+        ), f"Expected at least one entity type in partitioned data, but got no entities. In heterogeneous settings, \
+            please make sure you are registering entities with non-empty fields i.e. not an empty dictionary."
         # Heterogeneous case
         assert isinstance(
             partition_book, Mapping
@@ -1046,7 +1047,8 @@ def _prepare_feature_data(
             Got partition book of type {type(partition_book)}."
         assert (
             len(partition_book) > 0
-        ), f"Expected at least one entity type in partition book, but got {partition_book.keys()}."
+        ), f"Expected at least one entity type in partition book, but got no entities. In heterogeneous settings, \
+            please make sure you are registering entities with non-empty fields i.e. not an empty dictionary."
 
         # Extract features and IDs by type
         features_per_entity_type: dict[_EntityType, torch.Tensor] = {}


python/gigl/distributed/dist_dataset.py
```diff
@@ -783,6 +544,263 @@ def build(
...
+            partitioned_edge_index(Union[GraphPartitionData, dict[EdgeType, GraphPartitionData]]): The partitioned graph data
+        """
+
+        # Edge Index refers to the [2, num_edges] tensor representing pairs of nodes connecting each edge
+        # Edge IDs refers to the [num_edges] tensor representing the unique integer assigned to each edge
+        edge_index: Union[torch.Tensor, dict[EdgeType, torch.Tensor]]
+        edge_ids: Union[Optional[torch.Tensor], dict[EdgeType, Optional[torch.Tensor]]]
+        if isinstance(partitioned_edge_index, GraphPartitionData):
+            edge_index = partitioned_edge_index.edge_index
+            edge_ids = partitioned_edge_index.edge_ids
```
kmontemayor2-sc, line 569: Can we move away from these pre-declarations and instead inline them to the first time the variable is defined? 
```suggestion
        if isinstance(partitioned_edge_index, GraphPartitionData):
            edge_index: Union[torch.Tensor, dict[EdgeType, torch.Tensor]]
 = partitioned_edge_index.edge_index
            edge_ids: Union[Optional[torch.Tensor], dict[EdgeType, Optional[torch.Tensor]]]
 = partitioned_edge_index.edge_ids
```

python/gigl/distributed/dist_dataset.py
```diff
@@ -949,6 +967,90 @@ def add_clamped_counts(
...
+        feature_ids = partitioned_data.ids
+
+        if isinstance(partition_book, RangePartitionBook):
+            id_to_index = partition_book.id2index
+        else:
+            id_to_index = id2idx(feature_ids)
+
+        return features, id_to_index
+
+    elif isinstance(partitioned_data, Mapping) and len(partitioned_data) > 0:
```
kmontemayor2-sc, line 1023: should we be returning `None` if the mapping is empty?

python/gigl/distributed/dist_dataset.py
```diff
@@ -949,6 +967,90 @@ def add_clamped_counts(
...
+            dict[_EntityType, torch.Tensor],
+        ]
+    ],
+    Optional[
+        Union[
+            TensorDataType,
+            dict[_EntityType, TensorDataType],
+        ]
+    ],
+]:
```
kmontemayor2-sc, line 991: nit. considered using `@override` here so we can have more explicit type relations, like we do in [to_homogeneous](https://github.com/Snapchat/GiGL/blob/3a126845d209dd499d8edb84bc11381e34f040df/python/gigl/types/graph.py#L442-L462)

This would mainly help with downstream caller typing/narrowing but I think ours still needs to be the way it is so this isn't a big deal.

python/gigl/distributed/dist_dataset.py
```diff
@@ -386,319 +389,67 @@ def load(self, *args, **kwargs):
...
-    def build(
+    def _split_and_initialize_node_ids(
         self,
-        partition_output: PartitionOutput,
-        splitter: Optional[Union[NodeSplitter, NodeAnchorLinkSplitter]] = None,
+        node_partition_book: Union[PartitionBook, dict[NodeType, PartitionBook]],
+        splitter: Optional[Union[NodeSplitter, NodeAnchorLinkSplitter]],
+        partitioned_edge_index: Optional[
+            Union[GraphPartitionData, dict[EdgeType, GraphPartitionData]]
+        ],
```
kmontemayor2-sc, line 398: btw when would this be `None`? We always pass it in anyways, right?

kmontemayor2-sc: though maybe it's a little smelly to also pass in the `edge_index`, do you think we can avoid doing that and have two functions `split_node_ids` and `initialize_node_ids`?

python/gigl/distributed/dist_dataset.py
mkolodner-sc: This is a fair point, for node features and node labels we made the expectation be that we could never see an empty dictionary here, but for edge features our existing code base returns an empty dictionary for some reason instead of None. I've made a separate PR https://github.com/Snapchat/GiGL/pull/334 which addresses this, maybe we can aim to land that one first and then land this simplification work after that is in so that we can remove the `len(partitioned_data) > 0` requirement here.

python/gigl/distributed/dist_dataset.py
mkolodner-sc: Hmm, I see your point. I initially typed that as `Optional` since the `partitioned_edge_index` is not needed for node based splitting, although you are correct that it will always be passed in as a non-None value. 

> though maybe it's a little smelly to also pass in the edge_index, do you think we can avoid doing that and have two functions split_node_ids and initialize_node_ids?

Hmm, this wouldn't fix the issue though since `split_node_ids` would still need to take in an edge index field which would have the same smell. I've drafted up a solution where we just move the `split` generation generally back to `build` so we don't have to parameterize this function with this smell and decreased the remaining scope to `initialize_node_ids`.

python/gigl/distributed/dist_dataset.py
```diff
@@ -783,6 +495,303 @@ def build(
...
+
+        del node_ids_on_machine, splits
+        gc.collect()
+
+        # Initialize Graph and get edge data for splitting
+        self._initialize_graph(
+            partitioned_edge_index=partition_output.partitioned_edge_index
+        )
+        partition_output.partitioned_edge_index = None
+        gc.collect()
```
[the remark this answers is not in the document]
mkolodner-sc: I have a preference for leaving
```
partition_output.partitioned_edge_index = None
gc.collect()
```
outside of the function call so that we don't have to pass in the whole `partition_output` to the `_initialize_graph` function and also aren't secretly modifying the `partition_output` inside one of these functions, which would lead to more confusion and harder maintainability generally.

python/gigl/distributed/dist_dataset.py
```diff
@@ -783,6 +495,303 @@ def build(
...
+        self._edge_partition_book = partition_output.edge_partition_book
+
+        self._positive_edge_label = partition_output.partitioned_positive_labels
+        self._negative_edge_label = partition_output.partitioned_negative_labels
+
+        partition_output.node_partition_book = None
+        partition_output.edge_partition_book = None
+
+        partition_output.partitioned_positive_labels = None
+        partition_output.partitioned_negative_labels = None
```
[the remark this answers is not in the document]
mkolodner-sc: It'd be better to have the whole partition_output cleared for consistency than have only a subset of it cleared, resulting in an incomplete `partition_output` object.

PR #318, opened 2025-09-09
Reduce requirements for task metadata relative to graph metadata pb wrapper

mkolodner-sc:
**Scope of work done**

Reduces the restrictions of our task metadata to only enforce that the source and destination node types of the supervision edge types are in the graph metadata, rather than the edge types as well. This can unblock use cases where we may have `A -> B`, `B -> B` as graph metadata edge types but `B -> A` as our supervision edge type.

### DIFF
--- a/python/gigl/src/validation_check/libs/template_config_checks.py
+++ b/python/gigl/src/validation_check/libs/template_config_checks.py
@@ -109,17 +109,21 @@ def check_if_task_metadata_valid(
         assert (
             len(task_metadata_pb.supervision_edge_types) > 0
         ), "Must provide at least one supervision edge type."
-        graph_metadata_pb_edge_types = [
-            GbmlProtosTranslator.edge_type_from_EdgeTypePb(edge_type_pb=edge_type_pb)
-            for edge_type_pb in graph_metadata_pb.edge_types
+        graph_metadata_node_types = [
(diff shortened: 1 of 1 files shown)


python/tests/unit/src/validation/task_metadata_is_valid_test.py
kmontemayor2-sc, line 1: nit. this file should be `template_config_checks_test.py`?

python/tests/unit/src/validation/task_metadata_is_valid_test.py
```diff
@@ -0,0 +1,71 @@
+import unittest
+
+from gigl.src.validation_check.libs.template_config_checks import (
+    check_if_task_metadata_valid,
+)
+from snapchat.research.gbml import gbml_config_pb2, graph_schema_pb2
+from tests.test_assets.graph_metadata_constants import (
+    DEFAULT_HOMOGENEOUS_GRAPH_METADATA_PB,
+    DEFAULT_HOMOGENEOUS_NODE_TYPE_STR,
```
kmontemayor2-sc, line 9: unrelated, this shouldn't be done in this pr.

We should not have mutable constants like this [1]. If we do this, then one test *could* mutate them, and have other downstream tests fail.

Ideally, we have some `create_graph_metadata_pb()` that always creates a new object.

Can we add this to some task tracker?

PR #326, opened 2025-09-15
shutdown example inference loaders

kmontemayor2-sc:
Doing this is the preferred way as calling del on the loaders does not always clean up all memory, leading to `part or all of the requested memory range is already mapped` errors.


kmontemayor2-sc (commented):
/e2e_test

PR #327, opened 2025-09-16
Revert "Use `typing.Dict` for Beam type annotations. (#275)"

kmontemayor2-sc:
Turns out, this wasn't the problem, whoops!

Let's revert for consistency.


mkolodner-sc (approved):
Thanks Kyle!

xgao4-sc (approved):

PR #328, opened 2025-09-17
Shard nodes before selecting labels in ABLPLoader

kmontemayor2-sc:
Doing this is a bit more performant - we only care about he nodes we shard for.


yliu2-sc (approved):

mkolodner-sc (approved):
Thanks!

zfan3-sc (approved):

PR #336, opened 2025-09-23
Support multiple supervision edge types in ABLP Loader

kmontemayor2-sc:
**Scope of work done**

Added support for multiple supervision edge types.

There are two primary user-facing changes here:

1.  [Enable](https://github.com/Snapchat/GiGL/blob/ffe3d601836eb6b6f2af0c097628849d85635d92/python/gigl/distributed/dist_ablp_neighborloader.py#L61) multiple supervision edges types to be passed into the loader 
2. When more than one supervision edge types are provided, make y_positive/y_negative a `dict[EdgeType, dict[int, Tensor]]` [ex](https://github.com/Snapchat/GiGL/blob/ffe3d601836eb6b6f2af0c097628849d85635d92/python/tests/unit/distributed/dist_ablp_neighborloader_test.py#L790-L798). 

And then a lot of piping to go from 1 -> 2...

Where is the documentation for this feature?: 

In [docstring](https://github.com/Snapchat/GiGL/blob/ffe3d601836eb6b6f2af0c097628849d85635d92/python/gigl/distributed/dist_ablp_neighborloader.py#L117-L121) for ABLP loader

Added unit tests

***Updated Changelog.md?*** Yes

***Ready for code review?:*** Yes

### DIFF
--- a/python/gigl/common/beam/sharded_read.py
+++ b/python/gigl/common/beam/sharded_read.py
@@ -1,0 +1,98 @@
+from dataclasses import dataclass
+
+import apache_beam as beam
+from apache_beam.io.gcp.bigquery import BigQueryQueryPriority
+from apache_beam.io.gcp.internal.clients.bigquery import DatasetReference
+from apache_beam.pvalue import PBegin
+from google.cloud import bigquery
+
+from gigl.common.logger import Logger
+
+logger = Logger()
+
+
+@dataclass(frozen=True)
+class BigQueryShardedReadConfig:
+    # The key in the table that we will use to split the data into shards. This should be used if we are operating on
+    # very large tables, in which case we want to only read smaller slices of the table at a time to avoid oversized status update
+    # payloads.
+    shard_key: str
+    # The project ID to use for temporary datasets when running sharded reads.
+    project_id: str
+    # The temporary bigquery dataset name to use when running sharded reads.
+    temp_dataset_name: str
+    # The number of shards to split the data into. If not provided, the table will be shareded with a default
+    # value of 20 shards.
+    # TODO (mkolodner-sc): Instead of using this default, infer this value based on number of rows in table
+    num_shards: int = 20
+
+
+def _assert_shard_key_in_table(table_name: str, shard_key: str) -> None:
+    """
+    Validate that the shard key is a valid column in the BigQuery table.
+    """
+    client = bigquery.Client()
+    table_ref = bigquery.TableReference.from_string(table_name)
+
+    table = client.get_table(table_ref)
+    column_names = [field.name for field in table.schema]
+
+    if shard_key not in column_names:
+        raise ValueError(
+            f"Shard key '{shard_key}' is not a valid column in table '{table_name}'. "
+            f"Available columns: {column_names}"
+        )
+
+
+class ShardedExportRead(beam.PTransform):
+    def __init__(
+        self,
+        table_name: str,
+        sharded_read_info: BigQueryShardedReadConfig,
+        **kwargs,
+    ):
+        super().__init__()
+        self._table_name: str = table_name
+        self._num_shards: int = sharded_read_info.num_shards
+        if self._num_shards <= 0:
+            raise ValueError(
+                f"Number of shards specified must be greater than 0, got {self._num_shards}"
+            )
+        self._shard_key: str = sharded_read_info.shard_key
+        self._temp_dataset_reference: DatasetReference = DatasetReference(
+            projectId=sharded_read_info.project_id,
+            datasetId=sharded_read_info.temp_dataset_name,
+        )
+        self._kwargs = kwargs
+        logger.info(
+            f"Got ShardedExportRead arguments table_name={table_name}, sharded_read_info={sharded_read_info}, kwargs={kwargs}"
+        )
+
+        _assert_shard_key_in_table(self._table_name, self._shard_key)
+
+    def expand(self, pbegin: PBegin):
+        pcollection_list = []
+        for i in range(self._num_shards):
+            # We use farm_fingerprint as a determinstic hashing function which will allow us to shard
+            # on keys of any type (i.e. strings, integers, etc.) We take the MOD on the returned INT64 value first
+            # with the number of shards and then take the ABS value to ensure it is in range [0, num_shards-1].
+            # We do it in this order since ABS can error on the largest negative INT64 number, which has no positive equivalent.
+            query = (
+                f"SELECT * FROM `{self._table_name}` "
+                f"WHERE ABS(MOD(FARM_FINGERPRINT(CAST({self._shard_key} AS STRING)), {self._num_shards})) = {i}"
+            )
+
+            pcollection_list.append(
+                pbegin
+                | f"Read slice {i}/{self._num_shards}"
+                >> beam.io.ReadFromBigQuery(
+                    query=query,
+                    use_standard_sql=True,
+                    method=beam.io.ReadFromBigQuery.Method.EXPORT,
+                    query_priority=BigQueryQueryPriority.INTERACTIVE,
+                    temp_dataset=self._temp_dataset_reference,
+                    **self._kwargs,
+                )
+            )
+
+        return pcollection_list | "Flatten shards" >> beam.Flatten()
--- a/python/gigl/src/common/types/pb_wrappers/gbml_config.py
+++ b/python/gigl/src/common/types/pb_wrappers/gbml_config.py
@@ -469,6 +469,9 @@ class GbmlConfigPbWrapper:
         """
         Allows access to should_populate_predictions_path under GbmlConfig
 
+        This flag defaults to False, but can be set to True to populate the predictions path in the InferenceOutput for each entity type.
+        We default this to False since config populator currently does not populate the predictions path for link prediction tasks.
+
         This flag is a temporary workaround to populate the extra embeddings for the same entity type
 
         Returns:
@@ -478,6 +481,26 @@ class GbmlConfigPbWrapper:
             strtobool(
                 dict(self.gbml_config_pb.feature_flags).get(
                     "should_populate_predictions_path", "False"
+                )
+            )
+        )
+
+    @property
+    def should_populate_embeddings_path(self) -> bool:
+        """
+        Allows access to should_populate_embeddings_path under GbmlConfig
+
+        This flag defaults to True, but can be set to False to skip populating embedding paths in InferenceOutput.
+        We default this to True since config populator currently always by default populates the embeddings path
+        for both link prediction and node classification tasks.
+
+        Returns:
+            bool: Whether to populate embeddings path in the InferenceOutput for each entity type
+        """
+        return bool(
+            strtobool(
+                dict(self.gbml_config_pb.feature_flags).get(
+                    "should_populate_embeddings_path", "True"
                 )
             )
         )
--- a/python/gigl/src/config_populator/config_populator.py
+++ b/python/gigl/src/config_populator/config_populator.py
@@ -336,18 +336,21 @@ class ConfigPopulator:
             gbml_config_pb=self.template_gbml_config
         )
         for node_type in inferencer_node_types:
-            embeddings_path = bq_constants.get_embeddings_table(
-                applied_task_identifier=self.applied_task_identifier,
-                node_type=node_type,
-            )
+            embeddings_path: Optional[str] = None
             predictions_path: Optional[str] = None
+
+            if template_gbml_config_pb_wrapper.should_populate_embeddings_path:
+                embeddings_path = bq_constants.get_embeddings_table(
+                    applied_task_identifier=self.applied_task_identifier,
+                    node_type=node_type,
+                )
 
             if (
                 self.task_metadata_pb_wrapper.task_metadata_type
                 == TaskMetadataType.NODE_BASED_TASK
                 or template_gbml_config_pb_wrapper.should_populate_predictions_path
             ):
-                # TODO: currently, we are overloading the predictions path to store extra embeddings.
+                # TODO: currently, we are overloading the predictions path to store extra embeddings for link prediction.
                 # consider extending InferenceOutput's definition for this purpose.
                 predictions_path = bq_constants.get_predictions_table(
                     applied_task_identifier=self.applied_task_identifier,
--- a/python/gigl/src/data_preprocessor/data_preprocessor.py
+++ b/python/gigl/src/data_preprocessor/data_preprocessor.py
@@ -791,6 +791,7 @@ class DataPreprocessor:
                 reference_uri=node_data_reference.reference_uri,
                 node_type=node_data_reference.node_type,
                 identifier=node_data_preprocessing_spec.identifier_output,
+                sharded_read_config=node_data_reference.sharded_read_config,
             )
             node_refs_to_specs[
                 node_data_ref_with_identifier
@@ -810,6 +811,7 @@ class DataPreprocessor:
                 edge_usage_type=edge_data_reference.edge_usage_type,
                 src_identifier=edge_data_preprocessing_spec.identifier_output.src_node,
                 dst_identifier=edge_data_preprocessing_spec.identifier_output.dst_node,
+                sharded_read_config=edge_data_reference.sharded_read_config,
             )
             edge_refs_to_specs[
                 edge_data_ref_with_identifier
--- a/python/gigl/src/data_preprocessor/lib/enumerate/utils.py
+++ b/python/gigl/src/data_preprocessor/lib/enumerate/utils.py
@@ -235,6 +235,7 @@ class Enumerator:
                 reference_uri=bq_destination_enumerated_node_features_table_name,
                 node_type=node_data_ref.node_type,
                 identifier=node_data_ref.identifier,
+                sharded_read_config=node_data_ref.sharded_read_config,
             ),
             bq_unique_node_ids_enumerated_table_name=bq_unique_node_ids_enumerated_table_name,
             num_nodes=num_enumerated_nodes,
@@ -431,6 +432,7 @@ class Enumerator:
                 edge_usage_type=edge_data_ref.edge_usage_type,
                 src_identifier=edge_data_ref.src_identifier,
                 dst_identifier=edge_data_ref.dst_identifier,
+                sharded_read_config=edge_data_ref.sharded_read_config,
             ),
             num_edges=num_enumerated_edges,
         )
--- a/python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
+++ b/python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
@@ -1,43 +1,106 @@
-from typing import cast
+from dataclasses import dataclass
+from typing import Optional, cast
 
 import apache_beam as beam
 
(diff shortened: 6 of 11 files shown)


python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -170,6 +184,14 @@ def __init__(
         master_ip_address: str
         should_cleanup_distributed_context: bool = False
 
+        if supervision_edge_type is None:
+            self._supervision_edge_types: list[EdgeType] = []
+        elif isinstance(supervision_edge_type, list):
+            self._supervision_edge_types = supervision_edge_type
+        else:
+            self._supervision_edge_types = [supervision_edge_type]
+        del supervision_edge_type
```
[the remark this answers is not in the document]
kmontemayor2-sc: No reason, just trying to prevent mis-using this later (instead of self._supervision_edge_types which we are using)

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -117,13 +127,17 @@ def __init__(
...
-            local_process_rank (int): The local rank of the current process within a node.
-            local_process_world_size (int): The total number of processes within a node.
             input_nodes (Optional[torch.Tensor, tuple[NodeType, torch.Tensor]]):
                 Indices of seed nodes to start sampling from.
                 If set to `None` for homogeneous settings, all nodes will be considered.
                 In heterogeneous graphs, this flag must be passed in as a tuple that holds
                 the node type and node indices. (default: `None`)
+            supervision_edge_type (Optional[Union[EdgeType, list[EdgeType]]]):
+                The edge type(s) to use for supervision.
+                Must be None iff the dataset is labeled homogeneous.
```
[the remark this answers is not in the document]
kmontemayor2-sc: Good call! throwing now if an empty list is provided.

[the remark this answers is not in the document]
kmontemayor2-sc: also added some tests for empty inputs :P

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -497,82 +540,109 @@ def _get_labels(
             msg (SampleMessage): All possible results from a sampler, including subgraph data, features, and used defined metadata
         Returns:
             SampleMessage: Updated sample messsage with the label fields removed
-            torch.Tensor: Positive label ID tensor, where the ith row corresponds to the ith anchor node ID
-            Optional[torch.Tensor]: Negative label ID tensor, where the ith row corresponds to the ith anchor node ID, can be None if dataset has no negative labels
+            dict[EdgeType, torch.Tensor]: Positive label ID tensor, where the ith row corresponds to the ith anchor node ID
+            Optional[dict[EdgeType, torch.Tensor]]: Negative label ID tensor, where the ith row corresponds to the ith anchor node ID.
```
mkolodner-sc, line 544: Can we update these comments to reflect the heterogeneous nature here? Ditto elsewhere

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -497,82 +540,109 @@ def _get_labels(
...
         """
-        metadata = {}
+        metadata: dict[str, torch.Tensor] = {}
+        positive_labels: dict[EdgeType, torch.Tensor] = {}
+        negative_labels: dict[EdgeType, torch.Tensor] = {}
         for k in list(msg.keys()):
-            if k.startswith("#META."):
-                meta_key = str(k[6:])
+            if k.startswith(f"#META.{POSITIVE_LABEL_METADATA_KEY}."):
```
mkolodner-sc, line 550: Adding some comments here would be helpful to better understand the logic here

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -497,82 +540,109 @@ def _get_labels(
...
+
+                # Gets the indexes of the items in local_node_to_global_node which match any of the positive labels for the current anchor node
+                output_positive_labels[
+                    label_edge_type_to_message_passing_edge_type(edge_type)
+                ][local_anchor_node_id] = torch.nonzero(positive_mask)[:, 0].to(
+                    self.to_device
+                )
+                # Shape [X], where X is the number of indexes in the original local_node_to_global_node which match a node in the positive labels for the current anchor node
+
         if negative_labels is not None:
```
mkolodner-sc, line 620: Negative labels cannot be none anymore?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -497,82 +540,109 @@ def _get_labels(
...
+                    )  # shape [N, M], where N is the number of nodes and M is the number of negative labels for the current anchor node
+
+                    # Gets the indexes of the items in local_node_to_global_node which match any of the negative labels for the current anchor node
+                    output_negative_labels[
+                        label_edge_type_to_message_passing_edge_type(edge_type)
+                    ][local_anchor_node_id] = torch.nonzero(negative_mask)[:, 0].to(
+                        self.to_device
+                    )
+                    # Shape [X], where X is the number of indexes in the original local_node_to_global_node which match a node in the negative labels for the current anchor node
+        if len(output_positive_labels) == 1:
```
mkolodner-sc, line 637: Could positive labels ever be len(0) here? Should we raise a ValueError if this happens?

python/gigl/distributed/dist_neighbor_sampler.py
```diff
@@ -35,48 +39,42 @@ async def _sample_from_nodes(
...
+        metadata: dict[str, torch.Tensor] = {}
+        input_seeds_builder: dict[
+            Union[str, NodeType], list[torch.Tensor]
+        ] = defaultdict(list)
+        input_seeds_builder[input_type].append(input_seeds)
+        for edge_type, label_tensor in inputs.positive_label_by_edge_types.items():
+            input_seeds_builder[edge_type[label_edge_index]].append(
+                label_tensor[label_tensor != PADDING_NODE].to(self.device)
+            )
+            metadata[
```
mkolodner-sc, line 59: Rather than having to do all of this string parsing here so that each edge type is a different field in the metadata, could we just store the edge type information in the value itself (i.e. store the value as a dict[edge_type, label_tensor])?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -291,22 +319,41 @@ def __init__(
...
-        positive_labels, negative_labels = get_labels_for_anchor_nodes(
-            dataset=dataset,
-            node_ids=curr_process_nodes,
-            positive_label_edge_type=self._positive_label_edge_type,
-            negative_label_edge_type=self._negative_label_edge_type,
+
+        self._positive_label_edge_types: list[EdgeType] = []
+        self._negative_label_edge_types: list[EdgeType] = []
+        all_positive_labels: dict[EdgeType, torch.Tensor] = {}
+        all_negative_labels: dict[EdgeType, torch.Tensor] = {}
```
mkolodner-sc, line 332: Can you update these names to be more reflective of their dictionary typing?

python/gigl/distributed/sampler.py
```diff
@@ -16,24 +19,28 @@ def __init__(
...
+        self._positive_label_by_edge_types = positive_label_by_edge_types
+        self._negative_label_by_edge_types = negative_label_by_edge_types
+
+    @property
+    def positive_label_by_edge_types(self) -> dict[EdgeType, torch.Tensor]:
+        return self._positive_label_by_edge_types
+
+    @property
+    def negative_label_by_edge_types(self) -> dict[EdgeType, torch.Tensor]:
+        return self._negative_label_by_edge_types
```
mkolodner-sc, line 43: qq: where are these used?

python/gigl/distributed/sampler.py
```diff
@@ -45,9 +52,15 @@ def __getitem__(self, index: Union[torch.Tensor, Any]) -> "ABLPNodeSamplerInput"
         return ABLPNodeSamplerInput(
             node=self.node[index],
             input_type=self.input_type,
-            positive_labels=self.positive_labels[index],
-            negative_labels=self.negative_labels[index]
-            if self.negative_labels is not None
-            else None,
-            supervision_node_type=self.supervision_node_type,
+            positive_label_by_edge_types={
+                edge_type: self._positive_label_by_edge_types[edge_type][index]
```
mkolodner-sc, line 56: What is the behavior here if a given labeled edge type does not have that index?

mkolodner-sc (commented):
Thanks Kyle! Did a pass here -- will look through testing files shortly

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -296,6 +313,113 @@ def _run_toy_heterogeneous_ablp(
...
+            dim=0,
+        )
+    assert hasattr(datum, "y_positive")
+    assert set(datum.y_positive.keys()) == set(
+        expected_positive_labels.keys()
+    ), f"{datum.y_positive.keys()} != {expected_positive_labels.keys()}"
+    for edge_type in datum.y_positive.keys():
+        _assert_labels(
+            datum[edge_type[0]].node,
+            datum[edge_type[2]].node,
```
yliu2-sc, line 361: maybe assign 0/2 to in/out edge index variables so it's not magic

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -574,6 +698,260 @@ def test_toy_heterogeneous_ablp(
...
+                    graph_partition={},
+                    node_partition_book={},
+                ),
+                num_neighbors=[2, 2],
+                input_nodes=(NodeType("a"), torch.tensor([10])),
+                supervision_edge_type=[_A_TO_B, _A_TO_C],
+            ),
+        ]
+    )
+    def test_ablp_dataloder_invalid_inputs(
```
yliu2-sc, line 942:
```suggestion
    def test_ablp_dataloader_invalid_inputs(
```

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -574,6 +698,260 @@ def test_toy_heterogeneous_ablp(
...
+                    edge_dir="out",
+                    graph_partition={},
+                    node_partition_book={},
+                ),
+                num_neighbors=[2, 2],
+                input_nodes=(NodeType("a"), torch.tensor([10])),
+                supervision_edge_type=None,
+            ),
+            param(
+                "Mutiple supervision edge types, homogeneous sampling",
```
yliu2-sc, line 926: heterogeneous sampling?

yliu2-sc (approved):
In general LGTM

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -170,6 +184,18 @@ def __init__(
         master_ip_address: str
         should_cleanup_distributed_context: bool = False
 
+        if supervision_edge_type is None:
+            self._supervision_edge_types: list[EdgeType] = []
+        elif isinstance(supervision_edge_type, list):
+            if not supervision_edge_type:
+                raise ValueError(
```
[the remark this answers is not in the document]
kmontemayor2-sc: changed this up a bit, wdyt now?

python/gigl/distributed/dist_neighbor_sampler.py
```diff
@@ -35,48 +40,56 @@ async def _sample_from_nodes(
...
+            input_seeds_builder[edge_type[label_edge_index]].append(
+                filtered_label_tensor
+            )
+            # Update the metadata per negative label edge type.
+            # We do this because GLT only supports dict[str, torch.Tensor] for metadata.
+            metadata[
+                f"{NEGATIVE_LABEL_METADATA_KEY}.{str(tuple(edge_type))}"
+            ] = label_tensor
+        input_nodes: dict[Union[str, NodeType], torch.Tensor] = {
+            node_type: torch.cat(seeds, dim=0).to(self.device)
```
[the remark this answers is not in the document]
kmontemayor2-sc: Hmmm, think doing so is a perf optimization right? WDYT if I leave a TODO?

The way I see it is: we should only do this if we have multiple supervision edge types with *different* relations but the same src/dst nodes (or for "homogeneous" graphs), otherwise, we're doing the sort (which is slow) for no reason?

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -75,7 +89,8 @@ def _assert_labels(
     _assert_labels(node, y, expected)
 
     Args:
-        node (torch.Tensor): The node tensor, N, where N is the number of nodes in the batch,
+        src_node (torch.Tensor): The source (anchor) node tensor, [N], where N is the number of source nodes in the batch,
```
mkolodner-sc, line 92: These names are misleading -- this is not the anchor nodes tensor for the batch, but the nodes in the batch with the same type as the anchor node type. Could we rename and update the description here? Ditto for the supervision nodes.

nit: we should probably not use `N` variable for both of these parameters here, since their values may be different.

mkolodner-sc: Also, could we avoid using "src" and "dst" terms here? 

The current naming here is enforcing us to stick with the assumption that supervision edges are always directed outward, which we might want to change in the future.

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -132,9 +147,11 @@ def _run_distributed_ablp_neighbor_loader(
         expected_node,
         dim=0,
     )
-    _assert_labels(datum.node, datum.y_positive, expected_positive_labels)
+    _assert_labels(datum.node, datum.node, datum.y_positive, expected_positive_labels)
     if expected_negative_labels is not None:
-        _assert_labels(datum.node, datum.y_negative, expected_negative_labels)
+        _assert_labels(
+            datum.node, datum.node, datum.y_negative, expected_negative_labels
```
mkolodner-sc, line 153: a comment here on why we pass in `datum.node` twice could be helpful

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -296,6 +313,115 @@ def _run_toy_heterogeneous_ablp(
...
+            datum[node_type].node,
+            expected_node[node_type],
+            dim=0,
+        )
+    assert hasattr(datum, "y_positive")
+    assert set(datum.y_positive.keys()) == set(
+        expected_positive_labels.keys()
+    ), f"{datum.y_positive.keys()} != {expected_positive_labels.keys()}"
+    src_index = 0
+    dst_index = 2
```
mkolodner-sc, line 359: Same comment about avoiding use of "src" and "dst" terms here in favor of "anchor" and "supervision".

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -574,6 +700,260 @@ def test_toy_heterogeneous_ablp(
...
+                )
+                for etype, idx in edge_index.items()
+            },
+            partitioned_edge_features=None,
+            partitioned_node_features=None,
+            partitioned_negative_labels=None,
+            partitioned_positive_labels=None,
+            partitioned_node_labels=None,
+        )
+        dataset = DistDataset(rank=0, world_size=1, edge_dir="out")
```
mkolodner-sc, line 871: Should we test "in" as well here?

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -497,82 +544,121 @@ def _get_labels(
...
-        )
         return (msg, positive_labels, negative_labels)
 
     def _set_labels(
         self,
         data: Union[Data, HeteroData],
-        positive_labels: torch.Tensor,
-        negative_labels: Optional[torch.Tensor],
+        positive_labels: dict[EdgeType, torch.Tensor],
+        negative_labels: dict[EdgeType, torch.Tensor],
```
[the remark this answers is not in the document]
kmontemayor2-sc: done.

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
kmontemayor2-sc: done. found some bugs, thanks for asking for this!

python/tests/unit/distributed/dist_ablp_neighborloader_test.py
```diff
@@ -75,9 +91,8 @@ def _assert_labels(
     _assert_labels(node, y, expected)
 
     Args:
-        node (torch.Tensor): The node tensor, N, where N is the number of nodes in the batch,
-        where the local IDs are the index of a node in the tensor,
-        and the global IDs are the values of the tensor.
+        anchor_nodes (torch.Tensor): The source (anchor) node tensor, [N], where N is the number of supervision nodes in the batch,
+        supervision_nodes (torch.Tensor): The destination (supervision) node tensor, [M], where M is the number of supervision nodes in the batch,
```
mkolodner-sc, line 95: suggested rewording:
```
anchor_nodes (torch.Tensor): Tensor of nodes in the graph with the same type as the anchor node, shape [N] where N is the number of nodes in the batch with this node type
supervision_nodes (torch.Tensor): Tensor of nodes in the graph with the same type as the supervision node, shape [M] where M is the number of nodes in the batch with this node type

File: python/gigl/distributed/dist_ablp_neighborloader.py [hunk 22]
```diff
@@ -497,82 +545,127 @@ def _get_labels(
...
+            dict
+        )
+        output_negative_labels: dict[EdgeType, dict[int, torch.Tensor]] = defaultdict(
+            dict
+        )
+        # Since GLT swaps src/dst for edge_dir = "out",
+        # and GiGL assumes that supervision edge types are always (anchor_node_type, to, supervision_node_type),
+        # we need to index into supervision edge types accordingly.
+        # edge_index = 0 if self.edge_dir == "in" else 2
+        edge_index = 2
```
mkolodner-sc, line 624: Should we remove outdated comment and commented code here? Also add explanation for why this is always 2 for now would be helpful

mkolodner-sc (approved):
Thanks a lot Kyle! LGTM provided tests pass

python/gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -497,82 +546,139 @@ def _get_labels(
...
+                DEFAULT_HOMOGENEOUS_NODE_TYPE
+            ] = data.node
+        output_positive_labels: dict[EdgeType, dict[int, torch.Tensor]] = defaultdict(
+            dict
+        )
+        output_negative_labels: dict[EdgeType, dict[int, torch.Tensor]] = defaultdict(
+            dict
+        )
+        # We always have supervision edge types of the form (anchor_node_type, to, supervision_node_type)
+        # So we can index into the edge type accordingly.
```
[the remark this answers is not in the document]
mkolodner-sc: Agreed -- I was thinking that we could remove the edge flipping altogether. I had a POC for the work needed here, but we should probably first align at some point what the user should be providing across the the different strategies (SSL, UDL) and edge directions in this new paradigm.

PR #339, opened 2025-09-26
Allow users to specify custom master nodes for networking utils

kmontemayor2-sc:
**Scope of work done**

 We do this so in server/client mode we can have a "training" pg for both server and client, but have each server and client communicate with their respective leaders.
 
 e.g for us to determine the ip/port on "training cluster leader" (global rank = 2).

image

### DIFF
--- a/python/gigl/distributed/utils/networking.py
+++ b/python/gigl/distributed/utils/networking.py
@@ -44,12 +44,15 @@ def get_free_ports(num_ports: int) -> list[int]:
 
 
 def get_free_ports_from_master_node(
-    num_ports=1, _global_rank_override: Optional[int] = None
+    num_ports=1,
+    master_node_rank: int = 0,
+    _global_rank_override: Optional[int] = None,
 ) -> list[int]:
     """
     Get free ports from master node, that can be used for communication between workers.
     Args:
         num_ports (int): Number of free ports to find.
+        master_node_rank (int): Rank of the master node, default is 0.
         _global_rank_override (Optional[int]): Override for the global rank,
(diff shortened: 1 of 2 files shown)


python/gigl/distributed/utils/networking.py
```diff
@@ -44,21 +44,39 @@ def get_free_ports(num_ports: int) -> list[int]:
...
+        useful for testing or if global rank is not accurately available.
+    """
+    return get_free_ports_from_node(
+        num_ports, node_rank=0, _global_rank_override=_global_rank_override
+    )
+
+
+def get_free_ports_from_node(
+    num_ports=1,
+    node_rank: int = 0,
```
svij-sc, line 63: maybe we dont default `node_rank` for this function as the name implied we need to get a specifc node and being explicit is better here.

python/gigl/distributed/utils/networking.py
```diff
@@ -87,12 +104,24 @@ def get_internal_ip_from_master_node(
...
     """
     Get the internal IP address of the master node in a distributed setup.
+    """
+    return get_internal_ip_from_node(
+        node_rank=0, _global_rank_override=_global_rank_override
+    )
+
+
+def get_internal_ip_from_node(
+    node_rank: int = 0,
```
svij-sc, line 114: same comment re: generic function is not expecting node_rank to be explicitly passed in.
We already have a convenience function `get_internal_ip_from_master_node` that passes in `0`

mkolodner-sc (approved):

PR #341, opened 2025-09-27
[KGE] Add embedding export / unenumeration utils

nshah-sc:
Add some utils which are used to write embeddings to BQ and unenumerate the resulting tables for KGE.


python/gigl/experimental/knowledge_graph_embedding/lib/infer.py
```diff
@@ -0,0 +1,450 @@
...
+    )
+
+    # Initialize the distributed training pipeline for inference
+    pipeline = TrainPipelineSparseDist(
+        model=model_and_loss, optimizer=optimizer, device=device
+    )
+    logger.info(f"{rank_prefix_str} Initialized TrainPipelineSparseDist for inference.")
+
+    # Run inference in no_grad context to save memory and improve performance
+    with torch.no_grad():
```
svij-sc, line 233: suggestion / non-blocking: there is also `with torch.inference_mode():` which is faster than no_grad. It wasnt working w/ GNN inference due to some reason (@mkolodner-sc) - but maybe it works here.

python/gigl/experimental/knowledge_graph_embedding/lib/infer.py
```diff
@@ -0,0 +1,450 @@
...
+
+        # Process each edge type in the graph metadata
+        for (
+            condensed_edge_type,
+            edge_type,
+        ) in sorted(graph_metadata.condensed_edge_type_to_edge_type_map.items()):
+            logger.info(
+                f"""{rank_prefix_str} Running inference for edge type {edge_type} on
+                src node type {edge_type.src_node_type} and dst node type {edge_type.dst_node_type}."""
+            )
```
svij-sc, line 245: use textwrap.dedent

python/gigl/experimental/knowledge_graph_embedding/lib/infer.py
```diff
@@ -0,0 +1,450 @@
...
+        FROM
+            `{enumerated_assets_table}` enumerated_assets
+        INNER JOIN
+            `{mapping_table}` mapping
+        ON
+            mapping.int_id = enumerated_assets.{node_id_field}
+        QUALIFY RANK() OVER (PARTITION BY mapping.{original_node_id_field} ORDER BY RAND()) = 1
+    """
+
+    bq_utils = BqUtils(project=get_resource_config().project)
```
svij-sc, line 438: We have been trying to prevent coupling of `get_resource_config()` in low level / leaf apis/function calls.
It is better to dereference resource config as much as we can higher in the stack and pass those values through. Not doing this leads to usability and extensibility challanges.

PR #344, opened 2025-09-29
Partition DP

mkolodner-sc:
**Scope of work done**

### DIFF
--- a/python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
+++ b/python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
@@ -1,7 +1,11 @@
-from typing import cast
+from typing import Optional, cast
 
 import apache_beam as beam
+from apache_beam.io.gcp.bigquery import BigQueryQueryPriority
+from apache_beam.io.gcp.internal.clients.bigquery import DatasetReference
+from apache_beam.pvalue import PBegin
 
+from gigl.env.pipelines_config import get_resource_config
 from gigl.src.data_preprocessor.lib.ingest.reference import (
     EdgeDataReference,
     NodeDataReference,
@@ -9,19 +13,65 @@ from gigl.src.data_preprocessor.lib.ingest.reference import (
 from gigl.src.data_preprocessor.lib.types import InstanceDictPTransform
 
 
+class _SlicedExportRead(beam.PTransform):
+    def __init__(self, table_name: str, partition_key: str, **kwargs):
+        super().__init__()
+        self._table_name = table_name
+        # TODO (mkolodner-sc): Infer this value based on number of rows
+        self._num_slices = 20
+        self._partition_key = partition_key
+        resource_config = get_resource_config()
+        self._temp_dataset_reference = DatasetReference(
+            projectId=resource_config.project,
+            datasetId=resource_config.temp_assets_bq_dataset_name,
+        )
+        self._kwargs = kwargs
+
+    def expand(self, pbegin: PBegin):
+        pcollection_list = []
+        for i in range(self._num_slices):
+            query = (
+                f"SELECT * FROM `{self._table_name}` "
+                f"WHERE MOD(ABS(FARM_FINGERPRINT(CAST({self._partition_key} AS STRING))), {self._num_slices}) = {i}"
+            )
+
+            pcollection_list.append(
+                pbegin
+                | f"Read slice {i}/{self._num_slices}"
+                >> beam.io.ReadFromBigQuery(
+                    query=query,
+                    use_standard_sql=True,
+                    method=beam.io.ReadFromBigQuery.Method.EXPORT,
+                    query_priority=BigQueryQueryPriority.INTERACTIVE,
+                    temp_dataset=self._temp_dataset_reference,
+                    **self._kwargs,
+                )
+            )
+
+        return pcollection_list | "Flatten slices" >> beam.Flatten()
+
+
 def _get_bigquery_ptransform(
-    table_name: str, *args, **kwargs
+    table_name: str, partition_key: Optional[str], *args, **kwargs
 ) -> InstanceDictPTransform:
     table_name = table_name.replace(".", ":", 1)  # sanitize table name
-    return cast(
-        InstanceDictPTransform,
-        beam.io.ReadFromBigQuery(
-            table=table_name,
-            method=beam.io.ReadFromBigQuery.Method.EXPORT,  # type: ignore
-            *args,
-            **kwargs,
-        ),
-    )
+    if partition_key is not None:
+        return cast(
+            InstanceDictPTransform,
+            _SlicedExportRead(
+                table_name=table_name, partition_key=partition_key, **kwargs
+            ),
+        )
+    else:
+        return cast(
+            InstanceDictPTransform,
+            beam.io.ReadFromBigQuery(
+                table=table_name,
+                method=beam.io.ReadFromBigQuery.Method.EXPORT,  # type: ignore
+                *args,
+                **kwargs,
+            ),
+        )
 
 
 # Below type ignores are due to mypy star expansion issues: https://github.com/python/mypy/issues/6799
@@ -29,15 +79,15 @@ def _get_bigquery_ptransform(
 
 class BigqueryNodeDataReference(NodeDataReference):
     def yield_instance_dict_ptransform(self, *args, **kwargs) -> InstanceDictPTransform:
-        return _get_bigquery_ptransform(table_name=self.reference_uri, *args, **kwargs)  # type: ignore
+        return _get_bigquery_ptransform(table_name=self.reference_uri, partition_key=self.partition_key, *args, **kwargs)  # type: ignore
 
     def __repr__(self) -> str:
-        return f"BigqueryNodeDataReference(node_type={self.node_type}, identifier={self.identifier}, reference_uri={self.reference_uri})"
+        return f"BigqueryNodeDataReference(node_type={self.node_type}, identifier={self.identifier}, reference_uri={self.reference_uri}, partition_key={self.partition_key})"
 
 
 class BigqueryEdgeDataReference(EdgeDataReference):
     def yield_instance_dict_ptransform(self, *args, **kwargs) -> InstanceDictPTransform:
-        return _get_bigquery_ptransform(table_name=self.reference_uri, *args, **kwargs)  # type: ignore
+        return _get_bigquery_ptransform(table_name=self.reference_uri, partition_key=self.partition_key, *args, **kwargs)  # type: ignore
 
     def __repr__(self) -> str:
-        return f"BigqueryEdgeDataReference(edge_type={self.edge_type}, src_identifier={self.src_identifier}, dst_identifier={self.dst_identifier}, reference_uri={self.reference_uri})"
+        return f"BigqueryEdgeDataReference(edge_type={self.edge_type}, src_identifier={self.src_identifier}, dst_identifier={self.dst_identifier}, reference_uri={self.reference_uri}, partition_key={self.partition_key})"
--- a/python/gigl/src/data_preprocessor/lib/ingest/reference.py
+++ b/python/gigl/src/data_preprocessor/lib/ingest/reference.py
@@ -45,9 +45,10 @@ class NodeDataReference(DataReference, ABC):
 
     node_type: NodeType
     identifier: Optional[str] = None
+    partition_key: Optional[str] = None
 
     def __repr__(self) -> str:
-        return f"NodeDataReference(node_type={self.node_type}, identifier={self.identifier}, reference_uri={self.reference_uri})"
+        return f"NodeDataReference(node_type={self.node_type}, identifier={self.identifier}, reference_uri={self.reference_uri}, partition_key={self.partition_key})"
 
 
 @dataclass(frozen=True)  # type: ignore
@@ -60,6 +61,7 @@ class EdgeDataReference(DataReference, ABC):
     edge_usage_type: EdgeUsageType = EdgeUsageType.MAIN
     src_identifier: Optional[str] = None
     dst_identifier: Optional[str] = None
+    partition_key: Optional[str] = None
 
     def __repr__(self) -> str:
-        return f"EdgeDataReference(edge_type={self.edge_type}, src_identifier={self.src_identifier}, dst_identifier={self.dst_identifier}, reference_uri={self.reference_uri})"
+        return f"EdgeDataReference(edge_type={self.edge_type}, src_identifier={self.src_identifier}, dst_identifier={self.dst_identifier}, reference_uri={self.reference_uri}, partition_key={self.partition_key})"


python/tests/integration/main.py
```diff
@@ -21,5 +21,5 @@ def run(pattern: str = "*_test.py") -> bool:
 
 
 if __name__ == "__main__":
-    was_successful: bool = run(pattern=parse_args().test_file_pattern)
+    was_successful: bool = run(pattern="enumerator_test.py")
```
mkolodner-sc, line 24: This will be reverted for merge, this is just here for now so we can test using from github with `/integration_test`

python/tests/integration/pipeline/data_preprocessor/enumerator_test.py
```diff
@@ -461,13 +475,14 @@ def test_for_correctness(self):
             list_enumerator_edge_type_metadata,
         ) = enumerator.run(
             applied_task_identifier=AppliedTaskIdentifier(
-                self.__applied_task_identifier
+                f"{self.__applied_task_identifier}_{test_suffix}"
```
svij-sc, line 478: could we just pass in the applied_task_identifier and have it be diff for both tests?
This "hidden" indirection is introducing confusion.

python/tests/integration/pipeline/data_preprocessor/enumerator_test.py
```diff
@@ -493,20 +509,98 @@ def test_for_correctness(self):
...
+        self._run_enumeration_test_and_validate(
+            test_suffix="full_table_enumeration",
+            node_data_references=self.node_data_references,
+            edge_data_references=self.edge_data_references,
+        )
+
+    def test_partitioned_enumeration(self):
+        """Test that partitioned enumeration works correctly for both node and edge data."""
+
+        partitioned_node_reference = BigqueryNodeDataReference(
```
svij-sc, line 556: nit: re-use

```python
NUM_PARTITIONS = 20
partitioned_node_references = [
     BigqueryNodeDataReference(
           reference_uri=ref.reference_uri,
            node_type=ref.node_type,
            identifier=ref.identifier,
            partition_key=ref.identifier,
            num_partitions=NUM_PARTITIONS,
     )
     for ref in self.node_data_references
]
```

same for edge

svij-sc (commented):
Is it possible to change one of the e2e tests to also use these partitioned reads?

python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
```diff
@@ -1,43 +1,133 @@
...
+        self._table_name = table_name
+        # TODO (mkolodner-sc): Infer this value based on number of rows in table
+        if num_partitions is None:
+            logger.info(
+                f"No num_partitions provided for {table_name}, using default of 20"
+            )
+            num_partitions = 20
+        self._num_partitions = num_partitions
+        self._partition_key = partition_key
+        resource_config = get_resource_config()
```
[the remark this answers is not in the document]
mkolodner-sc: Discussed offline, we'll also add this as a parameter that can be configurable by users, but will fall back to `get_resource_config()` if this is not explicitly provided

mkolodner-sc:
> Is it possible to change one of the e2e tests to also use these partitioned reads?

Considering this has been suggested to be more performant, I wouldn't mind just applying this change to all our e2e tests.  I've updated our PassthroughDataPreprocessorConfig, which our e2e tests reference, to enable this partitioned read for all BigQueryEdgeDataReferences there and leave the BigQueryNodeDataReferences untouched so we have coverage of both partitioned and non-partitioned reads.

WDYT?

python/gigl/common/beam/partitioned_read.py
```diff
@@ -0,0 +1,93 @@
...
+                (
+                    self.temp_dataset_reference.projectId,
+                    self.temp_dataset_reference.datasetId,
+                )
+            )
+            if self.temp_dataset_reference is not None
+            else None
+        )
+
+        return hash((self.partition_key, self.num_partitions, temp_dataset_hash))
```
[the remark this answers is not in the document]
mkolodner-sc: We populate the temp_dataset_reference using `get_resource_config` during `_post_init__` if it was not explicitly provided, which doesn't work if we have a frozen dataclass.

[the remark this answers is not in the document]
kmontemayor2-sc: we could have it default with a `field` right?

python/gigl/common/beam/partitioned_read.py
```diff
@@ -0,0 +1,74 @@
...
+        self._kwargs = kwargs
+
+    def expand(self, pbegin: PBegin):
+        pcollection_list = []
+        for i in range(self._num_partitions):
+            # We use farm_fingerprint as a determinstic hashing function which will allow us to partition
+            # on keys of any type (i.e. strings, integers, etc.)
+            query = (
+                f"SELECT * FROM `{self._table_name}` "
+                f"WHERE MOD(ABS(FARM_FINGERPRINT(CAST({self._partition_key} AS STRING))), {self._num_partitions}) = {i}"
```
xgao4-sc, line 58: Per BigQuery's doc about [FARM_FINGERPRINT](https://cloud.google.com/bigquery/docs/reference/standard-sql/hash_functions#farm_fingerprint), the return type is INT64 and return value could be negative. Is there a risk of overflow when we take ABS? Per BigQuery's doc about [ABS](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#abs): "_...Returns an error if the argument is an integer and the output value can't be represented as the same type; this happens only for the largest negative input value, which has no positive representation._"

I am thinking of first MOD and then ABS. Per BigQuery's doc about [MOD](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#mod): "_MOD(X, Y) returns the remainder of the division of X by Y. Returned value has the same sign as X._". If X is negative, and Y is positive, MOD(X,Y) will be negative and within the range of 0, -1, -2, ..., -(Y-1). ABS(MOD(X,Y)) will still be within the range of [0, self._num_partitions). What do you think? cc: @svij-sc @kmontemayor2-sc for thoughts too.

mkolodner-sc: I don't see an issue with doing `ABS(MOD(X, Y))` to avoid that edge case here, will update the code and add a comment. Thanks!

python/gigl/src/data_preprocessor/lib/enumerate/utils.py
```diff
@@ -235,6 +235,7 @@ def __enumerate_node_reference(
                 reference_uri=bq_destination_enumerated_node_features_table_name,
                 node_type=node_data_ref.node_type,
                 identifier=node_data_ref.identifier,
+                partitioned_read_info=node_data_ref.partitioned_read_info,
```
xgao4-sc, line 238: Would you like to elaborate why we need the change here? Thanks! I guess it relates to how NodeDataReference gets modified and passed to the Dataflow job?

mkolodner-sc: Sure thing -- in our codebase there are places where we end up reconstructing the `BigQueryNodeDataReference`, for example, with a different `reference_uri`, such as [1](https://github.com/Snapchat/GiGL/blob/901a99a4ed9bcf9ce2e3be933ed3555eb28b6c0c/python/gigl/src/data_preprocessor/lib/enumerate/utils.py#L234-L235) where the `reference_uri` changes after enumeration. We need to ensure we pass the original `node_data_ref.partitioned_read_info` here -- otherwise it will be `None`. 

We need to make a similar change in `data_preprocessor` [[2](https://github.com/Snapchat/GiGL/blob/901a99a4ed9bcf9ce2e3be933ed3555eb28b6c0c/python/gigl/src/data_preprocessor/data_preprocessor.py#L793)] since we reconstruct a `BigQueryNodeDataReference` to populate the `identifier` fields. We need to reconstruct these dataclasses rather than modify them in place since they are frozen [[3](https://github.com/Snapchat/GiGL/blob/main/python/gigl/src/data_preprocessor/lib/ingest/reference.py#L40)].

svij-sc:
> @svij-sc Considering this has been suggested to be more performant, I wouldn't mind just applying this change to all our e2e tests. I've updated our PassthroughDataPreprocessorConfig, which our e2e tests reference, to enable this partitioned read for all BigQueryEdgeDataReferences there and leave the BigQueryNodeDataReferences untouched so we have coverage of both partitioned and non-partitioned reads.


@mkolodner-sc since we still have pipelines on the old pathway - we should keep tests for both.
Btw, is it more performant in all cases? I am imagining the user will need to tune this to work well for their data size.

mkolodner-sc:
> since we still have pipelines on the old pathway - we should keep tests for both.

@svij-sc Makes sense, all of our nodes and all of our UDL edges are still using references with the old pathway [[1](https://github.com/Snapchat/GiGL/blob/4f864b7556497500c7a698df9a083f5ccae88c54/python/gigl/src/mocking/mocking_assets/passthrough_preprocessor_config_for_mocked_assets.py#L67-L69)] [[2](https://github.com/Snapchat/GiGL/blob/4f864b7556497500c7a698df9a083f5ccae88c54/python/gigl/src/mocking/mocking_assets/passthrough_preprocessor_config_for_mocked_assets.py#L192-L196)]-- would this provide sufficient test coverage in this case for both?

> Btw, is it more performant in all cases? I am imagining the user will need to tune this to work well for their data size.

Good point, on larger dataset we have seen it improve, but I'd assume this might not always be the case especially on smaller datasets.

python/gigl/src/data_preprocessor/lib/enumerate/utils.py
xgao4-sc: Thanks a lot! To confirm my understanding: enumerator does not change column name, but just changes id column's values to INT, right?

xgao4-sc (approved):
The PR looks good to me. Thanks a lot for the work!

The [log](https://console.cloud.google.com/logs/query;query=resource.type%3D%22ml_job%22%0Aresource.labels.job_id%3D%225233207729354440704%22%0Atimestamp%3E%3D%222025-10-01T17:53:12.972Z%22%0Atimestamp%3C%3D%222025-10-01T19:38:27.947Z%22%20severity%3E%3DDEFAULT;storageScope=project;pinnedLogId=2025-10-01T17:57:14.609880915Z%2F1r9jrthf212lth;cursorTimestamp=2025-10-01T17:57:14.609880915Z;aroundTime=2025-10-01T17:57:14.609880915Z;duration=PT1H?project=external-snap-ci-github-gigl) in KFP component does not guarantee how Dataflow executes. To prove that partition read is effective, we could check the dataflow execution graph ([job page](https://console.cloud.google.com/dataflow/jobs/us-central1/2025-10-01_10_57_36-11845552290877035698;graphView=0?project=external-snap-ci-github-gigl&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))), which shows there are 5 partitions, and we flatten them afterwards.  
image

python/gigl/common/beam/partitioned_read.py
```diff
@@ -0,0 +1,80 @@
...
+
+import apache_beam as beam
+from apache_beam.io.gcp.bigquery import BigQueryQueryPriority
+from apache_beam.io.gcp.internal.clients.bigquery import DatasetReference
+from apache_beam.pvalue import PBegin
+
+from gigl.common.logger import Logger
+from gigl.env.pipelines_config import get_resource_config
+
+logger = Logger()
```
svij-sc, line 11: should we add some logs?
Atleast when initializing and the params used?

python/gigl/common/beam/partitioned_read.py
```diff
@@ -0,0 +1,80 @@
...
+        **kwargs,
+    ):
+        super().__init__()
+        self._table_name: str = table_name
+        self._num_partitions: int = partitioned_read_info.num_partitions
+        if self._num_partitions <= 0:
+            raise ValueError(
+                f"Number of partitions specified must be greater than 0, got {self._num_partitions}"
+            )
+        self._partition_key: str = partitioned_read_info.partition_key
```
svij-sc, line 48: nit: sanity check assert that partition column is a valid string.

python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
```diff
@@ -1,43 +1,70 @@
...
 )
 from gigl.src.data_preprocessor.lib.types import InstanceDictPTransform
 
+logger = Logger()
+
 
 def _get_bigquery_ptransform(
-    table_name: str, *args, **kwargs
+    table_name: str,
+    partitioned_read_info: Optional[BigQueryPartitionedReadInfo] = None,
```
svij-sc, line 22: I forgot to give this feedback earlier, but something I brought up w/ @xgao4-sc on the original POC design.
`partition` is already a concept in BQ: https://cloud.google.com/bigquery/docs/partitioned-tables

I think when first reading the POC I was confused about what we are doing vs something BQ supports.
Maybe we can rename to avoid other's confusion in the future:

```
BigQueryPartitionedReadInfo → BigQueryShardedReadConfig
partition_key → shard_key
num_partitions → num_shards
```
etc.

python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
```diff
@@ -1,43 +1,70 @@
...
+            ),
+        )
 
 
 # Below type ignores are due to mypy star expansion issues: https://github.com/python/mypy/issues/6799
-
-
+@dataclass(frozen=True)
 class BigqueryNodeDataReference(NodeDataReference):
+    partitioned_read_info: Optional[BigQueryPartitionedReadInfo] = None
```
svij-sc, line 53: Can we add docstring for this and edgeDataReference while we are at it?
And also provide documentation on how to optimize the reads w/ use of this new variable.

Probably good to include the base class params too.

svij-sc (approved):
Generally LGTM, some comments that should be addressed prior to merging.

python/gigl/src/data_preprocessor/lib/enumerate/utils.py
xgao4-sc: @mkolodner-sc friendly reminder on my question above. Thanks!

python/gigl/common/beam/partitioned_read.py
```diff
@@ -0,0 +1,80 @@
...
+    # TODO (mkolodner-sc): Instead of using this default, infer this value based on number of rows in table
+    num_partitions: int = 20
+    # The project ID to use for temporary datasets when running partitioned reads. If not provided, will default
+    # to use the project specified in the resource config environment variables.
+    project_id: str = field(default_factory=lambda: get_resource_config().project)
+    # The temporary bigquery dataset name to use when running partitioned reads. If not provided, will default
+    # to use the temp assets dataset name specified in the resource config environment variables.
+    temp_dataset_name: str = field(
+        default_factory=lambda: get_resource_config().temp_assets_bq_dataset_name
+    )
```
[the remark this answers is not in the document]
xgao4-sc: @svij-sc Do you mean let user call `get_resource_config()` in their code, and transfer the complexity there?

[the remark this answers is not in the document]
xgao4-sc: Discussed offline and resolved. Thanks!

python/gigl/src/data_preprocessor/lib/ingest/bigquery.py
mkolodner-sc: Done. 

When looking at the way these fields more closely I came across interesting finding, which is that the `identifier` field for nodes and `src_identifier`, `dst_identifier` fields for edges are always overridden by the corresponding fields in the `{Node/Edge}DataPreprocessingSpec` even if they were provided here. [[1](https://github.com/Snapchat/GiGL/blob/main/python/gigl/src/data_preprocessor/data_preprocessor.py#L811-L812)][[2](https://github.com/Snapchat/GiGL/blob/main/python/gigl/src/data_preprocessor/data_preprocessor.py#L793)]. This seems smelly to me that we are overriding the identifier value even if explicitly provided here. That being said, handling this smell seems out-of-scope for this current work, but I've left a TODO to look into addressing this.

PR #345, opened 2025-09-29
Add 'run_if_exists' so we can '''cache''' bq results

kmontemayor2-sc:
As discussed offline, we can add this to "easily" cache some bq table outputs. 

This is a very naive cacheing strategy, and needs to asusme the bq table doesn't get modified, but since many of our usecases are like:

```
CREATE IF NOT EXIST {date_table_name}
....
```

Using this should be "safe" in many circumstances.

### DIFF
--- a/python/gigl/src/common/utils/bq.py
+++ b/python/gigl/src/common/utils/bq.py
@@ -142,6 +142,15 @@ class BqUtils:
             logger.exception(f"Could not run query: {e}")
             raise e
 
+    def run_if_exists(
+        self, query: str, output_table: str, labels: dict[str, str], **job_config_args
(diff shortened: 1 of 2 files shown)


python/gigl/src/common/utils/bq.py
```diff
@@ -142,6 +142,15 @@ def run_query(
             logger.exception(f"Could not run query: {e}")
             raise e
 
+    def run_if_exists(
```
yliu2-sc, line 145: I think BQ already check this if you don't set your `write_disposition` to `WRITE_TRUNCATE` and it's default to `WRITE_EMPTY`

PR #347, opened 2025-09-29
Add utility to down_sample node ids based on label

mkolodner-sc:
**Scope of work done**

- Introduces a utility which downsamples node ids based on labels. This will get called in the training/inference loop by the user if they want to downsample their node IDs.
- Adds tests for introduced utilities

### DIFF
--- a/python/gigl/utils/down_sampler.py
+++ b/python/gigl/utils/down_sampler.py
@@ -11,42 +11,6 @@ from gigl.src.common.types.graph_data import NodeType
 logger = Logger()
 
 
-def down_sample_node_ids_from_labels(
-    node_ids: torch.Tensor,
-    id2idx: torch.Tensor,
-    node_label_feats: torch.Tensor,
-    label_filter_fn: Callable[[torch.Tensor], torch.Tensor] = lambda x: x >= 0,
-) -> torch.Tensor:
-    """
-    Downsample the provided node ids based on the provided node labels. Downsampling means that
-    we will sample a subset of the node ids to use for any downstream purpose. The node label values will be used to
-    determine what gets downsampled, and can be customized with the `label_filter_fn` argument.
-    By default, this will filter out negative label values from the node ids.
-
-    Args:
-        node_ids (torch.Tensor): The node ids to down_sample Should be a 1D tensor of shape (N,).
-        id2idx (torch.Tensor): Mapping from node IDs to indices in the feature tensor. Should be a 1D tensor of shape (max_node_id,).
-        node_label_feats (torch.Tensor): The node label features. Should be a 2D tensor of shape (N, 1), where N is the number of nodes.
-        label_filter_fn (Callable[[torch.Tensor], torch.Tensor]): A callable that takes a tensor of labels of shape [N, 1]
-            and returns a boolean mask of the same [N, 1] shape, indicating which labels should be included. By default,
-            filters out negative values.
-    Returns:
-        torch.Tensor: The down_sampled node ids of shape [M,], where M is the number of down_sampled node ids.
-            Only nodes with valid labels are included in the output.
-    """
-
-    # Use id2idx to get the feature indices directly
-    feature_indices = id2idx[node_ids]
-
-    # Apply the label filter function
-    labels_to_include = label_filter_fn(node_label_feats[feature_indices]).squeeze(-1)
-
-    # Return the down_sampled valid node ids
-    down_sampled_valid_node_ids = node_ids[labels_to_include]
-
-    return down_sampled_valid_node_ids
-
-
 def down_sample_node_ids_from_dataset_labels(
     dataset: DistDataset,
     node_ids: torch.Tensor,
@@ -80,13 +44,8 @@ def down_sample_node_ids_from_dataset_labels(
     else:
         labels = dataset.node_labels
 
-    assert labels.size(0) == node_ids.size(0)
+    labels_to_include = label_filter_fn(labels[node_ids]).squeeze(-1)
 
-    down_sampled_node_ids = down_sample_node_ids_from_labels(
-        node_ids=node_ids,
-        id2idx=labels.id2index,
-        node_label_feats=labels.feature_tensor,
-        label_filter_fn=label_filter_fn,
-    )
+    down_sampled_node_ids = node_ids[labels_to_include]
 
     return down_sampled_node_ids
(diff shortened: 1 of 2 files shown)


python/gigl/utils/down_sampler.py
```diff
@@ -0,0 +1,86 @@
...
+    if isinstance(dataset.node_labels, Mapping):
+        if node_type is None:
+            raise ValueError(
+                "node_type must be provided if the dataset is heterogeneous"
+            )
+        labels: Feature = dataset.node_labels[node_type]
+    else:
+        labels = dataset.node_labels
+
+    labels.lazy_init_with_ipc_handle()
```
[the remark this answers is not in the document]
mkolodner-sc: Added comment here.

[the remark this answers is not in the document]
mkolodner-sc: This is always the case and doesn't have anything to do with our codebase -- GLT does this under-the-hood whenever these fields are de-serialized and re-serialized to have better performance I presume. Generally, we'll need to do this when we are using these `Feature` fields after a dataset has been built. Note that this is a no-op if the fields have already been initialized [[1](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/data/feature.py#L243-L244)]

> I guess what I'm getting at is why don't we need this pattern with other parts of the dataset?

Many of the GLT operations on the dataset or features also end up calling this lazy init themselves [[2](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/data/feature.py#L147)]. The reason why we need to call this explicitly is because we need to extract the `id2index` tensor from the `Feature` object.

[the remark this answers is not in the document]
mkolodner-sc: Yes, outside of tests users will never need to interact with these Feature fields directly, as GLT takes care of it under-the-hood. 

> I think it will be very surprising if they need to call some init_... first before they access feature tensors.

Yes, I was surprised too the first time I tried inspecting these feature tensors and found these fields to be None. In the majority of cases though, users will not need to interact with these features. At least for this PR, this is still something we must do to safely acesss the values of the GLT `Feature` object.

python/gigl/utils/down_sampler.py
```diff
@@ -0,0 +1,94 @@
...
+import torch
+from graphlearn_torch.data import Feature
+
+from gigl.common.logger import Logger
+from gigl.distributed.dist_dataset import DistDataset
+from gigl.src.common.types.graph_data import NodeType
+
+logger = Logger()
+
+
```
yliu2-sc, line 13: better as an object class that handles downsampling, @kmontemayor2-sc for opinion as well

python/gigl/utils/down_sampler.py
```diff
@@ -0,0 +1,94 @@
...
+def down_sample_node_ids_from_labels(
+    node_ids: torch.Tensor,
+    id2idx: torch.Tensor,
```
yliu2-sc: this can benefit from more descirption naming, ex. `node_id_to_index`

python/gigl/utils/down_sampler.py
```diff
@@ -0,0 +1,94 @@
...
+    node_label_feats: torch.Tensor,
```
yliu2-sc: just `node_labels`?

python/gigl/utils/down_sampler.py
```diff
@@ -0,0 +1,51 @@
...
+from gigl.src.common.types.graph_data import NodeType
+
+logger = Logger()
+
+
+def down_sample_node_ids_from_dataset_labels(
+    dataset: DistDataset,
+    node_ids: torch.Tensor,
+    node_type: Optional[NodeType] = None,
+    label_filter_fn: Callable[[torch.Tensor], torch.Tensor] = lambda x: x >= 0,
```
yliu2-sc, line 18: does this actually achieve downsampling on majority class to minority class?

python/gigl/utils/down_sampler.py
kmontemayor2-sc: I prefer functions to objects generally - unless we have an explicit need for state.

python/gigl/utils/down_sampler.py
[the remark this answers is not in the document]
kmontemayor2-sc: Sorry I'm a little confused - is all your code pushed? below is the file as I see it rn. Did we intend to remove the other function here?

Are we relying on some part of `dist_dataset` now?


```
from collections.abc import Callable, Mapping
from typing import Optional

import torch
from graphlearn_torch.data import Feature

from gigl.distributed.dist_dataset import DistDataset
from gigl.src.common.types.graph_data import NodeType

logger = Logger()


def down_sample_node_ids_from_dataset_labels(
    dataset: DistDataset,
    node_ids: torch.Tensor,
    node_type: Optional[NodeType] = None,
    label_filter_fn: Callable[[torch.Tensor], torch.Tensor] = lambda x: x >= 0,
) -> torch.Tensor:
    """
    Down sample node ids using the provided dataset's labels. Downsampling means that
    we will sample a subset of the node ids to use for any downstream purpose. The node label values will be used to
    determine what gets downsampled, and can be customized with the `label_filter_fn` argument.
    By default, this will filter out negative label values from the node ids.

    Args:
        dataset (DistDataset): The dataset to down_sample.
        node_ids (torch.Tensor): The node ids to down_sample. Should be a 1D tensor of shape (N,).
        node_type (Optional[NodeType]): The node type to down_sample. If the dataset is heterogeneous, this must be provided.
        label_filter_fn (Callable[[torch.Tensor], torch.Tensor]): A callable that takes a tensor of
            labels of shape [N, 1] and returns a boolean mask of the same [N, 1] shape, indicating which labels should be included.
            By default, filters out negative values.
    Returns:
        torch.Tensor: The down_sampled node ids of shape [M,], where M is the number of down_sampled node ids.
            Only nodes with valid labels are included in the output.
    """

    if isinstance(dataset.node_labels, Mapping):
        if node_type is None:
            raise ValueError(
                "node_type must be provided if the dataset is heterogeneous"
            )
        labels: Feature = dataset.node_labels[node_type]
    else:
        labels = dataset.node_labels

    labels_to_include = label_filter_fn(labels[node_ids]).squeeze(-1)

    down_sampled_node_ids = node_ids[labels_to_include]

    return down_sampled_node_ids
```

PR #348, opened 2025-09-30
Add longer timeout to v1 trainer

kmontemayor2-sc:
**Scope of work done**

 
Vertex AI may start the training jobs in a scattered manner, e.g. rank 0 starts at time 0, rank 10 starts at time 11, and so rank 0 timesout and crashes.

We should probably parameterize this in the future but in practice the 45 minute timeout is sufficient here.

Tested internally succsefully.

### DIFF
--- a/python/gigl/src/training/v1/lib/training_process.py
+++ b/python/gigl/src/training/v1/lib/training_process.py
@@ -52,6 +52,12 @@ from gigl.src.training.v1.lib.base_trainer import BaseTrainer
 
 logger = Logger()
 
+# Without longer timeout, we can see the cluster failing as Vertex AI
+# will start the jobs in a scattered manner, so we may timeout waiting
+# for all machines to come online.
+# TODO(kmonte): We should parameterize this timeout.
+_PROCESS_GROUP_TIMEOUT = datetime.timedelta(minutes=45)
+
 
 @profileit(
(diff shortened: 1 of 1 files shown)


python/gigl/src/training/v1/lib/training_process.py
```diff
@@ -354,8 +355,17 @@ def __setup_training_env(self, device: torch.device):
         use_cuda = device.type != "cpu"
         if should_distribute():
             distributed_backend = get_distributed_backend(use_cuda=use_cuda)
-            logger.info(f"Using distributed PyTorch with {distributed_backend}")
-            torch.distributed.init_process_group(backend=distributed_backend)
+            # Without longer timeout, we can see the cluster failing as Vertex AI
+            # will start the jobs in a scattered manner, so we may timeout waiting
+            # for all machines to come online.
+            # TODO(kmonte): We should parameterize this timeout.
```
yliu2-sc, line 361: can we parameterize here already? Or at least make it a constant?

yliu2-sc (approved):

python/gigl/src/training/v1/lib/training_process.py
kmontemayor2-sc: hmmm, the problem is we'd need to expose this somehow through gbml config, we could add it as a CLI flag but I'm not sure it's worth piping through all of the code.

I can make it a constant though.

PR #350, opened 2025-10-01
Add `VertexAIService.launch_graph_store_job` 

kmontemayor2-sc:
**Scope of work done**

Add ability to launch heterogeneous VAI clusters via new `VertexAIService.launch_graph_store_job` API

Where is the documentation for this feature?: In doc comments

Did you add automated tests or write a test plan? Added unit tests


python/gigl/common/services/vertex_ai.py
```diff
@@ -355,3 +468,35 @@ def wait_for_run_completion(
                 f"Vertex AI run stopped with status: {run.state}. "
                 f"Please check the Vertex AI page to trace down the error."
             )
+
+
+def _get_machine_spec(job_config: VertexAiJobConfig) -> MachineSpec:
```
svij-sc, line 473:
```suggestion
def _create_machine_spec(job_config: VertexAiJobConfig) -> MachineSpec:
```

similar for other function

python/gigl/common/services/vertex_ai.py
```diff
@@ -245,6 +239,125 @@ def launch_job(self, job_config: VertexAiJobConfig) -> None:
...
+            - GIGL_COMPUTE_CLUSTER_MASTER_KEY
+        Whose values are the cluster ranks of the leaders for the storage and compute clusters respectively.
+        For example, if if there are 2 nodes in the storage cluster, and 3 nodes in the compute cluster,
+        Then,
+            - GIGL_STORAGE_CLUSTER_MASTER_KEY = 0
+            - GIGL_COMPUTE_CLUSTER_MASTER_KEY = 2 # e.g. the "first" worker in the computer cluster pool is the leader.
+
+        NOTE:
+            We use the job_name, timeout, and enable_web_access from the storage cluster.
+            These fields, if set on the compute cluster, will be ignored.
```
svij-sc, line 264: oh weird....
I guess this is a global flag eh?

python/gigl/common/services/vertex_ai.py
```diff
@@ -245,6 +239,125 @@ def launch_job(self, job_config: VertexAiJobConfig) -> None:
...
+        # Whenever `connect_worker_pool()` is called, the leader worker will
+        # write its internal IP address to this file. The other workers will
+        # read this file to get the leader worker's internal IP address.
+        # See connect_worker_pool() implementation for more details.
+        leader_worker_internal_ip_file_path = GcsUri.join(
+            self._staging_bucket,
+            storage_cluster.job_name,
+            datetime.datetime.now().strftime("%Y%m%d-%H%M%S"),
+            "leader_worker_internal_ip.txt",
+        )
```
svij-sc, line 288: there is plans for this to be deprecated in the launch_job setting as its not needed anymore with DDP + utilities to transfer free port information and main nodes ip address, et al.

I would recommend we don't do this in this new paradigm to begin with.

python/gigl/common/services/vertex_ai.py
```diff
@@ -245,6 +239,125 @@ def launch_job(self, job_config: VertexAiJobConfig) -> None:
...
+                value=leader_worker_internal_ip_file_path.uri,
+            ),
+            env_var.EnvVar(
+                name=STORAGE_CLUSTER_MASTER_KEY,
+                value="0",
+            ),
+            env_var.EnvVar(
+                name=COMPUTE_CLUSTER_MASTER_KEY,
+                value=str(storage_cluster.replica_count),
+            ),
```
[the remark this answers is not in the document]
mkolodner-sc: I have a preference for trying to reuse the existing `CLUSTER_SPEC` field -- otherwise we'd have two environment variables set both of which are sources of truth for the cluster information here. 

> Doing it this way also is less reliant on VAI if we want to move off in the future.

Seeing as this lives in `python/gigl/common/services/vertex_ai.py`, I don't think we'd be kicking ourselves in the foot by relying on VAI environment variables in this file specifically.

mkolodner-sc (approved):
Gotcha, thanks for the clarification. I don't feel too strongly here as that makes sense to me as well -- will defer to @svij-sc here. Approving to unblock

python/gigl/common/services/vertex_ai.py
```diff
@@ -245,6 +239,110 @@ def launch_job(self, job_config: VertexAiJobConfig) -> None:
...
+                disk_spec=storage_disk_spec,
+                replica_count=storage_cluster.replica_count - 1,
+            )
+            worker_pool_specs.append(worker_spec)
+        else:
+            worker_pool_specs.append({})
+        # For whatever reason, VAI errors out (indescriptly) if we put the computer cluster as anything other
+        # than the fourth worker pool.
+        # So we need to pad.
+        worker_pool_specs.append({})
```
svij-sc, line 314: For the existing setting, without graph store: we have the compute cluster as the second pool w/ first being the leader pool.

kmontemayor2-sc: VAI has a requirement that the leader pool be of size 1, otherwise this error gets thrown.

```
google.api_core.exceptions.InvalidArgument: 400 List of found errors:   1.Field: job_spec.worker_pool_specs; Message: Replica count for master worker pool (worker_pool_specs[0].replica_count) should be 1.     [field_violations {
  field: "job_spec.worker_pool_specs"
  description: "Replica count for master worker pool (worker_pool_specs[0].replica_count) should be 1."
}
```

As such we currently use `[leader pool], [worker pool]` for our existing distributed setups, where `len(worker_pool) == num_replicas - 1`.

I don't think we should really consider it a "computer" pool, it's just the only pool, it does computer and storage (and we have some implicit assumption the leader is also part of it).

kmontemayor2-sc:
> Could we try worker pools index 0, 1 continuing to be the main training/inference processes?
With either 2, or 3 being the storage pools? (Ideally 2; the the graph store is already the “server” lane \ and maps cleanly to paradigms off Parameter Servers / Reduction Server.


Sure, we can put the compute pool first.

For whatever reason, VAI rejects it when I put *anything* (gpu or not) into worker pool 2. idk why.

PR #352, opened 2025-10-01
Add FeatureFlag for not populating embedding field

mkolodner-sc:
**Scope of work done**

- Adds a GbmlConfigPbWrapper property for skipping embedding path population in the frozen config during `config_populator`
- Adds tests to ensure the embedding path is not populated with this feature flag

### DIFF
--- a/python/gigl/src/common/types/pb_wrappers/gbml_config.py
+++ b/python/gigl/src/common/types/pb_wrappers/gbml_config.py
@@ -469,6 +469,9 @@ class GbmlConfigPbWrapper:
         """
         Allows access to should_populate_predictions_path under GbmlConfig
 
+        This flag defaults to False, but can be set to True to populate the predictions path in the InferenceOutput for each entity type.
+        We default this to False since config populator currently does not populate the predictions path for link prediction tasks.
+
         This flag is a temporary workaround to populate the extra embeddings for the same entity type
 
         Returns:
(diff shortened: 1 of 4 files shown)


python/gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -481,3 +484,23 @@ def should_populate_predictions_path(self) -> bool:
                 )
             )
         )
+
+    @property
+    def should_populate_embeddings_path(self) -> bool:
+        """
+        Allows access to should_populate_embeddings_path under GbmlConfig
+
+        This flag defaults to True, but can be set to False to skip populating embedding paths in InferenceOutput.
```
mkolodner-sc, line 493: As mentioned in the description, we need to default this value to `True` to be backwards compatible with our existing pipelines which are expecting embedding table paths to be popualted

python/tests/unit/main.py
```diff
@@ -16,5 +16,5 @@ def run(pattern: str = "*_test.py") -> bool:
 
 
 if __name__ == "__main__":
-    was_successful: bool = run(pattern=parse_args().test_file_pattern)
+    was_successful: bool = run(pattern="config_populator_functionality_test.py")
```
mkolodner-sc, line 19: This will be removed prior to merging PR and is just here now for running testing.

svij-sc: reminder to revert

PR #355, opened 2025-10-06
Add get_graph_store_info to determine "Graph cluster" info from environment

kmontemayor2-sc:
**Scope of work done**

* Add get_graph_store_info to get information about graph store environment.
* Fix bug in `get_cluster_spec` to parse `job` correctly
* Bump GiGL version so that we can run this code in integration tests

### DIFF
--- a/python/tests/unit/distributed/utils/networking_test.py
+++ b/python/tests/unit/distributed/utils/networking_test.py
@@ -312,85 +312,118 @@ def _test_get_graph_store_info_in_dist_context(
         world_size=world_size,
         rank=rank,
     )
-    try:
-        # Call get_graph_store_info
-        graph_store_info = get_graph_store_info()
 
-        # Verify the result is a GraphStoreInfo instance
-        assert isinstance(
-            graph_store_info, GraphStoreInfo
-        ), "Result should be a GraphStoreInfo instance"
-        # Verify cluster sizes
-        assert (
-            graph_store_info.num_storage_nodes == storage_nodes
-        ), f"Expected {storage_nodes} storage nodes"
-        assert (
-            graph_store_info.num_compute_nodes == compute_nodes
-        ), f"Expected {compute_nodes} compute nodes"
-        assert (
-            graph_store_info.num_cluster_nodes == storage_nodes + compute_nodes
-        ), "Total nodes should be sum of storage and compute nodes"
+    if compute_nodes == 1:
+        worker_pool_sizes = [1, 0, storage_nodes]
+        if rank == 0:
+            worker_pool = "workerpool0"
+            index = 0
+        else:
+            worker_pool = "workerpool2"
+            index = rank - 1
+    else:
+        if rank == 0:
+            worker_pool = "workerpool0"
+            index = 0
+        elif rank < compute_nodes:
+            worker_pool = "workerpool1"
+            index = rank - 1
+        else:
+            worker_pool = "workerpool2"
+            index = rank - compute_nodes
+        worker_pool_sizes = [1, compute_nodes - 1, storage_nodes]
+    with patch.dict(
+        os.environ,
+        {
+            "RANK": str(rank),
+            "WORLD_SIZE": str(world_size),
+            "CLUSTER_SPEC": json.dumps(
+                _get_cluster_spec_for_test(worker_pool_sizes, worker_pool, index)
+            ),
+        },
+        clear=False,
+    ):
+        try:
+            # Call get_graph_store_info
+            graph_store_info = get_graph_store_info()
 
-        # Verify IP addresses are strings and not empty
-        assert isinstance(
-            graph_store_info.cluster_master_ip, str
-        ), "Cluster master IP should be a string"
-        assert (
-            len(graph_store_info.cluster_master_ip) > 0
-        ), "Cluster master IP should not be empty"
-        assert isinstance(
-            graph_store_info.storage_cluster_master_ip, str
-        ), "Storage cluster master IP should be a string"
-        assert (
-            len(graph_store_info.storage_cluster_master_ip) > 0
-        ), "Storage cluster master IP should not be empty"
-        assert isinstance(
-            graph_store_info.compute_cluster_master_ip, str
-        ), "Compute cluster master IP should be a string"
-        assert (
-            len(graph_store_info.compute_cluster_master_ip) > 0
-        ), "Compute cluster master IP should not be empty"
-
-        # Verify ports are positive integers
-        assert isinstance(
-            graph_store_info.cluster_master_port, int
-        ), "Cluster master port should be an integer"
-        assert (
-            graph_store_info.cluster_master_port > 0
-        ), "Cluster master port should be positive"
-        assert isinstance(
-            graph_store_info.storage_cluster_master_port, int
-        ), "Storage cluster master port should be an integer"
-        assert (
-            graph_store_info.storage_cluster_master_port > 0
-        ), "Storage cluster master port should be positive"
-        assert isinstance(
-            graph_store_info.compute_cluster_master_port, int
-        ), "Compute cluster master port should be an integer"
-        assert (
-            graph_store_info.compute_cluster_master_port > 0
-        ), "Compute cluster master port should be positive"
-
-        # Verify all ranks get the same result (since they should all get the same broadcasted values)
-        gathered_info: list[Optional[GraphStoreInfo]] = [None] * world_size
-        dist.all_gather_object(gathered_info, graph_store_info)
-
-        # All ranks should have the same GraphStoreInfo
-        for i, info in enumerate(gathered_info):
-            assert info is not None
+            # Verify the result is a GraphStoreInfo instance
+            assert isinstance(
+                graph_store_info, GraphStoreInfo
+            ), "Result should be a GraphStoreInfo instance"
+            # Verify cluster sizes
             assert (
-                info == graph_store_info
-            ), f"Rank {i} should have same GraphStoreInfo. Got {info} but expected {graph_store_info}"
-    finally:
-        dist.destroy_process_group()
+                graph_store_info.num_storage_nodes == storage_nodes
+            ), f"Expected {storage_nodes} storage nodes"
+            assert (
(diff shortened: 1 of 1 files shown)


python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -183,29 +182,35 @@ class ClusterSpec:
...
     @classmethod
     def from_json(cls, json_str: str) -> "ClusterSpec":
         """Instantiates ClusterSpec from a JSON string."""
         cluster_spec_json = json.loads(json_str)
         if "job" in cluster_spec_json and cluster_spec_json["job"] is not None:
-            job_spec = CustomJobSpec(**cluster_spec_json.pop("job"))
+            logger.info(f"Job spec: {cluster_spec_json['job']}")
+            job_spec = json.loads(cluster_spec_json.pop("job"))
         else:
             job_spec = None
```
svij-sc: should we raise error here instead?

python/gigl/distributed/utils/networking.py
```diff
@@ -179,3 +181,50 @@ def get_internal_ip_from_all_ranks() -> list[str]:
...
+    # We setup the VAI cluster such that the compute nodes come first, followed by the storage nodes.
+    if "workerpool1" in cluster_spec.cluster:
+        num_compute_nodes = len(cluster_spec.cluster["workerpool0"]) + len(
+            cluster_spec.cluster["workerpool1"]
+        )
+    else:
+        num_compute_nodes = len(cluster_spec.cluster["workerpool0"])
+    num_storage_nodes = len(cluster_spec.cluster["workerpool2"])
+
+    cluster_master_ip = get_internal_ip_from_master_node()
```
svij-sc: Is this gauranteed to return from `workerpool0` ?

python/gigl/distributed/utils/networking.py
```diff
@@ -179,3 +181,50 @@ def get_internal_ip_from_all_ranks() -> list[str]:
...
+    # If we want to ever support other (non-VAI) environments,
+    # we must switch here depending on the environment.
+    cluster_spec = get_cluster_spec()
+    # We setup the VAI cluster such that the compute nodes come first, followed by the storage nodes.
+    if "workerpool1" in cluster_spec.cluster:
+        num_compute_nodes = len(cluster_spec.cluster["workerpool0"]) + len(
+            cluster_spec.cluster["workerpool1"]
+        )
+    else:
+        num_compute_nodes = len(cluster_spec.cluster["workerpool0"])
```
svij-sc: do we make assumptions that this has to be length 1?
Should we assert?

python/gigl/distributed/utils/networking.py
```diff
@@ -179,3 +181,50 @@ def get_internal_ip_from_all_ranks() -> list[str]:
...
+    num_storage_nodes = len(cluster_spec.cluster["workerpool2"])
+
+    cluster_master_ip = get_internal_ip_from_master_node()
+    # We assume that the compute cluster nodes come first, followed by the storage nodes.
```
svij-sc: Is this documented somewhere?
I.e. why can we make this assumption?

python/gigl/distributed/utils/networking.py
```diff
@@ -179,3 +181,50 @@ def get_internal_ip_from_all_ranks() -> list[str]:
...
+    compute_cluster_master_ip = get_internal_ip_from_node(node_rank=0)
+    storage_cluster_master_ip = get_internal_ip_from_node(node_rank=num_compute_nodes)
+
+    cluster_master_port = get_free_ports_from_node(num_ports=1, node_rank=0)[0]
+    compute_cluster_master_port = get_free_ports_from_node(num_ports=1, node_rank=0)[0]
```
svij-sc: Calling `get_free_ports_from_node` twice without binding the port might return the same port number.
Id recommend making just one call: 
```python
cluster_master_port, compute_cluster_master_port = get_free_ports_from_node(num_ports=2, node_rank=0)
```

python/gigl/distributed/utils/networking.py
```diff
@@ -179,3 +181,50 @@ def get_internal_ip_from_all_ranks() -> list[str]:
...
+        num_compute_nodes = len(cluster_spec.cluster["workerpool0"]) + len(
+            cluster_spec.cluster["workerpool1"]
+        )
+    else:
+        num_compute_nodes = len(cluster_spec.cluster["workerpool0"])
+    num_storage_nodes = len(cluster_spec.cluster["workerpool2"])
+
+    cluster_master_ip = get_internal_ip_from_master_node()
+    # We assume that the compute cluster nodes come first, followed by the storage nodes.
+    compute_cluster_master_ip = get_internal_ip_from_node(node_rank=0)
```
svij-sc: isn't `compute_cluster_master_ip == cluster_master_ip` ?

python/tests/integration/distributed/utils/networking_test.py
```diff
@@ -0,0 +1,83 @@
...
+                storage_nodes=1,
+            ),
+            param(
+                "Test with 2 compute nodes and 2 storage nodes",
+                compute_nodes=2,
+                storage_nodes=2,
+            ),
+        ]
+    )
+    def test_get_graph_store_info(self, _, storage_nodes, compute_nodes):
```
svij-sc: Great idea to test like this! 😍 

thought: we are going to need to figure out some way to parallelize our test executions at some point.

python/tests/unit/distributed/utils/networking_test.py
```diff
@@ -289,3 +294,205 @@ def test_get_internal_ip_from_master_node_fails_if_process_group_not_initialized
...
+    ):
+        """Test successful execution of get_graph_store_info in a real distributed context."""
+        init_process_group_init_method = get_process_group_init_method()
+        world_size = storage_nodes + compute_nodes
+        if compute_nodes == 1:
+            worker_pool_sizes = [1, 0, storage_nodes]
+        else:
+            worker_pool_sizes = [1, compute_nodes - 1, storage_nodes]
+        with patch.dict(
+            os.environ,
```
svij-sc: any reason we need to capture os.environ?

this is potentially dangerous as capturing os.environ github secrets may also be captured here and could be a vector to circumvent safety that github runners provide against secrets logging.

python/tests/unit/common/utils/vertex_ai_context_test.py
```diff
@@ -150,11 +144,11 @@ def test_parse_cluster_spec_success(self):
                 },
                 environment="cloud",
                 task=TaskInfo(type="workerpool0", index=1, trial="trial-123"),
-                job=CustomJobSpec(
-                    worker_pool_specs=[
+                job={
+                    "worker_pool_specs": [
```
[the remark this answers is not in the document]
kmontemayor2-sc: we could but since the docs here are wrong I don't think we should be reliant on this field really. 

If anything, I'd prefer to remove the `job` bit all together from our parsed fields.

python/tests/integration/distributed/utils/networking_test.py
kmontemayor2-sc: Thanks, and created https://github.com/Snapchat/GiGL/issues/372

python/tests/unit/distributed/utils/networking_test.py
```diff
@@ -289,3 +295,176 @@ def test_get_internal_ip_from_master_node_fails_if_process_group_not_initialized
...
+
+
+class TestGetGraphStoreInfo(unittest.TestCase):
+    """Test suite for get_graph_store_info function."""
+
+    def tearDown(self):
+        """Clean up after each test."""
+        if dist.is_initialized():
+            dist.destroy_process_group()
+
```
svij-sc:
```suggestion
        for i, info in enumerate(gathered_info):
            self.assertEquals(info, graph_store_info)

```

python/tests/unit/distributed/utils/networking_test.py
```diff
@@ -289,3 +295,176 @@ def test_get_internal_ip_from_master_node_fails_if_process_group_not_initialized
...
+        graph_store_info = get_graph_store_info()
+
+        # Verify the result is a GraphStoreInfo instance
+        assert isinstance(
+            graph_store_info, GraphStoreInfo
+        ), "Result should be a GraphStoreInfo instance"
+        # Verify cluster sizes
+        assert (
+            graph_store_info.num_storage_nodes == storage_nodes
+        ), f"Expected {storage_nodes} storage nodes"
```
svij-sc: nit: use TestCase.assert (self.assert) instead of python asserts - when possible.  Python test engines usually give better error messages when using `TestCase.assert` methods vs. python asserts.

python/tests/unit/common/utils/vertex_ai_context_test.py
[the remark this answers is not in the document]
svij-sc: sure, I am okay removing this.
and marking a comment where we parse it / why we ignore it.

python/tests/unit/distributed/utils/networking_test.py
```diff
@@ -289,3 +294,205 @@ def test_get_internal_ip_from_master_node_fails_if_process_group_not_initialized
...
+        ), "Compute cluster master port should be an integer"
+        assert (
+            graph_store_info.compute_cluster_master_port > 0
+        ), "Compute cluster master port should be positive"
+
+        # Verify all ranks get the same result (since they should all get the same broadcasted values)
+        gathered_info: list[Optional[GraphStoreInfo]] = [None] * world_size
+        dist.all_gather_object(gathered_info, graph_store_info)
+
+        # All ranks should have the same GraphStoreInfo
```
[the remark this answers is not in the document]
svij-sc: thanks for making the change

python/gigl/common/utils/vertex_ai_context.py
kmontemayor2-sc: Sure, removed it.

python/tests/unit/distributed/utils/networking_test.py
kmontemayor2-sc: hmmm, we can't (easily) do that here since we don't have a test cast. (and `TestCase.assert*` are instance methods.)

PR #358, opened 2025-10-07
Add `get_cluster_spec` to get cluster spec on VAI

kmontemayor2-sc:
**Scope of work done**

Doing this per offline discussions to get server/client setup from env variables. I have https://github.com/Snapchat/GiGL/pull/355 as a bigger change but we can break this out so I did :)

### DIFF
--- a/python/gigl/common/utils/vertex_ai_context.py
+++ b/python/gigl/common/utils/vertex_ai_context.py
@@ -188,10 +188,10 @@ class ClusterSpec:
     # https://cloud.google.com/vertex-ai/docs/reference/rest/v1/CustomJobSpec
     job: Optional[CustomJobSpec] = None
 
-    # We use a custom method for parsing, because we need to handle the DictConfig -> Proto conversion
+    # We use a custom method for parsing, because CustomJobSpec is a protobuf message.
     @classmethod
     def from_json(cls, json_str: str) -> "ClusterSpec":
-        """Instantiates ClusterSpec from an OmegaConf DictConfig."""
+        """Instantiates ClusterSpec from a JSON string."""
         cluster_spec_json = json.loads(json_str)
         if "job" in cluster_spec_json and cluster_spec_json["job"] is not None:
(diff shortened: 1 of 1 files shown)


python/tests/unit/common/utils/vertex_ai_context_test.py
```diff
@@ -113,6 +113,112 @@ def test_connect_worker_pool_worker(
...
+            }
+        )
+
+        with patch.dict(
+            os.environ, self.VAI_JOB_ENV | {"CLUSTER_SPEC": cluster_spec_json}
+        ):
+            cluster_spec = get_cluster_spec()
+
+            # Test cluster data
+            self.assertEqual(len(cluster_spec.cluster), 4)
```
svij-sc, line 142: more readable / maintainable alternative:

expected_cluster_spec = ClusterSpec(...)
cluster_spec = get_cluster_spec()

self.assertEqual(cluster_spec, expected_cluster_spec)

If assertions are off we can implement `__eq__` on respective classes.

python/gigl/common/utils/vertex_ai_context.py
```diff
@@ -156,6 +161,85 @@ def connect_worker_pool() -> DistributedContext:
...
+    """
+    if not is_currently_running_in_vertex_ai_job():
+        raise ValueError("Not running in a Vertex AI job.")
+
+    cluster_spec_json = os.environ.get("CLUSTER_SPEC")
+    if not cluster_spec_json:
+        raise ValueError("CLUSTER_SPEC not found in environment variables.")
+
+    try:
+        cluster_spec_data = json.loads(cluster_spec_json)
```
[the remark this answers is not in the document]
kmontemayor2-sc: Sure, it's not so straight forward because `CustomJobSpec` is a proto but this way is cleaner overall.

python/tests/unit/common/utils/vertex_ai_context_test.py
kmontemayor2-sc: Done.

PR #361, opened 2025-10-09
Break up unit tests into separate runners for python and scala

kmontemayor2-sc:
**Scope of work done**

Break up python and scala unit tests into separete runners to speed up tests so we can run them in parallel

Tested with: 
image

I want to test by submitting, if it breaks we can force revert :)


nshah-sc (approved):
thanks!

PR #363, opened 2025-10-16
Add Distributed Model Parallel Tests for models_test.py

swong3-sc:
**Scope of work done**

- Adds a single DMP-related tests to `models_test.py` just to make sure we can use DMP to wrap a model. Forward and gradient flow testing is combined into this one multi-processed test.
- Slight fix, to deal with tensors being cast as awaitable objects, which didn't allow for DMP wrapping

Yes, added one unit test.

### DIFF
--- a/python/tests/unit/nn/models_test.py
+++ b/python/tests/unit/nn/models_test.py
@@ -7,7 +7,9 @@ import torch.multiprocessing as mp
 import torch.nn as nn
 from torch_geometric.data import Data, HeteroData
 from torch_geometric.nn.models import LightGCN as PyGLightGCN
-from torchrec.distributed.model_parallel import DistributedModelParallel as DistributedModelParallel
+from torchrec.distributed.model_parallel import (
+    DistributedModelParallel as DistributedModelParallel,
+)
 
 from gigl.nn.models import LightGCN, LinkPredictionGNN
 from gigl.src.common.types.graph_data import NodeType


python/tests/unit/module/models_test.py
```diff
@@ -283,6 +284,99 @@ def test_gradient_flow(self):
         self.assertIsNotNone(embedding_table.weight.grad)
         self.assertTrue(torch.any(embedding_table.weight.grad != 0))
 
+    def test_dmp_wrapped_model_produces_correct_output(self):
+        """
+        Test that DMP-wrapped LightGCN produces the same output as non-wrapped model. Note: We only test with a single process for unit test.
+        """
+        from torchrec.distributed.model_parallel import DistributedModelParallel as DMP
```
kmontemayor2-sc: nit. import at the top of the file?

python/tests/unit/module/models_test.py
```diff
@@ -283,6 +284,99 @@ def test_gradient_flow(self):
...
+
+        # Initialize distributed
+        if not dist.is_initialized():
+            dist.init_process_group(
+                backend="gloo",
+                init_method="tcp://localhost:29500",
+                rank=0,
+                world_size=1,  # Single process for unit test
+            )
```
kmontemayor2-sc: lets cleanup the process group after every test? Like we do [here](https://github.com/Snapchat/GiGL/blob/0d64f20e361e71c9ac8221caf390a5b67f695f40/python/tests/unit/utils/data_splitters_test.py#L73-L79)

kmontemayor2-sc: This way we can get rid of the `try/catch` here.

python/gigl/module/models.py
```diff
@@ -293,6 +294,10 @@ def _forward_homogeneous(
             key, global_ids
         )  # shape [N_sub, D], where N_sub is number of nodes in subgraph and D is embedding_dim
 
+        # When using DMP, EmbeddingBagCollection returns Awaitable that needs to be resolved
+        if isinstance(embeddings_0, Awaitable):
+            embeddings_0 = embeddings_0.wait()
```
kmontemayor2-sc: It does seem unfortunate/surprising that the rest of our code and/or pyg code doesn't support this type of tensor.

Can we add a TODO to look into this?

python/tests/unit/module/models_test.py
```diff
@@ -283,6 +284,99 @@ def test_gradient_flow(self):
...
+        """
+        Test that gradients flow properly through DMP-wrapped model.
+        """
+        from torchrec.distributed.model_parallel import DistributedModelParallel as DMP
+
+        # Initialize distributed
+        if not dist.is_initialized():
+            dist.init_process_group(
+                backend="gloo",
+                init_method="tcp://localhost:29500",
```
kmontemayor2-sc: let's also use `get_process_group_init_method` [1] so we can always have a free port?

[1]: https://github.com/Snapchat/GiGL/blob/0d64f20e361e71c9ac8221caf390a5b67f695f40/python/tests/unit/distributed/utils/networking_test.py#L224C42-L224C71

python/gigl/module/models.py
swong3-sc: I will add a TODO. This seems like an expected result of TorchRec sharding, as introduced by TorchRec itself.

https://docs.pytorch.org/tutorials/intermediate/torchrec_intro_tutorial.html#gpu-training-with-lazyawaitable

python/tests/unit/module/models_test.py
swong3-sc: Yeah, changed.

python/tests/unit/module/models_test.py
swong3-sc: Thanks, I added a tear down method

python/tests/unit/module/models_test.py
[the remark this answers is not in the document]
swong3-sc: Added a test to do so. I think we should have some discussion when you get back about the nature of this test, ie. use CPU vs CUDA. I went ahead and did a world size of 2, but with CPU, so we weren't really testing the sharding here, just that it works with a larger world size.

python/tests/unit/module/models_test.py
```diff
@@ -283,6 +293,197 @@ def test_gradient_flow(self):
...
+        model = self._create_lightgcn_model(self.num_nodes)
+
+        # Wrap with DMP
+        dmp_model = DMP(
+            module=model,
+            device=self.device,
+        )
+
+        # Set embeddings AFTER DMP wrapping (required for CPU/Gloo)
+        self._set_embeddings(model, "default_homogeneous_node_type")
```
kmontemayor2-sc: for my own understanding, why don't we set these before we wrap? What's the error in this case?

python/tests/unit/module/models_test.py
```diff
@@ -283,6 +293,197 @@ def test_gradient_flow(self):
...
+            fn=_run_dmp_multiprocess_test,
+            args=(
+                world_size,
+                process_group_init_method,
+                self.num_nodes,
+                self.embedding_dim,
+                self.num_layers,
+                self.edge_index,
+                self.test_embeddings,
+                self.expected_output,
```
kmontemayor2-sc: let's add comments for what each of these are, like in https://github.com/Snapchat/GiGL/blob/main/examples/link_prediction/heterogeneous_inference.py#L407-L425

python/tests/unit/module/models_test.py
```diff
@@ -283,6 +293,197 @@ def test_gradient_flow(self):
...
+    @parameterized.expand(
+        [
+            ("world_size_2", 2),
+        ]
+    )
+    def test_dmp_multiprocess(self, _name, world_size):
+        """
+        Test DMP with multiple processes to verify embedding sharding works correctly.
+
+        Note: Uses CPU/Gloo backend for unit testing. For GPU/NCCL testing, see integration tests.
```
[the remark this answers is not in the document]
swong3-sc: Good catch, will remove

python/tests/unit/module/models_test.py
swong3-sc: Added comments for test params

python/tests/unit/module/models_test.py
```diff
@@ -283,6 +293,197 @@ def test_gradient_flow(self):
...
+        """
+        process_group_init_method = get_process_group_init_method()
+        # Initialize distributed
+        if not dist.is_initialized():
+            dist.init_process_group(
+                backend="gloo",
+                init_method=process_group_init_method,
+                rank=0,
+                world_size=1,
+            )
```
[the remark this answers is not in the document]
swong3-sc: sg

python/tests/unit/nn/models_test.py
```diff
@@ -283,6 +293,197 @@ def test_gradient_flow(self):
...
+        model = self._create_lightgcn_model(self.num_nodes)
+
+        # Wrap with DMP
+        dmp_model = DMP(
+            module=model,
+            device=self.device,
+        )
+
+        # Set embeddings AFTER DMP wrapping (required for CPU/Gloo)
+        self._set_embeddings(model, "default_homogeneous_node_type")
```
kmontemayor2-sc: ping :)

python/tests/unit/nn/models_test.py
```diff
@@ -283,6 +293,193 @@ def test_gradient_flow(self):
...
+        self.assertTrue(
+            torch.any(embedding_table.weight.grad != 0),
+            "Gradients should be non-zero",
+        )
+
+    @parameterized.expand(
+        [
+            ("world_size_2", 2),
+        ]
+    )
```
kmontemayor2-sc: nit. we don't need to parameterize here, right?

python/tests/unit/nn/models_test.py
```diff
@@ -283,6 +293,193 @@ def test_gradient_flow(self):
...
+            # Set all embeddings (CPU/Gloo replicates, doesn't shard)
+            table.weight[:] = test_embeddings.to(device)
+
+        # Create test data
+        data = Data(edge_index=edge_index.to(device), num_nodes=num_nodes)
+        data.node = torch.arange(num_nodes, dtype=torch.long, device=device)
+
+        # Forward pass - DMP will fetch embeddings across ranks as needed
+        with torch.no_grad():
+            output = dmp_model(data=data, device=device)
```
kmontemayor2-sc: hmmm shouldn't we test that we can go backwards?

python/gigl/nn/models.py
```diff
@@ -136,6 +137,7 @@ def unwrap_from_ddp(self) -> "LinkPredictionGNN":
 # TODO(swong3): Move specific models to gigl.nn.models whenever we restructure model placement.
 # TODO(swong3): Abstract TorchRec functionality, and make this LightGCN specific
 # TODO(swong3): Remove device context from LightGCN module (use meta, but will have to figure out how to handle buffer transfer)
+# TODO(swong3): Look into why we have to call wait() on the Awaitable returned by the embedding bag collection.
```
kmontemayor2-sc: nit. the TODO should be above the *code* where this belong, e.g. around like 299.

Otherwise, it's hard to tell to other people where this fix should go :)

python/tests/unit/nn/models_test.py
swong3-sc: Correct

python/tests/unit/nn/models_test.py
swong3-sc: Yeah, we already check for this in `test_dmp_gradient_flow` but not in a multi-process env, so I will add it anyway,

python/gigl/nn/models.py
```diff
@@ -293,6 +294,11 @@ def _forward_homogeneous(
             key, global_ids
         )  # shape [N_sub, D], where N_sub is number of nodes in subgraph and D is embedding_dim
 
+        # TODO(swong3): Look into why we have to call wait() on the Awaitable returned by the embedding bag collection.
+        # When using DMP, EmbeddingBagCollection returns Awaitable that needs to be resolved
```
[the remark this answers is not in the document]
nshah-sc: FWIW _if_ you use specialized torchrec train pipelines, like https://github.com/meta-pytorch/torchrec/blob/f758d0800eab052ba7bba89d23a11fe085625f71/examples/golden_training/train_dlrm.py#L144 then you don't need to manually call .wait() etc. because these pipelines manage all the asynchronous stuff under the hood.  however, i don't think our plan is to adopt these pipelines currently so it's not really relevant

python/tests/unit/nn/models_test.py
```diff
@@ -283,6 +292,215 @@ def test_gradient_flow(self):
...
+            init_method=process_group_init_method,
+            rank=0,
+            world_size=1,
+        )
+
+        # Create model
+        model = self._create_lightgcn_model(self.num_nodes)
+
+        # Wrap with DMP
+        dmp_model = DMP(
```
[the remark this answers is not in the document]
nshah-sc: should we test the meta-device sharding setting rather than assuming CPU device sharding here?

python/tests/unit/nn/models_test.py
```diff
@@ -283,6 +292,215 @@ def test_gradient_flow(self):
...
+
+        dmp_model = DMP(
+            module=model,
+            device=self.device,
+        )
+
+        self._set_embeddings(model, "default_homogeneous_node_type")
+
+        dmp_model.train()
+        output = dmp_model(self.data, self.device)
```
nshah-sc: kwargs

python/tests/unit/nn/models_test.py
```diff
@@ -283,6 +292,215 @@ def test_gradient_flow(self):
...
+        # Wrap with DMP - this will shard embeddings across ranks
+        dmp_model = DMP(
+            module=model,
+            device=device,
+        )
+
+        # Set embeddings AFTER DMP wrapping
+        # Note: On CPU/Gloo, DMP doesn't actually shard - it replicates the full table
+        # So we need to set ALL embeddings on each rank
+        with torch.no_grad():
```
nshah-sc: not actually sure you need this no_grad when you're just setting values manually

python/tests/unit/nn/models_test.py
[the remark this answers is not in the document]
swong3-sc: From a prev [PR](https://github.com/Snapchat/GiGL/pull/338/files#:~:text=I%20will%20make%20a%20ToDo%20to%20look%20into%20other%20solutions%20for%20this%2C%20because%20I%20agree%2C%20traditionally%20we%20shouldn%27t%20give%20the%20modules%20context%20for%20a%20device.), I added a TODO for the `meta` device stuff because I couldn't get it to work with some of the buffers we needed to set like `layer_weights`. Specifically, this buffer doesn't get properly materialized when the model is init. on meta device. Then, when we do forward passes, we get a runtime error.

python/tests/unit/nn/models_test.py
swong3-sc: Good call

python/tests/unit/nn/models_test.py
```diff
@@ -283,6 +292,215 @@ def test_gradient_flow(self):
...
+                self.num_layers,  # number of LightGCN layers
+                self.edge_index,  # edge connectivity
+                self.test_embeddings,  # test embedding values
+                self.expected_output,  # expected model output
+            ),
+            nprocs=world_size,
+        )
+
+
+def _run_dmp_multiprocess_test(
```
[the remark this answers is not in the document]
swong3-sc: Yeah, this is a good point, I will just remove the single process tests.

PR #370, opened 2025-10-31
Add Bipartite Implementation for LightGCN

swong3-sc:
**Scope of work done**

- Added Heterogeneous functionality for LightGCN, using `HeteroData` input to `forward`

- Added unit test to check final embeddings for bipartite user-item graph
- Added unit test to check anchor-node functionality for a bipartite user-item graph

### DIFF
--- a/python/gigl/nn/models.py
+++ b/python/gigl/nn/models.py
@@ -416,23 +416,19 @@ class LightGCN(nn.Module):
 
             node_type_to_embeddings_0[node_type] = embeddings
 
-        # LightGCN propagation across node types
-        # Sort node types for deterministic ordering across machines
-        all_node_types = sorted(node_type_to_embeddings_0.keys())
-
         # For heterogeneous graphs, we need to create a unified edge representation
         # Collect all edges and map node indices to a combined space
         # E.g., node type 0 gets indices [0, num_type_0), node type 1 gets [num_type_0, num_type_0 + num_type_1)
         node_type_to_offset: dict[NodeType, int] = {}
         offset = 0
-        for node_type in all_node_types:
+        for node_type in all_node_types_in_data:
             node_type_to_offset[node_type] = offset
             node_type_str = str(node_type)
             offset += data[node_type_str].num_nodes
 
         # Combine all embeddings into a single tensor
         combined_embeddings_0 = torch.cat(
-            [node_type_to_embeddings_0[nt] for nt in all_node_types], dim=0
+            [node_type_to_embeddings_0[nt] for nt in all_node_types_in_data], dim=0
         )  # shape [total_nodes, D]
 
         # Combine all edges into a single edge_index
@@ -471,7 +467,7 @@ class LightGCN(nn.Module):
 
         # Split back into per-node-type embeddings
         final_embeddings: dict[NodeType, torch.Tensor] = {}
-        for node_type in all_node_types:
+        for node_type in all_node_types_in_data:
             start_idx = node_type_to_offset[node_type]
             node_type_str = str(node_type)
             num_nodes = data[node_type_str].num_nodes
@@ -483,7 +479,7 @@ class LightGCN(nn.Module):
         if anchor_node_ids is not None:
             # Only return embeddings for node types specified in anchor_node_ids
             filtered_embeddings: dict[NodeType, torch.Tensor] = {}
-            for node_type in all_node_types:
+            for node_type in all_node_types_in_data:
                 if node_type in anchor_node_ids:
                     anchors = anchor_node_ids[node_type].to(device).long()
                     filtered_embeddings[node_type] = final_embeddings[node_type][anchors]
--- a/python/tests/unit/nn/models_test.py
+++ b/python/tests/unit/nn/models_test.py
@@ -506,8 +506,8 @@ class TestLightGCN(unittest.TestCase):
             anchor_node_ids=anchor_node_ids,
         )
 
-        # Check shapes - should only return embeddings for anchor nodes
-        self.assertEqual(output_with_anchors[NodeType("user")].shape, (1, self.embedding_dim))
+        # Check that only user embeddings are returned
+        self.assertEqual(output_with_anchors.keys(), set([NodeType("user")]))
 
         # Check values - should match the corresponding rows from full output
         self.assertTrue(


python/gigl/module/models.py
```diff
@@ -217,38 +217,48 @@ def forward(
...
+        anchor_node_ids: Optional[Union[torch.Tensor, dict[NodeType, torch.Tensor]]] = None,
     ) -> Union[torch.Tensor, dict[NodeType, torch.Tensor]]:
         """
         Forward pass of the LightGCN model.
 
         Args:
-            data (Union[Data, HeteroData]): Graph data (homogeneous or heterogeneous).
+            data (Union[Data, HeteroData]): Graph data.
+                - For homogeneous: Data object with edge_index and node field
+                - For bipartite: HeteroData with 2 node types and edge_index_dict
```
kmontemayor2-sc: Why are we restricting this to bipartite? Does it get easier if we do?

python/gigl/module/models.py
```diff
@@ -324,6 +337,128 @@ def _forward_homogeneous(
...
+
+        for node_type in output_node_types:
+            node_type_str = str(node_type)
+            key = f"{node_type_str}_id"
+
+            assert hasattr(data[node_type_str], "node"), (
+                f"Subgraph must include .node field for node type {node_type_str}"
+            )
+
+            global_ids = data[node_type_str].node.to(device).long()  # shape [N_type]
```
kmontemayor2-sc: qq why do we call `.long` here?

python/gigl/module/models.py
```diff
@@ -324,6 +337,128 @@ def _forward_homogeneous(
...
+            node_type_str = str(node_type)
+            offset += data[node_type_str].num_nodes
+
+        # Combine all embeddings into a single tensor
+        combined_embeddings_0 = torch.cat(
+            [node_type_to_embeddings_0[nt] for nt in all_node_types], dim=0
+        )  # shape [total_nodes, D]
+
+        # Combine all edges into a single edge_index
+        combined_edge_list = []
```
kmontemayor2-sc: nit. add type.

python/gigl/module/models.py
```diff
@@ -324,6 +337,128 @@ def _forward_homogeneous(
...
+            # Handle DMP Awaitable
+            if isinstance(embeddings, Awaitable):
+                embeddings = embeddings.wait()
+
+            node_type_to_embeddings_0[node_type] = embeddings
+
+        # LightGCN propagation across node types
+        all_node_types = list(node_type_to_embeddings_0.keys())
+
+        # For bipartite, we need to create a unified edge representation
```
kmontemayor2-sc: why do we need to do this? Don't we have separate embedding tables for the different nodes?

Or does the pyg convolution require all nodes be in the same space?

python/gigl/module/models.py
```diff
@@ -324,6 +337,128 @@ def _forward_homogeneous(
...
+            end_idx = start_idx + num_nodes
+
+            final_embeddings[node_type] = combined_final_embeddings[start_idx:end_idx]  # shape [num_nodes, D]
+
+        # Extract anchor nodes if specified
+        if anchor_node_ids is not None:
+            for node_type in all_node_types:
+                if isinstance(anchor_node_ids, dict) and node_type in anchor_node_ids:
+                    anchors = anchor_node_ids[node_type].to(device).long()
+                    final_embeddings[node_type] = final_embeddings[node_type][anchors]
```
kmontemayor2-sc: If we have all node types: `a, b, c` and anchor node ids `{a: [1, 2], b, [3, 4]}` then the final embeddings would still contain all of the embeddings for `c`, right?

Should we fix this? (and also add some test for it?)

python/tests/unit/nn/models_test.py
```diff
@@ -403,6 +403,111 @@ def test_dmp_multiprocess(self, _name, world_size):
...
+            dtype=torch.float32,
+        )
+
+        # Check that bipartite output matches expected
+        self.assertTrue(
+            torch.allclose(output[NodeType("user")], expected_user_embeddings, atol=1e-4, rtol=1e-4)
+        )
+        self.assertTrue(
+            torch.allclose(output[NodeType("item")], expected_item_embeddings, atol=1e-4, rtol=1e-4)
+        )
```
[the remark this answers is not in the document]
swong3-sc: Same reason as [here](https://github.com/Snapchat/GiGL/pull/363#discussion_r2496672956), I need the approximate comparison functionality that `allclose` offers.

python/gigl/module/models.py
swong3-sc: Since we test with longs, this is a no-op, but if we get graph data that is not already of type `int64`, I think we could run into problems, especially with the KJT (maybe?). I just added it to be consistent with the rest of the code, and think we should keep.

python/gigl/module/models.py
swong3-sc: sg

python/gigl/module/models.py
swong3-sc: `LGConv`, unlike other HeteroConvs, doesn't support heterogeneous data, so we'd have to re-implement it. Thus, all nodes need to be in a single unified index space. Additionally, we could run multiple forward passes for each type, but this seems inefficient, as we'd have to aggregate the results anyway.

python/gigl/nn/models.py
```diff
@@ -323,6 +359,132 @@ def _forward_homogeneous(
...
+        data: HeteroData,
+        device: torch.device,
+        output_node_types: Optional[list[NodeType]] = None,
+        anchor_node_ids: Optional[dict[NodeType, torch.Tensor]] = None,
+    ) -> dict[NodeType, torch.Tensor]:
+        """
+        Forward pass for heterogeneous graphs using LightGCN propagation.
+
+        For heterogeneous graphs (e.g., user-item), we have
+        multiple node types. Note that we restrict to one edge type. LightGCN propagates embeddings across
```
kmontemayor2-sc: Do we restrict to one edge type?

swong3-sc: Yeah, it's more of a soft requirement I suppose. Will remove misleading comment

python/gigl/nn/models.py
```diff
@@ -323,6 +359,132 @@ def _forward_homogeneous(
...
+                to local anchor indices. If None, returns all nodes. Default: None.
+
+        Returns:
+            Dict[NodeType, torch.Tensor]: Dict mapping node types to their embeddings,
+                each of shape [num_nodes_of_type, embedding_dim].
+        """
+        # Determine which node types to process
+        if output_node_types is None:
+            # Sort node types for deterministic ordering across machines
+            output_node_types = sorted([NodeType(str(nt)) for nt in data.node_types], key=str)
```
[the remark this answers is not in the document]
swong3-sc: Ok, changed this and other instances of sorted.

python/tests/unit/nn/models_test.py
```diff
@@ -326,6 +326,210 @@ def test_dmp_multiprocess(self):
...
+        # Test with anchor nodes - select specific nodes from each type
+        anchor_node_ids = {
+            NodeType("user"): torch.tensor([0], dtype=torch.long),  # Select user 0
+            NodeType("item"): torch.tensor([1], dtype=torch.long),  # Select item 1
+        }
+
+        output_with_anchors = model(
+            data,
+            self.device,
+            output_node_types=[NodeType("user"), NodeType("item")],
```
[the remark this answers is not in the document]
swong3-sc: I found an issue where if you pass just one output node type, the code errors because it needs to know all the different node types. So I think we should just pass anchor nodes to determine the type of output we get. LMK your thoughts, but I don't see a need to specify the output node types anymore.

python/gigl/nn/models.py
```diff
@@ -134,6 +134,19 @@ def unwrap_from_ddp(self) -> "LinkPredictionGNN":
         return LinkPredictionGNN(encoder=encoder, decoder=decoder)
 
 
+def _get_feature_key(node_type: Union[str, NodeType]) -> str:
```
nshah-sc: i'd suggest a clearer name, e.g. `_build_node_embedding_table_name_by_node_type`.  i think we might actually use something like get_feature_key for other generic functionalities inside GiGL

python/gigl/nn/models.py
```diff
@@ -187,18 +200,29 @@ def __init__(
         )
 
         # Build TorchRec EBC (one table per node type)
-        # feature key naming convention: f"{node_type}_id"
+        # Sort node types for deterministic ordering across machines
         self._feature_keys: list[str] = [
-            f"{node_type}_id" for node_type in self._node_type_to_num_nodes.keys()
+            _get_feature_key(node_type) for node_type in sorted(self._node_type_to_num_nodes.keys())
```
nshah-sc: should this use e.g. SortedDict functionalities as introduced in e.g. #391?   i assume it was introduced for such usecases

python/gigl/nn/models.py
```diff
@@ -187,18 +200,29 @@ def __init__(
...
+            # TODO(kmonte, swong3): We should loosen this restriction and allow fully heterogenous graphs in the future.
+            raise ValueError(
+                f"LightGCN only supports homogeneous (1 node type) or bipartite (2 node types) graphs; "
+                f"got {num_node_types} node types: {self._feature_keys}"
+            )
+
         tables: list[EmbeddingBagConfig] = []
-        for node_type, num_nodes in self._node_type_to_num_nodes.items():
+        # Sort node types for deterministic ordering across machines
+        for node_type, num_nodes in sorted(self._node_type_to_num_nodes.items()):
```
nshah-sc: flagging in case sorteddict should be used here

python/gigl/nn/models.py
```diff
@@ -215,32 +239,44 @@ def forward(
...
+            data (Union[Data, HeteroData]): Graph data.
+                - For homogeneous: Data object with edge_index and node field
+                - For heterogeneous: HeteroData with node types and edge_index_dict
             device (torch.device): Device to run the computation on.
-            output_node_types (Optional[List[NodeType]]): List of node types to return
-                embeddings for. Required for heterogeneous graphs. Default: None.
-            anchor_node_ids (Optional[torch.Tensor]): Local node indices to return
-                embeddings for. If None, returns embeddings for all nodes. Default: None.
+            anchor_node_ids (Optional[Union[torch.Tensor, Dict[NodeType, torch.Tensor]]]):
+                Local node indices to return embeddings for.
```
nshah-sc: just clarifying -- "local" here refers specifically to the local index in the object, which is different from the global node ID, right?

python/gigl/nn/models.py
```diff
@@ -323,6 +359,134 @@ def _forward_homogeneous(
...
+            src_offset = node_type_to_offset[src_node_type]
+            dst_offset = node_type_to_offset[dst_node_type]
+
+            offset_edge_index = edge_index.clone()
+            offset_edge_index[0] += src_offset
+            offset_edge_index[1] += dst_offset
+
+            combined_edge_list.append(offset_edge_index)
+
+        combined_edge_index = torch.cat(combined_edge_list, dim=1)  # shape [2, total_edges]
```
nshah-sc: I'm confused why we need to unify all these edge tensors and node tensors with this offset logic

why can't we use native PyG operations like `message` or `propagate` on the HeteroData or HomoData objects using e.g. [torch_geometric.nn.conv.LGConv](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/lg_conv.html#LGConv)?

I think the important part of your current implementation is that it needs to be able to fetch embeddings at large scale.  I am not sure you need to handroll your own convolution logic for messaging and aggregation too.  The advantage of PyG is that it enables you to not have to do that right?

nshah-sc: >I may be a pyg noob here but my understanding is that [part of] the highlighted code is to build node_type_to_offset so that we can do the one lookup into the torchrec tables and then later reference the embeddings by node type?

the block from line 399-417 already gets all the node embeddings for each node_type by making `len(all_node_types_in_data)` calls to the TorchRec tables IIUC.  

I think we already went through the bother of making these separate calls.  You are right that perhaps they could be batched and fetched in one lookup if we were using a different implementation, but @swong3-sc is relying on `_lookup_embeddings_for_single_node_type` being called multiple times so that isn't happening here.

> this makes sense actually! again a bit of a pyg noob so I'm not sure the "idiomatic" way to solve this problem? Would it be to have some base class and then have subclasses override a pool method on it?

I don't totally know the best "most idiomatic" way here as others have likely read more PyG code than me, but just looking at LGConv's implementation: https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/lg_conv.html#LGConv it seems like this operates on a single tensor of node features `x` (which contains multiple types of nodes all in a single `(num_nodes, F)` tensor, and a single `edge_index` tensor which contains the edges between src and dst nodes.  Presumably this `edge_index` is defined on a single edge type.  we could just apply this on `num_edge_types` in a for loop or something and accumulate the embeddings for each one using some pooling function.  I would probably say that our best bet when building these "torchrec-enhanced modules" is to rely on a pattern of "doing the TorchRec munging to get in PyG format" + "calling PyG" so we don't have to reinvent parts of the wheel that are already built and tested

PR #378, opened 2025-11-12
Update splitter names to have `Dist` prefix

mkolodner-sc:
**Scope of work done**

- Addressing tech debt from this comment https://github.com/Snapchat/GiGL/pull/273#discussion_r2283701400, where it was proposed to rename splitters with a `Dist` prefix so that it is more clear they require distributed setup

### DIFF
--- a/python/gigl/utils/data_splitters.py
+++ b/python/gigl/utils/data_splitters.py
@@ -160,7 +160,7 @@ def _fast_hash(x: torch.Tensor) -> torch.Tensor:
     return x
 
 
-class DistHashedNodeAnchorLinkSplitter:
+class DistNodeAnchorLinkSplitter:
     """Selects train, val, and test nodes based on some provided edge index.
 
     NOTE: This splitter must be called when a Torch distributed process group is initialized.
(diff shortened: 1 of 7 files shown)


mkolodner-sc:
> These names are getting a bit long. WDYT about DistNodeAnchorLinkSplitter? The Hashed part doesn't really matter too much here right?

@kmontemayor2-sc done, thanks!

python/gigl/utils/data_splitters.py
```diff
@@ -500,6 +500,24 @@ def __call__(
...
+        super(HashedNodeAnchorLinkSplitter, self).__init__(*args, **kwargs)
+
+
+class HashedNodeSplitter(DistNodeSplitter):
+    def __init__(self, *args, **kwargs):
+        logger.warning(
+            "gigl.utils.data_splitters.HashedNodeSplitter is deprecated and will be removed in a future release. "
+            "Please use the `gigl.utils.data_splitters.DistNodeSplitter` class instead."
+        )
+        super(HashedNodeSplitter, self).__init__(*args, **kwargs)
```
kmontemayor2-sc: There aren't used anywhere internally? Can we just delete them?

I understand this is """best practice""" but in this case they're literally unused? idk sometimes it's good to be pragmatic.

kmontemayor2-sc (approved):
Thanks Matt!

python/gigl/utils/data_splitters.py
```diff
@@ -500,6 +500,24 @@ def __call__(
             return splits[DEFAULT_HOMOGENEOUS_NODE_TYPE]
 
 
+class HashedNodeAnchorLinkSplitter(DistNodeAnchorLinkSplitter):
+    def __init__(self, *args, **kwargs):
+        logger.warning(
+            "gigl.utils.data_splitters.HashedNodeAnchorLinkSplitter is deprecated and will be removed in a future release. "
+            "Please use the `gigl.utils.data_splitters.DistNodeAnchorLinkSplitter` class instead."
```
svij-sc: Just confirming the follow up to deprecate this is being tracked as tech debt?

PR #380, opened 2025-11-13
Add `compute_cluster_local_world_size` to `VertexAiGraphStoreConfig` and `GraphStoreInfo`

kmontemayor2-sc:
**Scope of work done**

We need to expose this somehow in our code, I am planning on updating our VAI service job launch code to set the env variable in a subsequent PR. 

We need this as we need "number clients" which is client machines x "processes / client machine" when [setting up](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/examples/distributed/server_client_mode/sage_supervised_client.py#L67) [server code](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/examples/distributed/server_client_mode/sage_supervised_server.py#L32) glt server/client mode. 

I want to put this in the config as I feel like it's more centralized and we require the config anyways to launch graph store jobs.
The alternative here is to have some cross/cluster information sharing with torch rpc but I feel like that's more involved, especially as we are going to control the server main (for now). 

### DIFF
--- a/python/gigl/distributed/utils/networking.py
+++ b/python/gigl/distributed/utils/networking.py
@@ -5,7 +5,11 @@ from typing import Optional
 import torch
 
 from gigl.common.logger import Logger
-from gigl.common.utils.vertex_ai_context import ClusterSpec, get_cluster_spec
+from gigl.common.utils.vertex_ai_context import (
+    ClusterSpec,
+    get_cluster_spec,
+    is_currently_running_in_vertex_ai_job,
+)
 from gigl.env.distributed import (
     COMPUTE_CLUSTER_LOCAL_WORLD_SIZE_ENV_KEY,
     GraphStoreInfo,
@@ -207,6 +211,8 @@ def get_graph_store_info() -> GraphStoreInfo:
     """
     # If we want to ever support other (non-VAI) environments,
     # we must switch here depending on the environment.
+    if not is_currently_running_in_vertex_ai_job():
+        raise ValueError("get_graph_store_info must be called in a Vertex AI job.")
     cluster_spec = get_cluster_spec()
     _validate_cluster_spec(cluster_spec)
 
--- a/python/tests/unit/distributed/utils/networking_test.py
+++ b/python/tests/unit/distributed/utils/networking_test.py
@@ -460,7 +460,7 @@ class TestGetGraphStoreInfo(unittest.TestCase):
             get_graph_store_info()
 
         self.assertIn(
-            "Not running in a Vertex AI job",
+            "get_graph_store_info must be called in a Vertex AI job.",
             str(context.exception),
         )
 


python/gigl/distributed/utils/networking.py
```diff
@@ -227,48 +233,18 @@ def get_graph_store_info() -> GraphStoreInfo:
...
         cluster_master_ip=cluster_master_ip,
         storage_cluster_master_ip=storage_cluster_master_ip,
         compute_cluster_master_ip=compute_cluster_master_ip,
         cluster_master_port=cluster_master_port,
         storage_cluster_master_port=storage_cluster_master_port,
         compute_cluster_master_port=compute_cluster_master_port,
     )
-
-
-def _validate_cluster_spec(cluster_spec: ClusterSpec) -> None:
```
svij-sc: any reason we are removing this?

python/gigl/env/distributed.py
```diff
@@ -1,6 +1,11 @@
 """Information about distributed environments."""
 
 from dataclasses import dataclass
+from typing import Final
+
+GRAPH_STORE_PROCESSES_PER_COMPUTE_VAR_NAME: Final[
+    str
+] = "GRAPH_STORE_PROCESSES_PER_COMPUTE"
```
svij-sc: Do we expect to have both `GRAPH_STORE_PROCESSES_PER_COMPUTE` and `GRAPH_STORE_PROCESSES_PER_COMPUTE_VAR_NAME` ?

If not `_VAR_NAME` is redundant - let's drop.

Also, we might need to rename according to above comment.

python/gigl/distributed/utils/networking.py
```diff
@@ -187,11 +195,6 @@ def get_internal_ip_from_all_ranks() -> list[str]:
 def get_graph_store_info() -> GraphStoreInfo:
```
svij-sc: Just noticing this now but this is very specific to `VAI` implementation of GraphStore vs. rest of the functions in this file are not.
Consider erroring out early if VAI context not detected - with comment to support other contexts i.e. local later; or moving this elsewhere? 


Context: One of the line items for next year likely will be to support training / inference on K8s directly - bypassing VAI. It would be good to keep this in mind so we can make tradeoffs when designing here and beyond.

python/gigl/env/distributed.py
kmontemayor2-sc: Sure renamed. I added `_VAR_NAME` to indicate this was the key, not the value, but maybe it's not needed.

Actually, WDYT about `_ENV_KEY` as a suffix?

svij-sc (approved):

python/gigl/env/distributed.py
kmontemayor2-sc: I'd prefer to distinguish between this being some pointer to the value (e.g. they key in os.environ) vs the value itself.

mkolodner-sc (approved):

PR #383, opened 2025-11-13
Update resource config wrapper / validation check to support graph store mode

kmontemayor2-sc:
**Scope of work done**

As mentioned, support VertexAIGraphStore config message in our wrapper and validation check now.

Also slop-up some tests ;)

### DIFF
--- a/python/gigl/src/validation_check/libs/resource_config_checks.py
+++ b/python/gigl/src/validation_check/libs/resource_config_checks.py
@@ -223,36 +223,34 @@ def _validate_machine_config(
         assert_proto_field_value_is_truthy(proto=config, field_name="num_workers")
     elif isinstance(config, gigl_resource_config_pb2.DataflowResourceConfig):
         _check_if_dataflow_resource_config_valid(dataflow_resource_config_pb=config)
-    else:
-        # Case where trainer config is gigl_resource_config_pb2.VertexAiResourceConfig or gigl_resource_config_pb2.KFPResourceConfig
-        if isinstance(config, gigl_resource_config_pb2.VertexAiResourceConfig):
-            _validate_vertex_ai_resource_config(vertex_ai_resource_config_pb=config)
-            _validate_accelerator_type(proto_config=config)
-            _validate_cloud_machine_config(config=config)
-        elif isinstance(config, gigl_resource_config_pb2.VertexAiGraphStoreConfig):
-            _validate_vertex_ai_resource_config(
-                vertex_ai_resource_config_pb=config.graph_store_pool
-            )
-            _validate_accelerator_type(proto_config=config.graph_store_pool)
-            _validate_cloud_machine_config(config=config.graph_store_pool)
+    elif isinstance(config, gigl_resource_config_pb2.KFPResourceConfig):
+        for field in [
+            "cpu_request",
+            "memory_request",
+        ]:
+            assert_proto_field_value_is_truthy(proto=config, field_name=field)
+        _validate_cloud_machine_config(config=config)
+    elif isinstance(config, gigl_resource_config_pb2.VertexAiResourceConfig):
+        _validate_vertex_ai_resource_config(vertex_ai_resource_config_pb=config)
+        _validate_accelerator_type(proto_config=config)
+        _validate_cloud_machine_config(config=config)
+    elif isinstance(config, gigl_resource_config_pb2.VertexAiGraphStoreConfig):
+        _validate_vertex_ai_resource_config(
+            vertex_ai_resource_config_pb=config.graph_store_pool
+        )
+        _validate_accelerator_type(proto_config=config.graph_store_pool)
+        _validate_cloud_machine_config(config=config.graph_store_pool)
 
-            _validate_vertex_ai_resource_config(
-                vertex_ai_resource_config_pb=config.compute_pool
-            )
-            _validate_accelerator_type(proto_config=config.compute_pool)
-            _validate_cloud_machine_config(config=config.compute_pool)
-        elif isinstance(config, gigl_resource_config_pb2.KFPResourceConfig):
-            for field in [
-                "cpu_request",
-                "memory_request",
-            ]:
-                assert_proto_field_value_is_truthy(proto=config, field_name=field)
-            _validate_cloud_machine_config(config=config)
-        else:
-            raise ValueError(
-                f"""Expected distributed config to be one of {gigl_resource_config_pb2.LocalResourceConfig.__name__},
-                {gigl_resource_config_pb2.VertexAiResourceConfig.__name__},
-                or {gigl_resource_config_pb2.KFPResourceConfig.__name__}.
-                or {gigl_resource_config_pb2.VertexAiGraphStoreConfig.__name__}.
-                Got {type(config)}"""
-            )
+        _validate_vertex_ai_resource_config(
+            vertex_ai_resource_config_pb=config.compute_pool
+        )
+        _validate_accelerator_type(proto_config=config.compute_pool)
+        _validate_cloud_machine_config(config=config.compute_pool)
+    else:
+        raise ValueError(
+            f"""Expected distributed config to be one of {gigl_resource_config_pb2.LocalResourceConfig.__name__},
+            {gigl_resource_config_pb2.VertexAiResourceConfig.__name__},
+            or {gigl_resource_config_pb2.KFPResourceConfig.__name__}.
+            or {gigl_resource_config_pb2.VertexAiGraphStoreConfig.__name__}.
+            Got {type(config)}"""
+        )
--- a/python/tests/unit/src/validation/lib/resource_config_checks_test.py
+++ b/python/tests/unit/src/validation/lib/resource_config_checks_test.py
@@ -4,6 +4,7 @@ from gigl.src.validation_check.libs.resource_config_checks import (
     _check_if_dataflow_resource_config_valid,
     _check_if_spark_resource_config_valid,
     _validate_accelerator_type,
+    _validate_machine_config,
     check_if_inferencer_resource_config_valid,
     check_if_preprocessor_resource_config_valid,
     check_if_shared_resource_config_valid,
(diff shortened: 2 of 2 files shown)


python/gigl/src/common/types/pb_wrappers/gigl_resource_config.py
```diff
@@ -35,22 +36,34 @@
 _VERTEX_AI_TRAINER_CONFIG = "vertex_ai_trainer_config"
 _KFP_TRAINER_CONFIG = "kfp_trainer_config"
 _LOCAL_TRAINER_CONFIG = "local_trainer_config"
+_VERTEX_AI_GRAPH_STORE_TRAINER_CONFIG = "vertex_ai_graph_store_trainer_config"
```
mkolodner-sc: More context in this file about differences between `_VERTEX_AI_TRAINER_CONFIG` and `_VERTEX_AI_GRAPH_STORE_TRAINER_CONFIG` would be helpful, ditto for inferencer

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,596 @@
...
+    def test_shared_resource_config_direct(self):
+        """Test loading SharedResourceConfig directly."""
+        config = self._create_gigl_resource_config_with_direct_shared_config()
+        wrapper = GiglResourceConfigWrapper(resource_config=config)
+
+        shared_config = wrapper.shared_resource_config
+        self.assertEqual(shared_config.common_compute_config.project, "test-project")
+        self.assertEqual(shared_config.common_compute_config.region, "us-central1")
+        self.assertEqual(shared_config.resource_labels["env"], "test")
+
```
svij-sc: w/ https://github.com/Snapchat/GiGL/pull/383#discussion_r2525459969 applied

```suggestion
        wrapper = GiglResourceConfigWrapper(resource_config=self._test_resource_config)
        self.assertEqual(wrapper.shared_resource_config, self._test_resource_config.shared_config)
```

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,596 @@
...
+        wrapper = GiglResourceConfigWrapper(resource_config=config)
+        split_gen_config = wrapper.split_generator_config
+
+        assert isinstance(
+            split_gen_config, gigl_resource_config_pb2.SparkResourceConfig
+        )
+        self.assertEqual(split_gen_config.machine_type, "n1-standard-8")
+        self.assertEqual(split_gen_config.num_replicas, 5)
+
+    def test_component_to_shortened_cost_label_map(self):
```
svij-sc: is this a valuable test?

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,596 @@
...
+        config.subgraph_sampler_config.num_local_ssds = 4
+
+        wrapper = GiglResourceConfigWrapper(resource_config=config)
+        sampler_config = wrapper.subgraph_sampler_config
+
+        assert isinstance(sampler_config, gigl_resource_config_pb2.SparkResourceConfig)
+        self.assertEqual(sampler_config.machine_type, "n1-standard-16")
+        self.assertEqual(sampler_config.num_replicas, 10)
+
+    def test_split_generator_config(self):
```
svij-sc: why dont we just create one giant yaml with all the properties filled and we can do a proto / proto assertEquals where possible, backfalling to individual field assertEquals where necessary. For fields like trainer where there are multiple possiblities we can overwrirde the input proto to the wrapper object and still do proto / proto assertEquals

svij-sc (commented):
Some comments.

Generally I am conflicted to see the value of adding all these tests. A lot of these tests  are essentially testing asserts that an object is correctly housing another object it was initialized with.

If anything, there may be more value in testing the config validator raising issues when configs are not correctly formulated - which is higher order and will implicitly capture most tests here.

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
kmontemayor2-sc: Done.

Creating new objects to ensure we don't mutate by accident.

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,596 @@
...
+
+    def test_get_resource_labels_no_component(self):
+        """Test getting resource labels without component replacement."""
+        config = self._create_gigl_resource_config_with_direct_shared_config()
+        wrapper = GiglResourceConfigWrapper(resource_config=config)
+
+        labels = wrapper.get_resource_labels()
+        self.assertEqual(labels["env"], "test")
+        self.assertEqual(labels["cost_resource_group_tag"], "unittest_na")
+        self.assertEqual(labels["cost_resource_group"], "gigl_test")
```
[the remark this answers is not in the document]
kmontemayor2-sc: Done.

svij-sc (approved):
I would prefer that we dont add new* tests to test supposedly deprecated code.

Also, unsure if you had a chance to look at https://github.com/Snapchat/GiGL/pull/383#pullrequestreview-3462404529 yet.
Nonetheless, approving since overall quality is still improving here for our testing suite.
Thanks for the work!

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,596 @@
...
+        wrapper = GiglResourceConfigWrapper(resource_config=config)
+        trainer_config = wrapper.trainer_config
+
+        assert isinstance(
+            trainer_config, gigl_resource_config_pb2.VertexAiGraphStoreConfig
+        )
+        self.assertEqual(trainer_config.compute_pool.machine_type, "n1-highmem-8")
+        self.assertEqual(trainer_config.graph_store_pool.machine_type, "n1-standard-4")
+
+    def test_trainer_config_deprecated_vertex_ai(self):
```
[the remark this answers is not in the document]
kmontemayor2-sc: Hmm, I again I feel like these tests should have always been here, and keeping them here is not that costly.

But you seem to want to remove them so I can do so.

kmontemayor2-sc:
> Generally I am conflicted to see the value of adding all these tests. A lot of these tests are essentially testing asserts that an object is correctly housing another object it was initialized with.

I think that there is actually non-trivial logic in the wrapper, see `trainer_config`,  `get_resource_labels`, and `shared_resource_config`, and that is worth testing.

As for testing the pass through stuff, agreed it's less useful but I don't think it hurts - it may catch a reggression in the future and ime these sorts of unit tests are not that costly to maintain. And if they are costly, we can always delete them :P

> If anything, there may be more value in testing the config validator raising issues when configs are not correctly formulated - which is higher order and will implicitly capture most tests here.

We do this in this PR, too!

python/tests/unit/src/validation/lib/resource_config_checks_test.py
```diff
@@ -0,0 +1,709 @@
...
+    def test_invalid_local_inferencer_config(self):
+        """Test that invalid local inferencer config raises an assertion error."""
+        config = _create_valid_local_inferencer_config()
+        config.inferencer_resource_config.local_inferencer_config.num_workers = 0
+        with self.assertRaises(AssertionError):
+            check_if_inferencer_resource_config_valid(config)
+
+
+class TestVertexAiResourceConfigValidation(unittest.TestCase):
+    """Test suite for Vertex AI resource configuration validation helper."""
```
mkolodner-sc: Is this supposed to be empty?

mkolodner-sc (commented):
Thanks Kyle! Did a pass

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,537 @@
...
+
+    def _create_gigl_resource_config_with_direct_shared_config(
+        self,
+    ) -> gigl_resource_config_pb2.GiglResourceConfig:
+        """Helper to create a GiglResourceConfig with direct SharedResourceConfig."""
+        config = gigl_resource_config_pb2.GiglResourceConfig()
+        config.shared_resource_config.CopyFrom(_create_shared_resource_config())
+        return config
+
+    def _create_gigl_resource_config_with_shared_config_uri(
```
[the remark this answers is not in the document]
kmontemayor2-sc: Sure, removed.

python/tests/unit/src/common/utils/types/pb_wrappers/gigl_resource_config_test.py
```diff
@@ -0,0 +1,537 @@
...
+        self.assertIn("SharedResourceConfig", str(context.exception))
+
+    def test_get_resource_labels(self):
+        """Test getting resource labels without component replacement."""
+        config = self._create_gigl_resource_config_with_direct_shared_config()
+        wrapper = GiglResourceConfigWrapper(resource_config=config)
+
+        # Test without component
+        labels = wrapper.get_resource_labels()
+        expected_labels = {
```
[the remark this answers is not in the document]
kmontemayor2-sc: Sure update these.

For the rest of this file - I don't think it'd be helpful to have the different trainer configs as constants. It's more readable to have them defined in the test.

python/gigl/src/common/types/pb_wrappers/gigl_resource_config.py
kmontemayor2-sc: Eh, frankly I don't think we should be even defining these constants like this. They're used one place each.

And they're only used in this file I don't think it's really helpful to add comments here it'd just clutter the code even more imo.

python/gigl/src/validation_check/libs/resource_config_checks.py
```diff
@@ -216,21 +224,54 @@ def check_if_inferencer_resource_config_valid(
...
+
+
+def _validate_vertex_ai_resource_config(
+    vertex_ai_resource_config_pb: gigl_resource_config_pb2.VertexAiResourceConfig,
+) -> None:
+    """
+    Checks if the provided Vertex AI resource configuration is valid.
+    """
+    assert_proto_field_value_is_truthy(
+        proto=vertex_ai_resource_config_pb, field_name="machine_type"
```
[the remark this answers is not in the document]
kmontemayor2-sc: This is the existing code - we can look ingot making the validation stricter in the future but this PR is already big enough let's no make more changes.

python/gigl/src/validation_check/libs/resource_config_checks.py
```diff
@@ -165,28 +178,23 @@ def check_if_trainer_resource_config_valid(
...
-            assert (
-                trainer_config.gpu_limit == 0
-            ), f"""gpu_limit must be equal to 0 for cpu training, indicated by provided gpu_type {trainer_config.gpu_type}. 
-                Got gpu_limit {trainer_config.gpu_limit}"""
-        else:
-            assert (
-                trainer_config.gpu_limit > 0
-            ), f"""gpu_limit must be greater than 0 for gpu training, indicated by provided gpu_type {trainer_config.gpu_type}. 
-                Got gpu_limit {trainer_config.gpu_limit}. Use gpu_type {AcceleratorType.ACCELERATOR_TYPE_UNSPECIFIED.name} for cpu training."""  # type: ignore
+        if not isinstance(
```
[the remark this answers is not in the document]
kmontemayor2-sc: re-factored a bit, lmk whjat you think!

PR #392, opened 2025-11-20
Add pre-commit validator for gh actions

kmontemayor2-sc:
**Scope of work done**

I found this helpful while developing locally.

We could make this a make/etc rule instead if you'd prefer but IMO that's more setup and having pre-commit do it is convenient :) 


mkolodner-sc (commented):
Can you elaborate more on what this is doing?

PR #394, opened 2025-11-25
Add "server_main.py" for graph store mode

kmontemayor2-sc:
**Scope of work done**

This just adds our (GiGL-specific) server code.

This is kind of a lot of code so it's somewhat barebones. I'll have some follow-up PRs to add our own init_client etc :)

The server code is adapting https://github.com/alibaba/graphlearn-for-pytorch/blob/main/examples/distributed/server_client_mode/sage_supervised_server.py for our own use cases.

NOTE: We only support one process per server for now. I forget why but having multiple processes per server is a bit of a hassle so let's just do one for now :)

### DIFF
--- a/python/gigl/distributed/graph_store/storage_main.py
+++ b/python/gigl/distributed/graph_store/storage_main.py
@@ -21,17 +21,17 @@ logger = Logger()
 
 
 def _run_storage_process(
-    server_rank: int,
+    storage_rank: int,
     cluster_info: GraphStoreInfo,
     dataset: DistDataset,
 ) -> None:
     logger.info(
-        f"Initializing server {server_rank} / {cluster_info.num_storage_nodes } on {cluster_info.cluster_master_ip}:{cluster_info.cluster_master_port}. Cluster rank: {os.environ.get('RANK')}"
+        f"Initializing storage node {storage_rank} / {cluster_info.num_storage_nodes } on {cluster_info.cluster_master_ip}:{cluster_info.cluster_master_port}. Cluster rank: {os.environ.get('RANK')}"
     )
     register_dataset(dataset)
     glt.distributed.init_server(
         num_servers=cluster_info.num_storage_nodes,
-        server_rank=server_rank,
+        server_rank=storage_rank,
         dataset=dataset,
         master_addr=cluster_info.cluster_master_ip,
         master_port=cluster_info.cluster_master_port,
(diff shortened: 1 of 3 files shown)


python/gigl/distributed/graph_store/server_main.py
```diff
@@ -0,0 +1,113 @@
...
+    parser.add_argument("--task_config_uri", type=str, required=True)
+    parser.add_argument("--resource_config_uri", type=str, required=True)
+    parser.add_argument("--is_inference", action="store_true")
+    args = parser.parse_args()
+    logger.info(f"Arguments: {args}")
+
+    is_inference = args.is_inference
+    torch.distributed.init_process_group()
+    cluster_info = get_graph_store_info()
+    server_rank = int(os.environ["RANK"]) - cluster_info.num_compute_nodes
```
svij-sc: The orientation of nodes/ranks is being exposed in another layer.

use dist util instead so all rank computations are contained in a single place?
One option is `GraphStoreInfo` could expose `storage_rank: Optional[int], compute_rank: Optional[int]` ?

python/gigl/distributed/graph_store/server_main.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Suppress TensorFlow logs
+os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"  # isort: skip
+
+import argparse
+
+import graphlearn_torch as glt
+import torch
+
+import gigl.distributed as gd
```
svij-sc: We are introducing a new pattern here of import aliasing our own code.
Maybe better to keep status quo of everything internal use explicit namespaces and not aliasing?

python/gigl/distributed/graph_store/server_main.py
```diff
@@ -0,0 +1,113 @@
...
+    )
+
+    logger.info(
+        f"Waiting for server rank {server_rank} / {cluster_info.num_storage_nodes} to exit"
+    )
+    glt.distributed.wait_and_shutdown_server()
+    logger.info(f"Server rank {server_rank} exited")
+
+
+def run_servers(
```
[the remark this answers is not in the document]
kmontemayor2-sc: Sure done.

python/gigl/distributed/graph_store/server_main.py
kmontemayor2-sc: Hmmm, good point, updated :)

PR #396, opened 2025-11-25
[1/6 Deps Update] Remove Existing Locked Deps.

svij-sc:
Background Context: https://github.com/Snapchat/GiGL/issues/405
Some guidance on stack of PRs: https://github.com/Snapchat/GiGL/issues/405#issuecomment-3603498914

**Scope of PR**

This PR simply removes the existing locked dep files - which will no longer be needed in `uv` world as `uv` maintains its own global lock file: https://docs.astral.sh/uv/concepts/projects/sync/


kmontemayor2-sc:
Actually - is there a reason you don't want to merge this into main? 
image

svij-sc:
> Actually - is there a reason you don't want to merge this into main?

I may need to address conflicts, and need to do final scale testing. Thus, holding off merging into main.

PR #404, opened 2025-12-01
Add `RemoteDistDataset` to operate on the dataset from compute nodes in graph store mode

kmontemayor2-sc:
**Scope of work done**

As we discussed offline - creating these class(es) so that client code can operate on some object instead of calling helper functions.

The purpose of  `RemoteDataset` is to allow client code to access information stored in the storage cluster. See below for example *client* (e.g. user-controlled) code.

```py
def inference_process(local_rank, cluster_info):  
  gigl.distribtued.graph_store.init_client_process(local_rank, cluster_info)
  remote_dataset = RemoteDistDataset(local_rank, cluster_info)
  inference_input: list[Tensor] = remote_dataset.get_node_ids()

  loader = DistNeighborLoader(
     dataset=remote_dataset,
     input_nodes=inference_input,
  )
  for data in loader: ...

# DistNeighbtheorLoader will also use `remote_dataset` to get required info about the dataset
class DistNeighbtheorLoader:
    def __init__(self, dataset: RemoteDataset | DistDataset, ...):
       ...
       self._node_feature_info = dataset.get_node_feature_info() # Needs to be added
       self._edge_feature_info = dataset.get_edge_feature_info() # Needs to be added
```

NOTE: I am *intentionally* making this a different class from `DistDataset` - I don't think we should directly conflate the two, they are different objects and I don't think we should treat them as the same thing (or as a subclass). For the most part, I don't think users are going to need to access the partition books or edge indices, and if they do we can always expose them via this class :)

### DIFF
--- a/python/gigl/distributed/dist_ablp_neighborloader.py
+++ b/python/gigl/distributed/dist_ablp_neighborloader.py
@@ -343,7 +343,7 @@ class DistABLPLoader(DistLoader):
 
             positive_labels, negative_labels = get_labels_for_anchor_nodes(
                 dataset=dataset,
-                node_ids=anchor_node_ids,
+                node_ids=curr_process_nodes,
                 positive_label_edge_type=positive_label_edge_type,
                 negative_label_edge_type=negative_label_edge_type,
             )
--- a/python/gigl/src/common/vertex_ai_launcher.py
+++ b/python/gigl/src/common/vertex_ai_launcher.py
@@ -1,0 +1,301 @@
+"""Shared functionality for launching Vertex AI jobs for training and inference."""
+
+from collections.abc import Mapping
+from typing import Optional
+
+from google.cloud.aiplatform_v1.types import Scheduling, accelerator_type, env_var
+
+from gigl.common import Uri
+from gigl.common.constants import (
+    DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU,
+    DEFAULT_GIGL_RELEASE_SRC_IMAGE_CUDA,
+)
+from gigl.common.logger import Logger
+from gigl.common.services.vertex_ai import VertexAiJobConfig, VertexAIService
+from gigl.env.distributed import COMPUTE_CLUSTER_LOCAL_WORLD_SIZE_ENV_KEY
+from gigl.src.common.constants.components import GiGLComponents
+from gigl.src.common.types.pb_wrappers.gigl_resource_config import (
+    GiglResourceConfigWrapper,
+)
+from snapchat.research.gbml.gigl_resource_config_pb2 import (
+    VertexAiGraphStoreConfig,
+    VertexAiResourceConfig,
+)
+
+logger = Logger()
+
+_LAUNCHABLE_COMPONENTS: frozenset[GiGLComponents] = frozenset(
+    [GiGLComponents.Trainer, GiGLComponents.Inferencer]
+)
+
+
+def launch_single_pool_job(
+    vertex_ai_resource_config: VertexAiResourceConfig,
+    job_name: str,
+    task_config_uri: Uri,
+    resource_config_uri: Uri,
+    process_command: str,
+    process_runtime_args: Mapping[str, str],
+    resource_config_wrapper: GiglResourceConfigWrapper,
+    cpu_docker_uri: Optional[str],
+    cuda_docker_uri: Optional[str],
+    component: GiGLComponents,
+    vertex_ai_region: str,
+) -> None:
+    """Launch a single pool job on Vertex AI.
+
+    Args:
+        vertex_ai_resource_config: The Vertex AI resource configuration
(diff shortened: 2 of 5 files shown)


svij-sc:
cc @mkolodner-sc  for review as well.

python/gigl/distributed/graph_store/remote_dist_dataset.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @property
+    def cluster_info(self) -> GraphStoreInfo:
+        return self._cluster_info
+
+    def get_node_ids(
+        self,
+        node_type: Optional[NodeType] = None,
+    ) -> list[torch.Tensor]:
+        """Get the node IDs for a given node type.
```
svij-sc: Maybe this needs a little more documentation.
As I am understanding is that this function gets node_ids stored for this compute rank across all storage ranks. Is this true?

svij-sc: Maybe a good idea to make it more readable for the next user.

python/gigl/distributed/graph_store/remote_dist_dataset.py
```diff
@@ -0,0 +1,87 @@
...
+
+from gigl.common.logger import Logger
+from gigl.distributed.graph_store.remote_dataset import get_node_ids_for_rank
+from gigl.env.distributed import GraphStoreInfo
+from gigl.src.common.types.graph_data import NodeType
+
+logger = Logger()
+
+
+class RemoteDataset(abc.ABC):
```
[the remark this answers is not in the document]
svij-sc: Hmm... It might be speculative generality i.e. I am not sure if get_node_ids is needed for other implementations of this, et al.

Would pref we dont have this.

python/gigl/distributed/graph_store/remote_dist_dataset.py
```diff
@@ -0,0 +1,87 @@
...
+    @abc.abstractmethod
+    def get_node_ids(self, node_type: Optional[NodeType] = None) -> list[torch.Tensor]:
+        """Get the node IDs for a given node type.
+
+        Args:
+            node_type (Optional[NodeType]): The type of nodes to get.
+        """
+
+
+class RemoteDistDataset:
```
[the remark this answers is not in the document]
svij-sc: sg, follow up.

nit: 

> Yeah I was thinking about this, I was going to rename remote_dataset.py to server_utils.py in a followup pr.

I am okay w/ renaming, but id imagine the server/storage cluster code is closely coupled with this specific remoteDistDataset right?
Maybe we have `RemoteDistDataset` , and `RemoteDistDatasetServer`  if not in same file. ?
This is a nit.

svij-sc (approved):
Id pref to get @mkolodner-sc 's review on the PR too.

python/gigl/distributed/graph_store/remote_dist_dataset.py
```diff
@@ -0,0 +1,157 @@
...
+
+        Returns:
+            The edge direction.
+        """
+        return request_server(
+            self._get_storage_shard(),
+            get_edge_dir,
+        )
+
+    def get_node_ids(
```
mkolodner-sc: What is the plan here if we want to fetch train, val, test node ids?

python/tests/integration/distributed/graph_store/graph_store_integration_test.py
```diff
@@ -83,6 +107,61 @@ def _run_server_processes(
     )
 
 
+def _get_expected_sampler_input(
```
mkolodner-sc: SamplerInput in the function name is confusing since that is conflated with [these](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/distributed/dist_neighbor_loader.py#L96) sort of classes.

mkolodner-sc (commented):
Thanks Kyle! A few comments from me

python/tests/integration/distributed/graph_store/graph_store_integration_test.py
kmontemayor2-sc: Sure, done.

python/gigl/distributed/graph_store/remote_dist_dataset.py
```diff
@@ -0,0 +1,157 @@
...
+)
+from gigl.distributed.utils.networking import get_free_ports_from_master_node
+from gigl.env.distributed import GraphStoreInfo
+from gigl.src.common.types.graph_data import EdgeType, NodeType
+from gigl.types.graph import FeatureInfo
+
+logger = Logger()
+
+
+class RemoteDistDataset:
```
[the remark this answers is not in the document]
kmontemayor2-sc: Re, for graph I was planning on not exposing that if we don't need to.

mkolodner-sc (approved):
Thanks Kyle!

PR #413, opened 2025-12-13
Rename `remote_dataset` to `storage_utils`

kmontemayor2-sc:
**Scope of work done**

As discussed prior - rename this so we can better distinguish between `storage_utils` which is for use on the server side and `remote_dist_dataset` which is meant to be used on the client side.


mkolodner-sc (approved):
Thanks!

PR #426, opened 2025-12-19
Unreferenced partitioned_node_labels fix

swong3-sc:
**Scope of work done**

In the `partition` method of the `DistPartitioner` class, the following statement exists:
`if self._node_feat is not None or self._node_labels is not None:`

If this statement is false, the `else` block doesn't initialize `partitioned_node_labels`, and this causes a variable reference before assignment error when we try to return this value.

No


mkolodner-sc (approved):
Thanks Sam!

PR #434, opened 2026-01-13
Add util to build balanced range partition book

kmontemayor2-sc:
**Scope of work done**

Doing this mainly so I can leverage the exact same logic for the GS integration test :P

We can't use this for the edges, as they are not guaranteed to be evenly balanced, as they are sharded by src node id.

### DIFF
--- a/python/gigl/distributed/graph_store/remote_dist_dataset.py
+++ b/python/gigl/distributed/graph_store/remote_dist_dataset.py
@@ -9,6 +9,7 @@ from gigl.common.logger import Logger
 from gigl.distributed.graph_store.storage_utils import (
     get_edge_dir,
     get_edge_feature_info,
+    get_edge_types,
     get_node_feature_info,
     get_node_ids_for_rank,
 )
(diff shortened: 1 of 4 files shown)


python/tests/unit/distributed/utils/partition_book_test.py
```diff
@@ -0,0 +1,65 @@
...
+        self.assertEqual(get_total_ids(range_pb), 10)
+
+    def test_invalid_tensor_partition_book(self):
+        invalid_pb = torch.tensor([[0, 1], [0, 1]])  # 2D tensor
+        with self.assertRaises(ValueError):
+            get_total_ids(invalid_pb)
+
+
+class TestBuildBalancedRangePartitionBook(unittest.TestCase):
+    def test_divides_evenly(self):
```
mkolodner-sc: Just curious, I see there isn't any use of @ parameterized here where in some cases I might have used it to reduce duplicity. Are we moving away from @ parameterized generally?

PR #436, opened 2026-01-14
Add local integration test for graph store with heterogeneous datasets

kmontemayor2-sc:
**Scope of work done**

Sadly, we still need to disable this on cloud build - I can have a followup so this still runs locally.

I think it's still worth having the code in-place so that we can test this locally during dev. (and by we I mean me)

### DIFF
--- a/python/tests/integration/distributed/graph_store/graph_store_integration_test.py
+++ b/python/tests/integration/distributed/graph_store/graph_store_integration_test.py
@@ -59,7 +59,7 @@ def _assert_sampler_input(
     torch.distributed.barrier()
 
 
-def _run_client_process(
+def _run_compute_tests(
     client_rank: int,
     cluster_info: GraphStoreInfo,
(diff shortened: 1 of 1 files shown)


python/tests/integration/distributed/graph_store/graph_store_integration_test.py
```diff
@@ -166,6 +177,7 @@ def _client_process(
                 i,  # client_rank
                 cluster_info,  # cluster_info
                 mp_sharing_dict,  # mp_sharing_dict
+                node_type,  # node_type
```
mkolodner-sc: I just noticed there is a `_run_client_process` and `_client_process` -- these are a bit hard for a reader to distinguish from each other. Would it be possible to rename and/or add docstring to these functions described their functionality?

kmontemayor2-sc: Sure, updated.

PR #438, opened 2026-01-16
Add get_training_input to storage_utils

kmontemayor2-sc:
**Scope of work done**

Add server-side util so we can remotely fetch the training input. 

Since this is kind of a big PR not adding the client-side equivalent in this one :P

Again this is server-side code, and it's really meant to be called by users.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -34,7 +34,7 @@ extensions = [
 ]
 
 autoapi_type = 'python'
-autoapi_dirs = ['python']
+autoapi_dirs = ['gigl', 'snapchat']
 autoapi_root = "docs/api"
 autoapi_ignore = [
     "*migrations*", # Default value: https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#confval-autoapi_ignore
--- a/docs/user_guide/getting_started/installation.md
+++ b/docs/user_guide/getting_started/installation.md
@@ -8,6 +8,11 @@ These are the current environments supported by GiGL
 | ------ | --------------- | --------- | ---------- | ------- | --- |
 | 3.9    | Partial Support | Supported | 12.1       | 2.5     | 2.5 |
 
+## Available Versions
+
+You can see the available wheels for GiGL
+[here](https://console.cloud.google.com/artifacts/python/external-snap-ci-github-gigl/us-central1/gigl/gigl?project=external-snap-ci-github-gigl)
+
 ## Install Prerequisites - setting up your dev machine
 
 Below we provide two ways to bootstrap an environment for using and/or developing GiGL


python/gigl/distributed/graph_store/storage_utils.py
```diff
@@ -155,3 +161,59 @@ def get_edge_types() -> Optional[list[EdgeType]]:
         return list(_dataset.graph.keys())
     else:
         return None
+
+
+def get_training_input(
+    split: Union[Literal["train", "val", "test"], str],
```
mkolodner-sc: Is this sufficient? What if we want "all" nodes (i.e. dataset.node_ids)

python/tests/test_assets/distributed/utils.py
```diff
@@ -67,3 +69,89 @@ def create_test_process_group() -> None:
...
+
+def destroy_test_process_group() -> None:
+    """
+    Destroys the test process group if it exists.
+    """
+    if torch.distributed.is_initialized():
+        torch.distributed.destroy_process_group()
+
+
+class MockNodeAnchorLinkSplitter:
```
mkolodner-sc: Hmm, I have some concern about the duplicity this is creating, in that having this independent splitter here may add overhead to our codebase in the future if we need to modify the NodeAnchorLinkSplitter to add a property to the class, for example, in that we would have to modify this as well.

Is this class strictly necessary for our case? IIRC in the past we just had the splitting be controlled using some identify function as the hashing function. Would that work in this case?

python/tests/unit/distributed/graph_store/storage_utils_test.py
```diff
@@ -96,6 +107,129 @@ def _create_homogeneous_dataset(self) -> DistDataset:
...
+                _USER: (
+                    torch.tensor(train_user_ids, dtype=torch.int64),
+                    torch.tensor(val_user_ids, dtype=torch.int64),
+                    torch.tensor(test_user_ids, dtype=torch.int64),
+                ),
+            },
+            should_convert_labels_to_edges=False,  # Labels are already edges in graph
+        )
+
+        dataset = DistDataset(rank=0, world_size=1, edge_dir="out")
```
mkolodner-sc: Since it's come up in the past, should we test for multiple edge directions?

mkolodner-sc (commented):
Thanks Kyle! Did a pass here and left some comments/questions

python/gigl/distributed/graph_store/storage_utils.py
```diff
@@ -155,3 +161,59 @@ def get_edge_types() -> Optional[list[EdgeType]]:
...
+    elif split == "val":
+        anchors = _dataset.val_node_ids
+    elif split == "test":
+        anchors = _dataset.test_node_ids
+    else:
+        raise ValueError(f"Invalid split: {split}")
+
+    if isinstance(anchors, torch.Tensor):
+        raise ValueError(
+            f"dataset.node_ids should be a dict[NodeType, torch.Tensor] for getting training input for datasets, got a torch.Tensor for split {split}"
```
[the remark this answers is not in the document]
kmontemayor2-sc: Hmmmm - sorry it's been a while since I looked at the SNC code, do we every distinguish (explicetely) that we're in a SNC setting? If we do then it'd be easy to verify/swap out.

Otherwise, I guess we'd need to look at the `labels` and infer?

Though in conjunction with your other comment, does it sound good to leave a TODO here to support SNC here in a later PR?

python/tests/test_assets/distributed/utils.py
kmontemayor2-sc: > we need to modify the NodeAnchorLinkSplitter to add a property to the class

I really don't think we should be adding properties to the splitters, we should keep them as simple as possible (and dissuading us from doing so is imo a good thing).

but...

> Is this class strictly necessary for our case? IIRC in the past we just had the splitting be controlled using some identify function as the hashing function. Would that work in this case?

I don't think we want identity here but we should be able to use a modified hash_fn to accomplish our goals here - let me do so :)

python/tests/unit/distributed/graph_store/storage_utils_test.py
```diff
@@ -310,6 +444,206 @@ def test_get_edge_types_heterogeneous(self) -> None:
...
+        negative_labels = {
+            0: [2],
+            1: [3],
+            2: [4],
+            3: [0],
+            4: [1],
+        }
+        train_user_ids = [0, 1, 2, 3]
+
+        dataset = self._create_heterogeneous_dataset_with_labels(
```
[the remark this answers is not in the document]
kmontemayor2-sc: > What is the expected behavior if we call get_training_input but specify world size or ranks beyond what was provided by the dataset?

Maybe we should throw - but that's something to investigate in another thread I think.

In this case there may be two different "worlds" here, the storage world (which in this case is size 1) and the compute world (which is size 2).

And when we "fetch" the labels we're fetching for the compute world.

I can clarify this in comments.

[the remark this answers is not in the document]
mkolodner-sc: Gotcha, yeah adding a comment even here that the dataset's world_size is specifically for the storage and the compute's world size is for the compute world would be helpful even here, thanks!

python/gigl/distributed/graph_store/storage_utils.py
kmontemayor2-sc: Hmmmm, I do think that the `tuple[Tensor, Tensor, Tensor | None]` is important for the ABLP.

I guess I could rename this to `get_ablp_input` or something? Would that ameliorate your concerns?

PR #556, opened 2026-03-24
Enable C++ PPR Sampling

mkolodner-sc:
**Scope of work done**

# PPR-Based Neighbor Sampling for GiGL

Adds C++ based PPR (Personalized PageRank) based neighbor sampling as an alternative to k-hop sampling in GiGL's distributed training stack, switching from the previous pythonic implementation.

## C++ Kernel (`gigl-core/csrc/sampling/`)

- Implements the **Forward Push** algorithm (Andersen et al., 2006) as `PPRForwardPushState` — a pybind11-wrapped C++ class that owns all hot-loop state (scores, residuals, queue, neighbor cache).
- Python drives the async RPC neighbor fetches; C++ handles the residual push loop, keeping the asyncio event loop unblocked via `run_in_executor`.
- `gigl-core` is a separately installable package (`uv pip install -e gigl-core/`) with a `py.typed` marker and `.pyi` stub for full mypy coverage.

## Python Sampler (`gigl/distributed/dist_ppr_sampler.py`)

- `DistPPRNeighborSampler` extends `BaseGiGLSampler` and integrates with the existing `DistNeighborLoader` / Graph Store infrastructure.
- **Output format:** `(seed_type, "ppr", neighbor_type)` edge types with `edge_index` (`[2, N]` int64) and `edge_attr` (`[N]` float PPR scores).
- `max_fetch_iterations: Optional[int] = None` caps RPC calls per batch; the algorithm continues to convergence using cached neighbor lists after the budget is exhausted.
- `num_neighbors_per_hop` defaults to `1000` — high-degree hub nodes receive diminishing residual per neighbor, so capping the fetch has negligible effect on PPR accuracy.

## Configuration

- `PPRSamplerOptions` added to `sampler_options.py` alongside existing `KHopNeighborSamplerOptions`; pass via `sampler_options=` on `DistNeighborLoader`.

## C++ Tests (`gigl-core/tests/ppr_forward_push_test.cpp`)

- 6 GoogleTest cases covering: initial queue drain, convergence, score absorption, residual distribution, cross-seed deduplication, and top-k limiting.
- `tests/CMakeLists.txt` updated to auto-link torch and auto-discover kernel sources under `csrc/` (excluding `python_*` entry points).


tests/unit/distributed/dist_ppr_sampler_test.py
```diff
@@ -290,7 +291,7 @@ def _assert_ppr_scores_match_reference(
         for node_id in reference_ppr[ntype_str]:
             ref_score = reference_ppr[ntype_str][node_id]
             sam_score = ntype_to_sampler_ppr[ntype_str][node_id]
-            assert abs(sam_score - ref_score) < 1e-6, (
+            assert abs(sam_score - ref_score) < 1e-5, (
```
mkolodner-sc: This needed to be updating 1e-6 -> 1e-5 for the margin of error. With the C++ conversions, my understanding is we have lost some float precision here, but the 1e-5 margin should still be sufficient for validating that the PPR algorithm is correct here.

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
+#include "ppr_forward_push.h"
```
kmontemayor2-sc, line 1: Do we want to setup some `gigl` namespace?

gigl-core/csrc/sampling/ppr_forward_push.h
```diff
@@ -0,0 +1,76 @@
...
+    // Graph structure (read-only after construction)
+    std::vector<std::vector<int32_t>> _nodeTypeToEdgeTypeIds;
+    std::vector<int32_t> _edgeTypeToDstNtypeId;
+    std::vector<torch::Tensor> _degreeTensors;
+
+    // Per-seed, per-node-type PPR state [seed_idx][ntype_id]
+    std::vector<std::vector<std::unordered_map<int32_t, double>>> _pprScores;
+    std::vector<std::vector<std::unordered_map<int32_t, double>>> _residuals;
+    std::vector<std::vector<std::unordered_set<int32_t>>> _queue;
+    std::vector<std::vector<std::unordered_set<int32_t>>> _queuedNodes;  // snapshot from drainQueue
```
kmontemayor2-sc: can we document what each "vector" is here? (or in the algorithm, I guess.)

A meta note here is that when reading CPP code I look at the header files first, I "feel" like they should be containing most of the documentation, as opposed to the .cpp files (where we should document *what* the algorithm is doing, if it's complicated.

But .h files should have a *why* and also in the case of these nested data structures explain a bit?

gigl-core/core/sampling/ppr_forward_push.h
kmontemayor2-sc, line 1: structure comment: I think if we have a `csrc` folder, it should not contain the header files, if we look at [GLT](https://github.com/alibaba/graphlearn-for-pytorch/tree/main/graphlearn_torch) they have both `csrc/` and `include/`

Image

Alternatively [torch](https://github.com/pytorch/pytorch/tree/main/c10/core) just puts everything into one  folder.

Image

I think I have a slight preference towards the torch style here but I wont push back if you want GLT style. But either way we should probably chose one and stick with it?

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc, line 1: Hmmm I think we should include all headers needed in the cpp files, even if they're also included in the .h files as well.

The Google style guide has  some [argument](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) as for why (let's us safely remove includes)

gigl-core/csrc/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
+#include "ppr_forward_push.h"
+
+// Pack (node_id, etype_id) into a single uint64 for use as a hash key.
+// Inputs are cast through uint32_t to avoid sign-extension of negative int32 values.
+static inline uint64_t packKey(int32_t nodeId, int32_t etypeId) {
```
kmontemayor2-sc: is there a reason we add `inline` here? from my reading `inline` was more intended to so people can define functions in header files. I imagine the compiler will optimize this away either way, see https://godbolt.org/z/s1za57hP4

Note: fine to keep this as-is, just wanted to bring this up as a learning :)

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+    _pprScores.assign(_batchSize, std::vector<std::unordered_map<int32_t, double>>(_numNodeTypes));
+    _residuals.assign(_batchSize, std::vector<std::unordered_map<int32_t, double>>(_numNodeTypes));
+    _queue.assign(_batchSize, std::vector<std::unordered_set<int32_t>>(_numNodeTypes));
+    _queuedNodes.assign(_batchSize, std::vector<std::unordered_set<int32_t>>(_numNodeTypes));
+
+    // accessor<dtype, ndim>() returns a typed view into the tensor's data that
+    // supports [i] indexing with bounds checking in debug builds.
+    auto acc = seedNodes.accessor<int64_t, 1>();
+    _numNodesInQueue = _batchSize;
+    for (int32_t i = 0; i < _batchSize; ++i) {
```
kmontemayor2-sc: should this be `int64_t`? I guess it's *unlikely* to have more than 1b nodes in a batch...

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+    int32_t totalDrainedThisRound = 0;
+    for (int32_t s = 0; s < _batchSize; ++s) {
+        for (int32_t nt = 0; nt < _numNodeTypes; ++nt) {
+            if (_queue[s][nt].empty()) {
+                continue;
+            }
+
+            // Move the live queue into the snapshot (no data copy — O(1)).
+            _queuedNodes[s][nt] = std::move(_queue[s][nt]);
+            _queue[s][nt].clear();
```
kmontemayor2-sc: Hmmm, seems a bit odd to move and then work on the moved variable?  Is there a reason we do that?

gigl-core/csrc/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+        for (int32_t nt = 0; nt < _numNodeTypes; ++nt) {
+            if (_queue[s][nt].empty()) {
+                continue;
+            }
+
+            // Move the live queue into the snapshot (no data copy — O(1)).
+            _queuedNodes[s][nt] = std::move(_queue[s][nt]);
+            _queue[s][nt].clear();
+            totalDrainedThisRound += static_cast<int32_t>(_queuedNodes[s][nt].size());
+            _numNodesInQueue -= static_cast<int32_t>(_queuedNodes[s][nt].size());
```
kmontemayor2-sc: nit. would it be more readable to have some `numDrained = static_cast<>...;` and then use that on these two lines?

gigl-core/csrc/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    _nodesDrainedPerIteration.push_back(totalDrainedThisRound);
+
+    std::unordered_map<int32_t, torch::Tensor> result;
+    for (auto& [eid, nodeSet] : nodesToLookup) {
```
kmontemayor2-sc: `const auto`?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+            }
+        }
+    }
+
+    _nodesDrainedPerIteration.push_back(totalDrainedThisRound);
+
+    std::unordered_map<int32_t, torch::Tensor> result;
+    for (auto& [eid, nodeSet] : nodesToLookup) {
+        std::vector<int64_t> ids(nodeSet.begin(), nodeSet.end());
+        result[eid] = torch::tensor(ids, torch::kLong);
```
kmontemayor2-sc: btw is it possible to directly construct a tensor from the begin /end? Or are we required to also create a vector?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+      _requeueThresholdFactor(requeueThresholdFactor),
+      // std::move transfers ownership of each vector into the member variable
+      // without copying its contents — equivalent to Python's list hand-off
+      // when you no longer need the original.
+      _nodeTypeToEdgeTypeIds(std::move(nodeTypeToEdgeTypeIds)),
+      _edgeTypeToDstNtypeId(std::move(edgeTypeToDstNtypeId)),
+      _degreeTensors(std::move(degreeTensors)) {
+    TORCH_CHECK(seedNodes.dim() == 1, "seedNodes must be 1D");
+    _batchSize = static_cast<int32_t>(seedNodes.size(0));
+    _numNodeTypes = static_cast<int32_t>(_nodeTypeToEdgeTypeIds.size());
```
kmontemayor2-sc, line 39: general question: is there a reason we use the singed types here? instead of `uint32_t` and friends?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+    if (_numNodesInQueue == 0) {
+        return std::nullopt;
+    }
+
+    // Reset the snapshot from the previous iteration.
+    for (int32_t s = 0; s < _batchSize; ++s) {
+        for (auto& qs : _queuedNodes[s]) {
+            qs.clear();
+        }
+    }
```
kmontemayor2-sc: is the size of _quededNodes guaranteed to be the same as _batchSize?

kmontemayor2-sc: Also, I think we can do this in parallel using this pattern https://stackoverflow.com/questions/45069888/c17-c1z-parallel-usage-of-stdfor-each

kmontemayor2-sc: Or maybe some vectorization hints to the compiler? https://blog.minhazav.dev/guide-compiler-to-auto-vectorise/

Though I think we're satisfied with performance rn? So maybe no need to do either of these (and the parallel may be slower actually due to thread overhead.

gigl-core/csrc/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+                // distributed among known neighbors rather than leaking to unfetched
+                // ones (which matters when num_neighbors_per_hop < true_degree).
+                int32_t totalFetched = 0;
+                for (int32_t eid : _nodeTypeToEdgeTypeIds[nt]) {
+                    auto fi = fetched.find(packKey(src, eid));
+                    if (fi != fetched.end()) {
+                        totalFetched += static_cast<int32_t>(fi->second.size());
+                    } else {
+                        auto ci = _neighborCache.find(packKey(src, eid));
+                        if (ci != _neighborCache.end()) {
```
kmontemayor2-sc: sorry what are `fi` and `ci` here? Can we give them better names?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+                }
+
+                double resPerNbr = (1.0 - _alpha) * res / static_cast<double>(totalFetched);
+
+                for (int32_t eid : _nodeTypeToEdgeTypeIds[nt]) {
+                    // Invariant: fetched and _neighborCache are mutually exclusive for
+                    // any given (node, etype) key within one iteration.  drainQueue()
+                    // only requests a fetch for nodes absent from _neighborCache, so a
+                    // key is in at most one of the two.
+                    const std::vector<int32_t>* nbrList = nullptr;
```
kmontemayor2-sc: why do we have a raw pointer here?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+                        if (_queue[s][dstNt].find(nbr) == _queue[s][dstNt].end() &&
+                            _residuals[s][dstNt][nbr] >= threshold) {
+                            _queue[s][dstNt].insert(nbr);
+                            ++_numNodesInQueue;
+
+                            // Promote neighbor lists to the persistent cache: this node will
+                            // be processed next iteration, so caching avoids a re-fetch.
+                            for (int32_t peid : _nodeTypeToEdgeTypeIds[dstNt]) {
+                                uint64_t pk = packKey(nbr, peid);
```
kmontemayor2-sc: Can we use more descript variable names?

gigl-core/csrc/sampling/python_ppr_forward_push.cpp
```diff
@@ -0,0 +1,73 @@
...
+    auto result = self.drainQueue();
+    if (!result) {
+        return py::none();
+    }
+    py::dict d;
+    for (auto& [eid, tensor] : *result) {
+        d[py::int_(eid)] = tensor;
+    }
+    return d;
+}
```
kmontemayor2-sc: is there a reason we name this `self`? Does this follow some common pybind practice?

gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -0,0 +1,73 @@
...
+                      int32_t,
+                      double,
+                      double,
+                      std::vector<std::vector<int32_t>>,
+                      std::vector<int32_t>,
+                      std::vector<torch::Tensor>>())
+        .def("drain_queue", drainQueueWrapper)
+        .def("push_residuals", pushResidualsWrapper)
+        .def("extract_top_k", extractTopKWrapper)
+        .def("get_nodes_drained_per_iteration", &PPRForwardPushState::getNodesDrainedPerIteration);
```
kmontemayor2-sc: likewise is there a reason we don't have `drainQueueWrapper` and friends as member functions on `PPRForwardPushState`, like `        .def("get_nodes_drained_per_iteration", &PPRForwardPushState::getNodesDrainedPerIteration);`?

gigl/distributed/utils/neighborloader.py
```diff
@@ -190,6 +190,10 @@ def strip_non_ppr_edge_types(
     for edge_type in list(data.edge_types):
         if edge_type not in ppr_edge_types:
             del data[edge_type]
+            # num_sampled_edges may not exist at all (e.g. in tests or when GLT doesn't
+            # populate it), and may lack entries for edge types with zero samples.
+            if hasattr(data, "num_sampled_edges"):
+                data.num_sampled_edges.pop(edge_type, None)
```
kmontemayor2-sc: why do we add this? When does GLT not populate this?

tests/unit/distributed/dist_ppr_sampler_test.py
kmontemayor2-sc: It's a little odd we lost precision, have we looked into why?

I think it's fine to update this I'm just concerned if we don't know why.

gigl/distributed/sampler_options.py
```diff
@@ -54,18 +54,27 @@ class PPRSamplerOptions:
...
+            no fetch limit.
     """
 
     alpha: float = 0.5
     eps: float = 1e-4
     max_ppr_nodes: int = 50
-    num_neighbors_per_hop: int = 100_000
+    num_neighbors_per_hop: int = 1_000
     total_degree_dtype: torch.dtype = torch.int32
+    max_fetch_iterations: Optional[int] = None
```
kmontemayor2-sc, line 77: Is this ever used? The robots seem to think we need to plumb this in.

gigl-core/src/gigl_core/py.typed
kmontemayor2-sc, line 1: QQ what does this file do?

gigl-core/csrc/sampling/ppr_forward_push.h
kmontemayor2-sc: what are these int32? Are they node ids? Can we make them `uint32`? just int32 is pretty close to our actual use cases.

gigl-core/csrc/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+    if (t.numel() == 0) {
+        return 0;
+    }
+    TORCH_CHECK(nodeId >= 0,
+                "Node ID ", nodeId, " is negative, which indicates a sampler bug.");
+    TORCH_CHECK(nodeId < static_cast<int32_t>(t.size(0)),
+                "Node ID ", nodeId, " out of range for degree tensor of ntype_id ",
+                ntypeId, " (size=", t.size(0), "). This indicates corrupted graph data or a sampler bug.");
+    // data_ptr<int32_t>() returns a raw C pointer to the tensor's int32 data buffer.
+    return t.data_ptr<int32_t>()[nodeId];
```
kmontemayor2-sc: Robot comment:

undefined memory reads when total_degree_dtype != int32

  getTotalDegree calls t.data_ptr<int32_t>() unconditionally. But
  PPRSamplerOptions.total_degree_dtype is a public, user-facing knob
  (sampler_options.py:76), and its docstring (lines 61-63) explicitly says
  "Use a larger dtype if nodes have exceptionally high aggregate degrees."

  If a user follows that guidance and passes torch.int64, modern PyTorch
  throws on the data_ptr<int32_t>() call with a generic dtype-mismatch
  error (older PyTorch returns garbage data).

  Worse: dist_ppr_sampler.py:198 passes torch.zeros(0, dtype=torch.int32)
  as the fallback for destination-only types regardless of user dtype, so
  even if you fix getTotalDegree for int64, you'd still have a mixed-dtype
  list (_degree_tensors_for_cpp contains int64 real tensors plus int32
  fallbacks).

  Fix options (pick one):
  1. Dispatch on dtype in C++:

  int32_t PPRForwardPushState::getTotalDegree(int32_t nodeId, int32_t ntypeId) const {
      ...
      if (t.scalar_type() == torch::kInt) {
          return t.data_ptr<int32_t>()[nodeId];
      } else if (t.scalar_type() == torch::kLong) {
          // saturate on return — see issue #6 for return-type discussion
          return static_cast<int32_t>(std::min<int64_t>(t.data_ptr<int64_t>()[nodeId], INT32_MAX));
      }
      TORCH_CHECK(false, "Unsupported degree dtype: ", t.scalar_type());
  }

     …and align the destination-only fallback in dist_ppr_sampler.py:198
     to use the user's dtype: torch.zeros(0, dtype=total_degree_dtype).

  2. Remove total_degree_dtype from the public API and hard-pin
  int64 internally. Cleaner, but breaks the option you just exposed.

  The current state is the worst of both worlds — option exposed, ignored.

gigl-core/csrc/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+                        if (ci != _neighborCache.end()) {
+                            totalFetched += static_cast<int32_t>(ci->second.size());
+                        }
+                    }
+                }
+                // Destination-only nodes (or nodes with no fetched neighbors) absorb
+                // residual but do not push further.
+                if (totalFetched == 0) {
+                    continue;
+                }
```
kmontemayor2-sc: Robot review - is it possible this is a cause of us needing to loosen the tolerances in our tests? 

incorrect PPR semantics, not a graceful approximation

  When the fetch budget is exhausted (dist_ppr_sampler.py:401-412),
  push_residuals({}) is called. For any queued node not in _neighborCache,
  totalFetched == 0 here, so the if (totalFetched == 0) continue; short-
  circuits at line 175 — but only AFTER the absorb step at line 152
  (_pprScores[s][nt][src] += res;) has already run.

  That means the residual r is added to the node's PPR score in full, but the
  (1-α)·r mass that should have been distributed to neighbors is silently
  dropped. PPR scores at frontier nodes are biased upward by approximately
  (1-α)/α · r_should_have_pushed.

  The docstring at sampler_options.py:64-69 claims "the algorithm still runs
  to convergence — re-enqueued nodes propagate through cached neighbors at
  negligible cost." That's only true for the cache-hit subset; cache-miss
  queued nodes silently convert residual to PPR score with the wrong scale.

  Fix options:

  1. Reset residual without absorbing, when no neighbors are available
  under fetch-budget exhaustion. Loses information but doesn't fabricate it.
  2. Track a separate "fetch-exhausted" flag on the state and treat absorb +
  no-push as a deliberate truncation — and document the bias in the option's
  docstring with the formula.
  3. Add a regression test: set max_fetch_iterations=1 on a graph wide enough
  that frontier nodes appear, and compare against the unbounded-fetch result.
  Currently no test exercises this code path.

  The current docstring is misleading and should be corrected regardless of
  which fix is chosen.

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // supports [i] indexing with bounds checking in debug builds.
+    auto acc = seedNodes.accessor<int64_t, 1>();
+    _numNodesInQueue = _batchSize;
+    for (int32_t i = 0; i < _batchSize; ++i) {
+        auto seed = static_cast<int32_t>(acc[i]);
+        // PPR initialisation: each seed starts with residual = alpha (the
+        // restart probability).  The first push will move alpha into ppr_score
+        // and distribute (1-alpha)*alpha to the seed's neighbors.
+        _residuals[i][seedNodeTypeId][seed] = _alpha;
+        _queue[i][seedNodeTypeId].insert(seed);
```

silent UB on invalid type IDs

  The constructor at line 45 indexes _residuals[i][seedNodeTypeId][seed] and
  _queue[i][seedNodeTypeId].insert(seed) without verifying
  0 <= seedNodeTypeId < _numNodeTypes. An out-of-range value is undefined
  behavior (vector::operator[] is unchecked).

  Same lack of validation applies in pushResiduals at lines 161, 199, 215
  where _nodeTypeToEdgeTypeIds[nt] and _edgeTypeToDstNtypeId[eid] are
  indexed without checks.

  Fix: add TORCH_CHECK guards in the constructor and at the top of each
  method:

  TORCH_CHECK(seedNodeTypeId >= 0 && seedNodeTypeId < _numNodeTypes,
              "seedNodeTypeId ", seedNodeTypeId, " out of range [0, ",
              _numNodeTypes, ")");

  Apply to constructor, pushResiduals (validate dst types in
  _edgeTypeToDstNtypeId), and any other indexed lookup.

tests/unit/distributed/dist_ppr_sampler_test.py
kmontemayor2-sc: robots have some thoughts about this, if it is floating point ordering then it's fine imo.


```
  The tolerance in both assertions was loosened from 1e-6 (Python original)
  to 1e-5. The justification given in the comments at lines 273-275 and
  375-378 — "theoretical bound is ~1.5e-6, so 1e-5 provides a safety margin"
  — is true for the new C++ implementation, but the Python implementation
  passed at 1e-6 with observed deltas around 1e-7.

  The C++ kernel is supposed to be mathematically equivalent. A 10x increase
  in observed error suggests the floating-point summation order changed —
  likely because unordered_map iteration order differs from Python dict
  insertion order, and the residual push at ppr_forward_push.cpp:178-225
  sums in nondeterministic order.
````

File: gigl-core/csrc/sampling/ppr_forward_push.cpp [hunk 22]
```diff
@@ -0,0 +1,290 @@
...
+std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPushState::extractTopK(
+    int32_t maxPprNodes) {
+    std::unordered_set<int32_t> active;
+    for (int32_t s = 0; s < _batchSize; ++s) {
+        for (int32_t nt = 0; nt < _numNodeTypes; ++nt) {
+            if (!_pprScores[s][nt].empty()) {
+                active.insert(nt);
+            }
+        }
+    }
```
kmontemayor2-sc: Robot review - is it possible this is why we need to update the neighborloader utils?

```
semantic divergence, downstream impact unclear

  The C++ side only emits entries for node types that have nonempty
  _pprScores for at least one seed. The Python translation at
  dist_ppr_sampler.py:425-431 iterates only those keys, so node types
  that were reachable in the schema but unreached in this batch get no
  entry in ntype_to_flat_ids.

  For the hetero path, _sample_from_nodes (line 568) iterates
  ntype_to_flat_ids.items(), so missing keys mean missing PPR edge
  types (seed_type, "ppr", ntype) in the final HeteroData metadata.

  Whether this breaks downstream consumers depends on whether they expect
  those edge types to always be present. The destination-only test
  (dist_ppr_sampler_test.py:_run_ppr_destination_only_node_type) covers
  the case where the destination IS reached in the batch, but not the
  case where a destination type goes unvisited in some batch but should
  still appear.

  Uncertain: this might be intentional, but if downstream relies on a
  stable set of PPR edge types per batch (e.g., for a fixed model
  architecture), this is a regression. Worth verifying with the author
  and adding a test that exercises the unreached-type case.
```

gigl-core/csrc/sampling/python_ppr_forward_push.cpp
```diff
@@ -0,0 +1,73 @@
...
+        auto tup = item.second.cast<py::tuple>();
+        cppMap[eid] = {tup[0].cast<torch::Tensor>(), tup[1].cast<torch::Tensor>(), tup[2].cast<torch::Tensor>()};
+    }
+    // C++ push only uses tensor accessor/data_ptr APIs — GIL-safe to release.
+    // Releasing here lets the asyncio event loop process RPC completion callbacks
+    // from other concurrent PPR coroutines while this push runs.
+    {
+        py::gil_scoped_release release;
+        self.pushResiduals(cppMap);
+    }
```
kmontemayor2-sc: robot review - did we look into using this other approach? 

Manual `gil_scoped_release` rather than `py::call_guard`

  Severity: Medium — style, not correctness

  {
      py::gil_scoped_release release;
      self.pushResiduals(cppMap);
  }

  This is correct — the destructor reacquires the GIL on scope exit,
  including on exception. But pybind11's idiomatic approach is to put
  the guard on def:

  .def("push_residuals", pushResidualsWrapper, py::call_guard<py::gil_scoped_release>())

  The current form requires the dict-conversion loop above (lines 33-37)
  to run holding the GIL, which is correct, but mixes two concerns in
  one wrapper. If you wanted to release earlier, you'd have to convert
  the dict to a C++ map under the GIL anyway, so the current structure
  is defensible.

  Uncertain because this works fine; flagging as a refactor target
  rather than a defect.

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -22,9 +13,12 @@
 from graphlearn_torch.typing import EdgeType, NodeType
 from graphlearn_torch.utils import merge_dict
 
+from gigl.common.logger import Logger
 from gigl.distributed.base_sampler import BaseDistNeighborSampler
 from gigl.types.graph import is_label_edge_type
 
+_logger = Logger()
```
kmontemayor2-sc: Is this unsed anymore?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -152,6 +149,56 @@ def __init__(
...
 
+        # Build integer ID mappings for the C++ forward-push kernel.  String
+        # NodeType / EdgeType keys are only used at the Python boundary
+        # (translating to/from _sample_one_hop); all hot-loop state inside
+        # PPRForwardPushState is indexed by int32 IDs.
+        #
+        # We include both source types (have outgoing edges) and destination-only
+        # types (no outgoing edges, but may accumulate PPR score during the walk)
+        # so the kernel can index residual/ppr_score tables for any node it sees.
+        _all_node_types: list[NodeType] = sorted(
```
kmontemayor2-sc: should we remove these `_underscore_prefixes`?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -204,105 +251,69 @@ def _build_total_degree_tensors(
...
+                    # _sample_one_hop expects None for homogeneous graphs, not the PPR sentinel.
+                    etype=(
+                        self._etype_id_to_etype[eid]
+                        if self._etype_id_to_etype[eid] != _PPR_HOMOGENEOUS_EDGE_TYPE
+                        else None
+                    ),
+                )
+                for eid in eids
+            ]
+        )
```
kmontemayor2-sc: Can we simplify this code? It's kind of hard to read imo

kmontemayor2-sc (commented):
Took a first pass, thanks Matt.

I was leaving a good amount of perf-related comments but we probably don't need to address those.

I do think we should address the file structure and variable names, it's hard to know what `pk` is, for instance.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Done, thanks!

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: Good call, renamed csrc -> core to take the torch style here.

gigl-core/csrc/sampling/ppr_forward_push.cpp
mkolodner-sc: Good call, removed

gigl-core/csrc/sampling/ppr_forward_push.cpp
mkolodner-sc: This isn't the cause of us having a looser tolerance -- this is a known phenomenon when we are doing an early stopping approximation (to minimize the number of fetches we need to do for better throughput). For the tests that have a different tolerance, we aren't using this early stopping feature. 

I've updated the docstring to better explain that this behavior is expected. When this parameter is set, the tradeoff is that we have much better sampling speed at the tradeoff of losing some of the most theoretically correct PPR solution. However, when we set the early stopping by max_fetch_iterations to be large enough, this difference from the theoretical is negligible.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: PyG/GLT represent node and type IDs as signed integers (torch.int32 / torch.int64) throughout. Using int32_t keeps the types consistent with what we receive from tensors, and means bounds checks like nodeId >= 0 are meaningful rather than tautological (an unsigned type can never be negative, hiding bugs where a bad cast produces a large positive value instead).

Added a comment explaining this.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Keeping as int32_t — batch sizes approaching 2B seeds aren't a realistic concern, and all our loop variables are int32_t anyway. Added a comment to make the intent explicit.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes — all four tables are .assign(_batchSize, ...) in the constructor and never resized. Added a comment to the header to make this explicit.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes — std::move transfers ownership of the set's internal data in O(1) rather than copying all node IDs (O(n)), which matters for large queues. The .clear() after is defensive since the standard only guarantees the moved-from container is  "valid but unspecified", not necessarily empty.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: torch::tensor() doesn't accept iterator pairs — it needs a contiguous buffer, so the intermediate vector is required. We could inline it but it doesn't improve readability.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: `nbrList` needs to either point at a neighbor list (from fetched or the cache) or be empty — a raw pointer is the simplest way to express "this might point at something or might be null." It doesn't own or manage any memory, just borrows a creference to an existing vector. Added a comment to make this clear.

gigl-core/csrc/sampling/python_ppr_forward_push.cpp
mkolodner-sc: renamed to `state`

gigl-core/core/sampling/python_ppr_forward_push.cpp
mkolodner-sc: Good catch — drain_queue and extract_top_k are now direct member function pointers; pybind11/stl.h handles the type conversions automatically. push_residuals still needs a wrapper, but only to explicitly release the GIL during the C++ computation so other async coroutines aren't blocked while it runs.

gigl/distributed/utils/neighborloader.py
mkolodner-sc: num_sampled_edges is set by GLT's standard k-hop sampler but not by PPR sampling, which constructs HeteroData manually. Added a comment explaining this.

gigl/distributed/sampler_options.py
mkolodner-sc: Good catch — it was wired up originally but got accidentally dropped at some point. Added back in.

gigl-core/src/gigl_core/py.typed
mkolodner-sc: This is a PEP 561 marker — its presence tells mypy that gigl_core supports type checking. Without it, mypy would ignore the package's annotations and treat all imports from it as Any. Added a note inside the file explaining this.

tests/unit/distributed/dist_ppr_sampler_test.py
mkolodner-sc: The original pure-Python implementation used Python float (float64) for scores end-to-end. The C++ kernel keeps internal scores as double but extractTopK returns float32 tensors to match PyG's edge weight convention — that's should be the new source of precision loss. 1e-5 accounts for it.

gigl-core/csrc/sampling/ppr_forward_push.h
mkolodner-sc: Good point — updated the header to document each level of nesting: outer vector = per-seed, inner vector = per-node-type.

mkolodner-sc: Similar response to https://github.com/Snapchat/GiGL/pull/556#discussion_r3169402760

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Added a TODO to investigate this if performance becomes a blocker

gigl-core/csrc/sampling/python_ppr_forward_push.cpp
mkolodner-sc: Py::call_guard releases the GIL for the entire function, but the dict-conversion loop at the top touches Python objects and must hold the GIL. The manual scoped release is intentional: it holds the GIL for the Python→C++  conversion, then releases it only for the pure C++ push.

gigl-core/csrc/sampling/ppr_forward_push.cpp
mkolodner-sc: Good catch. extractTopK now emits an entry for every node type including unreachable ones (empty tensors, all-zero valid_counts), so downstream model architectures see a consistent set of PPR edge types every batch.

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc: Any reason to not do `for (const auto& queuedNodeTypeCounts: _quededNodes){...}`?

kmontemayor2-sc: Thanks!

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+            TORCH_CHECK(eid >= 0,
+                        "nodeTypeToEdgeTypeIds[", nt, "] contains negative edge type id ", eid, ".");
+            TORCH_CHECK(eid < numEdgeTypes,
+                        "nodeTypeToEdgeTypeIds[", nt, "] contains edge type id ", eid,
+                        " out of range [0, ", numEdgeTypes, ").");
+        }
+    }
+
+    // Allocate per-seed, per-node-type tables.
+    // .assign(n, val) fills a vector with n copies of val — like [val] * n in Python.
```
kmontemayor2-sc: nit. It's more like `[val for _ in range(n)]`. `[val] * 10` has the same object at all indices.


```
>>> a = [{}] * 10
>>> a[0]["foo"] = "bar"
>>> a
[{'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}, {'foo': 'bar'}]
```

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    // nodesToLookup[eid] = set of node IDs that need a neighbor fetch for
+    // edge type eid this round.  Using a set deduplicates nodes that appear
+    // in multiple seeds' queues: we only fetch each (node, etype) pair once.
+    std::unordered_map<int32_t, std::unordered_set<int32_t>> nodesToLookup;
+
+    int32_t totalDrainedThisRound = 0;
+    for (int32_t s = 0; s < _batchSize; ++s) {
+        for (int32_t nt = 0; nt < _numNodeTypes; ++nt) {
```
kmontemayor2-sc: Can we give these loop vars better names too? `batchNode` and `nodeType` maybe? `nt` is better but `s` is kinda magical (in a bad way)

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+            // The explicit clear() after move is defensive: the standard only guarantees
+            // a moved-from container is "valid but unspecified", not necessarily empty.
+            _queuedNodes[s][nt] = std::move(_queue[s][nt]);
+            _queue[s][nt].clear();
+            auto numDrained = static_cast<int32_t>(_queuedNodes[s][nt].size());
+            totalDrainedThisRound += numDrained;
+            _numNodesInQueue -= numDrained;
+
+            for (int32_t nodeId : _queuedNodes[s][nt]) {
+                for (int32_t eid : _nodeTypeToEdgeTypeIds[nt]) {
```
kmontemayor2-sc: this is like `edgeTypeID` right? or `etid`? we don't have a concept of edge ids in this code

kmontemayor2-sc: ditto this is `nodeTypeID`?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+        // Walk the flat neighbor list, slicing out each node's neighbors using
+        // the running offset into the concatenated flat buffer.
+        int64_t offset = 0;
+        for (int64_t i = 0; i < nodeIdsT.size(0); ++i) {
+            auto nid = static_cast<int32_t>(nodeAcc[i]);
+            int64_t count = cntAcc[i];
+            std::vector<int32_t> nbrs(count);
+            for (int64_t j = 0; j < count; ++j) {
+                nbrs[j] = static_cast<int32_t>(nbrAcc[offset + j]);
```
kmontemayor2-sc: hmmm, is it possible that casting from int64_t (e.g. `auto nbrAcc = flatNbrsT.accessor<int64_t, 1>();` to the int32_t looses data here? I guess it'd require for us to have more than 2b fetched nodes in a batch?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+        }
+    }
+
+    // Step 2: For every node that was in the queue (captured in _queuedNodes
+    // by drainQueue()), apply one PPR push step:
+    //   a. Absorb residual into the PPR score.
+    //   b. Distribute (1-alpha) * residual equally to each neighbor.
+    //   c. Enqueue any neighbor whose residual now exceeds the requeue threshold.
+    for (int32_t s = 0; s < _batchSize; ++s) {
+        for (int32_t nt = 0; nt < _numNodeTypes; ++nt) {
```
kmontemayor2-sc: These variables are also hard to understand.

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc: why not use `std::optional` here? If we don't need the raw pointer semantics let's avoid using them?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+                    if (!nbrList || nbrList->empty()) {
+                        continue;
+                    }
+
+                    int32_t dstNt = _edgeTypeToDstNtypeId[eid];
+
+                    // c. Accumulate residual for each neighbor and re-enqueue if threshold
+                    // exceeded.
+                    for (int32_t nbr : *nbrList) {
+                        _residuals[s][dstNt][nbr] += resPerNbr;
```
kmontemayor2-sc: ditto on the vars, `_residuals[s][dstNt][nbr] += resPerNbr;` is very hard to understand without a lot of context.

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc: not resolved?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPushState::extractTopK(
+    int32_t maxPprNodes) {
+    std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> result;
+    // Emit an entry for every node type, even if unreachable in this batch (empty tensors,
+    // all-zero valid_counts).  This keeps the output shape consistent across batches so
+    // downstream model architectures see a fixed set of PPR edge types every iteration.
+    for (int32_t nt = 0; nt < _numNodeTypes; ++nt) {
+        std::vector<int64_t> flatIds;
+        std::vector<float> flatWeights;
```
kmontemayor2-sc: hmmm any reason to use float not double here? Maybe part of the reason we saw the need to loosen the tolerances earlier?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,343 @@
...
+                      torch::tensor(flatWeights, torch::kFloat),
+                      torch::tensor(validCounts, torch::kLong)};
+    }
+    return result;
+}
+
+int32_t PPRForwardPushState::getTotalDegree(int32_t nodeId, int32_t ntypeId) const {
+    if (ntypeId >= static_cast<int32_t>(_degreeTensors.size())) {
+        return 0;
+    }
```
kmontemayor2-sc: should this be an error?

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -0,0 +1,105 @@
...
+    //
+    // _pprScores:   absorbed PPR mass — the final output scores written by pushResiduals().
+    // _residuals:   unabsorbed mass waiting to be pushed to neighbors.
+    // _queue:       nodes whose residual exceeds the requeue threshold; drained each iteration.
+    // _queuedNodes: snapshot of _queue captured by drainQueue() so pushResiduals() knows
+    //               which nodes to process in the current iteration.
+    std::vector<std::vector<std::unordered_map<int32_t, double>>> _pprScores;
+    std::vector<std::vector<std::unordered_map<int32_t, double>>> _residuals;
+    std::vector<std::vector<std::unordered_set<int32_t>>> _queue;
+    std::vector<std::vector<std::unordered_set<int32_t>>> _queuedNodes;
```
kmontemayor2-sc: hmmmm, since we generally reference to these things together, would it make sense to create some new class that contains `map<int, double>, map<int, double>, set, set` and then we just have a double vector of those? 

I bring this up because it's a bit hard to read through and understand the code, though maybe better variable names could be useful here.

Side node, if we access all these things together then we may get better per if we don't need to keep re-indexing into different arrays.

gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <torch/extension.h>
+
+#include <cstdint>
+#include <tuple>
+#include <unordered_map>
+
+#include "ppr_forward_push.h"
+
+namespace py = pybind11;
+using gigl::PPRForwardPushState;
```
kmontemayor2-sc: we could just `namespace gigl` here right? Any reason we couldn't

gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+// pushResiduals: a wrapper is needed solely to release the GIL during the C++ push.
+// pybind11/stl.h handles all type conversions automatically; the other methods use
+// direct member function pointers for the same reason.
+static void pushResidualsWrapper(PPRForwardPushState& state, const py::dict& fetchedByEtypeId) {
+    std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> cppMap;
```
kmontemayor2-sc: can we give the map a better name?

gigl-core/Makefile
kmontemayor2-sc: Should we remove these chcks now?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -1,18 +1,9 @@
...
-# TODO (mkolodner-sc): Investigate whether concurrency for _sample_one_hop and _compute_ppr_scores will
-# yield performance benefits.
-
-import heapq
+import asyncio
 from collections import defaultdict
 from typing import Optional, Union
 
 import torch
+from gigl_core import PPRForwardPushState
```
kmontemayor2-sc: hmmm should we introduce this as part of some `gigl.core` namespace?

can leave as a TODO for later.

tests/unit/distributed/dist_ppr_sampler_test.py
kmontemayor2-sc: Wait did we use to always use float64 / double? Or never convert to float?

Makefile
```diff
@@ -22,7 +22,7 @@ DOCKER_IMAGE_MAIN_CPU_NAME_WITH_TAG?=${DOCKER_IMAGE_MAIN_CPU_NAME}:${DATE}
 DOCKER_IMAGE_DEV_WORKBENCH_NAME_WITH_TAG?=${DOCKER_IMAGE_DEV_WORKBENCH_NAME}:${DATE}
 
 PYTHON_DIRS:=.github/scripts examples gigl tests snapchat scripts
-CPP_SOURCES:=$(shell find gigl-core/csrc \( -name "*.cpp" -o -name "*.cu" \) 2>/dev/null)
```
kmontemayor2-sc, line 25: We don't need these vars in top-level MAKEFILE anymore right?

gigl/distributed/utils/neighborloader.py
kmontemayor2-sc: Hmmm, why did this only surface now?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -152,6 +147,56 @@ def __init__(
...
+        # types (no outgoing edges, but may accumulate PPR score during the walk)
+        # so the kernel can index residual/ppr_score tables for any node it sees.
+        all_node_types: list[NodeType] = sorted(
+            {nt for nt in self._node_type_to_edge_types}
+            | {
+                self._get_destination_type(et)
+                for etypes in self._node_type_to_edge_types.values()
+                for et in etypes
+            }
+        )
```
kmontemayor2-sc, line 162: this bit is also a little complicated and hard to read, do you think it'd be easier to have some individual variables.

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -152,6 +147,56 @@ def __init__(
...
+        # dict.fromkeys preserves insertion order while deduplicating.
+        all_edge_types: list[EdgeType] = list(
+            dict.fromkeys(
+                et for etypes in self._node_type_to_edge_types.values() for et in etypes
+            )
+        )
```
kmontemayor2-sc: should this be sorted?

gigl/distributed/utils/neighborloader.py
mkolodner-sc: The `num_sampled_edges` having an extra edge type is not a functionality error but moreso an inconsistency, since it's not used by PyG and is moreso there for debugging purposes by GLT. I noticed this inconsistency explicitly when I was running my experiments and logging all the fields of the resulting HeteroData object. Can add a test to the PPR sampler to ensure that this is consistent with the rest of the HeteroData object.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Good catch — fixed to `[val for _ in range(n)]`.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes — `_queuedNodes` is `.assign(_batchSize, ...)` in the constructor so the size is guaranteed. Switched to range-for. Note: needs `auto&` not `const auto&` since we call `.clear()` on the inner sets.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Renamed throughout: `s` → `seedIdx`, `nt` → `nodeTypeId`, `eid` → `edgeTypeId`, `dstNt` → `dstNodeTypeId`, `peid` → `neighborEdgeTypeId`, `pk` → `packedKey`.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Renamed throughout: `s` → `seedIdx`, `nt` → `nodeTypeId`, `eid` → `edgeTypeId`, `dstNt` → `dstNodeTypeId`, `peid` → `neighborEdgeTypeId`, `pk` → `packedKey`.

mkolodner-sc: Renamed throughout: `s` → `seedIdx`, `nt` → `nodeTypeId`, `eid` → `edgeTypeId`, `dstNt` → `dstNodeTypeId`, `peid` → `neighborEdgeTypeId`, `pk` → `packedKey`.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Renamed throughout: `s` → `seedIdx`, `nt` → `nodeTypeId`, `eid` → `edgeTypeId`, `dstNt` → `dstNodeTypeId`, `peid` → `neighborEdgeTypeId`, `pk` → `packedKey`.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Renamed throughout: `s` → `seedIdx`, `nt` → `nodeTypeId`, `eid` → `edgeTypeId`, `dstNt` → `dstNodeTypeId`, `peid` → `neighborEdgeTypeId`, `pk` → `packedKey`.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Resolved — `peid` → `neighborEdgeTypeId`, `pk` → `packedKey`.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: The relevant bound is partition size < INT32_MAX (~2.1B nodes), not batch size — so this is safe in practice. Happy to add a TORCH_CHECK on `nid` if you'd prefer an explicit guard, but skipping it in the inner neighbor loop to avoid hot-path overhead.

mkolodner-sc: Yes, that is not a realistic concern - if we have more than 2b fetched nodes there are other problems to be concerned about 😆

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Sure, used `std::optional<std::reference_wrapper<const std::vector<int32_t>>>` — optional can't hold a reference directly so `std::cref` wraps it. Access becomes `nbrList->get()` instead of `*nbrList`.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Sure, switched to a double, although this isn't sufficient to fix the tolerance issue (it still appears with this change).

tests/unit/distributed/dist_ppr_sampler_test.py
mkolodner-sc: Hmm, I believe we used to use doubles here. I've investigated this a bit further and tried changing all the floats to doubles in the cpp code, but it still is not within the 1e-6 tol. The robots came up with a potential reason similar to yours above:

```
The delta is purely a consequence of std::unordered_set iterating in a different order than CPython's set for small integer node IDs — different processing order within each push step leads to a different (but valid) residual distribution at convergence. 
```

An example of this:
```
  Iteration 2 queue = {1, 2}, both with residual 0.125                                                                                                                                                                                              
   
  - Order: process 1 before 2 — when 1 pushes, it adds residual to 2 (bumping 2 to 0.15625). Then 2 pushes with 0.15625, so ppr[2] = 0.15625.                                                                                                       
  - Order: process 2 before 1 — 2 pushes first with 0.125, so ppr[2] = 0.125. The extra 0.03125 from node 1 lands back in 2's residual for a later round.
```                                                                                                                                                                                                                                           
Both paths converge to the same true PPR eventually, but when convergence is declared (residuals just drop below threshold), the absorbed ppr_hat values can differ by up to the remaining residual magnitude. C++ unordered_set ordering for int keys differs from CPython (both libraries have undefined behavior for ordering of an unordered object), giving a different residual distribution at termination — hence the ~1.38e-6 delta.  

Given the theoretical bound is 1.5 e-6, we should probably have been setting it to 2e-6 in the first place. Running the tests with this leads to success.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Good call, updated.

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: Done, thanks!

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Added a TODO for this

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+namespace gigl {
+
+// Pack (node_id, etype_id) into a single uint64 for use as a hash key.
+// Inputs are cast through uint32_t to avoid sign-extension of negative int32 values.
+static uint64_t packKey(int32_t nodeId, int32_t edgeTypeId) {
+    return (static_cast<uint64_t>(static_cast<uint32_t>(nodeId)) << 32) | static_cast<uint32_t>(edgeTypeId);
+}
+
+PPRForwardPushState::PPRForwardPushState(const torch::Tensor& seedNodes,
```
kmontemayor2-sc: naming nit: I don't think `PPRForwardPushState` is really about "state" is it" It's more like, the tool that runs forward push? It has some state, but I'm not sure that naming it "state" is the best here?

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -0,0 +1,102 @@
...
+#include <optional>
+#include <tuple>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace gigl {
+
+// Per-seed, per-node-type PPR algorithm state.
+// Grouping all four tables into one struct keeps related data co-located in memory:
```
kmontemayor2-sc: I don't think this is true, my understanding is that `unordered_map` stores a pointer to some memory, so the pointers would be colocated, not the actual data.

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -0,0 +1,102 @@
...
+    //   outer vector  [_batchSize]    — one entry per seed node in the batch
+    //   inner vector  [_numNodeTypes] — one SeedNodeTypeState per node type
+    //
+    // int32_t is used for node and type IDs throughout to match PyG/GLT's signed-integer
+    // convention (torch.int32 / torch.int64).  Signed types also make nodeId >= 0 checks
+    // meaningful — an unsigned type would make that guard tautological.
+    //
+    // Sized [_batchSize][_numNodeTypes] at construction and never resized,
+    // so [seedIdx][nodeTypeId] indexing is always safe within the loop bounds.
+    std::vector<std::vector<SeedNodeTypeState>> _state;
```
kmontemayor2-sc, line 99: btw is there a reason we use the map<uint64_t` for _neighborCache and then the `vector<vector>>`?

It seems like we could also pack _state as `(seedNodeId, NodeTypeId)`?

kmontemayor2-sc: Might make the code more understandable if we are more consistent here :)

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+                for (int32_t edgeTypeId : _nodeTypeToEdgeTypeIds[nodeTypeId]) {
+                    // Invariant: fetched and _neighborCache are mutually exclusive for
+                    // any given (node, etype) key within one iteration.  drainQueue()
+                    // only requests a fetch for nodes absent from _neighborCache, so a
+                    // key is in at most one of the two.
+                    // Neighbor list for this (src, edgeTypeId) pair, if one exists —
+                    // either from `fetched` (new this iteration) or `_neighborCache`
+                    // (seen in a previous iteration).
+                    std::optional<std::reference_wrapper<const std::vector<int32_t>>> neighborList;
```
kmontemayor2-sc, line 241: hmmm, why don't we just use `std::optional<std::vector<>` here? Do we need a `const vector`?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -0,0 +1,357 @@
...
+    // all-zero valid_counts).  This keeps the output shape consistent across batches so
+    // downstream model architectures see a fixed set of PPR edge types every iteration.
+    for (int32_t nodeTypeId = 0; nodeTypeId < _numNodeTypes; ++nodeTypeId) {
+        std::vector<int64_t> flatIds;
+        std::vector<double> flatWeights;
+        std::vector<int64_t> validCounts;
+
+        for (int32_t seedIdx = 0; seedIdx < _batchSize; ++seedIdx) {
+            const auto& scores = _state[seedIdx][nodeTypeId].pprScores;
+            int32_t topK = std::min(maxPprNodes, static_cast<int32_t>(scores.size()));
```
kmontemayor2-sc, line 303: Can we add a note somewhere in these file that all of this code will break if we ever get > 2b nodes / batch?

gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -0,0 +1,58 @@
...
+    std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> neighborTensorsByEtypeId;
+    // Dict iteration touches Python objects — GIL must be held here.
+    for (auto item : fetchedByEtypeId) {
+        auto edgeTypeId = item.first.cast<int32_t>();
+        auto neighborTensors = item.second.cast<py::tuple>();
+        neighborTensorsByEtypeId[edgeTypeId] = {neighborTensors[0].cast<torch::Tensor>(),
+                                                neighborTensors[1].cast<torch::Tensor>(),
+                                                neighborTensors[2].cast<torch::Tensor>()};
+    }
+    // C++ push only uses tensor accessor/data_ptr APIs — GIL-safe to release.
```
kmontemayor2-sc, line 33: We also have a hard requirement here that noone else modifies the input `fetchedByEtypeId` right? Let's document that

Makefile
```diff
@@ -178,7 +178,7 @@ check_lint_cpp: build_cpp_extensions
 # WarningsAsErrors: '*', so the warning must be silenced at the compiler level before
 # clang-tidy ever sees it.
 fix_lint_cpp: build_cpp_extensions
-	$(if $(CPP_SOURCES_NO_CUDA),clang-tidy-15 --fix --extra-arg=-Wno-ignored-optimization-argument -p gigl-core/.cache/cmake_build/compile_commands.json $(CPP_SOURCES_NO_CUDA))
+	clang-tidy-15 --fix --extra-arg=-Wno-ignored-optimization-argument -p gigl-core/.cache/cmake_build/compile_commands.json $(CPP_SOURCES_NO_CUDA)
```
kmontemayor2-sc: I guess why not delegate teh `uv run ... `and `clang-tidy` checks to the Makefile in gigl-core?

kmontemayor2-sc (commented):
Thanks Matt! just nits now the only thing that's a bit odd is why we can't do `std::optional<std::vector>`

kmontemayor2-sc (approved):
Thanks Matt! just nits now the only thing that's a bit odd is why we can't do `std::optional<std::vector>`

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: The difference is intentional here. 

  - _state keys (seedIdx, nodeTypeId) are dense, sequential integers bounded at construction — the natural fit is a vector                                                                                                                          
  - _neighborCache keys (nodeId, edgeTypeId) have sparse nodeId from a large graph — the natural fit is a hash map       

Making the _neighborCache conform to a vector would incur unnecessary memory, while updating _state to be a hash map will incur a hashing overhead for what is currently a O(1) vector operation.

Updated the documentation to reflect this.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Discussed offline, we need a `reference_wrapper` or else it will copy memory from the cache or fetched list into the neighborList object, which hurts the performance.

yliu2-sc (approved):
Second stamp

PR #558, opened 2026-03-25
Setup C++ Infra in GiGL

mkolodner-sc:
**Scope of work done**

# C++ Infrastructure for GiGL

This PR adds the foundational C++ infrastructure to GiGL — everything needed to write, build, lint, and test C++ extensions alongside the existing Python codebase, and the release pipeline to ship them.

## Two-wheel model

GiGL is now distributed as two wheels installed together:

- **gigl** — pure Python (same wheel for CPU and CUDA users)
- **gigl-core** — compiled C++/CUDA pybind11 extensions, ABI-bound to the torch variant. Currently contains no C++ sources; the package builds an empty wheel now and gets populated as extensions are added.

Both wheels are versioned and released together. `gigl-core` is a dependency of `gigl` and is resolved automatically from the same GCP registry — no separate install step for users.

## Build system

`gigl-core/` is a standalone scikit-build-core + CMake package. `CMakeLists.txt` auto-discovers all `csrc/python_*.cpp` (and `python_*.cu`) files as pybind11 extension modules — adding a new extension requires only dropping source files, no CMake edits. pybind11 is fetched via `FetchContent` so it does not need to be pre-installed. Torch is located by asking the target Python directly (`torch.utils.cmake_prefix_path`).

`no-build-isolation-package = ["gigl-core"]` in `pyproject.toml` disables build isolation for `gigl-core` so cmake can find torch (not on PyPI) from the ambient venv when real extensions are added. The `build-backend` dependency group provides scikit-build-core as the PEP 517 build backend in both dev and Docker environments.

## Release pipeline

`release.yml` builds and publishes both wheels via a matrix build — one CPU runner, one GPU runner — to two GCP Artifact Registry repos (`gigl` for CPU, `gigl-cu128` for CUDA 12.8). `RELEASING.md` documents the two-workflow release process: *Create Release* (version bump, branch, tag, PR) followed by *Release GiGL* (wheel build and publish).

## C++ tooling

- **Formatting** — `.clang-format` establishes style conventions. `make format_cpp` formats in-place.
- **Linting** — `.clang-tidy` config. `make check_lint_cpp` runs `clangd --check` in parallel (faster than `clang-tidy` due to preamble caching). `make fix_lint_cpp` applies auto-fixable violations via `clang-tidy --fix` (intentionally separate from `make format` since it rewrites logic, not just style).
- **Dependency installation** — `requirements/install_cpp_deps.sh` installs `clang-format-15`, `clang-tidy-15`, `clangd-15`, `clang++-15`, and `libstdc++-12-dev` on Linux. (`clang++-15` requires `libstdc++-12-dev` on Ubuntu 22.04 to find standard headers.)
- **Style guide** — `docs/cpp_style_guide.md` documents formatting rules, linting rationale, and disabled checks.

## C++ unit tests

`gigl-core/tests/` with GoogleTest (via `FetchContent`). Any `*_test.cpp` or `*_test.cu` file is automatically discovered and registered — no CMake edits needed to add a test. Run via `make unit_test_cpp`; included in `make unit_test`. A stamp file (`.cache/cpp_tests/.configured`) avoids re-running cmake configure when nothing changed; `CONFIGURE_DEPENDS` auto-invalidates when source files are added or removed.


gigl/scripts/post_install.py
```diff
@@ -3,64 +3,56 @@
...
+- Builds pybind11 C++ extensions in-place so they are importable without a separate build step.
 """
 
 import subprocess
 import sys
 from pathlib import Path
-from typing import Optional
-
-
-def run_command_and_stream_stdout(cmd: str) -> Optional[int]:
```
mkolodner-sc, line 16: Removed the custom streaming helper — subprocess.run without a stdout argument already streams output to the terminal in real time (it just inherits the parent's stdout). The old approach using stdout=PIPE was actually worse since piped subprocesses buffer their output more aggressively. We also get free failure detection via check=True, which the old function didn't provide (it returned the exit code but never checked it).

docs/cpp_style_guide.md
```diff
@@ -0,0 +1,153 @@
...
+| `bugprone-string-constructor.LargeLengthThreshold`         | `8388608` (8 MB)  | Strings larger than 8 MB from a length argument are flagged                                                                                                                           |
+| `modernize-loop-convert.NamingStyle`                       | `CamelCase`       | Auto-generated loop variable names use CamelCase                                                                                                                                      |
+| `readability-function-size.LineThreshold`                  | `1000`            | Functions over 1000 lines are flagged                                                                                                                                                 |
+| `readability-braces-around-statements.ShortStatementLines` | `0`               | Braces required for all control-flow bodies, even single-line                                                                                                                         |
```
kmontemayor2-sc: This is good and useful! Do you think it'd be better if we could put it in `.clang-format` or `.clang-tidy` and then say "look at those files"?

That way it's easier for these things to stay synced.

docs/cpp_style_guide.md
```diff
@@ -0,0 +1,153 @@
...
+
+Enforced via `readability-identifier-naming`:
+
+| Identifier kind                                           | Convention                  | Example           |
+| --------------------------------------------------------- | --------------------------- | ----------------- |
+| Classes, enums, unions                                    | `CamelCase`                 | `DistDataset`     |
+| Type template parameters                                  | `CamelCase`                 | `NodeType`        |
+| Functions, methods                                        | `camelBack`                 | `sampleNeighbors` |
+| Variables, parameters, members                            | `camelBack`                 | `numNodes`        |
+| Private/protected members                                 | `camelBack` with `_` prefix | `_nodeFeatures`   |
```
kmontemayor2-sc: fwiw I prefer these to be `snake_case` (which is also what we do in Python).

Any reason to do otherwise?

gigl/csrc/cpp_compile_constants.py
kmontemayor2-sc, line 1: Any reason we have this in `csrc` instead of where the build script lives?

scripts/build_cpp_extensions.py
```diff
@@ -0,0 +1,55 @@
...
+        if impl_file.exists():
+            sources.append(str(impl_file))
+        extensions.append(
+            CppExtension(
+                name=module_name,
+                sources=sources,
+                extra_compile_args=COMPILE_ARGS,
+            )
+        )
+    return extensions
```
kmontemayor2-sc: any reason to do this vs just hard code? Might be nice to be intentional but I can see the appeal of auto-detecting.

scripts/generate_compile_commands.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+        return []
+
+
+def write_compile_commands() -> None:
+    """Write .cache/compile_commands.json for clangd."""
+    # Suppress PyTorch's CUDA-not-found warning emitted at import time.
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore")
+        from torch.utils.cpp_extension import include_paths as torch_include_paths
```
kmontemayor2-sc: any reason we do this?

Makefile
```diff
@@ -143,12 +153,30 @@ format_md:
 	@echo "Formatting markdown files..."
 	uv run mdformat ${MD_FILES}
 
-format: format_py format_scala format_md
+format_cpp:
+	$(if $(CPP_SOURCES), clang-format -i --style=file $(CPP_SOURCES))
```
kmontemayor2-sc: Should we just always run this?

Makefile
```diff
@@ -143,12 +153,30 @@ format_md:
...
 
-lint_test: check_format assert_yaml_configs_parse
+build_cpp_extensions:
+	uv run --no-sync python -m scripts.build_cpp_extensions build_ext --inplace
+
+generate_compile_commands:
+	uv run python -m scripts.generate_compile_commands
+
+check_lint_cpp:
+	$(if $(CPP_SOURCES), uv run python -m scripts.run_cpp_lint $(CPP_SOURCES))
```
kmontemayor2-sc: ditto on always running.

Makefile
```diff
@@ -143,12 +153,30 @@ format_md:
...
+
+# Not part of `make format`: clang-tidy --fix rewrites logic (renames identifiers,
+# changes expressions, adds/removes keywords), not just style. Run manually and
+# review the diff before committing.
+fix_lint_cpp: generate_compile_commands
+	$(if $(CPP_SOURCES), clang-tidy --fix -p .cache/compile_commands.json $(CPP_SOURCES))
```
kmontemayor2-sc: ditto on always running.

containers/Dockerfile.src
kmontemayor2-sc, line 1: Do we want to create a new release after this goes in?

Makefile
```diff
@@ -143,12 +153,30 @@ format_md:
...
 	uv run mypy ${PYTHON_DIRS} --check-untyped-defs
 
-lint_test: check_format assert_yaml_configs_parse
+build_cpp_extensions:
+	uv run --no-sync python -m scripts.build_cpp_extensions build_ext --inplace
+
+generate_compile_commands:
+	uv run python -m scripts.generate_compile_commands
+
+check_lint_cpp:
```
kmontemayor2-sc: Should we add a note about this also not being a part of `check_format` or what not.

kmontemayor2-sc (commented):
Thanks Matt! Nice work, did a pass here :)

requirements/install_cpp_deps.sh
```diff
@@ -0,0 +1,55 @@
...
+    # shell or make process. Open a new terminal (or run `source ~/.zshrc`) after
+    # install_dev_deps to pick up the PATH change.
+    export PATH="$LLVM_BIN:$PATH"
+else
+    # On Linux, apt-get installs versioned binaries (e.g. clang-format-15) directly
+    # into /usr/bin. update-alternatives wires up the bare names (clang-format,
+    # clang-tidy) so callers don't need to specify the version suffix. No PATH
+    # changes are needed since /usr/bin is already on PATH.
+    apt-get install -y "clang-format-${CLANG_VERSION}" "clang-tidy-${CLANG_VERSION}" "clangd-${CLANG_VERSION}" cmake
+    update-alternatives --install /usr/bin/clang-format clang-format "/usr/bin/clang-format-${CLANG_VERSION}" 100
```
svij-sc: can we pin the versions where-ever we use them instead?
I think we will run into pains in the future when folks either:
1. working with different cpp repos i.e. GLT
2. trying to upgrade these

Using clangd-15 directly inside our code base preventsthis.

svij-sc (commented):
A few things (all related):

1. `post_install.py` is script that we want to deprecate as this is not a common pattern for packaging python wheels - and its really only here till we allocate time to fixthe GLT wheels being available. Thus relying on it is not ideal here.
2. The **manual step** of `pip install -e` does not produce a usable extension on its own - which is counter intuitive to an isolated well packaged python package.
3. **Setuptools already has native support for cpp extensions.** - i.e. using setup.py and pyproject.toml we should already be able to build cpp extensions and package them without need for much extra tooling. Leveraging this has a lot of benifits, especially of just packaging the built c++  package with our wheel.

Given ^, some recos:
* Reference to uv guidance in enabling extension modules: https://docs.astral.sh/uv/concepts/projects/init/#projects-with-extension-modules
*  Some quick claude assisted sample of what the diff may look like:


```diff
 # ===================== Build/Project Configurations ===========================
 [build-system]
-requires      = ["setuptools>=61.0.0", "wheel"]
-build-backend = "setuptools.build_meta"
-
-
-[tool.setuptools.packages.find]
-where = ["."]  # list of folders that contain the packages
-include = ["gigl*", "snapchat*"]  # Include only packages that match the specified patterns
+requires = [
+    "scikit-build-core>=0.10",
+    # Build-time torch. Specifying here since I think you specify as a dependency for the tests in cmakelist.txt
+    "torch==2.8",
+    "pybind11>=2.12",
+]
+build-backend = "scikit_build_core.build"
+
+[tool.scikit-build]
+cmake.version        = ">=3.26"
+wheel.packages       = ["gigl", "snapchat"]
+editable.mode        = "redirect"
+editable.rebuild     = true
```



I'd recommend try refining the contributions here with following prompt in plan mode:

```
Goal: build C++ extensions as part of the normal `pip install` flow via scikit-build-core + CMake.
Reference `uv` documentation : https://docs.astral.sh/uv/concepts/projects/init/#projects-with-extension-modules to build the c++ extension module
Keep `gigl-post-install` for GLT only.
Ensure that we are able to release wheel(s) to support both cuda and cpu installations.
```


One small note:
This will affect our release process, I think we will need to build and release two wheels.
1. `pyg27-torch28-cpu`
2. `pyg27-torch28-cu128`

docs/cpp_style_guide.md
mkolodner-sc: I took this approach initially, but the readability of the clang-format and clang-tidy files, as well as the descriptions of what we are checking for, was pretty negative of an experience IMO. I opted to move these descriptions here instead as a result. Do you feel strongly about putting them in the same file?

docs/cpp_style_guide.md
mkolodner-sc: Following general guidance, the style guide here is based on LLVM, which leverages camelCase for several reasons, one of which is to distinguish custom functions (camelCase) from the standard `std::` library types functions.

Makefile
mkolodner-sc: Done, ditto elsewhere

gigl/csrc/cpp_compile_constants.py
mkolodner-sc: This has been removed as part of the refactor

Makefile
mkolodner-sc: `check_lint_cpp` is part of `lint_test`

```
lint_test: check_format assert_yaml_configs_parse check_lint_cpp
	@echo "Lint checks pass!"
```

`check_format` checks the cpp format

```
check_format_cpp:
	clang-format-15 --dry-run --Werror --style=file $(CPP_SOURCES)

check_format: check_format_py check_format_cpp check_format_scala check_format_md
```

scripts/build_cpp_extensions.py
mkolodner-sc: Updated as part of the refactor -- PTAL. Some similar logic now lives in the CMakeLists.txt

containers/Dockerfile.src
mkolodner-sc: Will create a release as an immediate follow-up to this PR going in

.github/cloud_builder/run_command_on_active_checkout.yaml
```diff
@@ -29,7 +29,7 @@ steps:
         docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
         gcloud auth configure-docker us-central1-docker.pkg.dev
         # Install GiGL
-        uv pip install -e .
+        uv pip install -e . --no-build-isolation
```
mkolodner-sc: ## Why `--no-build-isolation` is required

`--no-build-isolation` guarantees the C++ extensions are always built against the torch wheel that matches the target hardware — CPU or CUDA — rather than whatever PyPI resolves in an isolated environment.

Before this PR, the build backend was setuptools, which only copies Python files — no compilation step, no external tools needed. Build isolation was fine because the isolated environment had everything it needed.

This PR switches to `scikit-build-core`, which runs CMake to compile pybind11 C++ extensions against PyTorch's C++ headers. Those headers are provided by the `torch` package itself, so **torch must be present during the build** — not just at runtime.

With build isolation, `uv` creates a fresh throwaway environment and installs `[build-system].requires` from scratch. Since `torch` is not listed there, CMake's `find_package(Torch)` receives an empty prefix path and fails at configure time — before any compilation happens:

```
CMake Error at CMakeLists.txt (find_package):
  Could not find a package configuration file provided by "Torch" with
  any of the following names:
    TorchConfig.cmake
    torch-config.cmake
```

### Why not just add `torch==2.8` to `[build-system].requires`?

This was considered and done previously but has a fundamental problem. When pip/uv resolves `torch==2.8` from PyPI during the isolated build, on Linux x86_64 the default wheel is CUDA-enabled. CMake then finds a CUDA torch, expects the full CUDA toolkit, and fails on any CPU-only environment — CI runners, base Docker images, CPU developer machines. You'd need to point the isolated build at the CPU index via `UV_EXTRA_INDEX_URL` in every Dockerfile, CI script, and developer environment, adding a hidden requirement that's easy to miss and break.

`--no-build-isolation` skips the isolated env and uses the torch already installed in the active venv — the correct CPU or CUDA variant for the target hardware. This is the standard pattern across the PyTorch extension ecosystem: **torch must be pre-installed with the right variant before building against it.** `--no-build-isolation` makes that requirement explicit rather than hiding it behind environment variable conventions.

**Re: CUDA headers** — With the current solution with --no-build-isolation, this works correctly for both variants. `find_package(Torch)` resolves against whatever torch is in the active environment: the CPU wheel on CPU machines, the CUDA 12.8 wheel on GPU machines. The same `CMakeLists.txt` handles both.

Makefile
mkolodner-sc: Actually, revisiting this, it seems like clang-format and clang-tidy will error if there are no cpp files found here, as it would mistakenly try to read from stdin. There are no cpp files in csrc for this PR, so we should this check here to make sure we aren't incorrectly failing.

.github/cloud_builder/run_command_on_active_checkout.yaml
mkolodner-sc: Resolved by setting `no-build-isolation-package = ["gigl"]` in the pyproject.toml, allowing us to remove the less safe --no-build-isolation tags

containers/Dockerfile.dataflow.src
```diff
@@ -5,14 +5,17 @@ FROM $BASE_IMAGE
 # Copy the source
 WORKDIR /gigl
 
+COPY CMakeLists.txt CMakeLists.txt
 COPY MANIFEST.in MANIFEST.in
+COPY README.md README.md
```
kmontemayor2-sc: QQ any reason we copy this over? If we need it, can we add a comment saying why?

docs/cpp_style_guide.md
```diff
@@ -0,0 +1,176 @@
...
+header.
+
+### Include ordering
+
+Includes are sorted and split into three priority groups:
+
+| Priority | Pattern                              | Group                                 |
+| -------- | ------------------------------------ | ------------------------------------- |
+| 1        | `.*`                                 | Local project headers (first)         |
+| 2        | `^"(llvm\|llvm-c\|clang\|clang-c)/"` | LLVM/Clang internal headers           |
```
kmontemayor2-sc: hmmm, this seems very specific to projects relying on LLVM / clang internals, I don't think we're really going to use those internals?  I know we'll use the tools, but do we forsee needing their imports?

docs/cpp_style_guide.md
kmontemayor2-sc: Hmmm, I feel like being able to more easily differentiate between `MyClass` and `myFunction` my be useful (and is more inline with our Python code base). But I wont push back here anymore,

docs/cpp_style_guide.md
```diff
@@ -0,0 +1,176 @@
...
+| `readability-function-cognitive-complexity`           | Algorithmic code often requires nesting that is inherent to the problem structure. Enforcing an arbitrary complexity ceiling discourages clarity and encourages artificial decomposition.                                                              |
+| `readability-magic-numbers`                           | Literal constants are common in ML code (e.g. feature dimensions)                                                                                                                                                                                      |
+
+### Naming conventions
+
+Enforced via `readability-identifier-naming`:
+
+| Identifier kind                                           | Convention                  | Example           |
+| --------------------------------------------------------- | --------------------------- | ----------------- |
+| Classes, enums, unions                                    | `CamelCase`                 | `DistDataset`     |
```
kmontemayor2-sc: You know I was always under the impression first cap is `PascalCase`, and `camelCase` doesn't have an `initialCapital` but then wiki says it doesn't really matter. https://en.wikipedia.org/wiki/Camel_case#Description

Though since we use both (for different things) maybe we should differentiate?

docs/cpp_style_guide.md
kmontemayor2-sc: No strong preference, this is fine :)

requirements/install_py_deps.sh
```diff
@@ -138,20 +138,28 @@ install_gigl_lib_deps() {
         flag_use_inexact_match="--inexact"
     fi
 
+    # --no-install-project: skip building and installing the gigl package itself here.
+    # The project is installed separately via `uv pip install -e .`
+    # (see Makefile install_dev_deps/install_deps targets and the *.src Dockerfiles).
+    # This avoids scikit-build-core requiring all source files (e.g. README.md, CMakeLists.txt)
+    # to be present in environments that only need the dependencies (e.g. base Docker images).
```
kmontemayor2-sc: hmmm, our build is getting more and more and more bespoke. I'm worried this will get even more complicated to debug in the future.

I don't think we need to take action here, just something to keep in mind if we can figure out a way to simplify in the future :)

scripts/run_cpp_lint.py
```diff
@@ -0,0 +1,71 @@
...
+            f"--compile-commands-dir={_COMPILE_COMMANDS.parent}",
+        ],
+        capture_output=True,
+        text=True,
+    )
+    diagnostics: list[str] = []
+    for line in result.stderr.splitlines():
+        m = _DIAGNOSTIC_RE.match(line)
+        if m:
+            diagnostics.append(m.group(1))
```
kmontemayor2-sc, line 48: Does the `result` output everything to stderr? e.g. are we potentially missing anything by filtering like this?

scripts/run_cpp_lint.py
```diff
@@ -0,0 +1,71 @@
...
+    return source, diagnostics
```
kmontemayor2-sc: nit. Somewhat odd to be returning the input path? Can we simplify and not return the input?

scripts/run_cpp_lint.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def main() -> None:
+    sources = [Path(s) for s in sys.argv[1:]]
+    if not sources:
+        sys.exit(0)
```
kmontemayor2-sc, line 63: hmmm, should we use argparse here? Or would setting that up be even more complicated?

.clang-format
```diff
@@ -0,0 +1,83 @@
...
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+FixNamespaceComments: true
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeCategories:
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
+    Priority:        2
```
kmontemayor2-sc: if anything we should probably make torch / glt imports a different prio?

.clang-format
```diff
@@ -0,0 +1,83 @@
...
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
```
kmontemayor2-sc: qq is there no `Cpp17` standard?

Makefile
```diff
@@ -75,7 +80,7 @@ assert_yaml_configs_parse:
 # Ex. `make unit_test_py PY_TEST_FILES="eval_metrics_test.py"`
 # By default, runs all tests under tests/unit.
 # See the help text for "--test_file_pattern" in tests/test_args.py for more details.
-unit_test_py: clean_build_files_py type_check
+unit_test_py: clean_build_files_py build_cpp_extensions type_check
```
kmontemayor2-sc, line 83: hmmm,, should we also add a `clean_build_files_cpp`?

Makefile
```diff
@@ -108,13 +118,16 @@ check_format_md:
 	@echo "Checking markdown files..."
 	uv run mdformat --check ${MD_FILES}
 
-check_format: check_format_py check_format_scala check_format_md
+check_format_cpp:
+	$(if $(CPP_SOURCES),clang-format-15 --dry-run --Werror --style=file $(CPP_SOURCES))
```
kmontemayor2-sc: If we don't add some sort of `util_for_testing` in csrc, can we add a TODO to remove these $(IF)` when we actually add the cpp code in the followup PR?

Makefile
```diff
@@ -141,12 +154,30 @@ format_md:
...
+
+generate_compile_commands:
+	uv run python -m scripts.generate_compile_commands
+
+check_lint_cpp:
+	uv run python -m scripts.run_cpp_lint $(CPP_SOURCES_NO_CUDA)
+
+# Not part of `make format`: clang-tidy --fix rewrites logic (renames identifiers,
+# changes expressions, adds/removes keywords), not just style. Run manually and
+# review the diff before committing.
```
kmontemayor2-sc: this can't fix everything right? Should we note that this is not guaranteed to fix everything?

Makefile
kmontemayor2-sc: I more ment like, "Typing `make format_check` does not also check the cpp linter" but that's a small nit.

.github/workflows/on-pr-comment.yml
kmontemayor2-sc, line 1: Let's also add this to the merge workflow? https://github.com/Snapchat/GiGL/blob/main/.github/workflows/on-pr-merge.yml

.github/workflows/release.yml
```diff
@@ -12,8 +12,15 @@ permissions:
 
 jobs:
   build:
-    name: Build and release pip whl
+    name: Build and release pip whl (${{ matrix.torch-variant }})
```
kmontemayor2-sc, line 15: Do we need to update our install guidance for the correct name now?

scripts/generate_compile_commands.py
```diff
@@ -0,0 +1,65 @@
...
+    make generate_compile_commands
+"""
+
+import json
+import subprocess
+from pathlib import Path
+
+_REPO_ROOT: Path = Path(__file__).resolve().parent.parent
+_CMAKE_BUILD_DIR: Path = _REPO_ROOT / ".cache" / "cmake_build_lint"
+_COMPILE_COMMANDS: Path = _REPO_ROOT / ".cache" / "compile_commands.json"
```
kmontemayor2-sc: should this be `PUBLIC` if we import it later?

kmontemayor2-sc (commented):
Thanks for all the excellent work Matt! 

I think the only thing left blocking on my side is adding the cpp test to our merge suite https://github.com/Snapchat/GiGL/pull/558/changes#r3118534359

.github/workflows/release.yml
```diff
@@ -12,8 +12,15 @@ permissions:
...
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - torch-variant: cpu
+            torch-index: https://download.pytorch.org/whl/cpu
+          - torch-variant: cu128
+            torch-index: https://download.pytorch.org/whl/cu128
```
svij-sc: I setup some GPU instances for building these wheels - so we can probably get around the whole complication below.

New recommended matrix to run one job on gpu instance and one run on cpu instance:

```yaml
strategy:
  matrix:
    include:
      - runner: ubuntu-latest
        name: cpu
      - runner: gigl-gpu-instances. # Spec: Linux x64,  1 x NVIDIA Tesla T4 , 16 GB VRAM
        name: gpu

runs-on: ${{ matrix.runner }}
name: Build and release pip whl (${{ matrix.name }}) 
```

Also see [note below,](https://github.com/Snapchat/GiGL/pull/558/changes#r3126262806), we can probably skip specifying torch-index there.

pyproject.toml
```diff
@@ -99,6 +99,11 @@ required-environments = [
     "sys_platform == 'linux' and platform_machine == 'x86_64'",
     "sys_platform == 'darwin' and platform_machine == 'arm64'",
 ]
+# gigl's CMakeLists.txt requires torch headers at build time (find_package(Torch)).
+# Disabling build isolation for gigl lets the build use whatever torch variant is
+# already installed (CPU or CUDA), rather than downloading a fresh wheel into an
+# isolated env that ignores the project's index routing.
+no-build-isolation-package = ["gigl"]
```
svij-sc: IIRC, I believe this is the same as no-build-isolation for the entire build - this could cause issues aforementioned? 


Did we try the following already and it didnt work?
```.toml
[tool.uv.extra-build-dependencies]
gigl = [{ requirement = "torch", match-runtime = true }]
```

.github/workflows/release.yml
```diff
@@ -40,8 +48,17 @@ jobs:
         # Pre-install keyring and Artifact Registry plugin from the public PyPI
         uv tool install keyring --with keyrings.google-artifactregistry-auth==1.1.2
 
+    # install_dev_deps auto-detects CUDA and installs the CPU torch variant on
+    # ubuntu-latest runners. Override here with the correct variant for this build
+    # so CMake picks up the right torch headers (CPU-only vs CUDA-enabled).
+    - name: Install torch variant for build
+      run: |
+        uv pip install "torch==2.8" --index-url ${{ matrix.torch-index }} --inexact
```
svij-sc: I am pretty sure you do not need to do this: `Install torch variant for build`.
Reason:  we are already installing `./.github/actions/setup-python-tools` above - which is already setting up a uv environment for you with torch installed.

pyproject.toml
svij-sc: With the new gpu machine in matrix, I would recommend try switching to extra-build-dependencies with match-runtime = true, drop no-build-isolation-package, and verify a clean uv sync on both CPU and  CU128 extras produces a gigl wheel linked against the right torch. 

If it doesn't bite, we can fall back to the current no-build-isolation-package config.

svij-sc (commented):
Some comments re building the wheel

Also, have we actually tested this flow?

CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
...
+    COMMAND "${Python_EXECUTABLE}" -c
+        "import torch, os; print(os.path.join(os.path.dirname(torch.__file__), 'lib'))"
+    OUTPUT_VARIABLE TORCH_LIB_DIR
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+find_library(TORCH_PYTHON_LIBRARY torch_python PATHS "${TORCH_LIB_DIR}" REQUIRED)
+
+# Preserve torch lib dir in the installed .so RPATH so the extension loads
+# libtorch.so without requiring `import torch` to pre-populate the symbol table.
+set(CMAKE_INSTALL_RPATH "${TORCH_LIB_DIR}")
```
svij-sc: is this CC generated or is there actually reasoning behind this?
That absolute path embedded here-in will likely be hard coded in the generated assets and will not exist on the other instances. Unless this is only used for building wheels and nothing is embedded?

pyproject.toml
mkolodner-sc: ## What `extra-build-dependencies` with `match-runtime = true` does

This lets uv inject a package into the isolated build environment it creates for PEP 517 builds. `match-runtime = true` tells uv "grab this package from your resolution of the runtime environment and make it available during the build."

## What `no-build-isolation-package = ["gigl"]` does

This tells uv to skip build isolation for `gigl` entirely — just build it using the active Python environment as-is. cmake's `find_package(Torch)` then picks up torch headers from wherever they happen to live, regardless of how they got there.

## Why `no-build-isolation-package` is the right call

For GiGL — where torch is always installed by uv — both approaches produce the same wheel. But `no-build-isolation-package` is still more correct, and it comes down to how cmake and scikit-build-core interact.

PEP 517 build isolation controls which packages are visible to the build backend (scikit-build-core). But cmake isn't a Python process — it's a subprocess that scikit-build-core spawns. Critically, scikit-build-core passes `-DPython_EXECUTABLE=sys.executable` to cmake, where `sys.executable` is the runtime env's Python — not the isolated build env's Python. This means cmake's `find_package(Python)` resolves to the runtime interpreter before it even looks at `VIRTUAL_ENV` or `PATH`. torch is then discovered via that interpreter, pointing at the runtime env regardless of what uv installed into the isolated build env.

We verified this experimentally by adding a `message(STATUS)` probe to `CMakeLists.txt` and building under both configurations. In both cases `Python_EXECUTABLE` was set to the runtime env's Python before `find_package(Python)` ran — confirming that the isolated env's torch is never reachable by cmake.

In short: the torch that `extra-build-dependencies` puts in the isolated env is invisible to cmake. `no-build-isolation-package` is just honest about this — it opts out of isolation rather than creating one that gets silently bypassed.

CMakeLists.txt
mkolodner-sc: Good catch — the RPATH was originally there so the extension could find `libtorch.so` without needing `import torch` called first, but this isn't likely to happen in practice. Every other use of the extension already imports torch.

.github/cloud_builder/run_command_on_active_checkout.yaml
```diff
@@ -3,7 +3,7 @@ substitutions:
 options:
   logging: CLOUD_LOGGING_ONLY
 steps:
-  - name: us-central1-docker.pkg.dev/external-snap-ci-github-gigl/gigl-base-images/gigl-builder:b34c863a2168c8df5a6da1f6385e5d374f0175d2.91.1
```
mkolodner-sc: f

containers/Dockerfile.dataflow.src
mkolodner-sc: Added

.clang-format
mkolodner-sc: Fixed, thanks!

Makefile
mkolodner-sc: Added!

scripts/generate_compile_commands.py
mkolodner-sc: Made COMPILE_COMMANDS public since this is imported elsewhere, but the rest can stay private here

.github/workflows/release.yml
mkolodner-sc: Updating the installation instructions here, thanks!

scripts/run_cpp_lint.py
mkolodner-sc: Yes — clangd --check sends all diagnostic output to stderr; stdout is unused for this subcommand. Filtering result.stderr captures everything.

scripts/run_cpp_lint.py
mkolodner-sc: argparse would add more complexity than value here — the interface is just a flat list of file paths, no flags or options. sys.argv[1:] is the right choice for this.

.github/workflows/release.yml
```diff
@@ -41,19 +56,9 @@ jobs:
         uv tool install keyring --with keyrings.google-artifactregistry-auth==1.1.2
 
     - name: Build Whl Distribution
-      run: uv build
+      run: uv build --wheel
 
     - name: Publish Package 🚀
-      env:
```
mkolodner-sc, line 47: The `PYPIRC_CONTENTS` and `PIP_CONF_CONTENTS` secrets were used by the old twine publishing workflow. twine reads `~/.pypirc` for registry credentials and pip reads `~/.pip/pip.conf` — so the old workflow would write those secrets to those files, then clean them up afterward in the "Post Publish Package" step.

We now use `uv publish --keyring-provider subprocess`, which authenticates via the GCP keyring (set up by `gcloud auth` in the setup-python-tools action). It doesn't read `.pypirc` or `pip.conf` at all — credentials come from the active GCP service account token passed directly as `--username oauth2accesstoken`.

Since nothing in the current workflow writes `~/.pypirc` or `~/.pip/pip.conf` anymore, the `PYPIRC_CONTENTS`/`PIP_CONF_CONTENTS` env vars are never consumed, and the cleanup step has nothing to remove. Both are dead code.

kmontemayor2-sc: thanks :)

requirements/install_py_deps.sh
```diff
@@ -138,20 +136,28 @@ install_gigl_lib_deps() {
...
     # Taken from https://stackoverflow.com/questions/59895/how-do-i-get-the-directory-where-a-bash-script-is-located-from-within-the-script
     # We do this so if `install_py_deps.sh` is run from a different directory, the script can still find the post_install.py file.
     if [[ "${SKIP_GLT_POST_INSTALL}" -eq 0 ]]
     then
         SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-        uv run python $SCRIPT_DIR/../gigl/scripts/post_install.py
+        # --no-project: prevents uv from trying to install the gigl project before running
+        # the script. We intentionally skipped installing it above (--no-install-project),
+        # and post_install.py only runs install_glt.sh — it does not need gigl installed.
+        uv run --no-project python $SCRIPT_DIR/../gigl/scripts/post_install.py
```
kmontemayor2-sc: Hmmm, do we ever install the "gigl" project now? Should we?

scripts/generate_compile_commands.py
```diff
@@ -0,0 +1,77 @@
+"""Generate .cache/compile_commands.json for clangd.
+
+Delegates to CMake (which already knows all include paths and compiler flags via
+find_package(Torch)) rather than manually constructing the database.
+
+The build uses the system C++ compiler (g++) so that cmake, nvcc, and Torch's
```
kmontemayor2-sc: Hmmm, is this correct? I don't think it matters too much but for my own learnings, why can't "the build" (e.g. cmake?) use llvm our of the box?

Also I don't think g++ and llvm have interchangeable flags, do they?

scripts/run_cpp_lint.py
```diff
@@ -0,0 +1,84 @@
...
+    return diagnostics
+
+
+def main() -> None:
+    sources = [Path(s) for s in sys.argv[1:]]
+    if not sources:
+        sys.exit(0)
+
+    failures: dict[Path, list[str]] = {}
+    with ThreadPoolExecutor(max_workers=min(os.cpu_count() or 1, 8)) as executor:
```
kmontemayor2-sc: any reason we added this? no arg uses all cpu already right why limit it?

.clang-format
```diff
@@ -0,0 +1,87 @@
...
+# This is the Google/PyTorch convention — the opposite of the LLVM default.
+# Result: project-local headers (1) → torch/pybind11 (2) → system/third-party (3).
+# Add graphlearn_torch to the Priority 2 regex when GLT headers appear in the codebase.
+IncludeCategories:
+  - Regex:           '^<(torch|pybind11)/'
+    Priority:        2
+  - Regex:           '^(<|"(gtest|isl|json)/)'
+    Priority:        3
+  - Regex:           '.*'
+    Priority:        1
```
kmontemayor2-sc: nit. cat we order these? 
```suggestion
  - Regex:           '.*'
    Priority:        1
  - Regex:           '^<(torch|pybind11)/'
    Priority:        2
  - Regex:           '^(<|"(gtest|isl|json)/)'
    Priority:        3
```

kmontemayor2-sc: Also why did we chose `gtest/isl/json` as the "sys" imports?

docs/cpp_style_guide.md
```diff
@@ -0,0 +1,209 @@
...
+
+```bash
+uv pip install -e . -C cmake.build-type=Debug
+```
+
+**C++ unit tests:**
+
+```bash
+make unit_test_cpp CMAKE_BUILD_TYPE=Debug
+```
```
kmontemayor2-sc: hmmm, I feel like for unit tests we should probably build with debug symbols?

IIRC o3 makes it kind of hard to debug these things? (e.g. stack traces get mangled?), maybe worth enabling debug for development purposes.

I'd also be curious as to the slowdown we get if we just always `-g`. If it's negligible then maybe worth it to always have debug on.

requirements/install_cpp_deps.sh
```diff
@@ -0,0 +1,62 @@
...
+    for rc_file in ~/.zshrc ~/.bashrc; do
+        if [ -f "$rc_file" ] && ! grep -qF "$LLVM_BIN" "$rc_file"; then
+            printf '\n# Added by GiGL install_cpp_deps.sh\nexport PATH="%s:$PATH"\n' "$LLVM_BIN" >> "$rc_file"
+            echo "NOTE: Added LLVM bin to PATH in $rc_file."
+            echo "      Open a new terminal (or run: source $rc_file) to pick up the change."
+        fi
+    done
+
+    export PATH="$LLVM_BIN:$PATH"
+else
```
kmontemayor2-sc: BTW have tested this path at all? Since we need to use CPP with the GLT code, and we don't support GLT on macos, is there a use for this branch?

scripts/generate_compile_commands.py
```diff
@@ -0,0 +1,77 @@
...
+    for entry in entries:
+        if not entry.get("file", "").endswith(".cu"):
+            if "command" in entry:
+                tokens = shlex.split(entry["command"])
+                tokens[0] = "clang++-15"
+                entry["command"] = shlex.join(tokens)
+            elif "arguments" in entry:
+                entry["arguments"][0] = "clang++-15"
+
+    COMPILE_COMMANDS.write_text(json.dumps(entries, indent=2))
```
kmontemayor2-sc: do we ever need to create the parent dir first? I guess if ci/cd passes (and runs this code?) then it should be fine?

kmontemayor2-sc (approved):
LGTM, thanks for the cycles Matt!

I strongly think we should look into always enabling `-g`, if we can do a profiling run for the perf difference there I think it'll save us many, many debugging headaches in the future.

.clang-format
mkolodner-sc: Good catch — isl and json were inherited from an LLVM config template and aren't used in our codebase at all. Only gtest is an actual third-party header we use (in tests/unit/cpp/). Removed isl and json, updated the style guide to match.

scripts/generate_compile_commands.py
mkolodner-sc: The parent directory is guaranteed to exist by the time this runs. We call `_CMAKE_BUILD_DIR.mkdir(parents=True, exist_ok=True)` which creates `.cache/` as a side effect.

requirements/install_cpp_deps.sh
mkolodner-sc: Agreed — removed the macOS branch. The script now exits with an error on macOS with a clear message that GLT doesn't support it.

docs/cpp_style_guide.md
mkolodner-sc: Good call, there seems to be minimal/no performance impact with -g, so we can aim to keep it on and simplify our codebase as a result.

scripts/run_cpp_lint.py
mkolodner-sc: Good call, removed!

scripts/generate_compile_commands.py
mkolodner-sc: Good catch. We were running cmake with g++ then post-processing to swap in clang++-15, but this cmake invocation is only for generating compile_commands.json, so we can just pass -DCMAKE_CXX_COMPILER=clang++-15 directly and skip the rewrite entirely. This makes it so that flag compatibility is no longer a concern.

requirements/install_py_deps.sh
mkolodner-sc: `install_py_deps.sh` skips it with --no-install-project, but make install_dev_deps runs `uv pip install -e .` immediately after on Makefile, which installs the gigl project.

pyproject.toml
mkolodner-sc: I've added some additional context here in the pyproject.toml file:

```
# WANT: [tool.uv.extra-build-dependencies]
# WANT: gigl = [{ requirement = "torch", match-runtime = true }]
#
# BLOCKED: scikit-build-core passes -DPython_EXECUTABLE from sys.executable,
# which points at the runtime env's Python — not the isolated build env.
# CMake then discovers torch via that executable, bypassing uv's isolation.
# Until scikit-build-core or uv resolves this for cmake-based builds,
# we disable isolation so the declared config matches actual behavior.
```

mkolodner-sc: One update here -- I was able to avoid this issue altogether — not needing to specify `gigl = [{ requirement = "torch", match-runtime = true }]` or `no-build-isolation-package = ["gigl"]` — by making it so that torch no longer needs to be installed in the isolated build environment. Instead, `scikit-build-core` writes `CMakeInit.txt` with `CMAKE_PREFIX_PATH` pointing at the runtime venv's `site-packages`. `cmake` then finds torch's cmake config and headers there natively, without requiring torch to be present in the build env's own packages. This allows the majority of the build to be isolated, with the exception of the `.so`, which is linked against the runtime venv's `libtorch` via `CMAKE_PREFIX_PATH` — an inherent and unavoidable property of any PyTorch C++ extension.

gigl/orchestration/Dockerfile.customer_src
```diff
@@ -10,4 +10,4 @@ WORKDIR /gigl
 COPY . .
 
 # Find out if there is 'setup.py' or 'pyproject.toml' in the current directory, if so install it
-RUN if [ -f "setup.py" ] || [ -f "pyproject.toml" ]; then pip install -e .; fi
+RUN if [ -f "setup.py" ] || [ -f "pyproject.toml" ]; then uv pip install scikit-build-core pybind11 && uv pip install -e .; fi
```
svij-sc: ideally we should be packaging the c source so they dont need to rebuild it here. 
Can be a follow up

requirements/install_cpp_deps.sh
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+# On Linux, apt-get installs versioned binaries (e.g. clang-format-15) directly
+# into /usr/bin. No PATH changes are needed since /usr/bin is already on PATH.
+# Callers use the versioned names (clang-format-15, clang-tidy-15, clangd-15)
+# directly so the version is explicit and greppable across the codebase.
+# clang++-15 requires libstdc++-12-dev: on Ubuntu 22.04, clang++-15 looks for GCC 12
+# headers. Without this package clang++-15 cannot find standard headers like <cstddef>.
+# clang++-15 itself is needed because generate_compile_commands.py rewrites
+# compile_commands.json to use it so clangd natively understands the commands.
```
svij-sc: nit: can we make the comment more succinct ?
Traditionally, we would only have `clang++-15 requires libstdc++-12-dev: on Ubuntu 22.04, clang++-15 looks for GCC 12` as that is a gotcha. Everything else makes sense.

requirements/install_py_deps.sh
```diff
@@ -45,12 +45,10 @@ has_cuda_driver() {
 
 is_running_on_mac() {
     [ "$(uname)" == "Darwin" ]
-    return $?
 }
 
 is_running_on_m1_mac() {
     [ "$(uname)" == "Darwin" ] && [ $(uname -m) == 'arm64' ]
-    return $?
```
svij-sc, line 53: any reason to take these out?

requirements/install_py_deps.sh
```diff
@@ -138,20 +136,28 @@ install_gigl_lib_deps() {
...
+    # --no-install-project: skip building and installing the gigl package itself here.
+    # The project is installed separately via `uv pip install -e .`
+    # (see Makefile install_dev_deps/install_deps targets and the *.src Dockerfiles).
+    # This avoids scikit-build-core requiring all source files (e.g. README.md, CMakeLists.txt)
+    # to be present in environments that only need the dependencies (e.g. base Docker images).
     if [[ $DEV -eq 1 ]]
     then
         # https://docs.astral.sh/uv/reference/cli/#uv-sync
-        uv sync ${extra_deps_clause[@]} --group dev --locked ${flag_use_inexact_match}
+        uv sync ${extra_deps_clause[@]} --group dev --locked ${flag_use_inexact_match} --no-install-project
```
svij-sc: I think we might have a chicken and egg problem.
i.e. because we have this, we have to do some other thing, and because we do some other thing, we need to do something else thats also not ideal.

Looking into how we can simplify...

Makefile
```diff
@@ -49,6 +53,7 @@ install_dev_deps: check_if_valid_env
 	gcloud auth configure-docker us-central1-docker.pkg.dev
 	bash ./requirements/install_py_deps.sh --dev
 	bash ./requirements/install_scala_deps.sh
+	bash ./requirements/install_cpp_deps.sh
```
svij-sc: On my vm `make install_dev_deps` does not work:

```
...
...
+ echo 'Finished installing C++ tooling'
Finished installing C++ tooling
uv pip install -e .
Resolved 155 packages in 294ms
  × Failed to build `gigl @ file:///...`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `scikit_build_core.build.build_editable` failed (exit status: 1)

      [stdout]
      *** scikit-build-core 0.12.2 using CMake 3.22.1 (editable)
      *** Configuring CMake...
      loading initial cache file .cache/cmake_build/CMakeInit.txt
      -- The CXX compiler identification is GNU 11.4.0
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /usr/bin/c++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      -- Looking for a CUDA compiler
      -- Looking for a CUDA compiler - /usr/local/cuda/bin/nvcc
      -- The CUDA compiler identification is NVIDIA 12.8.61
      -- Detecting CUDA compiler ABI info
      -- Detecting CUDA compiler ABI info - done
      -- Check for working CUDA compiler: /usr/local/cuda/bin/nvcc - skipped
      -- Detecting CUDA compile features
      -- Detecting CUDA compile features - done
      -- Found Python: .../bin/python (found version
      "3.11.14") found components: Interpreter Development.Module
      -- Configuring incomplete, errors occurred!
      See also
      "/.../CMakeFiles/CMakeOutput.log".

      [stderr]
      CMake Error at CMakeLists.txt:58 (message):
        Cannot find torch cmake config in CMAKE_PREFIX_PATH:
      .../site-packages.
        Run: make install_dev_deps

      

      *** CMake configuration failed

      hint: This usually indicates a problem with the package or the build environment.
```

pyproject.toml
```diff
@@ -177,12 +182,26 @@ explicit = true
...
 [[tool.uv.index]]
-name = "gcp-release-registry"
+name = "gcp-release-registry-cpu"
 url = "https://us-central1-python.pkg.dev/external-snap-ci-github-gigl/gigl/simple/"
 publish-url = "https://us-central1-python.pkg.dev/external-snap-ci-github-gigl/gigl"
 explicit = true
 
+[[tool.uv.index]]
+name = "gcp-release-registry-cu128"
+url = "https://us-central1-python.pkg.dev/external-snap-ci-github-gigl/gigl-cu128/simple/"
```
svij-sc, line 208: do we need to create this?

pyproject.toml
```diff
@@ -101,7 +101,12 @@ required-environments = [
 ]
 
 [dependency-groups]
+cpp-build = [
+  "scikit-build-core>=0.10",
+  "pybind11>=2.12",
+]
 dev = [
+  {include-group = "cpp-build"},
```
svij-sc: i guess this is needed because we rely on preinstalled libs to build vs. the build system requires ?

Makefile
```diff
@@ -94,7 +99,16 @@ unit_test_scala: clean_build_files_scala
 # Eventually, we should look into splitting these up.
 # We run `make check_format` separately instead of as a dependent make rule so that it always runs after the actual testing.
 # We don't want to fail the tests due to non-conformant formatting during development.
-unit_test: precondition_tests unit_test_py unit_test_scala
+.cache/cpp_tests/.configured: CMakeLists.txt tests/unit/cpp/CMakeLists.txt .cache/cmake_build/CMakeInit.txt
+	uv run cmake -C .cache/cmake_build/CMakeInit.txt \
+		-S . -B .cache/cpp_tests -DGIGL_BUILD_TESTS=ON
+	touch .cache/cpp_tests/.configured
```
svij-sc: btw, make has a prebuilt caching mechanism.
This rule will only run once and wont run again unless you delete `touch .cache/cpp_tests/.configured`

You can test this using:
```Makefile
marco:
	echo "polo"
	touch marco
```
```bash
(base) ➜  cpp-infrastructure git:(mkolodner-sc/cpp-infrastructure) ✗ make marco
echo "polo"
polo
touch marco
(base) ➜  cpp-infrastructure git:(mkolodner-sc/cpp-infrastructure) ✗ make marco
make: 'marco' is up to date.
```


Note that second time the entire code block is skipped because a file name with the same name as the rule exists.

pyproject.toml
svij-sc: Thanks for the detailed response.
Although my understanding slightly deviates from yours above, I think its fine if we do `no-build-isolation-package = ["gigl"]` given the cycles - thank you!

.github/workflows/release.yml
svij-sc: ummm... have we actually tested the release flow post this change? It works?

Makefile
svij-sc: I guess this is what is expected? i.e. u are trying to just do a one time setup?
How does one develop cpp post this change; would they have to manually erase this file?

svij-sc (commented):
We may need to update the .gitignore file as working with the relevent make commands I quickly realized that there are a lot of files that are generated not being git ignored.

svij-sc (commented):
I am okay to approve if comments are addressed, and we tested the wheel building flow - i.e. lets not break it.
We can iterate on the final design here post merge - I do see CC is suggesting a lot of things here and/or restructuring code without completely understanding the wide spanning impact that it may have, so we need to be a little careful here in ensuring we manually test all flows including internal usage.

svij-sc:
Btw, given the restrictions, comments, and keeping the code base clean I have an alternate suggestion to have cpp build as a separate workspace member with its own `pyproject.toml` and all.

We should be able to just run 
`uv init --build-backend scikit-build-core gigl-core`

And that should automatically:
1. bootstrap a new dir in the project that houses our cpp code
Image

2. adding a new member to the main pyproject.toml:
```
[tool.uv.workspace]
members = [
    "gigl-core",
]
```

I also ended up drafting what this might look like in the future: https://github.com/Snapchat/GiGL/pull/596
(Haven't tested - mostly CC)

requirements/install_py_deps.sh
svij-sc: Btw, does taking this --no-install-project out actually cause issues?

requirements/install_py_deps.sh
mkolodner-sc: This isn't necessary since with bash a function's return value is implicitly the exit status of the last command executed — so return $? is always a no-op at the end of a function. That being said, this is out of the scope of this PR, so I'll restore this to the previous state.

requirements/install_py_deps.sh
svij-sc: > Looking into how we can simplify...

Provided suggestion here for a future direction - https://github.com/Snapchat/GiGL/pull/558#issuecomment-4314428358

(non blocking)

.github/workflows/release.yml
mkolodner-sc: Yep, the steps above work.

pyproject.toml
mkolodner-sc: This has been created and verified

Makefile
mkolodner-sc: No, cmake automatically re-configures when new .cpp/.cu files are added by watching the directory modification timestamp. When a file is added or removed, the parent directory's mtime changes, cmake detects that at the start of the next build, re-runs the glob, and triggers re-configuration if the result changed. Normal development never requires touching the cached file.

requirements/install_py_deps.sh
mkolodner-sc: We are able to remove the --no-install-project now with the above changes, thanks! I tested this locally -- uv sync correctly installs gigl as an editable install and builds the gigl-core stub without it.

pyproject.toml
mkolodner-sc: This can be removed now with the gigl-core solution.

gigl/orchestration/Dockerfile.customer_src
mkolodner-sc: the concern is resolved — gigl-core now ships as a pre-compiled wheel for both CPU and CUDA; customers install the wheel from the registry and get no compilation at all. Dockerfile.customer_src remains for the explicit source-build case but is no longer the customer default.

Makefile
mkolodner-sc: The installation work should with the recent changes.

pyproject.toml
mkolodner-sc: Actually, we still need to have `"scikit-build-core>=0.10"` be included here, since uv does not install [build-system].requires automatically, so it must be present in the environment before any gigl-core build

.github/workflows/release.yml
```diff
@@ -23,6 +32,9 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
 
+    - name: Install cmake
+      run: sudo apt-get install -y cmake
```
svij-sc: move it to install_cpp_deps.sh?

.github/workflows/release.yml
```diff
@@ -40,20 +53,26 @@ jobs:
...
       run: |
-        uv publish --index gcp-release-registry --username oauth2accesstoken --keyring-provider subprocess
+        # Remove stale cmake cache from previous runs on self-hosted runners.
+        rm -rf gigl-core/.cache/cmake_build
+        uv build --wheel gigl-core/
 
-    - name: Post Publish Package
-      if: always()
-      # Clean up files created during Publish Package step.
+    - name: Publish gigl-core wheel
```
svij-sc, line 64: Should we be building and publishing the gigl-core wheel first and then referencing it in the gigl wheel?

requirements/install_py_deps.sh
```diff
@@ -143,7 +143,7 @@ install_gigl_lib_deps() {
         # https://docs.astral.sh/uv/reference/cli/#uv-sync
         uv sync ${extra_deps_clause[@]} --group dev --locked ${flag_use_inexact_match}
     else
-        uv sync ${extra_deps_clause[@]} --locked ${flag_use_inexact_match}
+        uv sync ${extra_deps_clause[@]} --group build-backend --locked ${flag_use_inexact_match}
```
svij-sc: why?
this is the prod path, meaning we should not need build-backend since gigl-core should be prebuilt.

scripts/generate_compile_commands.py
```diff
@@ -0,0 +1,78 @@
...
+    COMPILE_COMMANDS.write_text(raw_path.read_text())
+
+
+def main() -> None:
+    write_compile_commands()
+    print(f"Wrote {COMPILE_COMMANDS}")
+
+
+if __name__ == "__main__":
+    main()
```
svij-sc: is this file still needed with the changes?

scripts/run_cpp_lint.py
```diff
@@ -0,0 +1,86 @@
...
+            for d in failures[source]:
+                print(f"        {d}")
+        print(
+            "\nRun \033[1mmake fix_lint_cpp\033[0m to auto-fix violations where possible."
+        )
+        sys.exit(1)
+
+
+if __name__ == "__main__":
+    main()
```
svij-sc, line 91: is this file still needed with the changes?

Makefile
svij-sc: can we identify where there is a side effect of invalidating this file?

Makefile
```diff
@@ -311,7 +365,11 @@ clean_build_files_scala:
 	( cd scala; sbt clean; find . -type d -name "target" -prune -exec rm -rf {} \; )
 	( cd scala_spark35; sbt clean; find . -type d -name "target" -prune -exec rm -rf {} \; )
 
-clean_build_files: clean_build_files_py clean_build_files_scala
+clean_build_files_cpp:
+	rm -rf .cache/cpp_tests
+	rm -f .cache/compile_commands.json
+
+clean_build_files: clean_build_files_py clean_build_files_scala clean_build_files_cpp
```
svij-sc, line 356: nit: Thoughts on. moving this to a Makefile in the cpp directory itself?

pyproject.toml
```diff
@@ -221,30 +247,34 @@ torch_spline_conv = [
...
+# wheel (see gigl-core/pyproject.toml). Release builds two variants per package:
+# cpu and cu128. gigl's build stays fast because no CMake runs here.
 [build-system]
-requires      = ["setuptools>=61.0.0", "wheel"]
+requires      = ["setuptools>=68"]
 build-backend = "setuptools.build_meta"
 
-
 [tool.setuptools.packages.find]
-where = ["."]  # list of folders that contain the packages
```
svij-sc, line 234: i guess we dont need this anymore? It works without - i.e. validated that new wheel when installed we can import gigl, et al in a fresh python intepreter.

pyproject.toml
```diff
@@ -99,9 +102,18 @@ required-environments = [
...
+# gigl-core must build against torch, which is not on PyPI and cannot be installed
+# into an isolated build environment. Disabling isolation lets cmake find torch from
+# the ambient venv (installed by make install_dev_deps / base Docker images).
+no-build-isolation-package = ["gigl-core"]
 
 [dependency-groups]
+# scikit-build-core is gigl-core's PEP 517 build backend. With no-build-isolation-package
+# set, uv does not install [build-system].requires automatically, so it must be present
+# in the ambient environment before any gigl-core build (uv sync, uv build, Dockerfiles).
+build-backend = ["scikit-build-core>=0.10"]
```
svij-sc: nit: Rename `gigl-core-build-backend`

gigl-core/requirements/install_cpp_deps.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
svij-sc, line 1: nit: should we move this to inside gigl-core?

.github/workflows/release.yml
svij-sc: thanks, lets add a note to remove those secrets as a follow up?

svij-sc (commented):
Thanks, this is looking great!
Last few comments/questions

.github/workflows/release.yml
mkolodner-sc: Good callout. This can actually be removed altogether with a re-ordering of the installation of cpp_deps and the py_deps in the makefile.

.github/workflows/release.yml
mkolodner-sc: The build wasn't failing with the order we had previously since gigl is a pure-Python package. uv build --wheel just packages the Python source files and embeds gigl-core as dependency metadata; nothing imports or compiles against gigl-core at build time.

That being said, there was a potential race condition if a user installs gigl in a window where gigl-core has been published but gigl hasn't, so it's still more correct to swap the order here. This has been done.

requirements/install_py_deps.sh
mkolodner-sc: We need this currently since  `gigl-core = { path = "gigl-core" }` in `[tool.uv.sources]` causes uv sync to build gigl-core from source even in base images — the stub build (no C++ files) still invokes scikit-build-core as the PEP 517 build backend. Without it, uv sync --locked fails.

scripts/run_cpp_lint.py
mkolodner-sc: This is a convenience wrapper around clangd to make the lint check more parsable. Preference to keep this for ease of debugging lint errors in cpp.

docs/cpp_style_guide.md
```diff
@@ -0,0 +1,194 @@
...
+| `modernize-loop-convert.NamingStyle`                       | `camelBack`            | Auto-generated loop variable names use camelBack, matching `readability-identifier-naming.VariableCase`                                                                               |
+| `readability-function-size.LineThreshold`                  | `1000`                 | Functions over 1000 lines are flagged                                                                                                                                                 |
+| `readability-braces-around-statements.ShortStatementLines` | `0`                    | Braces required for all control-flow bodies, even single-line                                                                                                                         |
+
+______________________________________________________________________
+
+## pybind11 Extension Modules
+
+Extension modules live under `gigl/csrc/`.
```
kmontemayor2-sc: How does `gigl-core/src/gigl_core` ( ex `gigl-core/src/gigl_core/__init__.py`) interplace with this?

docs/cpp_style_guide.md
```diff
@@ -0,0 +1,194 @@
...
+| -------------------------- | ---------------------------------------------------------------- |
+| `python_<name>.cpp`        | pybind11 bindings — contains the `PYBIND11_MODULE` definition    |
+| `<name>.cpp` / `<name>.cu` | Implementation — function and class definitions                  |
+| `<name>.h`                 | Declarations (function signatures, class definitions, constants) |
+
+Example: to add a `my_op` extension under `gigl/csrc/sampling/`:
+
+```
+gigl/csrc/sampling/python_my_op.cpp   ← pybind11 bindings
+gigl/csrc/sampling/my_op.cpp          ← implementation
```
kmontemayor2-sc: ditto on updating paths?

RELEASING.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Two-wheel model
+
+GiGL is distributed as two wheels that are always installed together:
+
+- **`gigl`** — pure Python (same wheel for CPU and CUDA users)
+- **`gigl-core`** — compiled C++/CUDA extensions, ABI-bound to the torch variant
+
+Both wheels are always versioned and released together. `bump_version.py` updates both versions and keeps the
+`gigl-core` pin in `gigl/pyproject.toml` in sync automatically.
```
kmontemayor2-sc: nit. just `pyproject.toml`? Since we this pyproject lives in project root, and not in `GiGL/gigl`?

scripts/generate_compile_commands.py
mkolodner-sc: Identified a way to remove our need for this file, thanks for the callout!

Makefile
```diff
@@ -94,7 +99,16 @@ unit_test_scala: clean_build_files_scala
 # Eventually, we should look into splitting these up.
 # We run `make check_format` separately instead of as a dependent make rule so that it always runs after the actual testing.
 # We don't want to fail the tests due to non-conformant formatting during development.
-unit_test: precondition_tests unit_test_py unit_test_scala
+.cache/cpp_tests/.configured: gigl-core/CMakeLists.txt gigl-core/tests/CMakeLists.txt gigl-core/.cache/cmake_build/CMakeInit.txt
```
kmontemayor2-sc: it this a make rule? IMO it's a bit odd to have `/` in the make rule names?

kmontemayor2-sc (commented):
Thanks for all the hard work here, Matt :)

pyproject.toml
mkolodner-sc: yup, where = ["."] was redundant — setuptools defaults to searching from the project root when where is omitted. The include filter is doing all the real work here. This was validated with new wheel installation and importing gigl in a new python interpreter.

Makefile
mkolodner-sc: For file targets, where make checks whether that file exists on disk and compares its timestamp against the prerequisites, this pattern is expected.

Makefile
mkolodner-sc: There should not be any unexpected side effects here. To clarify what each piece does:

  - .cache/cpp_tests/.configured is the file target — make checks if this file exists and is newer than the prerequisites                                                                                                                           
  - touch .cache/cpp_tests/.configured at the end of the recipe creates/updates that file's timestamp, telling make "this target is now current"
  - Without the touch, cmake would re-run on every make unit_test_cpp because the file would never exist (cmake configures into .cache/cpp_tests/ but doesn't create a file named .configured)                                                      
                                                                                                                                                                                                                                                    
  The difference from the reviewer's marco example is that our target has prerequisites (CMakeLists.txt, tests/CMakeLists.txt, CMakeInit.txt) defined above:

```
.cache/cpp_tests/.configured: CMakeLists.txt tests/CMakeLists.txt .cache/cmake_build/CMakeInit.txt   
```
So the rule re-runs not just when .configured is deleted, but also when any of those files are newer than .configured — which is what triggers re-configuration when you change CMakeLists.txt. The marco example has no prerequisites so it truly only runs once.

.github/workflows/release.yml
```diff
@@ -12,8 +12,17 @@ permissions:
 
 jobs:
   build:
-    name: Build and release pip whl
-    runs-on: ubuntu-latest
+    name: Build and release pip whl (${{ matrix.torch-variant }})
+    runs-on: ${{ matrix.runner }}
```
svij-sc: slight nit: move runs-on below the definition of strategy/matrix
(variable used before reference)

gigl-core/pyproject.toml
```diff
@@ -0,0 +1,36 @@
+[project]
```
svij-sc, line 1: we should persist the lock file as well and ensure any uv commands for this target is using `--locked` accordingly.

svij-sc (approved):
Thanks!
Once the subsequent PR is in, lets do a formal release and test everything

.github/workflows/on-pr-merge.yml
```diff
@@ -70,6 +70,17 @@ jobs:
         service_account:  ${{ secrets.gcp_service_account_email }}
         project:  ${{ vars.GCP_PROJECT_ID }}
 
+  ci-unit-test-cpp:
```
kmontemayor2-sc, line 73: BTW we need to add this to our merge checks. https://github.com/Snapchat/GiGL/settings/rules/5062740 "Require status checks to pass"

kmontemayor2-sc (approved):

PR #577, opened 2026-04-06
Add SharedDistSamplingBackend — multi-channel sampling backend

kmontemayor2-sc:
Part of a larger change to support multiple cleints reading from the same backend.
We do this for graph store mode so that we can have the "full" storage cluster backend rpc, which is required by GLT, but we can only "read" from one rank per client (this goes along with CONTIGUOUS). 

Introduce SharedDistSamplingBackend which manages a pool of worker
processes servicing multiple compute-rank channels through a fair-queued
round-robin scheduler. This replaces the per-channel producer model in
graph-store mode with a shared backend + lightweight per-channel state.

Includes tests for pure business logic helpers (_compute_num_batches,
_epoch_batch_indices, _compute_worker_seeds_ranges), shuffle behavior,
and completion reporting.

### DIFF
--- a/gigl/distributed/graph_store/shared_dist_sampling_producer.py
+++ b/gigl/distributed/graph_store/shared_dist_sampling_producer.py
@@ -1,0 +1,814 @@
+"""Shared graph-store sampling backend and fair-queued worker loop.
+
+This module implements the multi-channel sampling backend used in graph-store
+mode.  A single ``SharedDistSamplingBackend`` per loader instance manages a
+pool of worker processes that service many compute-rank channels through a
+fair-queued scheduler (``_shared_sampling_worker_loop``).
+"""
+
+import datetime
+import queue
+import threading
+import time
+from collections import defaultdict, deque
+from dataclasses import dataclass
+from enum import Enum, auto
+from multiprocessing.process import BaseProcess
+from threading import Barrier
+from typing import Optional, Union, cast
+
+import torch
+import torch.multiprocessing as mp
+from graphlearn_torch.channel import ChannelBase
+from graphlearn_torch.distributed import (
+    DistDataset,
+    RemoteDistSamplingWorkerOptions,
+    get_context,
+    init_rpc,
+    init_worker_group,
+    shutdown_rpc,
+)
+from graphlearn_torch.distributed.dist_sampling_producer import MP_STATUS_CHECK_INTERVAL
+from graphlearn_torch.sampler import (
+    EdgeSamplerInput,
+    NodeSamplerInput,
+    SamplingConfig,
+    SamplingType,
+)
+from graphlearn_torch.typing import EdgeType
+from torch._C import _set_worker_signal_handlers
+
+from gigl.common.logger import Logger
+from gigl.distributed.sampler_options import PPRSamplerOptions, SamplerOptions
+from gigl.distributed.utils.dist_sampler import (
+    SamplerInput,
+    SamplerRuntime,
+    create_dist_sampler,
+)
+
+logger = Logger()
+
+
+def _prepare_degree_tensors(
+    data: DistDataset,
+    sampler_options: SamplerOptions,
+) -> Optional[Union[torch.Tensor, dict[EdgeType, torch.Tensor]]]:
+    """Materialize PPR degree tensors before worker spawn when required."""
+    degree_tensors: Optional[Union[torch.Tensor, dict[EdgeType, torch.Tensor]]] = None
+    if isinstance(sampler_options, PPRSamplerOptions):
+        degree_tensors = data.degree_tensor
+        if isinstance(degree_tensors, dict):
+            logger.info(
+                "Pre-computed degree tensors for PPR sampling across "
+                f"{len(degree_tensors)} edge types."
+            )
+        elif degree_tensors is not None:
+            logger.info(
+                "Pre-computed degree tensor for PPR sampling with "
+                f"{degree_tensors.size(0)} nodes."
+            )
+    return degree_tensors
+
+
+EPOCH_DONE_EVENT = "EPOCH_DONE"
+SCHEDULER_TICK_SECS = 0.05
+SCHEDULER_STATE_LOG_INTERVAL_SECS = 10.0
+SCHEDULER_STATE_MAX_CHANNELS = 6
+SCHEDULER_SLOW_SUBMIT_SECS = 1.0
+
+
+class SharedMpCommand(Enum):
+    REGISTER_INPUT = auto()
+    UNREGISTER_INPUT = auto()
+    START_EPOCH = auto()
+    STOP = auto()
+
+
+@dataclass(frozen=True)
+class RegisterInputCmd:
+    channel_id: int
+    worker_key: str
+    sampler_input: SamplerInput
+    sampling_config: SamplingConfig
+    channel: ChannelBase
+
+
+@dataclass(frozen=True)
+class StartEpochCmd:
+    channel_id: int
+    epoch: int
+    seeds_index: Optional[torch.Tensor]
+
+
+@dataclass
+class ActiveEpochState:
+    channel_id: int
+    epoch: int
+    input_len: int
+    batch_size: int
+    drop_last: bool
+    seeds_index: Optional[torch.Tensor]
+    total_batches: int
+    submitted_batches: int = 0
+    completed_batches: int = 0
+    cancelled: bool = False
+
+
+def _command_channel_id(command: SharedMpCommand, payload: object) -> Optional[int]:
+    """Extract the channel id from a worker command payload."""
+    if command == SharedMpCommand.STOP:
+        return None
+    if isinstance(payload, RegisterInputCmd):
+        return payload.channel_id
+    if isinstance(payload, StartEpochCmd):
+        return payload.channel_id
+    if isinstance(payload, int):
+        return payload
+    return None
+
+
+def _compute_num_batches(input_len: int, batch_size: int, drop_last: bool) -> int:
+    """Compute the number of batches emitted for an input length."""
+    if input_len <= 0:
+        return 0
+    if drop_last:
+        return input_len // batch_size
+    return (input_len + batch_size - 1) // batch_size
+
+
+def _epoch_batch_indices(state: ActiveEpochState) -> Optional[torch.Tensor]:
+    """Return the next batch of indices for an active epoch.
+
+    Returns the index tensor for the next batch, or None if no more batches
+    should be submitted (epoch cancelled, all batches already submitted, or
+    incomplete final batch with drop_last=True).
+    """
+    if state.cancelled or state.submitted_batches >= state.total_batches:
+        return None
+
+    batch_start = state.submitted_batches * state.batch_size
+    batch_end = min(batch_start + state.batch_size, state.input_len)
+    if state.drop_last and batch_end - batch_start < state.batch_size:
+        return None
+
+    if state.seeds_index is None:
+        return torch.arange(batch_start, batch_end, dtype=torch.long)
+    return state.seeds_index[batch_start:batch_end]
+
+
+def _compute_worker_seeds_ranges(
+    input_len: int, batch_size: int, num_workers: int
+) -> list[tuple[int, int]]:
+    """Distribute complete batches across workers like GLT's producer does."""
+    num_worker_batches = [0] * num_workers
+    num_total_complete_batches = input_len // batch_size
+    for rank in range(num_workers):
+        num_worker_batches[rank] += num_total_complete_batches // num_workers
+    for rank in range(num_total_complete_batches % num_workers):
+        num_worker_batches[rank] += 1
+
+    index_ranges: list[tuple[int, int]] = []
+    start = 0
+    for rank in range(num_workers):
+        end = start + num_worker_batches[rank] * batch_size
+        if rank == num_workers - 1:
+            end = input_len
+        index_ranges.append((start, end))
+        start = end
+    return index_ranges
+
+
+def _shared_sampling_worker_loop(
+    rank: int,
+    data: DistDataset,
+    worker_options: RemoteDistSamplingWorkerOptions,
+    task_queue: mp.Queue,
+    event_queue: mp.Queue,
+    mp_barrier: Barrier,
+    sampler_options: SamplerOptions,
+    degree_tensors: Optional[Union[torch.Tensor, dict[EdgeType, torch.Tensor]]],
+) -> None:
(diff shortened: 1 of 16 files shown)


gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+    batch_size: int
+    drop_last: bool
+    seeds_index: Optional[torch.Tensor]
+    total_batches: int
+    submitted_batches: int = 0
+    completed_batches: int = 0
+    cancelled: bool = False
+
+
+def _command_channel_id(command: SharedMpCommand, payload: object) -> Optional[int]:
```
mkolodner-sc: nit: payload: object --> Can we be more specific on the possible inputs here?

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+            if state.submitted_batches < state.total_batches and not state.cancelled:
+                runnable_channel_ids.append(channel_id)
+                runnable_channel_id_set.add(channel_id)
+
+        sampler_input = channel_input[batch_indices]
+
+        callback = lambda _: _on_batch_done(channel_id, current_epoch)
+        if cfg.sampling_type == SamplingType.NODE:
+            sampler.sample_from_nodes(
+                cast(NodeSamplerInput, sampler_input), callback=callback
```
mkolodner-sc: nit: any way we can avoid using cast here?

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+            if state.completed_batches == state.total_batches:
+                active_epoch_by_channel_id.pop(channel_id, None)
+                event_queue.put((EPOCH_DONE_EVENT, channel_id, epoch, rank))
+            if (
+                channel_id in removing_channel_ids
+                and state.completed_batches == state.submitted_batches
+            ):
+                _clear_registered_input_locked(channel_id)
+
+    def _submit_one_batch(channel_id: int) -> bool:
```
mkolodner-sc, line 518: What's the behavior if sampler.sample_from_nodes() raises an exception? Ideally it shouldn't be, but we should still make sure we fail gracefully in that setting if possible. 

Currently, `_submit_one_batch` releases state_lock before calling sampler.sample_from_nodes(), which is correct for avoiding long lock holds. 

However, it re-enqueues the channel into runnable_channel_ids inside the lock, before the batch is actually submitted. If the sampler call raises an exception, submitted_batches has already been incremented and the channel has been re-enqueued, but no callback will ever fire. This means _on_batch_done will never be called for that batch, so the epoch will never complete — it silently hangs.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+            sampler.sample_from_edges(
+                cast(EdgeSamplerInput, sampler_input), callback=callback
+            )
+        elif cfg.sampling_type == SamplingType.SUBGRAPH:
+            sampler.subgraph(cast(NodeSamplerInput, sampler_input), callback=callback)
+        else:
+            raise RuntimeError(f"Unsupported sampling type: {cfg.sampling_type}")
+        return True
+
+    def _pump_runnable_channel_ids() -> bool:
```
mkolodner-sc, line 576: Curious if throughput be improved here

Between reading num_candidates and the loop iterations, _on_batch_done callbacks (running on sampler threads) can re-enqueue channels via _submit_one_batch's lock-held section. So the channels may grow during the iteration, but we only process up to num_candidates entries meaning that newly re-enqueued channels from this round could be skipped until the next pump cycle.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+            self._drain_events()
+            sampling_config = self._channel_sampling_config[channel_id]
+            if self._channel_epoch[channel_id] >= epoch:
+                return
+            previous_epoch = self._channel_epoch[channel_id]
+            self._channel_epoch[channel_id] = epoch
+            stale_keys = [
+                k
+                for k in self._completed_workers
+                if k[0] == channel_id and k[1] <= epoch
```
mkolodner-sc, line 1012: Should this be k[1] < epoch?

Currently, this deletes completion records for the current epoch (<= epoch, not < epoch). If some workers have already finished the current epoch before the main thread gets here, their completion records are wiped. Then is_channel_epoch_done would never see those workers as complete — the epoch would appear stuck.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+        while True:
+            try:
+                event = self._event_queue.get_nowait()
+            except queue.Empty:
+                return
+            if event[0] == EPOCH_DONE_EVENT:
+                _, channel_id, epoch, worker_rank = event
+                self._completed_workers[(channel_id, epoch)].add(worker_rank)
+
+    def start_new_epoch_sampling(self, channel_id: int, epoch: int) -> None:
```
mkolodner-sc, line 969: Are there guarantees here that the previous epoch has finished by the time this is called? Otherwise, we could accidentally overwrite `active_epoch_by_channel_id[channel_id]`.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+            return
+        with state_lock:
+            scheduler_state = _format_scheduler_state_locked()
+        logger.info(
+            f"shared_sampling_scheduler worker_rank={rank} reason={reason} "
+            f"{scheduler_state}"
+        )
+        last_state_log_time = now
+
+    def _on_batch_done(channel_id: int, epoch: int) -> None:
```
mkolodner-sc, line 491: One potential race condition identified by the robots:

This runs on a sampler callback thread. Meanwhile _handle_command for a new REGISTER_INPUT with the same channel_id could be executing on the main worker thread. Since _clear_registered_input_locked pops from all the dicts, and _handle_command(REGISTER_INPUT) inserts into them, there's a window where the cleanup wipes out a freshly registered channel. The state_lock is held, so it won't be concurrent, but the ordering could be: register → unregister arrives → callback fires and clears the newly registered data.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+                        None,
+                    )
+                for worker in self._workers:
+                    worker.join(timeout=MP_STATUS_CHECK_INTERVAL)
+                for queue_ in self._task_queues:
+                    queue_.cancel_join_thread()
+                    queue_.close()
+                if self._event_queue is not None:
+                    self._event_queue.cancel_join_thread()
+                    self._event_queue.close()
```
mkolodner-sc, line 1175: Another potential race condition identified here:

If the event queue has unread items when close() is called, this can cause a BrokenPipeError in worker processes that are still trying to put() into the event queue (e.g., a final EPOCH_DONE_EVENT). The cancel_join_thread() prevents the feeder thread from blocking, but the workers may not have exited yet — there's only a join(timeout=MP_STATUS_CHECK_INTERVAL) which is non-blocking after the timeout. Workers that are still in sampler callbacks could hit this.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+    runnable_channel_ids: deque[int] = deque()
+    runnable_channel_id_set: set[int] = set()
+    removing_channel_ids: set[int] = set()
+    state_lock = threading.RLock()
+    last_state_log_time = 0.0
+    current_device: Optional[torch.device] = None
+
+    # --- Scheduler helper functions ---
+
+    def _enqueue_channel_if_runnable_locked(channel_id: int) -> None:
```
mkolodner-sc, line 399: nit: should we have a check here to ensure that the caller has the state lock?

gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -0,0 +1,1105 @@
...
+                break
+
+            # Phase 3: If idle (no commands, no batches), block until next command.
+            if not (processed_command or made_progress):
+                try:
+                    command, payload = task_queue.get(timeout=SCHEDULER_TICK_SECS)
+                except queue.Empty:
+                    continue
+                keep_running = _handle_command(command, payload)
+    except KeyboardInterrupt:
```
mkolodner-sc, line 739: Why do we need this except?

mkolodner-sc (commented):
Thanks Kyle! Took a pass and left some comments. A general note: while there are useful docstrings, some of these functions still have some complex logic inside of them -- some more in-line comments would be helpful for improving readability/understanding of the mechanisms here.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: `START_EPOCH` commands are dispatched later in this same method (after the cleanup). No worker can have completed an epoch that hasn't been dispatched yet, so there can't be any `_completed_workers` entries for the new epoch at this point. `<= epoch` is equivalent to `< epoch` here and is slightly more defensive.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: Fixed — introduced a `CommandPayload = Union[RegisterInputCmd, StartEpochCmd, int, None]` type alias and updated `_command_channel_id`, `_handle_command`, and `_enqueue_worker_command` signatures.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: Valid theoretical issue. If UNREGISTER defers cleanup (in-flight batches) and then REGISTER arrives for the same `channel_id`, the deferred `_clear_registered_input_locked` call from `_on_batch_done` would wipe the new registration.

The proper fix is a per-registration generation token — but that's a larger design change. In practice, `DistServer` allocates `channel_id` from a monotonically increasing counter, so reuse doesn't happen with current callers. Added defensive comments documenting this assumption in both `_clear_registered_input_locked` and the `REGISTER_INPUT` handler. Will address with a generation token in a follow-up if needed.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: Good catch. If the sampler call raises, `submitted_batches` is already incremented but the callback never fires — silent hang.

Fixed with two changes:
1. **Worker side**: wrapped the sampler call in try/except. On failure, logs the error with full context (worker_rank, channel_id, epoch, batch index) and re-raises. The exception kills the worker process (cleanup runs via `finally`).
2. **Parent side**: `is_channel_epoch_done` now checks worker liveness via `_check_workers_alive()`. If any worker process has died, it raises `RuntimeError` immediately instead of returning False and letting the caller poll forever.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: We could type narrow with `assert isinstance` but tbch I'd prefer to avoid adding more code to our hot loop.

If you strongly prefer we can do the assert but it would be a bit slower (I haven't benchmarked.).

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: Hmmm, it seems like you can't really tell if a lock is owned (`.locked` is only exposed in py >= 3.14).

Added a todo to add a check later :/

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: We have the same except in the existing loop. [except KeyboardInterrupt:](https://github.com/Snapchat/GiGL/blob/kmonte/shared-backend-decomp-2/gigl/distributed/dist_sampling_producer.py#L165-L166)

I suppose it's not strictly needed but it's nice for cleaning up when we run this locally?

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: The snapshot behavior is intentional for fairness. Processing newly re-enqueued channels within the same pump cycle would let fast channels get multiple submissions per cycle, starving slower ones. The outer event loop runs immediately after (no sleep between phases), so the "delay" is one loop iteration (~microseconds). Added an inline comment explaining this.


Also note that we will have a follow up to weight the different channels by input size, (otherwise we can see starvation too)

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: The guarantee comes from the caller: `BaseDistLoader.__iter__` consumes all batches via `__next__` before re-entering the for-loop, which gates the next `__iter__` call. But this wasn't documented or enforced.

Fixed:
1. Added docstring note documenting the caller contract.
2. Added `logger.warning` that fires if `start_new_epoch_sampling` is called while the previous epoch hasn't completed on all workers.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
kmontemayor2-sc: aThis is inherent to multiprocessing shutdown. Workers are daemon processes. The `finally` block (lines 1102-1105) terminates any still-alive workers after the join timeout. A BrokenPipeError in a daemon worker during teardown is expected and harmless — the process exits regardless. The `cancel_join_thread()` call is specifically to prevent the main process from deadlocking on queue cleanup during shutdown.

mkolodner-sc (approved):
Thanks Kyle! TBH this one was a bit hard for me to wrap my head around, but I'll stamp to unblock. In the future, it might be helpful if we could have a doc to go over all of the moving pieces here, or a knowledge sharing session could be useful as well.

svij-sc (approved):
stamping

PR #578, opened 2026-04-06
Add two-phase sampling API to DistServer

kmontemayor2-sc:
Replace the single-step create_sampling_producer with a two-phase API:
- init_sampling_backend: creates/reuses a SharedDistSamplingBackend
- register_sampling_input: registers a lightweight per-channel input

We do this so we can re-use the sampler backends across the storage cluster, this greatly improves on cluster stability and lets us save on process_start_gap_seconds time.

I will have a followup PR (https://github.com/Snapchat/GiGL/pull/579) where we have BaseDistLoader use the new two-phase API, but for now we just delegate to the two phases in DistServer.
Note, we really should have that follow up, as this approach means we'd be creating one dist sampling process tree per input still, which we are trying to avoid (but should be fine as a temporary standin to help make the reviews easier).

The existing create_sampling_producer/destroy_sampling_producer methods
are preserved as bridge methods that delegate to the new API, keeping
existing loaders working without changes.

Also adds InitSamplingBackendRequest and RegisterBackendRequest message
dataclasses, and per-channel fetch stats logging.


gigl/distributed/graph_store/dist_server.py
```diff
@@ -412,6 +467,143 @@ def get_ablp_input(
...
+                    sampling_config=opts.sampling_config,
+                    sampler_options=opts.sampler_options,
+                    # We only need degree tensor for PPR sampling
+                    degree_tensors=self.dataset.degree_tensor
+                    if isinstance(opts.sampler_options, PPRSamplerOptions)
+                    else None,
+                ),
+            )
+            self._backend_id_by_backend_key[opts.backend_key] = backend_id
+            self._backend_state_by_backend_id[backend_id] = backend_state
```
mkolodner-sc: Potential race condition: The backend is inserted into _backend_id_by_backend_key while holding _lock, but init_backend() is called outside it. 

A second concurrent caller for the same backend_key will find the ID in the map and return immediately — before initialization completes. That caller can then call register_sampling_input against a half-initialized backend.

gigl/distributed/graph_store/dist_server.py
```diff
@@ -421,136 +613,163 @@ def create_sampling_producer(
...
+                channel_state.backend_id
+            )
+        if backend_state is None:
+            raise RuntimeError(
+                f"start_new_epoch_sampling: backend for channel_id={channel_id} "
+                f"backend_id={channel_state.backend_id} not found"
+            )
+
+        if channel_state.epoch >= epoch:
+            return
```
mkolodner-sc: qq -- why do we return here?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -421,136 +613,163 @@ def create_sampling_producer(
...
+        channel_state.epoch = epoch
         logger.info(
-            f"Request to create producer for worker key {worker_options.worker_key} took {request_end_time - request_start_time:.2f}s"
+            f"Starting epoch channel_id={channel_id} backend_id={channel_state.backend_id} "
+            f"epoch={epoch}"
         )
-        return producer_id
+        backend_state.runtime.start_new_epoch_sampling(channel_id, epoch)
```
mkolodner-sc: Do we need to use the channel's lock when running this?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -421,136 +613,163 @@ def create_sampling_producer(
...
+            while True:
+                try:
+                    msg = channel_state.channel.recv(timeout_ms=100)
+                    self._log_fetch_stats_if_due(
+                        channel_id,
+                        channel_state.worker_key,
+                        time.monotonic() - request_start_time,
+                    )
+                    return msg, False
+                except QueueTimeoutError:
```
mkolodner-sc, line 845: When is this expected to happen?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -421,136 +613,163 @@ def create_sampling_producer(
...
-                return msg, False
-            except QueueTimeoutError as e:
-                if producer.is_all_sampling_completed():
-                    return None, True
+
+        with channel_state.lock:
+            while True:
+                try:
+                    msg = channel_state.channel.recv(timeout_ms=100)
+                    self._log_fetch_stats_if_due(
```
mkolodner-sc, line 839: Should we make the logging of the stats configurable, rather than always logging here?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -412,6 +467,143 @@ def get_ablp_input(
...
+    def destroy_sampling_input(self, channel_id: int) -> None:
+        """Destroy one registered sampling channel and maybe its backend.
+
+        If this is the last channel on the backend, the backend is shut down
+        and removed.
+
+        Args:
+            channel_id: The ID of the channel to destroy.
+        """
+        self._fetch_stats_by_channel_id.pop(channel_id, None)
```
mkolodner-sc: Any reason this isn't in the lock?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -66,34 +118,37 @@ class DistServer:
...
     def shutdown(self) -> None:
-        for producer_id in list(self._producer_pool.keys()):
-            self.destroy_sampling_producer(producer_id)
-        assert len(self._producer_pool) == 0
-        assert len(self._msg_buffer_pool) == 0
+        with self._lock:
+            backends = list(self._backend_state_by_backend_id.values())
+            self._backend_id_by_backend_key.clear()
+            self._backend_state_by_backend_id.clear()
+            self._channel_state_by_channel_id.clear()
```
mkolodner-sc, line 146: should we clear the `_fetch_stats_by_channel_id` here as well

gigl/distributed/graph_store/dist_server.py
```diff
@@ -412,6 +467,143 @@ def get_ablp_input(
...
+        except Exception:
+            with self._lock:
+                self._channel_state_by_channel_id.pop(channel_id, None)
+                backend_state.active_channels.discard(channel_id)
+            raise
+
+        logger.info(
+            f"Registered sampling input backend_id={opts.backend_id} "
+            f"channel_id={channel_id} worker_key={opts.worker_key} "
+            f"active_channels={len(backend_state.active_channels)} "
```
mkolodner-sc: should we save the value of `len(backend_state.active_channels)` within the lock, and then log that here instead, in case it changes after the lock is dropped?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -66,34 +118,37 @@ class DistServer:
     Args:
       dataset (DistDataset): The ``DistDataset`` object of a partition of graph
         data and feature data, along with distributed patition books.
```
mkolodner-sc: oops just noticed this --> partition

mkolodner-sc (commented):
Thanks Kyle! Did a pass here

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Expected and frequent — same `recv(timeout_ms=100)` + `except QueueTimeoutError` loop pattern as upstream `graphlearn_torch/distributed/dist_server.py:216-222`. `ShmChannel.recv` raises `QueueTimeoutError` every 100ms the producer has nothing ready, and we loop back to check `is_channel_epoch_done`.
  Added an inline comment at the `except` documenting this.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Fixed — also fixed the same typo in `init_server`'s docstring at L807.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Fixed. Added `self._fetch_stats_by_channel_id.clear()` inside the existing `with self._lock:` block in `shutdown()`, next to the other dict clears.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: no reason - done.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Thanks — worth digging into.

  Minor nuance first: the "half-initialized backend" phrasing doesn't quite match runtime behavior. `SharedDistSamplingBackend.init_backend()` and `register_input()` both take the backend's internal `_lock` (shared_dist_sampling_producer.py:816-851, 907), so a concurrent second caller's
  `register_sampling_input` would block on that internal lock until init completed. Today's loaders also always generate per-rank `worker_key`s (`compute_rank_{N}_worker_{M}` / `compute_ablp_loader_rank_{N}_worker_{M}`), so the duplicate-key path isn't reachable from the current loader stack.

  That said, fixed anyway because:
  1. **Failure-path bug.** If the first caller's `init_backend` raises, the rollback removes the backend from both maps — but the second caller already returned a `backend_id` that no longer exists, and then hits `KeyError` at `register_sampling_input`'s bracket access.
  2. **Fragile invariant.** Relying on the internal lock of `SharedDistSamplingBackend` isn't obvious from reading `DistServer`; a future refactor could silently break it.
  3. **Prerequisite** for the PR's stated goal of enabling backend reuse across channels.

  Fix: lock-overlap pattern on `backend_state.lock`. The first caller acquires `backend_state.lock` while still holding `self._lock`, then releases `self._lock` before running `init_backend()` (can't hold `self._lock` across init — it spawns workers and blocks on a Barrier, which would stall all other
  RPCs). The second caller observes the existing backend under `self._lock`, then blocks on `backend_state.lock`; when it acquires, it checks `init_error`/`init_complete` and either re-raises (via `raise ... from` so the original traceback survives) or returns the `backend_id`.

  Regression tests added in `tests/unit/distributed/dist_server_test.py`:
  - `test_concurrent_init_same_key_reuses_backend` — 2 concurrent inits return same id, `init_backend` called once.
  - `test_concurrent_init_failure_propagates_to_second_caller` — second caller sees `RuntimeError` with `__cause__` set to the original exception, not `KeyError`.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Idempotency — if the channel already started this epoch (or a later one), silently skip. Matches the same check in `SharedDistSamplingBackend.start_new_epoch_sampling` at `shared_dist_sampling_producer.py:992-993`. Agreed it wasn't documented — added to the docstring: "No-op if this channel has
  already started `epoch` or a later epoch (idempotent — safe to call repeatedly from retries)."

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Good catch — fixed. `channel_state.epoch` was being read/written at L686-688 with no `channel_state.lock` held, so two concurrent callers with different epoch values could interleave the check-then-set and dispatch to `runtime.start_new_epoch_sampling` in non-monotonic order.

  Wrapped the full check/update/dispatch under `with channel_state.lock:`. Also wrapped the pop + `unregister_input` region of `destroy_sampling_input` under `channel_state.lock`: I went back and checked the pre-two-phase code (`git show 720681e9~1:gigl/distributed/graph_store/dist_server.py`), and the
  old `destroy_sampling_producer` and `start_new_epoch_sampling` both held the same `_producer_lock`, so destroy/start_epoch were serialized before this PR. This PR had inadvertently dropped that serialization; wrapping destroy under `channel_state.lock` restores the pre-refactor invariant via the
  two-phase API's per-channel lock.

  Caller-contract wrinkle: `fetch_one_sampled_message` holds `channel_state.lock` for the entire recv loop, so a destroy issued mid-epoch blocks until fetch exits (or forever if the producer is stuck). In practice this is fine — `BaseDistLoader` always drains the channel (observes `is_done=True`) before
   teardown. Documented this contract in the `destroy_sampling_input` docstring.

  Regression test: `test_concurrent_start_epoch_dispatches_monotonically` — two concurrent `start_new_epoch_sampling` calls with different epochs, verifying monotonic dispatch under the lock.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Slight pushback: logging rate is already configurable via `DistServer.__init__(..., log_every_n: int = 50)` (condition at L716). Callers can set it arbitrarily high to effectively mute.

  A true on/off switch would need plumbing through `init_server()` (`dist_server.py:831` — which doesn't accept a `log_every_n` today) and `run_storage_server()` (`storage_utils.py:177-184` — only threads `num_rpc_threads`/`rpc_timeout`). Reasonable follow-up, but it adds a couple new parameters to
  already-deep signatures on this PR, which I'd rather not widen further. Happy to file a follow-up that lands an end-to-end disable knob as its own focused change if you think it's worth having.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Fixed — capturing `len(backend_state.active_channels)` into `active_channels_at_register` right after `add(channel_id)` under `self._lock`, logging the local.

mkolodner-sc (approved):
Thanks Kyle!

svij-sc (approved):

PR #579, opened 2026-04-06
Switch loaders to two-phase init and remove bridge methods

kmontemayor2-sc:
Switch loaders to two-phase init and remove bridge methods
Refactor BaseDistLoader to use the two-phase sampling API directly:
- init_sampling_backend (shared across all ranks per loader instance)
- register_sampling_input (unique per compute rank)

Key changes:
- Add GroupLeaderInfo, _compute_group_leader, _dispatch_grouped_graph_store_phase
  for generic leader-elected grouped RPC dispatch
- Add _init_graph_store_sampling_backends and _register_graph_store_sampling_inputs
- Replace _producer_id_list with _backend_id_list + _channel_id_list
- Remove create_sampling_producer/destroy_sampling_producer bridge methods
- Keep per-class _counter in each loader (not a global counter) since
  type-prefixed _backend_key already prevents cross-type collisions
- Fix test_multiple_loaders_in_graph_store to use num_compute_nodes=2
  so backend-sharing assertions are exercised across ranks


gigl/distributed/base_dist_loader.py
```diff
@@ -78,6 +83,117 @@ class DistributedRuntimeInfo:
...
 
 
+@dataclass(frozen=True)
+class GroupLeaderInfo:
+    """Leader election result for one graph-store dispatch group.
+
+    Args:
+        leader_rank: The global rank elected as leader for this key group.
+        is_leader: Whether the current rank is the leader.
+        my_batch: The stagger batch index for this key.
```
mkolodner-sc: nit -- can we give a better name than my_batch?

gigl/distributed/base_dist_loader.py
```diff
@@ -78,6 +83,117 @@ class DistributedRuntimeInfo:
...
+        num_batches: Total number of stagger batches.
```
mkolodner-sc: ditto -- num_batches is a bit unclear -- can we rename to `num_stagger_batches`

gigl/distributed/base_dist_loader.py
```diff
@@ -78,6 +83,117 @@ class DistributedRuntimeInfo:
...
+    leader_rank: int
+    is_leader: bool
+    my_batch: int
+    num_batches: int
+    stagger_sleep: float
+    group_size: int
+
+
+def _compute_group_leader(
+    my_key: str,
```
mkolodner-sc: meganit: slight pref to move away from `my` var naming in favor of `cur_key`

gigl/distributed/base_dist_loader.py
```diff
@@ -78,6 +83,117 @@ class DistributedRuntimeInfo:
...
+    all_keys: list[Optional[str]],
```
mkolodner-sc: qq -- why is this optional?

gigl/distributed/base_dist_loader.py
```diff
@@ -78,6 +83,117 @@ class DistributedRuntimeInfo:
...
+    """
+    key_to_ranks: dict[str, list[int]] = defaultdict(list)
+    for other_rank, key in enumerate(all_keys):
+        assert key is not None, f"Rank {other_rank} did not provide a key."
+        key_to_ranks[key].append(other_rank)
+
+    leader_rank = min(key_to_ranks[my_key])
+    unique_keys = sorted(key_to_ranks.keys())
+    my_key_index = unique_keys.index(my_key)
+    num_batches = math.ceil(len(unique_keys) / max_concurrent_producer_inits)
```
mkolodner-sc: nit: maybe add a comment explaining why we are staggering batches this way?

gigl/distributed/base_dist_loader.py
```diff
@@ -633,88 +759,159 @@ def _init_colocated_connections(
...
+            == len(backend_id_list)
+            == len(self._input_data_list)
+        ), (
+            f"Mismatched lengths: server_rank_list={len(self._server_rank_list)}, "
+            f"backend_id_list={len(backend_id_list)}, "
+            f"input_data_list={len(self._input_data_list)}"
+        )
+        worker_key = self.worker_options.worker_key
+
+        def issue_rpcs() -> list[int]:
```
mkolodner-sc: nit: slight pref to not define and pass in the issue_rpcs callable function this way -- can we declare this outside and not have to pass this through to subsequent functions? (i.e. the group leader will just call issue_rpcs directly, rather than a callable that is provided)

gigl/distributed/base_dist_loader.py
```diff
@@ -264,7 +381,11 @@ def __init__(
...
 
         # --- Mode-specific attributes and connection initialization ---
-        if isinstance(producer, DistSamplingProducer):
+        if (
+            isinstance(dataset, DistDataset)
+            and isinstance(worker_options, MpDistSamplingWorkerOptions)
+            and isinstance(producer, DistSamplingProducer)
+        ):
             assert isinstance(dataset, DistDataset)
             assert isinstance(worker_options, MpDistSamplingWorkerOptions)
```
mkolodner-sc: We can get rid of these asserts right

gigl/distributed/base_dist_loader.py
```diff
@@ -633,88 +759,159 @@ def _init_colocated_connections(
...
-
-          ┌─────────────────────────────────────────────────────────────────────────────┐
-          │  (1) Compute Node 0  →  Storage 0   (4 connections, one per GPU)            │
-          │  (2) Compute Node 0  →  Storage 1   (4 connections, one per GPU)            │
-          │  (3) Compute Node 1  →  Storage 0   (4 connections, one per GPU)            │
-          │  (4) Compute Node 1  →  Storage 1   (4 connections, one per GPU)            │
-          └─────────────────────────────────────────────────────────────────────────────┘
+        - ``_backend_id_list``: one backend per storage server, shared across all
+          compute ranks using the same loader instance (keyed by ``_backend_key``).
+          Use this when the operation targets the backend itself (e.g.
```
mkolodner-sc: supernit: `Use this` -> `Used when`, since I imagine users won't be exposed to this field, which is just an implementation detail

gigl/distributed/base_dist_loader.py
```diff
@@ -633,88 +759,159 @@ def _init_colocated_connections(
...
-                of this size.
+        - ``_channel_id_list``: one channel per storage server, unique to this
+          compute rank.  Use this for per-rank operations (e.g.
+          ``start_new_epoch_sampling``, ``fetch_one_sampled_message``,
+          ``destroy_sampling_input``).
+
+        Example with 2 storage servers and 4 compute ranks sharing one
+        DistNeighborLoader::
+
+            backend_id_list = [0, 1]        # same for all 4 ranks
```
mkolodner-sc: so the possible values here are the # of storage servers -- maybe we can be explicit about this? Is there ever a case where backend_id_list isn't equal to `[0 ... # of storage servers]`?

gigl/distributed/base_dist_loader.py
```diff
@@ -633,88 +759,159 @@ def _init_colocated_connections(
...
+            channel_id_list = [0, 1]        # rank 0
+            channel_id_list = [2, 3]        # rank 1
```
mkolodner-sc: Likewise, might be helpful to go further into detail about the possible values this can take on.  Just to double check. will there be a case ever where multiple ranks might have an overlapping value?

gigl/distributed/base_dist_loader.py
```diff
@@ -633,88 +759,159 @@ def _init_colocated_connections(
...
-                concurrent producer initializations.
-            max_concurrent_producer_inits: Maximum number of leader ranks that
-                may dispatch RPCs concurrently. Leaders are grouped into batches
-                of this size.
+        - ``_channel_id_list``: one channel per storage server, unique to this
+          compute rank.  Use this for per-rank operations (e.g.
+          ``start_new_epoch_sampling``, ``fetch_one_sampled_message``,
+          ``destroy_sampling_input``).
+
+        Example with 2 storage servers and 4 compute ranks sharing one
```
mkolodner-sc: Can you add a note about how compute ranks relates to the # of machines and training/inference processes on the compute resource?

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -624,13 +618,11 @@ def _setup_for_graph_store(
         edge_feature_info = dataset.fetch_edge_feature_info()
         edge_types = dataset.fetch_edge_types()
         compute_rank = torch.distributed.get_rank()
-        worker_key = (
-            f"compute_ablp_loader_rank_{compute_rank}_worker_{self._instance_count}"
-        )
+        self._backend_key = f"dist_ablp_loader_{self._instance_count}"
```
mkolodner-sc: Instead of doing the setting of `self._backend_key` here, are we able to provide it to super.__init__() instead for the base_dist_loader class

mkolodner-sc (commented):
Thanks Kyle! Left a couple small comments, but generally LGTM

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done — and then some. Deleted GroupLeaderInfo entirely as part of another refactor (the leader-election pattern turned out to be unnecessary once DistServer.init_sampling_backend handles dedup internally), so my_batch is gone along with the surrounding dataclass.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done.

gigl/distributed/base_dist_loader.py
```diff
@@ -78,6 +83,117 @@ class DistributedRuntimeInfo:
...
+        process_start_gap_seconds: Delay multiplier per stagger batch.
+        max_concurrent_producer_inits: Max concurrent leaders per batch.
+        issue_phase_rpcs: Callable that issues RPCs and returns a list of results.
+            Only called on the leader rank.
+
+    Returns:
+        The leader's RPC results, broadcast to all ranks in the group.
+    """
+    all_keys: list[Optional[str]] = [None] * runtime.world_size
+    torch.distributed.all_gather_object(all_keys, my_key)
```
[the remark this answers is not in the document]
kmontemayor2-sc: /ar internal distributed training runs inside a trusted VPC where ranks authenticate against torch.distributed's process group — an attacker who controls a rank already owns a training process. Every object-based collective in torch.distributed (all_gather_object, broadcast_object_list, gather_object,
   scatter_object_list) uses pickle, so the finding applies to our entire distributed stack and to GLT. Replacing per-string/int exchanges with byte-tensor plumbing adds substantial code for near-zero threat-model gain. Same posture as everywhere else we use object collectives.

  Also note: after the refactor below (thread 8), this specific call site is gone anyway.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Good catch, done.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Good push — once I re-read DistServer I realized the whole _dispatch_grouped_graph_store_phase helper is unnecessary, not just the callable plumbing:

  1. DistServer.register_sampling_input allocates a fresh channel ID per call and is not keyed on anything server-side. Each compute rank has a unique worker_key, so the all-gather in the current code elects every rank as leader of its own group of 1 and broadcasts that rank's own result back to itself.
   Pure round-trip.
  2. DistServer.init_sampling_backend is already thread-safe and idempotent by backend_key (first caller creates the backend under self._lock, acquires backend_state.lock before releasing self._lock at dist_server.py:517, and runs init_backend() outside the server lock; subsequent callers block on
  backend_state.lock until init completes or fails). So having every rank issue the RPC is safe — the server deduplicates and all callers get the same backend_id.

  Replaced the leader-elect + stagger + broadcast machinery with direct async_request_server calls in both phases. GroupLeaderInfo, _compute_group_leader, and _dispatch_grouped_graph_store_phase are gone; the max_concurrent_producer_inits parameter (which was only wired into the deleted machinery) is
  also removed from the public loader APIs.

  Net: loader __init__ lost ~130 lines, every `ranks × servers` RPC is now just `servers` RPCs on the common path, and the scaling caveat is that if/when we ever need to rate-limit hundreds of concurrent compute ranks hitting a single storage server, that concern deserves its own helper rather than
  living in every loader's hot path.

  Verified with test_multiple_loaders_in_graph_store (4 loaders × 2 compute ranks × 2 servers) — still passes, and the in-test assertions about all ranks sharing backend IDs for one logical loader now rely on server-side dedup rather than leader broadcast.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: If there are multiple instances of loaders (e.g. for training) then they will get sequential backend ids. Updated the comment.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Hmmm, they can have overlapping backend ids, but not channel ids.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: done.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: Done.

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Done. BaseDistLoader.__init__ now accepts a backend_key: Optional[str] = None parameter. _setup_for_graph_store in both DistNeighborLoader and DistABLPLoader returns the key along with the other setup info, and the caller passes it through super().__init__(). The hasattr(self, "_backend_key") fallback
   in the base class is gone, and the graph-store branch asserts that the key is set rather than silently defaulting to None.

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: done.

mkolodner-sc (approved):
Thanks Kyle!

svij-sc (approved):
stamp

PR #580, opened 2026-04-07
Add ability to prefix e2e tests from CLI

kmontemayor2-sc:
**Scope of work done**

Do this so we can more easily differentiate between different CI/CD runs :)


svij-sc (commented):
A slight design choice I made here when implementing this:
A fixed prefix keeps jobs sorted/grouped --> filterable by identity, while a configurable suffix cleanly captures the variable part (environment, version, run) without disrupting any of that. (Similar to how naming files / tables / etc - prefixes are static and suffixes change)

Thus, Id pref if we instead had something like "override_suffix" or something.

What is this for btw? We just want to differentiate jobs that are run by a single person i.e. adding ${whoami} as part of the name?

PR #585, opened 2026-04-13
Replace mypy with ty for static type checking

svij-sc:
ty (from Astral, makers of ruff/uv) replaces mypy as the project's static type checker. 

The ty fixes are introduced here: https://github.com/Snapchat/GiGL/pull/586

Benchmarked on the code base (631 source files):

  | Metric      | mypy     | ty    | Speedup |
  |-------------|----------|-------|---------|
  | Cold cache  | 184.6s   | 0.79s | ~234x   |
  | Warm cache  | 13.9s    | 0.83s | ~17x    |

Changes:
- Add ty to lint deps, remove mypy + mypy-extensions (keep mypy-protobuf for protoc stub generation)
- Migrate mypy.ini config to [tool.ty.*] sections in pyproject.toml
- Update Makefile type_check target: `uv run ty check`
- Downgrade 12 rule categories to "warn" for initial migration; a follow-up PR will fix the 226 warnings and promote them back to errors

**Scope of work done**


pyproject.toml
```diff
@@ -283,3 +282,84 @@ exclude =  [
...
 in-place = true
 recursive = true
+
+[tool.ty.environment]
+python-version = "3.11"
+
+[tool.ty.src]
+exclude = ["*_pb2.py", "*_pb2.pyi", "**/*.ipynb"]
+
+[tool.ty.analysis]
```
svij-sc, line 311: Migrated from mypy.ini  - removed any that could be fixed - thank you robots!

gigl/distributed/dataset_factory.py
```diff
@@ -606,7 +606,7 @@ def build_dataset_from_task_config_uri(
     )
 
     # Need to do this "backwards" so the parent class can be defined first.
-    # Otherwise, mypy complains that:
+    # Otherwise, the type checker complains that:
```
kmontemayor2-sc, line 620: Shall we check that `ty` still complains?

(I am going to leave identical comments on all of the changes like this - that way the robots can pick them up and check automatically)

gigl/distributed/dist_partitioner.py
```diff
@@ -1268,7 +1268,7 @@ def _label_pfn(source_node_ids, _):
         del label_edge_index
 
         if is_positive:
-            # This assert is added to pass mypy type check, in practice we will not see this fail
+            # This assert is added to pass the type checker, in practice we will not see this fail
```
kmontemayor2-sc: Shall we check that `ty` still complains?

(I am going to leave identical comments on all of the changes like this - that way the robots can pick them up and check automatically)

gigl/distributed/dist_partitioner.py
```diff
@@ -1277,7 +1277,7 @@ def _label_pfn(source_node_ids, _):
             if len(self._positive_label_edge_index) == 0:
                 self._positive_label_edge_index = None
         else:
-            # This assert is added to pass mypy type check, in practice we will not see this fail
+            # This assert is added to pass the type checker, in practice we will not see this fail
```
kmontemayor2-sc: Shall we check that `ty` still complains?

(I am going to leave identical comments on all of the changes like this - that way the robots can pick them up and check automatically)

gigl/src/common/vertex_ai_launcher.py
```diff
@@ -268,7 +268,7 @@ def _build_job_config(
         if vertex_ai_resource_config.timeout
         else None,
         # This should be `aiplatform.gapic.Scheduling.Strategy[inferencer_resource_config.scheduling_strategy]`
-        # But mypy complains otherwise...
+        # But the type checker complains otherwise...
```
kmontemayor2-sc, line 284: Shall we check that `ty` still complains?

(I am going to leave identical comments on all of the changes like this - that way the robots can pick them up and check automatically)

gigl/src/data_preprocessor/lib/ingest/bigquery.py
```diff
@@ -44,7 +44,7 @@ def _get_bigquery_ptransform(
         )
 
 
-# Below type ignores are due to mypy star expansion issues: https://github.com/python/mypy/issues/6799
+# Below type ignores are due to star expansion issues with the type checker: https://github.com/python/mypy/issues/6799
```
kmontemayor2-sc: Shall we check that `ty` still complains?

(I am going to leave identical comments on all of the changes like this - that way the robots can pick them up and check automatically)

gigl/src/data_preprocessor/lib/ingest/reference.py
```diff
@@ -7,7 +7,7 @@
 from gigl.src.common.types.graph_data import EdgeType, EdgeUsageType, NodeType
 from gigl.src.data_preprocessor.lib.types import InstanceDictPTransform
 
-# Type hints for abstract dataclasses are currently not supported. https://github.com/python/mypy/issues/5374
+# Type hints for abstract dataclasses may have limited support in type checkers. https://github.com/python/mypy/issues/5374
```
kmontemayor2-sc: Shall we check that `ty` still complains?

(I am going to leave identical comments on all of the changes like this - that way the robots can pick them up and check automatically)

Also update the TODO from python/mypy?

gigl/types/graph.py
```diff
@@ -387,8 +387,8 @@ def select_label_edge_types(
     str,
     int,
     # TODO(kmonte): Add GLT Partition book here
-    # We cannot at the moment as we mypy ignore GLT
-    # And adding it as a type here will break mypy.
+    # We cannot at the moment as we type-ignore GLT
+    # And adding it as a type here will break the type checker.
     # PartitionBook
```
kmontemayor2-sc: Shall we check that `ty` still complains?

(I am going to leave identical comments on all of the changes like this - that way the robots can pick them up and check automatically)

pyproject.toml
```diff
@@ -283,3 +282,35 @@ exclude =  [
 remove-all-unused-imports = true
 in-place = true
 recursive = true
+
+[tool.ty.environment]
+python-version = "3.11"
+
+[tool.ty.src]
+exclude = ["*_pb2.py", "*_pb2.pyi", "**/*.ipynb"]
```
kmontemayor2-sc, line 309: should we add a TODO to include our notebooks here eventually?

pyproject.toml
```diff
@@ -283,3 +282,35 @@ exclude =  [
...
+# on values imported from those modules.
+# https://docs.astral.sh/ty/reference/configuration/#replace-imports-with-any
+replace-imports-with-any = [
+  "apache_beam.**",
+  "google.cloud.**",
+  "graphlearn_torch.**",
+  "kfp.**",
+  "tensorflow.**",
+  "tensorflow_metadata.**",
+  "torch.**",
```
kmontemayor2-sc: hmmmm, we weren't type-ignoring torch before.

IIRC I saw some issues with torch when I looked into this earlier, the ty people suggested to enable some flag - did you try this? https://github.com/astral-sh/ty/issues/2244

Since we're so reliant on torch I'd rather not type-ignore it if possible.

pyproject.toml
```diff
@@ -283,3 +282,35 @@ exclude =  [
...
+exclude = ["*_pb2.py", "*_pb2.pyi", "**/*.ipynb"]
+
+[tool.ty.analysis]
+# Treat all types from unresolved third-party modules as Any instead of Unknown,
+# preventing cascading type errors (e.g. unresolved-attribute, invalid-argument-type)
+# on values imported from those modules.
+# https://docs.astral.sh/ty/reference/configuration/#replace-imports-with-any
+replace-imports-with-any = [
+  "apache_beam.**",
+  "google.cloud.**",
```
kmontemayor2-sc: likewise, is there a reason we're blanket ignoring all gcp imports?

pyproject.toml
svij-sc: Unforunately the flag is not the problem

 
From the robots analysis, there are 200+ issues that were not being caught before; I think we should reserve for a future change?
Seems might just be using the apis in an incorrect way according to types - event hough it seems to be working.

>  Group 1: Tensor.__getitem__ with NodeType/EdgeType/CondensedNodeType keys (~64 errors)           
>                                                                                                    
>   These happen because dict[NodeType, Tensor] access returns Tensor, and then a second indexing    
>   like result[tensor_index] hits Tensor's strict __getitem__ signature. The custom NewType keys    
>   (NodeType = str, CondensedNodeType = int, EdgeType = NamedTuple) don't satisfy Tensor's index    
>   types.                                                                                           
>                                                                                                    
>   Fix: These are dict-on-dict access patterns — the real issue is that ty is resolving the chained 
>   subscript incorrectly or the intermediate type needs explicit annotation. Likely fixable via
>   local variable annotations or targeted # type: ignore comments.                                  
>                                                                                          
>   Group 2: FloatTensor/LongTensor vs Tensor return type mismatches (~25 errors)                    
>                                       
>   Functions annotated to return FloatTensor/LongTensor but torch.gather(), torch.stack(),          
>   torch.cat(), torch.mean(), torch.tensor() all return generic Tensor.                   
>                                                                                                    
>   Fix: Change return type annotations from FloatTensor/LongTensor to Tensor. These deprecated
>   aliases are not what torch operations actually return.                                           
>                                                                                          
>   Group 3: __init__ argument type mismatches (~20 errors)                                          
>                                                                                                    
>   Parameters using torch.dtype, torch.device, etc. where ty resolves the stub types more strictly.
>                                                                                                    
>   Fix: Explicit casts or type annotations at call sites.                                 
>                                       
>   Group 4: HeteroData/dict access returning object/Unknown (~15 errors)                            
>    
>   When iterating over graph data structures, values come back as object rather than typed tensors. 
>                                                                                          
>   Fix: Add explicit type annotations to loop variables.                                            
>                                                                                          
>   Group 5: List invariance issues (~8 errors)                                                      
>    
>   list[Tensor | None] assigned to list[Tensor], or Optional[Tensor] appended to list[FloatTensor]. 
>                                                                                          
>   Fix: Widen the list type annotations.                                                            
>                                                                                          
>   Group 6: Misc (operators, spawn, call-non-callable, etc.) (~10 errors)                           
>                                       
>   Fix: Case-by-case — spawn import fix, operator type annotations, etc.

gigl/distributed/dataset_factory.py
svij-sc: I threw bot at it, looks like we still might need these.

gigl/src/data_preprocessor/lib/ingest/reference.py
svij-sc: this was fixed with ty - updated

gigl/src/data_preprocessor/lib/ingest/bigquery.py
svij-sc: fixed - nice

gigl/src/common/vertex_ai_launcher.py
svij-sc: needed

gigl/distributed/dist_partitioner.py
svij-sc: needed

gigl/distributed/dist_partitioner.py
svij-sc: needed

pyproject.toml
svij-sc: Tbh, I think its fine if we don't 
Notebooks need not have verbose static type checking

Most act as documentation

pyproject.toml
svij-sc: fixed

svij-sc (commented):
I updated the downstream PR where I addressed the ty fixes and to it added some new fixes identified by @kmontemayor2-sc below.

Ref: https://github.com/Snapchat/GiGL/pull/586

gigl/src/data_preprocessor/lib/ingest/bigquery.py
kmontemayor2-sc: Wait did we keep the link as is?

gigl/src/data_preprocessor/lib/ingest/reference.py
kmontemayor2-sc: ditto on link.

pyproject.toml
kmontemayor2-sc: I think I'd prefer to keep the torch typing around (if my reading is correct that this infact is equivalent to type-ignoring torch) and just add a bunch of `# type-ignore` and TODOs to fix.

IDK how feasible that'd be to setup tho.

gigl/src/data_preprocessor/lib/ingest/bigquery.py
svij-sc: This along with others are fixed in the follow up PR where I address the types in one go.
Since it was a long list of changes to fix type issues I left it as a separate scoped PR than this one - can merge if you think easier to review.

Fix here (comment is removed as ignore not needed): https://github.com/Snapchat/GiGL/blob/af3dc20a38d2d97c400d75a80713578b93dcb611/gigl/src/data_preprocessor/lib/ingest/bigquery.py#L69-L75

gigl/src/data_preprocessor/lib/ingest/reference.py
svij-sc: Was. removed in follow up: https://github.com/Snapchat/GiGL/blob/af3dc20a38d2d97c400d75a80713578b93dcb611/gigl/src/data_preprocessor/lib/ingest/reference.py#L9-L14

gigl/types/graph.py
svij-sc: Addressed - I added:
https://github.com/Snapchat/GiGL/blob/af3dc20a38d2d97c400d75a80713578b93dcb611/gigl/types/graph.py#L383-L394

pyproject.toml
svij-sc: Ye for sure.

its too wide of a scope change for 1 pr.
Id pref if we do it as a follow up - I can throw cc at it once we get these 2 prs merged in.

kmontemayor2-sc: I gave the robots a go at it and the generated https://github.com/Snapchat/GiGL/pull/597/changes - WDYT about merging it into your follow up PR?

I do feel pretty strongly about not type-ignoring torch, even if temporarily.

kmontemayor2-sc (commented):
I do feel pretty strongly about not type ignoring torch, WDYT about including https://github.com/Snapchat/GiGL/pull/597 as part of the follow up here? 

Otherwise this LGTM, thanks for the work :)

svij-sc:
> I gave the robots a go at it and the generated https://github.com/Snapchat/GiGL/pull/597/changes - WDYT about merging it into your follow up PR?

> I do feel pretty strongly about not type-ignoring torch, even if temporarily.


I pulled in the changes - thanks @kmontemayor2-sc

gigl/common/utils/gcs.py
```diff
@@ -402,10 +420,12 @@ def __batch_copy_blobs(
...
-            dst_blob_names: list[str] = [
-                src_blob.name.replace(src_prefix, dst_prefix, 1)
-                for src_blob in src_blobs
-            ]
+            dst_blob_names: list[str] = []
+            for src_blob in src_blobs:
+                assert src_blob.name is not None, (
+                    "Blob from list_blobs must have a name"
+                )
+                dst_blob_names.append(src_blob.name.replace(src_prefix, dst_prefix, 1))
```
kmontemayor2-sc, line 422: hmm is it easier to do `blobs  = [b for b in self.__list_file_blobs_at_gcs_path(gcs_path=gcs_path) if b.name is not None` on line 213?

gigl/src/common/utils/file_loader.py
```diff
@@ -266,12 +272,12 @@ def does_uri_exist(self, uri: Union[str, Uri]) -> bool:
...
+        if isinstance(_uri, GcsUri):
             exists = self.__gcs_utils.does_gcs_file_exist(gcs_path=_uri)
-        elif type(_uri) == LocalUri:
-            exists = does_path_exist(cast(LocalUri, _uri))
-        elif type(_uri) == HttpUri:
-            exists = HttpUtils.does_http_path_resolve(http_path=cast(HttpUri, _uri))
+        elif isinstance(_uri, LocalUri):
+            exists = does_path_exist(_uri)
+        elif isinstance(_uri, HttpUri):
+            exists = HttpUtils.does_http_path_resolve(http_path=_uri)
```
kmontemayor2-sc, line 280: isn't this a real logic change? 

I suppose this is better since it catches sub classes?

pyproject.toml
```diff
@@ -302,3 +301,33 @@ select = ["F401", "I"]
...
+[tool.ty.src]
+exclude = ["*_pb2.py", "*_pb2.pyi", "**/*.ipynb"]
+
+[tool.ty.analysis]
+# Treat all types from unresolved third-party modules as Any instead of Unknown,
+# preventing cascading type errors (e.g. unresolved-attribute, invalid-argument-type)
+# on values imported from those modules.
+# https://docs.astral.sh/ty/reference/configuration/#replace-imports-with-any
+replace-imports-with-any = [
+  "apache_beam.**",
```
kmontemayor2-sc, line 317: I don't think apache_beam was ignored before? I assume ty is unhappy with this? Can we test if we remove this?

kmontemayor2-sc (approved):
Awesome work Shubham! Really appreciate it :)

pyproject.toml
svij-sc: yea, ty doesnt like beam

gigl/src/common/utils/file_loader.py
svij-sc: correct.
We currently don't have any subclasses so its not truly a logic change; but I agree with you.

.claude/formatting.md
```diff
@@ -2,7 +2,7 @@
 
 Do not suppress errors with workarounds like `# type: ignore`:
```
mkolodner-sc, line 3: Does type: ignore still work, or should we remove this?

If not, I noticed there are still several type: ignores left across our code that we should be able to remove now?

mkolodner-sc (approved):
Stamped with a comment/question, thanks!

.claude/formatting.md
svij-sc: ty respectes `type: ignore`

gigl/common/utils/gcs.py
svij-sc: one throws an exception; this does not.
Which we should be in this case.

pyproject.toml
svij-sc: resolving for now, but we can revisit in the future.

PR #587, opened 2026-04-14
Add 'local' launch script for graph store mode - so we can simulate the cluster on dev boxes

kmontemayor2-sc:
…

**Scope of work done**


mkolodner-sc (commented):
Should this go in scripts? Looking our scripts folder [currently](https://github.com/Snapchat/GiGL/tree/main/scripts), the majority of scripts here are to help setup GiGL. This to me feels like something that would be better in an experimental folder since I imagine this is primarily used for testing/debugging -- WDYT?

kmontemayor2-sc:
I don't think this belongs in "experimental" as it's not experimental - it's production ready and meant to be used (potentially for experiments).

We could put it elsewhere, maybe `gigl/scripts/` as this is more buisness-logic-y? Happy to put it under `gigl/scripts` if you prefer

tests/unit/scripts/launch_graph_store_local_test.py
```diff
@@ -0,0 +1,366 @@
...
+                launch_delay_s: float,
+            ) -> int:
+                captured["nodes"] = nodes
+                captured["cwd"] = cwd
+                captured["job_dir"] = job_dir
+                captured["launch_delay_s"] = launch_delay_s
+                return 0
+
+            with (
+                patch.object(launcher, "_get_free_ports", return_value=[23456]),
```
mkolodner-sc: Is this line mocking anything? We'd want to mock `get_free_ports` rather, right?

scripts/launch_graph_store_local.py
```diff
@@ -0,0 +1,781 @@
...
+
+Example - training with 3 storage nodes and 2 compute nodes::
+
+    python scripts/launch_graph_store_local.py \\
+        --storage-nodes 3 \\
+        --compute-nodes 2 \\
+        --compute-procs-per-node 1 \\
+        --mode train \\
+        --task-config-uri path/to/task_config.yaml \\
+        --resource-config-uri path/to/resource_config.yaml
```
mkolodner-sc, line 27: It could be worth having instructions for how a user should generate these as part of the instructions, ditto elsewhere

scripts/launch_graph_store_local.py
```diff
@@ -0,0 +1,781 @@
...
+    """Build a Vertex AI-style ``CLUSTER_SPEC`` JSON object for one process."""
+
+    cluster: dict[str, list[str]] = {
+        "workerpool0": ["workerpool0-0:2222"],
+        "workerpool2": [f"workerpool2-{index}:2222" for index in range(storage_nodes)],
+    }
+    if compute_nodes > 1:
+        cluster["workerpool1"] = [
+            f"workerpool1-{index}:2222" for index in range(compute_nodes - 1)
+        ]
```
mkolodner-sc, line 363: qq -- any particular reason for these values?

scripts/launch_graph_store_local.py
```diff
@@ -0,0 +1,781 @@
...
+                "rank": node.rank,
+                "command": node.command,
+                "log_path": str(node.log_path),
+                "cuda_visible_devices": node.env.get("CUDA_VISIBLE_DEVICES", ""),
+                "cluster_spec": node.cluster_spec,
+            }
+            for node in nodes
+        ],
+    }
+    _write_json(job_dir / "launcher_manifest.json", manifest)
```
mkolodner-sc, line 792: Do we do cleanup of job_dir after the script is complete?

mkolodner-sc (approved):
Thanks!

scripts/launch_graph_store_local.py
kmontemayor2-sc: Good callout, done.

scripts/launch_graph_store_local.py
kmontemayor2-sc: No, we keep the logs around. They're useful for debugging / etc.

They're just local files it should be fine, do you have concerns here?

svij-sc (approved):

tests/unit/scripts/launch_graph_store_local_test.py
kmontemayor2-sc: oops, fixed thanks :)

scripts/launch_graph_store_local.py
kmontemayor2-sc: they were hard coded- we never actually use them but I swapped to "real" ports

PR #589, opened 2026-04-16
Enable max label per anchor in data splitter

yliu2-sc:
**Scope of work done**

In previous experiments, for non user defined labels where labels are generated from 1 hop, we need to enable limiting num max labels per anchor to reduce blow up in memory.

This PR updates to include the option through configs to limit max mum labels per anchor. 


gigl/utils/data_splitters.py
```diff
@@ -31,11 +31,44 @@
...
+    max_labels_per_anchor_node: Optional[int],
+) -> Optional[int]:
+    """Validate the optional per-anchor label cap."""
+    if max_labels_per_anchor_node is None:
+        return None
+    if max_labels_per_anchor_node <= 0:
+        raise ValueError(
+            "max_labels_per_anchor_node must be a positive integer when provided."
+        )
+    return max_labels_per_anchor_node
```
kmontemayor2-sc: I'm not sure this provides a ton of value (and returning from a "check" seems weird, It should probably just always return none and throw an error fi there's a problem?)

gigl/utils/data_splitters.py
```diff
@@ -642,9 +685,14 @@ def _get_padded_labels(
...
         allow_non_existant_node_ids (bool): If True, will allow anchor node ids that do not exist in the topology.
             This means that the returned tensor will be padded with `PADDING_NODE` for those anchor node ids.
+        max_labels_per_anchor_node (Optional[int]): If provided, caps the number of
+            labels materialized per anchor node.
     Returns:
         The shape of the returned tensor is [N, max_number_of_labels].
     """
+    max_labels_per_anchor_node = validate_max_labels_per_anchor_node(
+        max_labels_per_anchor_node
+    )
```
kmontemayor2-sc: I think we're probably calling this function too much, it really only need to be called here right?

I'm not sure it's super useful to call it everywhere when this is the only place it actually gets used, and otherwise it makes our code more complicated.

gigl/distributed/graph_store/storage_utils.py
```diff
@@ -71,26 +76,36 @@ def build_storage_dataset(
...
     Returns:
         A partitioned :class:`DistDataset` ready to be served.
     """
     gbml_config_pb_wrapper = GbmlConfigPbWrapper.get_gbml_config_pb_wrapper_from_uri(
         gbml_config_uri=task_config_uri
     )
+    if max_labels_per_anchor_node is None:
+        max_labels_per_anchor_node = get_max_labels_per_anchor_node_from_runtime_args(
+            dict(gbml_config_pb_wrapper.trainer_config.trainer_args)
+        )
```
kmontemayor2-sc: hmmm, I meant to leave a comment here about using `storage args` for graph store mode [1].

How about we just never call this function here, and always require it to be passed in? We can just parse it as a CLI flag e.g.  at [2]

[1]: https://github.com/Snapchat/GiGL/blob/main/proto/snapchat/research/gbml/gbml_config.proto#L185
[2]: https://github.com/Snapchat/GiGL/blob/main/examples/link_prediction/graph_store/storage_main.py#L184-L220

tests/unit/utils/data_splitters_test.py
```diff
@@ -810,6 +811,37 @@ def test_get_padded_labels(self, _, node_ids, topo, expected):
...
+                layout="CSR",
+            ),
+            max_labels_per_anchor_node=1,
+        )
+        assert_close(
+            labels,
+            torch.tensor([[1], [2]], dtype=torch.int64),
+            rtol=0,
+            atol=0,
+        )
```
kmontemayor2-sc: `nit, self.assert_tensor_close`?

yliu2-sc: this function doesn't exist?

kmontemayor2-sc: oh it's `assert_tensor_equality`. https://github.com/Snapchat/GiGL/blob/main/tests/test_assets/test_case.py#L96

gigl/distributed/graph_store/storage_utils.py
```diff
@@ -83,14 +90,16 @@ def build_storage_dataset(
         graph_metadata_pb_wrapper=gbml_config_pb_wrapper.graph_metadata_pb_wrapper,
         tfrecord_uri_pattern=tf_record_uri_pattern,
     )
-    return build_dataset(
+    dataset = build_dataset(
```
kmontemayor2-sc, line 94: Can we add a TODO here to pipe in `max_labels_per_anchor_node`? Doesn't need to be now.

kmontemayor2-sc (approved):
Thanks Yozen!

mkolodner-sc (approved):
Stamp

PR #590, opened 2026-04-17
Enable using reservations with VAI jobs

kmontemayor2-sc:
V2/GLT Vertex AI Trainer and Inferencer jobs (launched via
  vertex_ai_launcher.py) can now consume Compute Engine reservations.
  Adds a new `reservation_affinity` field on VertexAiResourceConfig that
  maps to MachineSpec.reservation_affinity on the worker pool.

  Valid `type` values: NO_RESERVATION, ANY_RESERVATION, SPECIFIC_RESERVATION.
  Leaving the block unset keeps the current (default) Vertex AI behavior.
  See https://docs.cloud.google.com/vertex-ai/docs/training/use-reservations

  Resource config YAML examples:

      # Pin a single-pool trainer to a specific reservation.
      trainer_config:
        vertex_ai_trainer_config:
          machine_type: "a2-highgpu-1g"
          gpu_type: "NVIDIA_TESLA_A100"
          gpu_limit: 1
          num_replicas: 1
          reservation_affinity:
            type: "SPECIFIC_RESERVATION"
            reservation_resource_names:
              - "projects/my-project/zones/us-central1-a/reservations/my-reservation"

      # Use any matching reservation, fall back to on-demand.
      inferencer_config:
        vertex_ai_inferencer_config:
          machine_type: "g2-standard-48"
          gpu_type: "NVIDIA_L4"
          gpu_limit: 4
          num_replicas: 1
          reservation_affinity:
            type: "ANY_RESERVATION"

      # Heterogeneous graph-store: compute pool on a reservation,
      # storage pool explicitly opted out.
      trainer_config:
        vertex_ai_graph_store_trainer_config:
          compute_pool:
            machine_type: "g2-standard-48"
            gpu_type: "NVIDIA_L4"
            gpu_limit: 4
            num_replicas: 1
            reservation_affinity:
              type: "SPECIFIC_RESERVATION"
              reservation_resource_names:
                - "projects/my-project/zones/us-central1-a/reservations/my-reservation"
          graph_store_pool:
            machine_type: "n1-standard-32"
            num_replicas: 1
            reservation_affinity:
              type: "NO_RESERVATION"

  Tested internally on Snap infrastructure with real Compute Engine reservations.


gigl/common/services/vertex_ai.py
```diff
@@ -89,6 +90,23 @@ def get_pipeline() -> int: # NOTE: `get_pipeline` here is  the Pipeline name
 
 @dataclass
 class VertexAiJobConfig:
+    """Configuration for a Vertex AI CustomJob worker pool.
+
+    ``reservation_affinity`` maps to ``MachineSpec.reservation_affinity``;
```
svij-sc: should we populate the rest of the args too?
Or just leave this out ?

gigl/src/common/vertex_ai_launcher.py
```diff
@@ -277,10 +288,74 @@ def _build_job_config(
...
         else None,
+        reservation_affinity=_build_reservation_affinity(
+            vertex_ai_resource_config.reservation_affinity
+        ),
     )
     return job_config
 
 
+def _build_reservation_affinity(
+    cfg: VertexAiReservationAffinity,
```
svij-sc: edit: cfg --> affinity

gigl/src/common/vertex_ai_launcher.py
```diff
@@ -277,10 +288,74 @@ def _build_job_config(
...
+    # ReservationAffinity.Type[cfg.type].
+    affinity_type = getattr(ReservationAffinity.Type, cfg.type)
+
+    if affinity_type == ReservationAffinity.Type.SPECIFIC_RESERVATION:
+        if not names:
+            raise ValueError(
+                "SPECIFIC_RESERVATION requires at least one entry in "
+                "reservation_resource_names "
+                "(format: projects/{p}/zones/{z}/reservations/{n})."
+            )
```
svij-sc, line 348: nit: We tend to keep runtime checks in GiglResourceConfigWrapper.
Keeps it consistent

svij-sc (approved):

gigl/src/common/vertex_ai_launcher.py
kmontemayor2-sc: we don't really do introspection of the protos in the wrapper, and these fields are not accessed directly by the wrapper which is the pattern when we throw [1].

Further more I'm not sure we should have the wrapper be explicitly responsible for checking the validity of the data it's wrapping. We could add some `.validate()` method to it but that seems out of scope for this PR?

I don't really think we do enough validation in the rc wrapper currently for us to put this in there? 

[1]: https://github.com/Snapchat/GiGL/blob/main/gigl/src/common/types/pb_wrappers/gigl_resource_config.py#L424-L431

gigl/src/common/vertex_ai_launcher.py
kmontemayor2-sc: done.

gigl/common/services/vertex_ai.py
kmontemayor2-sc: Added the rest.

zfan3-sc (approved):

PR #592, opened 2026-04-21
Create smaller ShmChannel for empty input in graph store mode

kmontemayor2-sc:
**Scope of work done**

Since we default to "CONTIGUOUS" now, many ranks will have inputs registered that are empty, so we should not allocated the full SHM channel for them.

Note we still need to register inputs for all ranks as the DistSampler is in charge of serving incoming RPC sampling requests.


mkolodner-sc (approved):
Thanks!

svij-sc (approved):
stamping

PR #594, opened 2026-04-22
Fix graph-store sampling teardown protocol

kmontemayor2-sc:
Publish destroy state before unregister completes, make fetch terminate cleanly during teardown, and fail fast if DistServer.shutdown() sees live channels or backends.

Wait for per-worker unregister acknowledgements before shutting down the last shared sampling backend, move deferred unregister finalization out of sampler callbacks to avoid deadlocks, and skip epoch starts for empty remote inputs.

Add unit and integration coverage for destroy/start races, buffered fetch teardown, worker-side drain and unregister acknowledgement, loader shutdown behavior, and graph-store training teardown ordering.

**Scope of work done**


gigl/distributed/graph_store/dist_server.py
```diff
@@ -59,13 +58,16 @@ class ChannelState:
         channel: The shared-memory channel for passing sampled messages.
         epoch: The last epoch started on this channel.
         lock: A reentrant lock guarding channel-level operations.
+        destroyed: Whether teardown has published terminal state for this
```
mkolodner-sc: nit: pref to name this is_destroyed to indicate boolean nature of this field

gigl/distributed/graph_store/dist_server.py
```diff
@@ -80,6 +82,12 @@ class SamplingBackendState:
         lock: A reentrant lock guarding backend-level operations.
         init_complete: Whether ``runtime.init_backend()`` has completed successfully.
         init_error: If ``runtime.init_backend()`` raised, the exception; otherwise ``None``.
+        shutting_down: Set to ``True`` while ``destroy_sampling_input`` is tearing
```
mkolodner-sc: At an initial glance, it seems strange that we have "shutting down" as a flag for one dataclass and "has been shut down" as a flag for the other. Is there a way we can achieve parity between the two, s.t they are tracking a same sort of behavior? Do we need to track if a sampling backend state has been shut down, for example, in the dataclass?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -124,25 +133,44 @@ def __init__(self, dataset: DistDataset, log_every_n: int = 50) -> None:
...
+        Raises:
+            RuntimeError: If any sampling channel or backend is still
+                registered when ``shutdown`` is called.
+        """
         with self._lock:
-            backends = list(self._backend_state_by_backend_id.values())
+            if (
+                self._channel_state_by_channel_id
+                or self._backend_state_by_backend_id
+                or self._backend_id_by_backend_key
```
mkolodner-sc, line 243: should we log self._backend_id_by_backend_key as well, in case this caused the runtime error?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -124,25 +133,44 @@ def __init__(self, dataset: DistDataset, log_every_n: int = 50) -> None:
...
+            ):
+                raise RuntimeError(
+                    "DistServer.shutdown() requires all channels and backends to be "
+                    "torn down first. "
+                    f"live_channels={sorted(self._channel_state_by_channel_id)} "
+                    f"live_backends={sorted(self._backend_state_by_backend_id)}"
+                )
             self._backend_id_by_backend_key.clear()
             self._backend_state_by_backend_id.clear()
             self._channel_state_by_channel_id.clear()
```
mkolodner-sc: Do we still need these lines if we are asserting them to be empty earlier?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -124,25 +133,44 @@ def __init__(self, dataset: DistDataset, log_every_n: int = 50) -> None:
         self._backend_id_by_backend_key: dict[str, int] = {}
         self._backend_state_by_backend_id: dict[int, SamplingBackendState] = {}
         self._channel_state_by_channel_id: dict[int, ChannelState] = {}
+        self._destroyed_channel_ids: set[int] = set()
         self._fetch_stats_by_channel_id: dict[int, _ChannelFetchStats] = {}
         self._log_every_n = log_every_n
 
     def shutdown(self) -> None:
```
mkolodner-sc, line 216: I feel like this name is a bit misleading now -- this is more-so doing a validation that assets have been cleaned up rather than an actual shutdown, which must have happened prior. With the current naming, one might be incentivized to call this still to try to clean up assets. Updating naming to the more validating nature of the function here would prevent it from being misused.

gigl/distributed/graph_store/dist_server.py
```diff
@@ -545,6 +574,12 @@ def init_sampling_backend(self, opts: InitSamplingBackendRequest) -> int:
             # first caller finishes init (or fails). Ordering is enforced
             # by the acquire-before-self._lock-release invariant above.
             with backend_state.lock:
+                if backend_state.shutting_down:
+                    raise RuntimeError(
```
mkolodner-sc, line 661: do we want to be raising an error in this case? This seems to be something we'd want to handle gracefully, rather than have a failure here.

gigl/distributed/graph_store/dist_server.py
```diff
@@ -625,103 +667,135 @@ def destroy_sampling_input(self, channel_id: int) -> None:
...
-        # Hold channel_state.lock to serialize destroy against in-flight
-        # start_new_epoch_sampling and any remaining
-        # fetch_one_sampled_message. Pre-PR #578, the old
-        # destroy_sampling_producer and start_new_epoch_sampling shared
-        # the same _producer_lock, so destroy/start_epoch were serialized;
-        # this restores that invariant via the two-phase API's
-        # per-channel lock.
-        with channel_state.lock:
-            with self._lock:
+            if backend_state is None:
```
mkolodner-sc, line 774: qq: What scenario is backend_state none here? Maybe adding a comment could help.

gigl/distributed/graph_store/dist_server.py
```diff
@@ -625,103 +667,135 @@ def destroy_sampling_input(self, channel_id: int) -> None:
...
-
         Args:
             channel_id: The ID of the channel to destroy.
         """
         with self._lock:
             channel_state = self._channel_state_by_channel_id.get(channel_id)
             if channel_state is None:
+                # Idempotent re-destroy: the channel was already torn
+                # down by a prior destroy_sampling_input call. Re-assert
+                # the tombstone so a concurrent start_new_epoch_sampling
```
mkolodner-sc: This says re-assert, but I don't see any assert happening here?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -625,103 +667,135 @@ def destroy_sampling_input(self, channel_id: int) -> None:
...
-        # fetch_one_sampled_message. Pre-PR #578, the old
-        # destroy_sampling_producer and start_new_epoch_sampling shared
-        # the same _producer_lock, so destroy/start_epoch were serialized;
-        # this restores that invariant via the two-phase API's
-        # per-channel lock.
-        with channel_state.lock:
-            with self._lock:
+            if backend_state is None:
+                self._destroyed_channel_ids.add(channel_id)
                 self._channel_state_by_channel_id.pop(channel_id, None)
```
mkolodner-sc, line 787: do we need to do any of these operations in the `if channel_state is None:` case?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -625,103 +667,135 @@ def destroy_sampling_input(self, channel_id: int) -> None:
...
-        with self._lock:
-            backend_state.active_channels.discard(channel_id)
-            if not backend_state.active_channels:
-                self._backend_state_by_backend_id.pop(backend_state.backend_id, None)
-                self._backend_id_by_backend_key.pop(backend_state.backend_key, None)
-                should_shutdown_backend = True
-        if should_shutdown_backend:
-            backend_state.runtime.shutdown()
+        with backend_state.lock:
+            with self._lock:
```
mkolodner-sc, line 813: Adding comments and context to explain why we have two locks here would be helpful. This is a bit hard to understand as is.

gigl/distributed/graph_store/dist_server.py
```diff
@@ -625,103 +667,135 @@ def destroy_sampling_input(self, channel_id: int) -> None:
...
+                )
+                if current_channel_state is None:
+                    return
+                if current_channel_state.backend_id != backend_state.backend_id:
+                    return
+                current_channel_state.destroyed = True
+                self._destroyed_channel_ids.add(channel_id)
+                self._channel_state_by_channel_id.pop(channel_id, None)
+                self._fetch_stats_by_channel_id.pop(channel_id, None)
+            backend_state.runtime.unregister_input(channel_id)
```
mkolodner-sc, line 833: Is there a reason this is excluded from the lock?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -625,103 +667,135 @@ def destroy_sampling_input(self, channel_id: int) -> None:
...
-        # concurrent callers with different epoch values cannot interleave
-        # the check-then-set on channel_state.epoch, and so a concurrent
-        # fetch cannot observe channel_state.epoch bumped before
-        # runtime.start_new_epoch_sampling has been dispatched.
-        with channel_state.lock:
-            if channel_state.epoch >= epoch:
-                return
-            channel_state.epoch = epoch
+        with backend_state.lock:
+            with self._lock:
```
mkolodner-sc, line 902: nit: can we add more comments here as well

gigl/distributed/graph_store/dist_server.py
```diff
@@ -893,9 +990,14 @@ def wait_and_shutdown_server() -> None:
...
-        _dist_server.shutdown()
+        try:
+            _dist_server.shutdown()
+        except Exception:
+            logger.exception(
+                "DistServer.shutdown() failed during cluster teardown; "
+                "continuing with barrier/shutdown_rpc anyway"
+            )
         _dist_server = None
-        # Also clear GLT's _dist_server
```
mkolodner-sc, line 898: meganit -- why'd we remove this comment?

gigl/distributed/graph_store/dist_server.py
```diff
@@ -893,9 +990,14 @@ def wait_and_shutdown_server() -> None:
     global _dist_server
     if _dist_server is not None:
         _dist_server.wait_for_exit()
-        _dist_server.shutdown()
+        try:
+            _dist_server.shutdown()
+        except Exception:
```
mkolodner-sc: Can we make the except more targeted beyond Exception? Also, if shutdown() raises, the server process logs it and then exits cleanly, meaning from the outside it may still seem like a normal shutdown. Should we re-raise the error after we do barrier/shutdown_rpc?

I guess trying to understand if it's safe if _dister_server.shutdown() fails.

mkolodner-sc (commented):
Thanks Kyle, left some comments/questions. For my learning, what was the behavior we were seeing prior to this change? 

Also, since there seems to be a new contract now in this PR with the order in which things are shut down, can we document somewhere order of steps needed to be taken here to do shutdown? Perhaps we can update the `shutdown_compute_proccess` (seems like theres a typo there) docstring.

kmontemayor2-sc:
> Thanks Kyle, left some comments/questions. For my learning, what was the behavior we were seeing prior to this change?

Three real bugs and one regression motivated this PR; all of them
showed up because the old protocol had no enforced ordering between
loader teardown and server shutdown:

1. A compute client that crashed mid-training left a sampling channel
   registered on the storage server. The old `DistServer.shutdown()`
   silently swallowed the leftover, called `runtime.shutdown()` on
   each backend, and returned. From the outside the cluster looked
   clean even though the runtime had been torn down under orphaned
   channels — the next bug from this state was an opaque GLT crash on
   the next training run that reused the same `backend_key`.
2. `destroy_sampling_input` popped the backend out of
   `_backend_id_by_backend_key` *before* calling `runtime.shutdown()`
   lock-free. A concurrent `init_sampling_backend` for the same key
   could then create a fresh runtime while the old one was still
   joining its workers, producing intermittent worker re-init hangs.
   Fixed in `5e04be9c` by holding `backend_state.lock` through
   `runtime.shutdown()` and setting a `tearing_down` flag the
   second-caller path observes.
3. `start_new_epoch_sampling` raised `RuntimeError: channel_id not
   found` if a compute peer's start RPC raced a destroy. Fixed in
   `772ed9ab` by tombstoning destroyed channel ids and silent-no-op'ing
   the start when the id is in the tombstone set.
4. Regression in the idempotent re-destroy path: the second
   `destroy_sampling_input(channel_id)` hit the `channel_state is None`
   early-return *without* re-adding the tombstone, so a subsequent
   `start_new_epoch_sampling` spuriously raised "unknown channel_id".
   Fixed in `46f92cf1` and covered by
   `test_start_new_epoch_sampling_silent_after_double_destroy`.
   
> Also, since there seems to be a new contract now in this PR with the order in which things are shut down, can we document somewhere order of steps needed to be taken here to do shutdown? Perhaps we can update the shutdown_compute_proccess (seems like theres a typo there) docstring.


   
   The new contract is now spelled out in the
`gigl/distributed/graph_store/dist_server.py` module docstring, with
back-pointers from `wait_and_shutdown_server` and from
`shutdown_compute_proccess` (the misspelling is preserved — it's used
by downstream training scripts, e.g.
`examples/link_prediction/graph_store/homogeneous_training.py:143` and
the integration tests, so a rename is a separate breaking change). TL;DR:

1. **Per-loader (compute side):** `DistLoader.shutdown()` →
   `DistServer.destroy_sampling_input` per channel. This is where the
   actual `runtime.unregister_input` and (last-channel)
   `runtime.shutdown` work happens.
2. **Per-compute-process:** `shutdown_compute_proccess` after every
   loader on this rank has been shut down.
3. **Per-storage-process:** `wait_and_shutdown_server` → strict-fail
   `DistServer.shutdown()` (validates that phase 1 ran for every
   channel/backend), then `barrier()` + `shutdown_rpc()`. Healthy
   peers clear the barrier even if one storage process saw leftover
   state; the failing process re-raises after the barrier so the
   orchestrator sees the failure.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Renamed the field to `tombstoned` instead of `is_destroyed`. The
underlying ask is "give this boolean a name that telegraphs what it
means," and `tombstoned` is the distributed-systems term for exactly
what this flag is — a terminal marker on an object that has briefly
outlived its registry entry so that an in-flight `fetch_one_sampled_message`
on another RPC thread can break out of its recv loop. Paired with a
matching rename on the backend flag (`shutting_down` → `tearing_down`
— see thread 2) the verb tense difference now telegraphs the
asymmetry at every call site:

* `tombstoned` (past participle) → terminal, "this has been done"
* `tearing_down` (present participle) → in-progress, "this is
  happening now"

Both names match the local "bare adjective / participle" convention
(`init_complete`, `init_error`).

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Good push on the naming. The two flags are intentionally asymmetric
because they're modeling different things, but the old names didn't
say so. Renamed:

* `ChannelState.destroyed` → `ChannelState.tombstoned` (terminal —
  past participle)
* `SamplingBackendState.shutting_down` → `SamplingBackendState.tearing_down`
  (in-progress — present participle)
* `DistServer._destroyed_channel_ids` → `DistServer._tombstoned_channel_ids`
  for consistency.

I left both flags rather than introducing parity, because they
genuinely model different states:

* `ChannelState.tombstoned` is a **terminal-state tombstone on a
  still-pinned object.** `fetch_one_sampled_message` may be holding a
  reference to `channel_state` on a different RPC thread when destroy
  lands; the tombstone is how that fetch loop learns to give up. The
  channel object outlives its registry entry briefly, so the flag
  has to live on the object itself.
* `SamplingBackendState.tearing_down` is an **in-progress marker on a
  still-registered backend.** The second-caller path in
  `init_sampling_backend` finds the backend by key, blocks on
  `backend_state.lock`, and needs to know it is being torn down so
  it does not reuse a half-shutdown runtime. Once teardown completes,
  the backend is popped from the registry entirely; a "tombstoned"
  flag on the popped object would never be observed.

A `backend_state.tombstoned` flag would have no observer — no path
keeps a backend reference past registry removal — so forcing parity
would mean either keeping shut-down backends in the registry
indefinitely or threading references through the second-caller path
for a flag that's always `True` when the field is reachable.

Both dataclass docstrings are updated to spell out the asymmetry and
why the verb tense difference matters.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Done, good catch :)

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Right — three of the five clears were no-ops once the strict-fail
check passes. Dropped `_backend_id_by_backend_key.clear()`,
`_backend_state_by_backend_id.clear()`, and
`_channel_state_by_channel_id.clear()`.

Kept `_tombstoned_channel_ids.clear()` and `_fetch_stats_by_channel_id.clear()`
since neither is covered by the strict-fail assert: tombstones
accumulate across destroys, and orphan stats can be created by an
in-flight fetch that re-runs `_log_fetch_stats_if_due` after a
destroy has popped the entry. Added a one-line comment explaining why
those two stay.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Agree the name is misleading in isolation. I held off on renaming
because `DistServer.shutdown` is dispatched by attribute name through
`_call_func_on_server` (`dist_server.py:1006-1021`), which routes
client RPCs to server methods by name. Renaming it would require
lockstep updates anywhere a client invokes it via RPC; the only
in-tree caller today is `wait_and_shutdown_server`, but the method is
exposed by name through GLT's RPC infrastructure and could be
invoked from out-of-tree storage scripts.

Tightened the docstring instead so the validator-not-shutdown nature
is loud where it matters. Lead line is now:

> Final post-teardown validator — does not perform any shutdown.

Plus an explicit sentence calling out that the actual
`runtime.shutdown()` work happens in `destroy_sampling_input` when
the last channel on a backend is destroyed. Strict-fail behavior also
protects against misuse in practice — calling this mid-flight raises
loudly rather than corrupting state. Happy to do the rename in a
follow-up if you'd rather take the breakage.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: I think raising is the right call here, though happy to revisit if you
have a concrete caller scenario in mind. Two ways to "handle gracefully"
both have downsides:

* **Auto-retry inside the call.** Release `backend_state.lock`,
  re-take `self._lock`, look up the key again, either reuse a
  (possibly new) backend or create a fresh one. This silently
  absorbs a real caller bug — the only way to hit this race is for a
  caller to `init_sampling_backend` for the same `backend_key` while
  another call is mid-`destroy_sampling_input` for the same backend.
  In normal flow loaders init at construction and destroy at
  shutdown, so reaching this path means a loader was rebuilt
  mid-teardown, exactly the kind of contract violation we want loud.
* **Block until teardown completes, then proceed.** Easy to
  implement, but it changes the call's latency profile non-locally:
  the second caller waits on a `runtime.shutdown()` (worker join +
  GLT internals) it didn't ask for. Worse, by the time it unblocks
  the registry has already been cleared
  (`dist_server.py:722-726`), so the second caller still has to
  re-query — same retry the caller could do externally, but now
  hidden inside the RPC.

The current `RuntimeError` lets the caller decide: surface the bug,
or retry the call (which by then will create a fresh backend). The
shutdown window is short (`runtime.shutdown` joins workers, typically
sub-second for the configurations we run), so caller-side retry is
cheap.

If there's a real caller scenario where the auto-retry would help,
let me know — I'd rather add a separate helper for it than build the
retry into `init_sampling_backend`.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: In current code this branch is unreachable — added a comment to that
effect. The invariant is that a channel's backend outlives the
channel: the backend registry entry is popped only inside the
`should_shutdown_backend` branch below, which fires only when
`backend_state.active_channels` is empty. Reaching this branch would
mean a future code path popped the backend without popping its
channels, or test fixture set up the state by hand
(e.g. `test_shutdown_fails_fast_with_live_backends` does this).

The branch is defense in depth: rather than blow up later in
`DistServer.shutdown()` from leaked channel state, treat it as a
partial teardown — tombstone the channel, drop bookkeeping, return.
Comment now spells out the invariant so a future reader doesn't have
to re-derive it from `init` / `register` / `destroy`.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Re-worded comment.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: added comments.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: `runtime.unregister_input` joins the per-channel worker thread —
bounded but not instantaneous. Holding `self._lock` across it would
block every other RPC handler in the server (fetches on other
channels, inits for other backends, status RPCs) until the join
completes. We've already removed the channel from the registry under
`self._lock` and set `tombstoned=True`, so other callers see the
channel as gone before we leave the registry-protecting lock; we're
still inside `backend_state.lock`, so any concurrent operation on
*this* backend is serialized correctly. Comment in the code now
spells this out.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: done.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: oops, removed.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Re-raising after barrier/shutdown_rpc — agreed. The whole point of
the strict-fail in `shutdown()` is to surface teardown bugs *loudly*,
and silently swallowing them in `wait_and_shutdown_server` defeated
that. The original reason for the try/except (commit `2eeb0c7d`) was
to keep healthy peers from wedging on the barrier when one storage
process saw leftover state. Re-raising *after* `barrier()` +
`shutdown_rpc()` preserves that — by the time we re-raise, the
barrier has been crossed and RPC torn down, so no peer is wedged. The
failing storage process now exits non-zero, which is exactly the
signal the orchestrator needs.

Kept the broad `except Exception`, deliberately. Narrowing to
`RuntimeError` (the only exception `shutdown()` is documented to
raise today) would mean a future bug inside `shutdown()` propagates
out *before* the barrier — re-introducing the wedge bug from
`2eeb0c7d`. The catch is defense in depth for cluster liveness;
broad is correct.

gigl/distributed/graph_store/compute.py
```diff
@@ -77,13 +77,21 @@ def init_compute_process(
...
+    :mod:`gigl.distributed.graph_store.dist_server` — call this *after*
+    every ``DistLoader.shutdown()`` on this rank has returned, so all
+    server-side channels are already destroyed.
 
-    Args:
-        None
+    Calls ``glt.distributed.shutdown_client`` and
+    ``torch.distributed.destroy_process_group`` exactly once per
+    compute process. (Function name is misspelled and kept for
+    back-compat with downstream training scripts.)
```
mkolodner-sc, line 90: Not super urgent but could we track a TODO for updating this. Will be easier to update this sooner rather than later when there are more adoptions of the server client logic.

gigl/distributed/graph_store/dist_server.py
```diff
@@ -6,9 +6,38 @@
...
+-----------------
+Cluster teardown is a strict three-phase sequence:
+
+1. **Per-loader (compute side).** Each ``DistLoader.shutdown()`` issues
+   ``DistServer.destroy_sampling_input(channel_id)`` against every
+   storage server it registered with. This call performs the actual
+   teardown work on the server: ``runtime.unregister_input`` for the
+   channel and, when the channel is the last on its backend,
+   ``runtime.shutdown`` for the backend.
+2. **Per-compute-process.** After all loaders are torn down, the
```
mkolodner-sc, line 20: Thanks! This is super helpful. 

The explanation covers it (by outling the `per-compute-process`, but just to be more explicit, maybe we can also give an example of the scenario where we have multiple dataloaders in sequence (train, val, test) in that we should only call the shutdown_compute_proccess once all the dataloaders have have gone through their initialization, iterate, and shutdown cycles.

gigl/distributed/graph_store/dist_server.py
mkolodner-sc: That sounds good -- my understanding is the current behavior here is to surface the bug directly (i didnt' see any retry logic from any caller). That's fine if we want to take that approach instead of retry, but in that setting, what fixes would a user who observes this bug have to resolve this? Would they need to add some sleep in between tearing down a previous sampling backend and initializing a new one? Some clarity on the fixes for resolving this error, perhaps in the error message, would be helpful for a user to identify next steps.

mkolodner-sc (approved):
Thanks Kyle!

gigl/distributed/graph_store/compute.py
kmontemayor2-sc: Added TODO.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Done!

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Honestly, users should never see this as backend key is auto-generated, I'll just mention that in the error.

yliu2-sc (approved):
Second stamp

PR #598, opened 2026-04-27
Move GT code to `gigl.nn`

kmontemayor2-sc:
**Scope of work done**

Let's try to put the modeling code in `gigl.nn` for now. 

Note this code does still preserve backwards compatability. 


gigl/src/common/models/graph_transformer/graph_transformer.py
```diff
@@ -1,1093 +1,43 @@
-"""Graph Transformer encoder for heterogeneous graphs.
+"""Deprecated module - kept for backwards compatibility.
```
yliu2-sc, line 1: we can honestly remove this file, only @zfan3-sc & I are using and it should be easy to migrate

yliu2-sc (approved):
Sounds good thanks. FYI @zfan3-sc

mkolodner-sc (approved):

gigl/src/common/models/graph_transformer/graph_transformer.py
kmontemayor2-sc: Hmmm, I'd prefer to not break existing workflows, we can do it as a follow up once your migrate?

PR #599, opened 2026-04-27
GT relation aware attn

yliu2-sc:
**Scope of work done**


gigl/nn/graph_transformer.py
```diff
@@ -252,23 +260,43 @@ def __init__(
...
         self._attention_norm = nn.LayerNorm(model_dim)
         self._query_projection = nn.Linear(model_dim, model_dim)
         self._key_projection = nn.Linear(model_dim, model_dim)
         self._value_projection = nn.Linear(model_dim, model_dim)
         self._output_projection = nn.Linear(model_dim, model_dim)
         self._dropout = nn.Dropout(dropout_rate)
+        self._relation_attention_matrices: Optional[nn.Parameter] = None
+        if relation_attention_mode == "edge_type_additive":
+            self._relation_attention_matrices = nn.Parameter(
+                torch.empty(num_relations, num_heads, self._head_dim, self._head_dim)
```
zfan3-sc, line 298: If we look at HGT's equation 3, the `W^(ATT)_{\phi(e)} ` does't have a super script i, so I think the edge type specific transformation is per edge type, not per (edge type, head index). The current implementation has more capacity but we are not doing apple-to-apple comparison with HGT

gigl/nn/graph_transformer.py
```diff
@@ -360,6 +405,102 @@ def forward(
...
+        attn_bias: Optional[Tensor],
+        pairwise_relation_mask: Optional[Tensor],
+    ) -> Tensor:
+        relation_attention_bias = self._build_relation_attention_bias(
+            query,
+            key,
+            pairwise_relation_mask=pairwise_relation_mask,
+        )
+        if relation_attention_bias is not None:
+            attn_bias = (
```
zfan3-sc, line 422: HGT uses K*W*Q^T, whereas we use K*Q^T + K*W*Q^T. So basically we reparametrized HGT's formula with W'^ATT = I + W^ATT. I think we have the same expressiveness, but maybe should initialize our W^ATT differently. zero-init or small sigma gaussian init could be an option, as xavier could make the variance a bit too big for bias. I'm open to discussion

gigl/nn/graph_transformer.py
zfan3-sc: Also it seems like we are missing the `\mu` attention multiplier per (source_type, edge_type, destination_type) in equation 3 if we compare with HGT

gigl/nn/graph_transformer.py
```diff
@@ -360,6 +405,102 @@ def forward(
...
+        relation_attention_bias = query.new_zeros(
+            (query.size(0), query.size(2), key.size(2), self._num_heads)
+        )
+        query_by_position = query.transpose(1, 2)
+        key_by_position = key.transpose(1, 2)
+        relation_matrices = self._relation_attention_matrices.to(dtype=query.dtype)
+        active_relation_ids = torch.unique(active_relation_positions[:, 3], sorted=True)
+
+        for relation_idx_tensor in active_relation_ids:
+            relation_idx = int(relation_idx_tensor.item())
```
zfan3-sc, line 480: this `.item()` can the above `active_relation_ids = torch.unique(active_relation_positions[:, 3], sorted=True)` causes GPU sync. Some AI suggestions to avoid it for performance
```
# remove the torch.unique line above
for relation_idx in range(self._num_relations):
    # remove the .item() line
    relation_positions = active_relation_positions[
        active_relation_positions[:, 3] == relation_idx
    ]
    ...
```

yliu2-sc:
#674

PR #606, opened 2026-05-04
[BUG][Fix] Not leveraging should_load_tensors_in_parallel when using dataset factory

svij-sc:
`should_load_tensors_in_parallel` is read from args but dropped it before calling `build_dataset` - fixed.


mkolodner-sc (approved):
Good find, thanks!

kmontemayor2-sc (approved):
Curious why this didn't get noticed until now

PR #612, opened 2026-05-05
Fix dataloading of multiple labels 

mkolodner-sc:
**Scope of work done**

<!-- Description of PR goes here -->
  GLT's sampler truncates node labels to a single column (nlabels.T[0]) when populating data.y for the DistFeature path, breaking multi-label node classification tasks where y is expected to be a 2-D tensor of shape [N, num_labels].            
                                                                                                                                                                                                                                                    
  Root cause: In GLTDistNeighborSampler._colloate_fn (note: GLT's typo), the DistFeature label-fetch path writes result_map['nlabels'] = nlabels.T[0], discarding all label columns beyond the first before the SampleMessage is sent to the main   
  process.                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
  Fix: Override _colloate_fn in BaseDistNeighborSampler with the full method body copied from GLT 0.2.4, removing the .T[0] truncation from both the homogeneous and heterogeneous DistFeature paths. Single-label datasets are unaffected — the    
  override preserves nlabels.T[0] when there is only one label column, keeping the existing 1-D shape contract. The non-DistFeature path (plain torch.Tensor labels) is unchanged since it never applied .T[0].
                                                                                                                                                                                                                                                    
  The method name typo is preserved on a minimal redirect stub (_colloate_fn → _collate_fn) so the actual implementation lives under the correctly-spelled name. 

<!-- Relevant screenshots go here (optional) -->

### DIFF
--- a/gigl/distributed/base_sampler.py
+++ b/gigl/distributed/base_sampler.py
@@ -238,6 +238,13 @@ class BaseDistNeighborSampler(GLTDistNeighborSampler):
         re-fetch approach would require.  The non-``DistFeature`` path (plain
         ``torch.Tensor`` labels) is unchanged — it never applied ``.T[0]``.
 
+        # TODO (mkolodner-sc): Now that GiGL owns this method, investigate whether
+        # post-processing steps in DistNeighborLoader._collate_fn can be folded in
+        # here and simplified — e.g. set_missing_features (populating empty tensors
+        # for node/edge features not fanned out to) and extract_metadata (stripping
+        # #META. keys before to_hetero_data to work around a GLT bug where those
+        # keys are misinterpreted as edge types).
+
         Args:
             output: The ``SamplerOutput`` or ``HeteroSamplerOutput`` returned by
                 ``_sample_from_nodes``.
@@ -281,9 +288,13 @@ class BaseDistNeighborSampler(GLTDistNeighborSampler):
                             output.node[input_type], input_type
                         )
                         nlabels = await wrap_torch_future(fut)
-                        # GLT writes nlabels.T[0], which truncates multi-column labels
-                        # to a single column. Preserve 1-D shape for single-label but
-                        # keep the full matrix for multi-label.
+                        # DistFeature always returns [N, K]. We collapse K=1 to 1-D
+                        # [N] to match GLT's convention and what downstream code
+                        # (e.g. CrossEntropyLoss) expects for data.y. Multi-label
+                        # (K>1) keeps the full 2-D matrix.
+                        # TODO (mkolodner-sc): Consider investigating always returning
+                        # 2-D — this may be a breaking change for single-label
+                        # training pipelines (e.g. CrossEntropyLoss expects 1-D data.y).
                         result_map[f"{as_str(input_type)}.nlabels"] = (
                             nlabels if nlabels.shape[1] > 1 else nlabels.T[0]
                         )
@@ -353,9 +364,13 @@ class BaseDistNeighborSampler(GLTDistNeighborSampler):
                 if isinstance(self.dist_node_labels, DistFeature):
                     fut = self.dist_node_labels.async_get(output.node)
                     nlabels = await wrap_torch_future(fut)
-                    # GLT writes nlabels.T[0], which truncates multi-column labels
-                    # to a single column. Preserve 1-D shape for single-label but
-                    # keep the full matrix for multi-label.
+                    # DistFeature always returns [N, K]. We collapse K=1 to 1-D
+                    # [N] to match GLT's convention and what downstream code
+                    # (e.g. CrossEntropyLoss) expects for data.y. Multi-label
+                    # (K>1) keeps the full 2-D matrix.
+                    # TODO (mkolodner-sc): Consider investigating always returning
+                    # 2-D — this may be a breaking change for single-label
+                    # training pipelines (e.g. CrossEntropyLoss expects 1-D data.y).
                     result_map["nlabels"] = (
                         nlabels if nlabels.shape[1] > 1 else nlabels.T[0]
                     )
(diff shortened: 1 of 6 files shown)


gigl/distributed/base_sampler.py
```diff
@@ -194,6 +197,55 @@ def _prepare_ablp_inputs(
             metadata=metadata,
         )
 
+    async def _colloate_fn(
```
kmontemayor2-sc: unrelated to this PR but can we add a TODO to fix the typo here?

kmontemayor2-sc: Did we consider just copying over the GLT code here (and also fixing the typo along the way)?

If we do that then we get to skip one rpc call, since it looks like the base class will do that anyways https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/distributed/dist_neighbor_sampler.py#L739-L741

gigl/distributed/base_sampler.py
```diff
@@ -194,6 +197,55 @@ def _prepare_ablp_inputs(
...
+        is_hetero = self.dist_graph.data_cls == "hetero"
+        if is_hetero:
+            input_type = output.input_type
+            if input_type is not None and not isinstance(input_type, tuple):
+                fut = self.dist_node_labels.async_get(
+                    output.node[input_type], input_type
+                )
+                nlabels = await wrap_torch_future(fut)
+                if nlabels.ndim == 2 and nlabels.shape[1] > 1:
+                    result[f"{as_str(input_type)}.nlabels"] = nlabels
```
kmontemayor2-sc: is the idea here to not overwrite the existing labels if already set? I feel like this is more reason to just copy over the code.

If we really don't want to copy over the code then can we say why we're doing this?

kmontemayor2-sc (commented):
Thanks Matt!

I have some preference for copying over the code just addressing the issue some of the time, but if you have strong reasons to not do I'm curious as to why :)

gigl/distributed/base_sampler.py
mkolodner-sc: No strong preference on copying over the code here -- I did it this way since it is less code we have to maintain, but am fine with copying the code (and fix the resulting bug) directly as well.

gigl/distributed/base_sampler.py
```diff
@@ -194,6 +198,185 @@ def _prepare_ablp_inputs(
...
+            if not isinstance(input_type, tuple):
+                if self.dist_node_labels is not None:
+                    if isinstance(self.dist_node_labels, DistFeature):
+                        fut = self.dist_node_labels.async_get(
+                            output.node[input_type], input_type
+                        )
+                        nlabels = await wrap_torch_future(fut)
+                        # GLT writes nlabels.T[0], which truncates multi-column labels
+                        # to a single column. Preserve 1-D shape for single-label but
+                        # keep the full matrix for multi-label.
```
kmontemayor2-sc: this behavior is a bit odd isn't it? Can we document it (and look into always having the full matrix returned?)

gigl/distributed/base_sampler.py
```diff
@@ -194,6 +198,185 @@ def _prepare_ablp_inputs(
...
+                fut = self.dist_node_feature.async_get(output.node)
+                nfeats = await wrap_torch_future(fut)
+                result_map["nfeats"] = nfeats
+            if self.dist_edge_feature is not None:
+                eids = result_map["eids"]
+                fut = self.dist_edge_feature.async_get(eids)
+                efeats = await wrap_torch_future(fut)
+                result_map["efeats"] = efeats
+            if output.batch is not None:
+                result_map["batch"] = output.batch
```
kmontemayor2-sc, line 391: should we also set `batch_size` here? IIRC there's some weirdness where it's not always set for hetero or something?

gigl/distributed/base_sampler.py
```diff
@@ -194,6 +198,185 @@ def _prepare_ablp_inputs(
...
+        (GLT 0.2.4) with the single change of ``_colloate_fn`` → ``_collate_fn``.
+        """
+        sampler_output = await async_func(*args, **kwargs)
+        res = await self._collate_fn(sampler_output)
+        if self.channel is None:
+            return res
+        self.channel.send(res)
+        return None
+
+    async def _collate_fn(
```
kmontemayor2-sc, line 223: We can also fix the need for `set_missing_features` now right? Though probably best as a follow up with a TODO

https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/utils/neighborloader.py#L224

kmontemayor2-sc (approved):
Thanks Matt! LGTM (though I'd prefer if you can add the TODOS here).

Also, are there any other patches we made to our code that we can do-away with now that this code is in our repo? Maybe `extract_metadata` too?

https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/utils/neighborloader.py#L303C5-L303C21

gigl/distributed/base_sampler.py
mkolodner-sc: Added a TODO to look into this

gigl/distributed/base_sampler.py
mkolodner-sc: I'm not sure if I remember what the weirdness there is here -- batch size is inferred automatically later on in `to_data` and `to_hetero_data` based on the length of the batch object [[1](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9/graphlearn_torch/python/loader/transform.py#L40)]. As of now, we shouldn't set this here.

gigl/distributed/base_sampler.py
mkolodner-sc: Its somewhat intentional by GLT I believe since some loss functions expect the labels to be one dimensional, but documenting this behavior and added a TODO to look into this further, as this would be a breaking change as well.

yliu2-sc (approved):

PR #615, opened 2026-05-06
launcher: fix inverted use_cuda flag and force CPU image on graph-store storage pool

kmontemayor2-sc:
## Summary

      Fixes two related bugs in `gigl/src/common/vertex_ai_launcher.py` that caused GPU/CPU misconfiguration on Vertex AI training jobs.

      ### Bug 1: Inverted `use_cuda` flag
      `_determine_if_cpu_execution(...)` returns `True` when the resource config asks for CPU, but its result was passed directly as `use_cuda`.
      The launcher now negates it in both `launch_single_pool_job` and the graph-store compute pool, so `use_cuda` correctly reflects GPU
      configuration.

      Effect of the bug: CPU-configured jobs ran with `--use_cuda` (then crashed/fell back), and GPU-configured jobs ran without `--use_cuda`
      (then ran on CPU at GPU cost).

      ### Bug 2: Storage pool tracking compute pool's CUDA-ness
      In `launch_graph_store_enabled_job`, the storage pool was being built with the same `use_cuda` and `container_uri` as the compute pool. The
       storage pool runs `gigl.distributed.graph_store.storage_main`, which is a CPU entrypoint that should never consume GPU resources. Storage
      now hardcodes `use_cuda=False` and `container_uri=cpu_docker_uri`.

      ### Clarity refactor
      Inside `launch_graph_store_enabled_job`: renamed `is_cpu_execution` → `is_compute_cpu_execution` and `container_uri` →
      `compute_container_uri` to make explicit that these describe only the compute pool. Updated the comment accordingly.

      ### Tests
      Re-added unit-test assertions (originally bundled with an unrelated TensorBoard PR that was reverted):
      - `test_launch_training_graph_store_cuda`: compute pool passes `--use_cuda`; storage pool uses the CPU image and omits `--use_cuda`.
      - `test_launch_inference_single_pool_cpu`: CPU job does not pass `--use_cuda`.

      ## Test plan

      - [x] `make unit_test_py PY_TEST_FILES="vertex_ai_launcher_test.py"` — both tests pass
      - [x] `make type_check` — no mypy issues
      - [x] `make format` — no formatting changes required for the touched files

      🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gigl/src/common/vertex_ai_launcher.py
+++ b/gigl/src/common/vertex_ai_launcher.py
@@ -85,7 +85,7 @@ def launch_single_pool_job(
         resource_config_uri=resource_config_uri,
         command_str=process_command,
         args=process_runtime_args,
-        use_cuda=is_cpu_execution,
+        use_cuda=not is_cpu_execution,
         container_uri=container_uri,
         vertex_ai_resource_config=vertex_ai_resource_config,
         env_vars=[env_var.EnvVar(name="TF_CPP_MIN_LOG_LEVEL", value="3")],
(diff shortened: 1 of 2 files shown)


mkolodner-sc (approved):
Thanks!

yliu2-sc (approved):

PR #616, opened 2026-05-06
Make graph-store shutdown tolerant of inactive-server channels

kmontemayor2-sc:
`BaseDistLoader.shutdown` was
  firing `destroy_sampling_input` to all 120 and `wait_all`-ing every
  future, so the slowest of 120 per-loader-per-rank set the latency floor
  and a ~57k-RPC cluster-wide burst reliably exceeded it.

  Changes:

  * base_dist_loader.py: BaseDistLoader.shutdown now honors `_remote_input_has_batches` (mirroring the existing filter in `__iter__` and `RemoteReceivingChannel`), collapsing per-loader destroy fan-out from 120 to 1 under CONTIGUOUS. Resolves futures one at a time with `shutdown destroy_start / destroy_ok / destroy_fail / destroy_done` log lines so a slow server is identifiable next time.

  * dist_server.py: `DistServer.shutdown` is now lenient on residual state — it warns with `live_channels` / `live_backends` / `live_backend_keys` and clears the registry instead of raising `RuntimeError`. The strict-fail behavior turned the documented optimization above into a guaranteed non-zero exit; the process is already terminating, so a warning is enough.

  * dist_server.py: `destroy_sampling_input` wrapped in try/finally with a slow-path warning at >1s for normal destroys and >30s for the last-channel destroy that triggers `runtime.shutdown()`. We had no visibility into server-side destroy timing before.

  * dist_server_test.py: updated `test_shutdown_warns_and_clears_with_live_backends` and `test_wait_and_shutdown_server_runs_barrier_when_state_remains` to match the new lenient contract.

**Scope of work done**


gigl/distributed/base_dist_loader.py
```diff
@@ -808,15 +808,58 @@ def shutdown(self) -> None:
...
+            )
+            for server_rank, fut in rpc_futures:
+                try:
+                    fut.wait()
+                    logger.info(
+                        f"shutdown destroy_ok worker_key={worker_key} "
+                        f"server_rank={server_rank} "
+                        f"elapsed={time.monotonic() - shutdown_start:.2f}s"
+                    )
+                except Exception as e:
```
mkolodner-sc, line 854: If one future fails, the remaining futures might be silently dropped here. Would it be better to collect all errors and raise at the end so every server gets a wait().

gigl/distributed/graph_store/dist_server.py
```diff
@@ -213,44 +214,35 @@ def __init__(self, dataset: DistDataset, log_every_n: int = 50) -> None:
...
                 self._channel_state_by_channel_id
                 or self._backend_state_by_backend_id
                 or self._backend_id_by_backend_key
             ):
-                raise RuntimeError(
-                    "DistServer.shutdown() requires all channels and backends to be "
-                    "torn down first. "
+                logger.warning(
+                    "DistServer.shutdown() called with live channels/backends; "
+                    "leaving residual state for process exit. "
```
mkolodner-sc, line 234: Can you elaborate more why we changed to only raising a warning here? What is the behavior if you leave it as a RuntimeError? Was it not able to do efficient training/inference in that setting?

mkolodner-sc (approved):
Thanks Kyle! LGTM provided comments are addressed.

gigl/distributed/graph_store/dist_server.py
kmontemayor2-sc: Since we don't de-register empty channels, there will be "live" channels now.

Since we shutdown anyways it's fine if it's not graceful..

We needed to add this change otherwise the servers would get overwhelmed with all the deregister calls at once.

svij-sc (approved):

gigl/distributed/base_dist_loader.py
kmontemayor2-sc: sure, done

PR #619, opened 2026-05-06
Fix CPP Installation

mkolodner-sc:
**Scope of work done**

`make install_dev_deps` failed on some VMs with:
 
```
CMake Error at CMakeLists.txt:76: Cannot find torch cmake config in CMAKE_PREFIX_PATH
```
 
`gigl-core` is a C++/CUDA extension that must be compiled against PyTorch (CMake needs to locate torch's cmake config). The `pyproject.toml` has `no-build-isolation-package = ["gigl-core"]` to let CMake find torch from the ambient venv, but this setting seems to be transiently ignored by `uv sync` for workspace members — it then builds them in an isolated temporary environment (`.cache/uv/builds-v0/.tmpXXXX`) where torch is absent.
 
**The root bootstrapping problem:** torch is a runtime dependency of `gigl`, not a declared build dependency of `gigl-core`, so `uv` has no signal to install torch before starting the `gigl-core` build.

## Fix
 
Two-phase install in `install_py_deps.sh`:
 
1. `uv sync --no-install-package gigl-core` — installs all packages (including torch) into the venv, skipping `gigl-core`
2. `uv pip install --no-build-isolation ./gigl-core/` — builds `gigl-core` explicitly without isolation, so CMake finds torch in the now-populated venv
`uv pip install --no-build-isolation` bypasses the workspace-member isolation behavior that makes `no-build-isolation-package` transiently ineffective in `uv sync`.

`uv cache clean` also has been identified as a potential fix here, allowing `make_install_dev_deps` to work, but still seemed to be leading to failures when running `build_cpp_extensions`

### DIFF
--- a/requirements/install_py_deps.sh
+++ b/requirements/install_py_deps.sh
@@ -140,7 +140,7 @@ install_gigl_lib_deps() {
 
     # gigl-core's CMake build requires torch to be present during the build, but
     # torch is a runtime dep of gigl, not a declared build dep of gigl-core.
-    # uv sync ignores no-build-isolation-package for workspace members and always
+    # uv sync may ignore no-build-isolation-package for workspace members and always
     # builds them in an isolated environment where torch is absent.
     # Workaround: phase 1 installs everything except gigl-core (so torch lands in
     # the venv); phase 2 uses `uv pip install --no-build-isolation` which explicitly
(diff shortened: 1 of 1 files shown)


kmontemayor2-sc (approved):
Can we also add a note somewhere about `uv cache clean` too?

yliu2-sc (approved):

PR #626, opened 2026-05-06
Add launch_custom subprocess dispatcher for CustomResourceConfig

kmontemayor2-sc:
Implements `launch_custom`, a thin shim that takes a populated
`CustomResourceConfig` and shells out via
`subprocess.run(shell=True, check=True)`. The proto's `command` is
a shell snippet (so leading `KEY=VALUE` env assignments parse
naturally) and `args[]` are individually `shlex.quote`-d before
joining, so values containing whitespace survive the shell pass.

The dispatcher performs no template substitution: `command` and
`args[]` are taken verbatim, and any placeholder text reaches
`subprocess.run` literally. Consumers that want runtime-context
substitution (e.g. ${gigl:foo}) should resolve it at YAML-load
time before the proto reaches this module. No call site in the
rest of the repo invokes `launch_custom` yet — wiring is added in
a follow-up PR.


tests/unit/src/common/custom_launcher_test.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_unsubstituted_gigl_placeholder_passes_through_verbatim(
+        self, mock_run: MagicMock
+    ) -> None:
+        # The launcher performs no template substitution: any
+        # ``${gigl:*}`` placeholder in command/args reaches subprocess
+        # unchanged. Consumers that want substitution must resolve at
+        # YAML-load time before the proto reaches this module.
+        config = self._build_config(
+            command="python -m my.cli",
+            args=["--foo=${gigl:bar}"],
```
svij-sc: what purpose will this be used for?

tests/unit/src/common/custom_launcher_test.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_invalid_component_raises_value_error(self, mock_run: MagicMock) -> None:
+        config = self._build_config(command="echo")
+        with self.assertRaises(ValueError):
+            launch_custom(
+                custom_launcher_config=config,
+                applied_task_identifier="job",
+                task_config_uri=Uri("gs://bucket/task.yaml"),
+                resource_config_uri=Uri("gs://bucket/resource.yaml"),
+                process_command="",
+                process_runtime_args={},
```
svij-sc, line 89: This test has two things invalid - it is not clear which one would fail i.e. maybe it fails on "test_empty_command_raises_value_error" and not "test_invalid_component_raises_value_error" - behaviour testing could be better containerized

svij-sc (approved):

mkolodner-sc (approved):
Stamp

tests/unit/src/common/custom_launcher_test.py
kmontemayor2-sc: oops, removed.

tests/unit/src/common/custom_launcher_test.py
kmontemayor2-sc: Updated!

PR #627, opened 2026-05-06
Validate CustomResourceConfig shape + GLT-backend compatibility

kmontemayor2-sc:
Adds two validators wired into the existing pre-run validation chain:

1. A bypass in `_validate_machine_config` (via the trainer /
   inferencer resource-config-valid checks) so a `CustomResourceConfig`
   trainer / inferencer skips machine-shape checks — the new oneof arm
   carries no machine spec to validate.
2. `check_custom_resource_config_shape`, which raises if a populated
   `CustomResourceConfig` has an empty `command`. Pure shape check, no
   subprocess spawn.
3. `check_custom_resource_config_requires_glt_backend`, an
   unconditional gate that pairs `CustomResourceConfig` with
   `feature_flags.should_run_glt_backend`. The v1 dispatchers do not
   route through `launch_custom`; this catches the misconfig at
   validate time.

Both new gates run unconditionally inside `kfp_validation_checks` for
both Trainer and Inferencer components.


gigl/src/validation_check/libs/gbml_and_resource_config_compatibility_checks.py
```diff
@@ -135,3 +135,63 @@ def check_inferencer_graph_store_compatibility(
...
+    resource_config_wrapper: GiglResourceConfigWrapper,
+) -> None:
+    """Enforce that ``CustomLauncherConfig`` is only used with the GLT (v2) backend.
+
+    The v1 trainer/inferencer dispatchers never consult the
+    ``custom_trainer_config`` / ``custom_inferencer_config`` oneof, so pairing
+    a ``CustomLauncherConfig`` with a task config that has
+    ``should_use_glt_backend=False`` would silently fall through the v1 path
+    and fail at runtime. Catch it up-front here so the failure is loud and
+    actionable at validation time.
```
svij-sc, line 151: Lets update the agent generated comments/docs so that they are meant for future readers and not just the author of the PR?

svij-sc: Probably the value error raised below is self documenting this function - this comment seems to verbose to add value.

gigl/src/validation_check/libs/gbml_and_resource_config_compatibility_checks.py
```diff
@@ -135,3 +135,63 @@ def check_inferencer_graph_store_compatibility(
...
+
+    if not gbml_config_pb_wrapper.should_use_glt_backend:
+        offending: list[str] = []
+        if trainer_is_custom:
+            offending.append("trainer_resource_config.custom_trainer_config")
+        if inferencer_is_custom:
+            offending.append("inferencer_resource_config.custom_inferencer_config")
+        raise ValueError(
+            "CustomLauncherConfig is only wired into the GLT (v2) dispatchers "
+            "(glt_trainer.py / glt_inferencer.py); the v1 trainer/inferencer "
```
svij-sc, line 188: v1 and v2 may be very arbritary to reader, and this message is too verbose and doesn't get to the point fast enough - a more useful message (that also simplifies above logic):
```python
raise ValueError("""
CustomLauncherConfig can only be used w/ GLT backend. Detected 
trainer_resource_config using custom implementation: {trainer_is_custom} ; 
inference_resource_config using custom implementation: {inference_is_custom}. 
Ensure feature_flags.should_run_glt_backend is set to 'True' in the task config.
""")
```

gigl/src/validation_check/libs/resource_config_checks.py
```diff
@@ -312,6 +313,57 @@ def _validate_machine_config(
...
+        ValueError: If ``component`` is not Trainer or Inferencer, or if a
+            populated ``CustomLauncherConfig`` has an empty ``command``.
+    """
+    if component not in {GiGLComponents.Trainer, GiGLComponents.Inferencer}:
+        raise ValueError(
+            f"check_custom_launcher_config_shape only supports "
+            f"Trainer and Inferencer components; got {component}."
+        )
+
+    wrapper = GiglResourceConfigWrapper(resource_config=resource_config_pb)
```
svij-sc, line 344: the caller already has access to the wrapper - why not just pass that through?

gigl/src/validation_check/config_validator.py
```diff
@@ -347,6 +355,15 @@ def kfp_validation_checks(
         resource_config_wrapper=resource_config_wrapper,
     )
 
+    # Validate any populated CustomLauncherConfig has a non-empty command.
+    # Unconditional — the check is shape-only and does not call out to any
+    # external service.
+    for component in (GiGLComponents.Trainer, GiGLComponents.Inferencer):
```
svij-sc, line 361: we already have an existing pattern to check for these in `resource_config_checks` - lets adopt here.

tests/unit/src/validation/lib/gbml_and_resource_config_compatibility_checks_test.py
```diff
@@ -214,5 +220,107 @@ def test_resource_has_inferencer_graph_store_template_does_not(self):
...
+
+    Because v1 trainer/inferencer dispatchers don't consult the custom oneof,
+    pairing a ``CustomLauncherConfig`` with
+    ``feature_flags.should_run_glt_backend = "False"`` must be caught at
+    validation time rather than at runtime.
+    """
+
+    def test_custom_trainer_without_glt_raises(self):
+        """CustomLauncherConfig trainer + glt=False raises a clear ValueError."""
+        gbml_config = _create_gbml_config_with_glt_flag("False")
```
svij-sc, line 273: nit: We are writing a function here to save 1 line of code but introducing a layer of indirection as a result.

```python
gbml_config = gbml_config_pb2.GbmlConfig()
gbml_config.feature_flags["should_run_glt_backend"] = False
```

much more readable.

tests/unit/src/validation/lib/gbml_and_resource_config_compatibility_checks_test.py
```diff
@@ -214,5 +220,107 @@ def test_resource_has_inferencer_graph_store_template_does_not(self):
...
+        resource_config = _create_resource_config_with_custom_trainer()
+        with self.assertRaises(ValueError) as ctx:
+            check_custom_launcher_config_requires_glt_backend(
+                gbml_config_pb_wrapper=gbml_config,
+                resource_config_wrapper=resource_config,
+            )
+        self.assertIn("should_run_glt_backend", str(ctx.exception))
+        self.assertIn("custom_trainer_config", str(ctx.exception))
```
svij-sc, line 281: Doing string matching in the error message feels too verbose to be helpful - unless you were testing on the error message creation logic itself, which it doesnt seem like we are.

tests/unit/src/validation/lib/resource_config_checks_test.py
```diff
@@ -791,6 +820,106 @@ def test_no_graph_store_config(self):
...
+        ) as mock_validate:
+            check_if_trainer_resource_config_valid(resource_config_pb=config)
+        mock_validate.assert_not_called()
+
+    def test_inferencer_custom_config_bypasses_machine_validation(self):
+        """CustomLauncherConfig inferencer bypasses _validate_machine_config entirely."""
+        config = _create_valid_custom_inferencer_config(args=["--cluster_size=4"])
+        with patch(
+            "gigl.src.validation_check.libs.resource_config_checks._validate_machine_config"
+        ) as mock_validate:
```
svij-sc, line 847: These mock patches seem like a new pattern in this file and are making testing very verbose - is there a way to not have these in the tests?

tests/unit/src/validation/lib/resource_config_checks_test.py
```diff
@@ -791,6 +820,106 @@ def test_no_graph_store_config(self):
...
+            check_if_inferencer_resource_config_valid(resource_config_pb=config)
+        mock_validate.assert_not_called()
+
+    def test_vertex_ai_trainer_still_calls_machine_validation(self):
```
svij-sc, line 851: needed?

tests/unit/src/validation/lib/resource_config_checks_test.py
```diff
@@ -791,6 +820,106 @@ def test_no_graph_store_config(self):
...
+
+    def test_populated_command_passes(self):
+        config = _create_valid_custom_trainer_config(command="python -m my.cli")
+        check_custom_launcher_config_shape(
+            resource_config_pb=config,
+            component=GiGLComponents.Trainer,
+        )
+
+    def test_empty_trainer_command_raises(self):
+        config = _create_valid_custom_trainer_config(command="")
```
svij-sc, line 885: nit: should we still call function `_create_valid_custom_trainer_config`  given the test setup here?

`_create_custom_trainer_config` ?

tests/unit/src/validation/config_validator_test.py
```diff
@@ -352,5 +352,61 @@ def test_resource_config_validation_failure_with_mock_configs(
             )
 
 
+class TestCustomLauncherConfigInValidationChain(TestCase):
```
svij-sc, line 355: This feels like a weird test to add.
i.e. (1) ensure that some function is calling some other function - (2) and do that using some knowledge of the error message that should be propagated.

If we start adopting this test pattern even if we just adopt (1); what is the stopping point? Our tests would be too verbose and would not be valuable. If instead you wanted to migrate our unit tests introduced above to an order higher i.e. here - that would be valuable.  But honestly, I would prefer not having this at all unless its a core library / performance functionality.

PR #630, opened 2026-05-07
Update URI join semantics

jchmura-sc:
Updates`Uri.join()` semantics so only the first token may be absolute. Additional join tokens now raise `TypeError` when they are _absolute_ local paths or URI-looking values such as `gs://...` / `http://...`. See more details on updated semantics in the inline comments.

**Scope of work done**

### DIFF
--- a/gigl/common/types/uri/http_uri.py
+++ b/gigl/common/types/uri/http_uri.py
@@ -10,8 +10,7 @@ class HttpUri(Uri):
     """Represents an HTTP URI."""
 
     def __init__(self, uri: Union[str, Path, HttpUri]) -> None:
-        self._has_valid_prefix(uri=uri)
-        self._has_no_backslash(uri=uri)
+        self.is_valid(uri=self._token_to_string(uri), raise_exception=True)
         super().__init__(uri=uri)
 
     @classmethod
--- a/gigl/common/types/uri/uri.py
+++ b/gigl/common/types/uri/uri.py
@@ -42,15 +42,18 @@ class Uri(object):
         cls, token: _URI_LIKE, allow_base_uri_join: bool = False
     ) -> bool:
         token_is_path_like = not isinstance(token, Uri)
-        if token_is_path_like: # str and Path tokens do not have a Uri type to match.
+        if token_is_path_like:
+            # str and Path tokens do not have a Uri type to match.
             return True
 
         token_matches_join_type = token.__class__ is cls
-        if token_matches_join_type: # Same concrete Uri type, e.g. GcsUri.join(GcsUri(...)).
+        if token_matches_join_type:
+            # Same concrete Uri type, e.g. GcsUri.join(GcsUri(...)).
             return True
 
         base_join_with_concrete_token = allow_base_uri_join and cls is Uri
-        if base_join_with_concrete_token: # Uri.join(GcsUri(...), "suffix") stays supported for compatibility.
+        if base_join_with_concrete_token:
+            # Uri.join(GcsUri(...), "suffix") stays supported for compatibility.
             return True
 
         # Remaining Uri tokens are cross-type mixes, e.g. GcsUri with HttpUri.
--- a/gigl/common/types/uri/uri.py
+++ b/gigl/common/types/uri/uri.py
@@ -29,13 +29,25 @@ class Uri(object):
             return str(token)
         return ""
 
-    # TODO(kmonte): You should not be able to join a Uri with a Uri of a different type.
-    # *or* join HTTP on HTTP or GCS on GCS.
-    # This is not backwards compatible, so come around to this later.
+    @classmethod
+    def _is_absolute_path(cls, token: _URI_LIKE) -> bool:
+        token_str = cls._token_to_string(token)
+
+        # Note: "://" is used to detect GcsUri and HttpUri prefixes, but will incorrectly
+        # classify relative LocalUri path's containing "://" as absolute
+        return Path(token_str).is_absolute() or "://" in token_str
+
     @classmethod
     def join(cls, token: _URI_LIKE, *tokens: _URI_LIKE) -> Self:
         """
         Join multiple tokens to create a new Uri instance.
+        The first token may be an absolute or relative path. Every additional token must be relative.
+
+        Note:
+            - Rejecting suffix strings containing "://" may break callers that
+              intentionally store URI-looking strings inside paths.
+            - Rejecting absolute LocalUri suffix tokens is stricter than os.path.join,
+              which implicitly discards earlier tokens on the join path.
 
         Args:
             token: The first token to join.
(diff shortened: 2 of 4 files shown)


gigl/common/types/uri/uri.py
```diff
@@ -29,13 +29,25 @@ def _token_to_string(token: _URI_LIKE) -> str:
...
+    @classmethod
+    def _is_absolute_path(cls, token: _URI_LIKE) -> bool:
+        token_str = cls._token_to_string(token)
+
+        # Note: "://" is used to detect GcsUri and HttpUri prefixes, but will incorrectly
+        # classify relative LocalUri path's containing "://" as absolute
+        return Path(token_str).is_absolute() or "://" in token_str
+
     @classmethod
     def join(cls, token: _URI_LIKE, *tokens: _URI_LIKE) -> Self:
```
jchmura-sc, line 63: The changes introduced here have 3 forms of backwards incompatibility.

1. Rejecting suffix strings containing "://" may break callers that intentionally store URI-looking strings inside paths. (e.g. join("foo", "bar:://baz") will raise a TypeError instead of "foo/bar:://baz"
2. Rejecting absolute `LocalUri` or `pathlib.Path` or str suffix tokens instead of implicitly discarding earlier tokens. (e.g. join("foo", "/bar") will raise a TypeError instead of becoming "/bar"
3. Reject joining an absolute `GcsUri` or a `HttpUri` onto anything.

Please let me know if any of these semantics are unclear, or undesirable.

gigl/common/types/uri/uri.py
```diff
@@ -29,13 +29,25 @@ def _token_to_string(token: _URI_LIKE) -> str:
...
         return ""
 
-    # TODO(kmonte): You should not be able to join a Uri with a Uri of a different type.
-    # *or* join HTTP on HTTP or GCS on GCS.
-    # This is not backwards compatible, so come around to this later.
+    @classmethod
+    def _is_absolute_path(cls, token: _URI_LIKE) -> bool:
+        token_str = cls._token_to_string(token)
+
+        # Note: "://" is used to detect GcsUri and HttpUri prefixes, but will incorrectly
```
jchmura-sc, line 36: I'm not thrilled about this but it seems like the simplest way to detect `GcsUri` or `HttpUri` without either (a) storing subclass prefixes in the base class or (b) checking against subclass type directly

gigl/common/types/uri/uri.py
```diff
@@ -88,10 +106,6 @@ def __eq__(self, other: Any) -> bool:
         return False
 
     def __truediv__(self, other: _URI_LIKE) -> Self:
```
jchmura-sc, line 143: Previously, `__truediv__` would reject joining a relative `LocalUri` onto a `GCSUri` or `HTTPUri`. In this implementation, we allow it.

jchmura-sc (commented):
/all_test

gigl/common/types/uri/uri.py
kmontemayor2-sc: Any reason to not check the subclasses here? e.g. `if isinstance(token, (GcsURI, HTTPURI): raise`

jchmura-sc: My thought was that it's not great to have to hardcode this, in case new subclasses appear. 

Also, we probably want to raise when the rhs token is a string like `"gs://foo"`?

svij-sc: I read this wrong initially the first time around.
IIUC,  what we want to do is `http_uri.join(gcs_uri)` , or `gcs_uri.join(local_uri)`


This change is maybe introducing some flexibility that although works in practice maybe we don't want to allow.

Similar to truedic, we could do something like:
```
if isinstance(other, Uri) and not isinstance(other, type(self)):
            raise TypeError(
                f"Cannot use '/' operator to join {type(self).__name__} with {type(other).__name__}"
            )
```

gigl/common/types/uri/uri.py
```diff
@@ -29,13 +29,25 @@ def _token_to_string(token: _URI_LIKE) -> str:
             return str(token)
         return ""
 
-    # TODO(kmonte): You should not be able to join a Uri with a Uri of a different type.
-    # *or* join HTTP on HTTP or GCS on GCS.
-    # This is not backwards compatible, so come around to this later.
+    @classmethod
+    def _is_absolute_path(cls, token: _URI_LIKE) -> bool:
```
svij-sc: `is_abs() ` could be an abstract method or property that subclasses could implement.

gigl/common/types/uri/uri.py
svij-sc: `is_abs() ` could be an abstract method or property that subclasses could implement.

svij-sc (commented):
IIUC,  what we want to prevent something like following:

```python
>>> from gigl.common import GcsUri, HttpUri, LocalUri, Uri
>>> http_uri = HttpUri("https://www.something.com")
>>> gcs_uri = GcsUri("gs://sdsds")
>>> Uri.join(gcs_uri, http_uri)
gs://sdsds/https://www.something.com
```

I think more generally whether abs path or not, we should:
1. disallow joining URI of two different types.
2. allow an abs path of any type to be joined against a string and for it to resolve correctly.

In regards to abs paths:
1. disallow two abs paths from joining
2. disallow a abs path on the right

gigl/common/types/uri/uri.py
svij-sc: 1. This is fine and expected - if we are doing this anywhere lets fix.
2. This makes sense and aligned with what I mentioned above.
3. We should allow for an absolute GcsUri and an absolute HttpUri to join with strings.

gigl/common/types/uri/uri.py
svij-sc: As per my comment above, Its probably safe to disallow this.
There is a small case where a GcSUri can be joined with string. (same with other URI) - which should be the only case this is allowed.

gigl/common/types/uri/http_uri.py
```diff
@@ -10,8 +10,7 @@ class HttpUri(Uri):
     """Represents an HTTP URI."""
 
     def __init__(self, uri: Union[str, Path, HttpUri]) -> None:
-        self._has_valid_prefix(uri=uri)
-        self._has_no_backslash(uri=uri)
+        self.is_valid(uri=self._token_to_string(uri), raise_exception=True)
```
jchmura-sc, line 13: Matching the behaviour of the [`GcsUri` constructor](https://github.com/Snapchat/GiGL/blob/3c28a8f6d985bf4ebda3463c73d9f993aad5017d/gigl/common/types/uri/gcs_uri.py#L18) which I suspect is the right thing to do.

This simplifies joining logic as now validly constructed `GcsUri` and `HttpUri` are _always_ absolute.

gigl/common/types/uri/uri.py
jchmura-sc: As discussed, the check against `://` is still there to prevent joins of the sort: `Uri.join(LocalUri("foo"), "gs://bar")`.

gigl/common/types/uri/uri.py
jchmura-sc: Understood. The integration tests have been updated accordingly.

PR #635, opened 2026-05-12
Enabled Weighted Sampling

mkolodner-sc:
## Summary

Adds native weighted edge sampling to GiGL's distributed training pipeline via GLT's `CPUWeightedSampler`. When enabled, neighbors are sampled proportionally to edge weights rather than uniformly.

## New API

- **`DistPartitioner.register_edge_weights(edge_weights)`** — registers a 1D per-edge weight tensor (homogeneous or `dict[EdgeType, Tensor]` for heterogeneous) before calling `partition_edge_index_and_edge_features()`. Weights are partitioned alongside edge features in the same pass (co-partitioned, mirroring the node features + labels pattern).
- **`load_torch_tensors_from_tf_record(weight_edge_feat_name=...)`** — accepts the name of an existing edge feature column to extract as sampling weights during TFRecord loading. The column is sliced out of the feature tensor and stored in `LoadedGraphTensors.edge_weights`; it is never duplicated in memory.
- **`build_dataset(weight_edge_feat_name=...)`** — threads `weight_edge_feat_name` through to TFRecord loading and then calls `register_edge_weights()` with the extracted weights.
- **`DistNeighborLoader(with_weight=True)`** / **`DistABLPLoader(with_weight=True)`** — enables weighted sampling. Defaults to `False`; must be set explicitly.
- **`BaseDistLoader.validate_with_weight()`** — shared validation: raises `ValueError` if `with_weight=True` but no weights are registered in the dataset; raises `NotImplementedError` if used with `PPRSamplerOptions` (weight-proportional PPR residual propagation is deferred to a future PR).

## Implementation notes

- `LoadedGraphTensors.edge_weights` — new field carrying extracted weights from TFRecord loading through to `register_edge_weights()`.
- `GraphPartitionData.weights` (field already existed) carries the partitioned weight tensor to `DistDataset._initialize_graph()`, which forwards it to GLT's `init_graph(edge_weights=...)`.
- `DistDataset.has_edge_weights` property reflects whether weights were registered at construction time.
- `SamplingConfig.with_weight` is now threaded through from the loader rather than hardcoded to `False`.
- **Graph Store mode:** `DistServer.get_edge_weights_registered()` and `RemoteDistDataset.fetch_edge_weights_registered()` propagate `has_edge_weights` across the RPC boundary so compute nodes can validate `with_weight` against the remote dataset.

## Tests

- **`tests/unit/distributed/distributed_weighted_sampling_test.py`** (8 new tests):
  - **Correctness (homogeneous + heterogeneous):** weight=0 edges to "bad" nodes are never traversed in sampled subgraphs — verified by encoding node class in features and asserting no bad node appears after weighted sampling.
  - **Partitioner edge cases:** features only, weights only, neither, both (with consistency check that `GraphPartitionData.edge_ids == FeaturePartitionData.ids`), and heterogeneous partial weights (one edge type weighted, another not).
- **`tests/unit/common/data/dataloaders_test.py`** (1 new test): `test_load_edge_weights_from_tf_record` — verifies that `load_torch_tensors_from_tf_record` correctly extracts a named column into `edge_weights`, removes it from `edge_features`, and returns the right shapes and values.


gigl/common/data/load_torch_tensors.py
```diff
@@ -161,6 +163,56 @@ def _data_loading_process(
...
+                feature_keys = list(serialized_tf_record_info[edge_type].feature_keys)
+                if col_name not in feature_keys:
+                    raise ValueError(
+                        f"weight_edge_feat_name '{col_name}' not found in edge feature keys "
+                        f"for edge type {edge_type}: {feature_keys}"
+                    )
+                col_idx = feature_keys.index(col_name)
+                weights[edge_type] = feat_tensor[:, col_idx]
+                keep_cols = [i for i in range(feat_tensor.shape[1]) if i != col_idx]
+                features[edge_type] = feat_tensor[:, keep_cols]
```
kmontemayor2-sc: nit. does it make sense to break this out into some separate helper if it's duplicated?

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -201,6 +202,10 @@ def __init__(
                 Defaults to `2` if set to `None` when using cpu training/inference.
             shuffle (bool): Whether to shuffle the input nodes. (default: ``False``).
             drop_last (bool): Whether to drop the last incomplete batch. (default: ``False``).
+            with_weight (bool): Whether to use edge weights for neighbor sampling.
+                Requires edge weights to be registered via
+                ``DistPartitioner.register_edge_weights()`` during dataset construction.
```
kmontemayor2-sc: Users don't really touch `DistPartitioner.register_edge_weights()`, right? Should we add some way that they'd enable it (via build_dataset?)

gigl/distributed/dist_dataset.py
```diff
@@ -584,12 +603,31 @@ def _initialize_graph(
                 edge_type: graph_partition_data.edge_ids
                 for edge_type, graph_partition_data in partitioned_edge_index.items()
             }
+            weights_by_type = {
+                edge_type: graph_partition_data.weights
+                for edge_type, graph_partition_data in partitioned_edge_index.items()
+                if graph_partition_data.weights is not None
+            }
+            edge_weights = weights_by_type if weights_by_type else None
```
kmontemayor2-sc: does this var do anything?  can we remove?

gigl/distributed/dist_partitioner.py
```diff
@@ -623,6 +624,59 @@ def register_edge_features(
         for edge_type in input_edge_features:
             self._edge_feat_dim[edge_type] = input_edge_features[edge_type].shape[1]
 
+    def register_edge_weights(
+        self, edge_weights: Union[torch.Tensor, dict[EdgeType, torch.Tensor]]
+    ) -> None:
+        """Registers per-edge sampling weights to the partitioner.
+
+        Weights must be a 1-D float tensor of shape ``[num_edges]``, one scalar per edge.
+        Must be called after ``register_edge_index()`` and before ``partition()``.
```
kmontemayor2-sc: Why is this the case? Where does this fail otherwise?

kmontemayor2-sc: is it. just so we can check that they are the same shape later? Because we do `self._edge_index is not None` which implies we'd skip this check.

gigl/distributed/dist_partitioner.py
```diff
@@ -1079,6 +1133,13 @@ def _partition_edge_index_and_edge_features(
         should_skip_edge_feats = (
             self._edge_feat is None or edge_type not in self._edge_feat
         )
+        has_weights_for_edge_type = (
+            self._edge_weights is not None and edge_type in self._edge_weights
+        )
```
kmontemayor2-sc: can we make both of these vars positive? e.g. `has_edge_feats` and `has_edge_weights` or similar? One being positive (has) and one being negative (should skip) can be a bit confusing.

gigl/types/graph.py
kmontemayor2-sc, line 1: Robots found this:

  Line 253 — treat_labels_as_edges() does not convert edge_weights


  self.node_ids = to_heterogeneous_node(self.node_ids)
  self.node_features = to_heterogeneous_node(self.node_features)

  self.edge_index = edge_index_with_labels
  self.edge_features = to_heterogeneous_edge(self.edge_features)


  This method converts node_ids, node_features, edge_index, and edge_features to heterogeneous form, but edge_weights is left as the original homogeneous tensor. For homogeneous ABLP training, dataset_factory calls this conversion before registering tensors, so register_node_ids() /
  register_edge_index() see heterogeneous dicts while register_edge_weights() receives a homogeneous tensor — DistPartitioner._convert_edge_entity_to_heterogeneous_format() rejects that mixed mode and weighted homogeneous ABLP dataset building fails.

  Fix: Convert edge_weights alongside edge_features:

  self.edge_features = to_heterogeneous_edge(self.edge_features)
  self.edge_weights = to_heterogeneous_edge(self.edge_weights)


  Please add a regression test that exercises homogeneous ABLP + weighted sampling end-to-end.

gigl/distributed/dist_partitioner.py
kmontemayor2-sc, line 1: More robot findings: 

  Lines 188–330 — _partition_edge_index_and_edge_features ignores _edge_weights

  The override never reads self._edge_weights, never partitions it, and never sets GraphPartitionData.weights. Because build_dataset_from_task_config_uri defaults to range partitioning (dataset_factory.py:643-646), any user who calls register_edge_weights while using the default config-driven path
  silently loses all weights. Downstream DistDataset._has_edge_weights becomes False, and validate_with_weight fails with a confusing "no edge weights registered" error even though weights were registered.


  Fix: Mirror DistPartitioner._partition_edge_index_and_edge_features. Roughly:


  # After line 223

  weight_tensor = None
  if self._edge_weights is not None and edge_type in self._edge_weights:
      weight_tensor = self._edge_weights[edge_type]


  # Augment input_data on lines 229/234 to include weight_tensor (e.g. (..., edge_feat, weight_tensor))

  # After _partition_by_chunk on line 246, slice partitioned weights back out of res_list
  # Set GraphPartitionData(..., weights=partitioned_weights) in both branches at lines 308/320
  # del self._edge_weights[edge_type] and clear when empty, mirroring the edge_feat cleanup at 253-262

  Add a regression test for DistRangePartitioner + register_edge_weights.

gigl/common/data/load_torch_tensors.py
```diff
@@ -72,6 +73,7 @@ def _data_loading_process(
     ],
     rank: int,
     tf_dataset_options: TFDatasetOptions = TFDatasetOptions(),
+    weight_edge_feat_name: Optional[Union[str, dict[EdgeType, str]]] = None,
```
kmontemayor2-sc, line 131: should we update the doc string for the new arg?

kmontemayor2-sc: Also it seems a little specific to put the `weight_edge_feat_name` here when `_data_loading_process` is kind of a generic function? Or do you think that's fine for now? I'm not sure how else we'd address this.

gigl/distributed/dist_dataset.py
kmontemayor2-sc, line 1: More robot findings: 

  Lines 606–621 — Warning promises uniform fallback for partial heterogeneous weights, but GLT will crash

  weights_by_type = {
      edge_type: graph_partition_data.weights

      for edge_type, graph_partition_data in partitioned_edge_index.items()
      if graph_partition_data.weights is not None

  }
  edge_weights = weights_by_type if weights_by_type else None
  if weights_by_type:

      missing = set(partitioned_edge_index.keys()) - set(weights_by_type.keys())
      if missing:

          logger.warning(
              f"... When with_weight=True, edge types without weights "

              f"will fall back to uniform sampling."
          )

  GLT does not fall back to uniform sampling in this case. graphlearn_torch/sampler/neighbor_sampler.py:104-113 unconditionally instantiates pywrap.CPUWeightedSampler for every edge type when with_weight=True. For unweighted types, Topology keeps edge_weights = torch.empty(0), Graph::InitCPUGraphFromCSR
   leaves edge_weight_ unset, and CPUWeightedSampler::Sample dereferences a null prob pointer in std::discrete_distribution<> — undefined behavior, typically a crash.


  Fix: Either tighten BaseDistLoader.validate_with_weight to require weights for every edge type when with_weight=True, or synthesize all-ones weight tensors here for missing edge types:


  if weights_by_type:

      for edge_type, gpd in partitioned_edge_index.items():
          if gpd.weights is None:
  if weights_by_type:
      missing = set(partitioned_edge_index.keys()) - set(weights_by_type.keys())
      if missing:
          logger.warning(
              f"... When with_weight=True, edge types without weights "
              f"will fall back to uniform sampling."
          )

  GLT does not fall back to uniform sampling in this case. graphlearn_torch/sampler/neighbor_sampler.py:104-113 unconditionally instantiates pywrap.CPUWeightedSampler for every edge type when with_weight=True. For unweighted types, Topology keeps edge_weights =
  torch.empty(0), Graph::InitCPUGraphFromCSR leaves edge_weight_ unset, and CPUWeightedSampler::Sample dereferences a null prob pointer in std::discrete_distribution<> — undefined behavior, typically a crash.

  Fix: Either tighten BaseDistLoader.validate_with_weight to require weights for every edge type when with_weight=True, or synthesize all-ones weight tensors here for missing edge types:

  if weights_by_type:
      for edge_type, gpd in partitioned_edge_index.items():
          if gpd.weights is None:
              weights_by_type[edge_type] = torch.ones(gpd.edge_index.shape[1], dtype=torch.float32)
      edge_weights = weights_by_type

  Either way, rewrite the warning to reflect actual behavior. Add a test with one weighted + one unweighted sampled edge type and with_weight=True.

gigl/common/data/load_torch_tensors.py
```diff
@@ -161,6 +163,56 @@ def _data_loading_process(
...
+                    )
+                col_name = weight_edge_feat_name
+                edge_type, feat_tensor = next(iter(features.items()))
+                feature_keys = list(serialized_tf_record_info[edge_type].feature_keys)
+                if col_name not in feature_keys:
+                    raise ValueError(
+                        f"weight_edge_feat_name '{col_name}' not found in edge feature keys "
+                        f"for edge type {edge_type}: {feature_keys}"
+                    )
+                col_idx = feature_keys.index(col_name)
```
kmontemayor2-sc: Robot finding - is this true? Do we support multi dim features? 

  Lines 185, 207 — feature_keys.index(col_name) assumes each feature key is exactly one column wide

  col_idx = feature_keys.index(col_name)
  weights[edge_type] = feat_tensor[:, col_idx]
  keep_cols = [i for i in range(feat_tensor.shape[1]) if i != col_idx]
  features[edge_type] = feat_tensor[:, keep_cols]

  _concatenate_features_by_names (gigl/common/data/dataloaders.py:155-187) supports multi-dim features, so a single feature key may contribute multiple columns. Example: feature_keys = ["embedding", "weight"] where embedding has shape [N, 16] and weight has shape [N, 1] →
   concat tensor has 17 columns and the weight lives at column 16, not column 1. tests/unit/common/data/dataloaders_test.py:494 only covers scalar features so doesn't catch this.

  Fix: Compute the actual column offset by summing the widths of preceding features (from feat_spec[feature_key].shape or by deriving from the concatenated tensor's per-feature widths), and assert the weight feature has width 1 before squeezing:

  col_widths = [serialized_tf_record_info[edge_type].feature_spec[k].shape[-1] or 1 for k in feature_keys]
  col_offset = sum(col_widths[: feature_keys.index(col_name)])
  weight_width = col_widths[feature_keys.index(col_name)]
  assert weight_width == 1, f"weight column '{col_name}' must be width 1, got {weight_width}"
  weights[edge_type] = feat_tensor[:, col_offset]
  keep_cols = [i for i in range(feat_tensor.shape[1]) if i != col_offset]
  features[edge_type] = feat_tensor[:, keep_cols]

  Add a test with at least one multi-dim feature alongside the weight column.

gigl/common/data/load_torch_tensors.py
kmontemayor2-sc: Robot finding 

  weights[edge_type] = feat_tensor[:, col_idx]
  keep_cols = [i for i in range(feat_tensor.shape[1]) if i != col_idx]
  features[edge_type] = feat_tensor[:, keep_cols]

  If the only feature key for an edge type is the weight column, slicing produces keep_cols = [] and features[edge_type] becomes shape [N, 0]. The outer if features: check (around line 222) still passes, so a zero-width feature tensor gets registered downstream and
  _edge_feat_dim[edge_type] = 0.

  Fix: After slicing, drop the edge type from features when no columns remain:

  if keep_cols:
      features[edge_type] = feat_tensor[:, keep_cols]
  else:
      del features[edge_type]

kmontemayor2-sc (commented):
Thanks Matt! Me and the robots did a first pass, it's possible they're imagining some of the issues here but I figured I'd flag :)

gigl/common/data/load_torch_tensors.py
mkolodner-sc: Makes sense to update the docstring. Re: the second comment, I see your concern, but since this is not exposed to users, I think it is fine for now, as any other solution here feels more unnecessarily complex.

gigl/common/data/load_torch_tensors.py
mkolodner-sc: We do technically support multi-dimensional features, but don't frequently use them. This is still a bug though and should be addressed, thanks!

gigl/distributed/dist_partitioner.py
mkolodner-sc: Yes, updated the docstring.

```
Must be called before ``partition()``. May be called before or after                                                                                                                                                      
``register_edge_index()``; shape validation against the edge index is performed                                                                                                                                           
only when ``register_edge_index()`` has already been called.   
```

gigl/distributed/dist_partitioner.py
mkolodner-sc: Good catch, thanks!

gigl/distributed/dist_dataset.py
mkolodner-sc: It's used subsequently: `self._has_edge_weights = edge_weights is not None`

gigl/distributed/dist_dataset.py
mkolodner-sc: Fixed, thanks!

gigl/common/data/load_torch_tensors.py
```diff
@@ -27,7 +27,63 @@
...
+    if col_name not in feature_keys:
+        raise ValueError(
+            f"weight_edge_feat_name '{col_name}' not found in edge feature keys "
+            f"for edge type {edge_type}: {feature_keys}"
+        )
+    key_idx = feature_keys.index(col_name)
+    col_widths = [
+        (spec.shape[-1] if spec.shape else 1)
+        for spec in (feature_spec[k] for k in feature_keys)
+    ]
```
kmontemayor2-sc: nit. I'd prefer to avoid the nested comprehensions I feel like they're a bit hard to read. I'd prefer something like:

```
col_widths = []
for key in feature_keys:
  for spec in feature_spec[keys]:
     ...
```

gigl/distributed/base_dist_loader.py
```diff
@@ -333,13 +333,50 @@ def __init__(
                 "for graph-store mode."
             )
 
+    @staticmethod
+    def validate_with_weight(
```
kmontemayor2-sc: nit. rename to `validate_for_weighted_sampling` or similar?

gigl/distributed/dist_partitioner.py
```diff
@@ -623,6 +624,63 @@ def register_edge_features(
...
+            # Shape validation is best-effort: only fires when register_edge_index()
+            # has already been called for this edge type.
+            if self._edge_index is not None and edge_type in self._edge_index:
+                local_num_edges = self._edge_index[edge_type].shape[1]
+                if weight_tensor.shape[0] != local_num_edges:
+                    raise ValueError(
+                        f"Edge weights for edge type {edge_type} have length "
+                        f"{weight_tensor.shape[0]} but the registered edge index has "
+                        f"{local_num_edges} edges on this rank."
+                    )
```
kmontemayor2-sc: is this worth validating if it's only sometimes?

gigl/distributed/dist_partitioner.py
```diff
@@ -1128,85 +1188,138 @@ def _edge_pfn(_, chunk_range):
...
+                input_parts.append(edge_feat)
+            if edge_weights_tensor is not None:
+                input_parts.append(edge_weights_tensor)
+            input_parts.append(edge_ids)
+
+            # Positional indices: features first, weights next, ids always last.
+            feat_idx: Optional[int] = 0 if has_edge_feats else None
+            weight_idx: Optional[int] = (
+                (1 if has_edge_feats else 0) if has_weights_for_edge_type else None
+            )
```
kmontemayor2-sc: can we put this on several lines like `weight_idx = None; if has_weights_for_edge_type: if has_edge_feats: weight_idx = 1`?

gigl/distributed/dist_dataset.py
```diff
@@ -336,6 +341,13 @@ def degree_tensor(
             self._degree_tensor = compute_and_broadcast_degree_tensor(self.graph)
         return self._degree_tensor
 
+    @property
+    def has_edge_weights(self) -> bool:
+        """True if edge weights were registered during dataset construction via
+        ``DistPartitioner.register_edge_weights()``.
+        """
+        return self._has_edge_weights
```
kmontemayor2-sc: does it make sense to expose some `edge_weights` property here? Can do so as a follow up if we want :)

gigl/distributed/dist_dataset.py
```diff
@@ -584,12 +603,34 @@ def _initialize_graph(
...
+                )
+                if missing:
+                    logger.warning(
+                        f"Edge weights are registered for {set(weights_by_type.keys())} but "
+                        f"not for {missing}. Filling missing edge types with uniform weights "
+                        f"(all 1s) so GLT does not segfault on partial-weight heterogeneous graphs."
+                    )
+                    for edge_type in missing:
+                        n_edges = partitioned_edge_index[edge_type].edge_index.size(1)
+                        weights_by_type[edge_type] = torch.ones(n_edges)
```
kmontemayor2-sc: optimization nit:

Can we do the following:

1. Find the largest edge type without weights, and generate a `torch.ones` for it.
2. For all other weightless edgetypes, generate views onto the tensor generated in 1.

kmontemayor2-sc (approved):
Thanks Matt!

gigl/distributed/dist_partitioner.py
mkolodner-sc: Hmm, probably not. I removed this validation block as a result, thanks!

yliu2-sc (approved):

PR #637, opened 2026-05-13
Move sbt override to cloud build

mkolodner-sc:
**Scope of work done**


mkolodner-sc:
`lint_test` doesn't set `use_cloud_run` to `True` (default `False`), meaning the changes in this PR can't be triggered on comment for this test, and will only be seen on merge. We see through the success of the scala unit test that the change is captured.

kmontemayor2-sc (approved):
Thanks matt! Can we ad comments to these overrides saying why we do them?

svij-sc (approved):
I would personally do:

`cp path/to/repositories ~/.sbt/repositories` once in `run_command_on_active_checkout`, then remove the -Dsbt.repository.config=repositories from .sbtopts file.
Or, just bake the repositories file inside the docker image for cloud build.

Then we have one source of truth not two.

But, this is fine too given we will deprecate scala code soon anyways.

PR #638, opened 2026-05-13
Make ParseDict call sites compatible with types-protobuf v7 stubs

kmontemayor2-sc:
## Summary

  - `types-protobuf` v7 tightened `google.protobuf.json_format.ParseDict`'s `js_dict` parameter from an `Any`-compatible alias to a strict `dict[str, Any]`.
  - The two `ParseDict` call sites in this repo pass the result of `OmegaConf.to_object(...)` / `OmegaConf.to_container(...)`, which mypy infers as a broad union (`dict[str | bytes | int | Enum | float | bool, Any] | list[Any] | str | Any | None`). With v7 stubs, both fail mypy.
  - At each call site, narrow with an explicit `isinstance(..., dict)` guard that raises `TypeError` (per the "Fail fast on invalid state" coding standard in `CLAUDE.md`), then `cast(dict[str, Any], ...)` before handing the value to `ParseDict`. The `isinstance` narrowing alone is insufficient because
  `dict` is invariant in its key type — the cast refines the key type from the broader union down to `str`.
  - A `grep -rn "ParseDict(" gigl snapchat --include="*.py" | grep -v _pb2.py` returns exactly the two call sites fixed here, so the surface area is fully enumerated.


gigl/common/utils/proto_utils.py
```diff
@@ -28,7 +28,12 @@ def read_proto_from_yaml(self, uri: Uri, proto_cls: Type[T]) -> T:
             omega_conf_obj = OmegaConf.create(raw_data)
         tfh.close()
         obj_dict = OmegaConf.to_object(omega_conf_obj)
-        proto = ParseDict(js_dict=obj_dict, message=proto_cls())
+        if not isinstance(obj_dict, dict):
+            raise TypeError(
+                f"ProtoUtils.read_proto_from_yaml expected a mapping at the YAML root for "
+                f"{uri}, got {type(obj_dict).__name__}."
+            )
+        proto = ParseDict(js_dict=cast(dict[str, Any], obj_dict), message=proto_cls())
```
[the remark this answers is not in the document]
svij-sc: @kmontemayor2-sc -worth upgrading?

svij-sc: Are proto yaml representations guaranteed to be strings?

svij-sc (commented):
a couple qqs

gigl/common/utils/proto_utils.py
kmontemayor2-sc: you mean the keys? I'm pretty sure but it's not a big deal here either way the important part is to ensure it's a dict.

kmontemayor2-sc:
> @kmontemayor2-sc -worth upgrading?

Is what worth upgrading?

gigl/common/utils/proto_utils.py
svij-sc: right we are casting to `dict[str, Any]` should it just be `cast(dict)` ?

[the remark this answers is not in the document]
kmontemayor2-sc: created https://github.com/Snapchat/GiGL/issues/640 :)

svij-sc (approved):

yliu2-sc (approved):

PR #642, opened 2026-05-18
Expose gigl vars as env vars for the custom luancher

kmontemayor2-sc:
Do this so custom launchers can get the task config, etc. 

The alternative approach here is that we can use omegaconf but as we discussed offline that may be more complicated as the resolution would change often.


gigl/env/constants.py
```diff
@@ -0,0 +1,16 @@
...
+These keys are set on the subprocess env (never on the parent
+``os.environ``) by ``gigl.src.common.custom_launcher.launch_custom`` so
+that receiving CLIs can ``os.environ.get(...)`` their runtime context.
+"""
+
+from typing import Final
+
+GIGL_APPLIED_TASK_IDENTIFIER_ENV_KEY: Final[str] = "GIGL_APPLIED_TASK_IDENTIFIER"
+GIGL_TASK_CONFIG_URI_ENV_KEY: Final[str] = "GIGL_TASK_CONFIG_URI"
+GIGL_RESOURCE_CONFIG_URI_ENV_KEY: Final[str] = "GIGL_RESOURCE_CONFIG_URI"
```
svij-sc, line 12: We have `RESOURCE_CONFIG_OS_ENV`  already defined in code. 
See: `gigl/env/pipelines_config.py`

Some opportunity to consolidate / refactor.

svij-sc (commented):
Might be good opportunity to also consolidate existing VAI launchers to use env vars too instead of injected runtime args: https://github.com/Snapchat/GiGL/blob/454b54b8cd1a2b20b6fa8d2347c089438ef7984b/gigl/src/common/vertex_ai_launcher.py#L256-L264

Adopting one distinct pattern will make it easier for us to maintain the different code paths, and I am happy to adapt the env variables given prior discussed arguments as the golden path, but lets also old our old injection paths in existing launchers.

gigl/env/constants.py
```diff
@@ -0,0 +1,16 @@
...
+GIGL_PROCESS_COMMAND_ENV_KEY: Final[str] = "GIGL_PROCESS_COMMAND"
+GIGL_CPU_DOCKER_URI_ENV_KEY: Final[str] = "GIGL_CPU_DOCKER_URI"
+GIGL_CUDA_DOCKER_URI_ENV_KEY: Final[str] = "GIGL_CUDA_DOCKER_URI"
```
svij-sc, line 14: Could you remind me what does cpu and cuda docker images mean in this case?

gigl/env/constants.py
```diff
@@ -0,0 +1,16 @@
...
+``os.environ``) by ``gigl.src.common.custom_launcher.launch_custom`` so
+that receiving CLIs can ``os.environ.get(...)`` their runtime context.
+"""
+
+from typing import Final
+
+GIGL_APPLIED_TASK_IDENTIFIER_ENV_KEY: Final[str] = "GIGL_APPLIED_TASK_IDENTIFIER"
+GIGL_TASK_CONFIG_URI_ENV_KEY: Final[str] = "GIGL_TASK_CONFIG_URI"
+GIGL_RESOURCE_CONFIG_URI_ENV_KEY: Final[str] = "GIGL_RESOURCE_CONFIG_URI"
+GIGL_PROCESS_COMMAND_ENV_KEY: Final[str] = "GIGL_PROCESS_COMMAND"
```
svij-sc: could you remind me what is the process command in this case? 
Adding comments could help.

tests/unit/src/common/custom_launcher_test.py
```diff
@@ -113,6 +123,107 @@ def test_args_with_spaces_are_shell_quoted(self, mock_run: MagicMock) -> None:
...
+            GIGL_RESOURCE_CONFIG_URI_ENV_KEY,
+            GIGL_PROCESS_COMMAND_ENV_KEY,
+            GIGL_CPU_DOCKER_URI_ENV_KEY,
+            GIGL_CUDA_DOCKER_URI_ENV_KEY,
+            GIGL_COMPONENT_ENV_KEY,
+        ):
+            self.assertNotIn(key, os.environ)
+
+    @patch("gigl.src.common.custom_launcher.subprocess.run")
+    def test_dispatch_preserves_inherited_env(self, mock_run: MagicMock) -> None:
```
svij-sc: cant this test and the first test be both combined to reduce verbosity?

tests/unit/src/common/custom_launcher_test.py
```diff
@@ -113,6 +123,107 @@ def test_args_with_spaces_are_shell_quoted(self, mock_run: MagicMock) -> None:
...
+        config = self._build_config(command="echo")
+        launch_custom(
+            custom_launcher_config=config,
+            applied_task_identifier="job",
+            task_config_uri=Uri("gs://bucket/task.yaml"),
+            resource_config_uri=Uri("gs://bucket/resource.yaml"),
+            process_command="echo",
+            process_runtime_args={},
+            cpu_docker_uri=None,
+            cuda_docker_uri=None,
```
svij-sc, line 155: In our code we usually default these
`cuda_docker_uri or DEFAULT_GIGL_RELEASE_SRC_IMAGE_CUDA`

Could you remind me the reason we need to deviate here?

gigl/env/constants.py
kmontemayor2-sc: Hmmm, how about a new `gigl/env/constants.py`? And then migrate the existing RC one in as a follow up?

I think there's two things here tho:

1. These env vars aren't always defined (though we can also add them to the VAI path I suppose which is probably a good idea?)
2. now we have both `dep_constants` and `constants` that may get confusing.


WDYT? I'm fine with adding `constants` and then moving in the existing var in a future PR (in additional to exposing in our VAI jobs)

gigl/env/constants.py
kmontemayor2-sc: This is like `python -m my.launcher.code` I guess we can skip this one?

gigl/env/constants.py
kmontemayor2-sc: they're the containers we launch on e.x vai for custom training.

tests/unit/src/common/custom_launcher_test.py
kmontemayor2-sc: No reason, we can use the default ones if you prefer.

gigl/env/constants.py
svij-sc: This plan sounds good to me.
One small suggestion; should we export these variables @ the start of each component too where possible? 

i.e. currently we have `initialize_metrics(task_config_uri=task_config_uri, service_name=args.job_name)` being called at the start of every component. We could refactor this to `initialize_gigl_env` which can both init the metrics and make available these env variables. Having the same behaviour throughout will bring down the long term complexity of the codebase.

svij-sc: It would also simplify our code below for "passing the env variables to subprocesses" as those env variables would already be available everywhere. 

But open to thoughts.

kmontemayor2-sc: > One small suggestion; should we export these variables @ the start of each component too where possible?

Yes! I can look into setting this up as a follow up :)

kmontemayor2-sc: I'd prefer to have this PR in first as the other changes may be somewhat complicated still (and I'm not sure we need all of these things to be exposed globally)

gigl/env/constants.py
kmontemayor2-sc: removed!

svij-sc (commented):
> I'd prefer to have this PR in first as the other changes may be somewhat complicated still (and I'm not sure we need all of these things to be exposed globally)

Approving conditionally - only for the purposes of unblocking. I am hoping we do a fast follow up here to address concerns outlined below:
1. https://github.com/Snapchat/GiGL/pull/642/changes#r3268620564
2. https://github.com/Snapchat/GiGL/pull/642/changes#r3269575120

svij-sc (approved):

tests/unit/src/common/custom_launcher_test.py
svij-sc: lets do it

kmontemayor2-sc: done.

mkolodner-sc (approved):

PR #643, opened 2026-05-18
Update GBML Config Wrapper to house {node|edge}_type_to_*_maps.

svij-sc:
Prior to this change one would need to manually convert `condensed_{node|edge}_type_*_maps` manually to `{node|edge}_type_to_*_maps` using information available in `graph_metadata`. This causes a lot of duplicity in code when we can just create convenience functions in the gbml wrapper.

### DIFF
--- a/gigl/src/common/types/pb_wrappers/gbml_config.py
+++ b/gigl/src/common/types/pb_wrappers/gbml_config.py
@@ -104,24 +104,30 @@ class GbmlConfigPbWrapper:
                 self,
                 "_node_type_to_feature_dim_map",
                 {
-                    graph_metadata.condensed_node_type_to_node_type_map[c]: d
-                    for c, d in preprocessed_metadata.condensed_node_type_to_feature_dim_map.items()
+                    graph_metadata.condensed_node_type_to_node_type_map[
+                        condensed_node_type
+                    ]: feature_dim
+                    for condensed_node_type, feature_dim in preprocessed_metadata.condensed_node_type_to_feature_dim_map.items()
                 },
             )
             object.__setattr__(
                 self,
                 "_node_type_to_feature_schema_map",
                 {
-                    graph_metadata.condensed_node_type_to_node_type_map[c]: s
-                    for c, s in preprocessed_metadata.condensed_node_type_to_feature_schema_map.items()
+                    graph_metadata.condensed_node_type_to_node_type_map[
+                        condensed_node_type
+                    ]: feature_schema
+                    for condensed_node_type, feature_schema in preprocessed_metadata.condensed_node_type_to_feature_schema_map.items()
                 },
             )
             object.__setattr__(
                 self,
                 "_edge_type_to_feature_dim_map",
                 {
-                    graph_metadata.condensed_edge_type_to_edge_type_map[c]: d
-                    for c, d in preprocessed_metadata.condensed_edge_type_to_feature_dim_map.items()
+                    graph_metadata.condensed_edge_type_to_edge_type_map[
+                        condensed_edge_type
+                    ]: feature_dim
+                    for condensed_edge_type, feature_dim in preprocessed_metadata.condensed_edge_type_to_feature_dim_map.items()
                 },
             )
         # Populate the _flattened_graph_metadata_pb_wrapper field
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -152,7 +152,7 @@ lint = [
   "ruff==0.15.10",
   "mdformat==0.7.22",
   "mdformat_tables==1.0.0",
-  "ty~=0.0.29",
+  "ty==0.0.31",
```
svij-sc, line 155: Newer ty is more restrictive; so actually locking it.

gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -80,6 +92,38 @@ def __post_init__(self):
...
+            self, "_preprocessed_metadata_pb_wrapper"
+        ):
+            graph_metadata = self._graph_metadata_pb_wrapper
+            preprocessed_metadata = self._preprocessed_metadata_pb_wrapper
+            object.__setattr__(
+                self,
+                "_node_type_to_feature_dim_map",
+                {
+                    graph_metadata.condensed_node_type_to_node_type_map[c]: d
+                    for c, d in preprocessed_metadata.condensed_node_type_to_feature_dim_map.items()
```
mkolodner-sc: nit: pref to use better var names than `c` and `d`. Ditto elsewhere.

gigl/src/common/types/pb_wrappers/gbml_config.py
```diff
@@ -365,6 +409,30 @@ def preprocessed_metadata_pb_wrapper(self) -> PreprocessedMetadataPbWrapper:
             )
         return self._preprocessed_metadata_pb_wrapper
 
+    @property
+    def node_type_to_feature_dim_map(self) -> dict[NodeType, int]:
```
mkolodner-sc, line 419: For better cohesion, I think these fields belong on `PreprocessedMetadataPbWrapper` rather than `GbmlConfigPbWrapper`. Many GiGL functions — like [this one](https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/utils/serialized_graph_metadata_translator.py#L49-L51) — accept `PreprocessedMetadataPbWrapper` and `GraphMetadataPbWrapper` directly (not `GbmlConfigPbWrapper`), so they can't benefit from this change as currently placed. Putting the typed-keyed maps on `PreprocessedMetadataPbWrapper` (via a method that accepts a `GraphMetadataPbWrapper`) would make them reachable anywhere the preprocessed metadata is available, and would let callers like the linked function simplify their internal condensed-typed translation.

svij-sc: I originally tried this but for a different reason. We already have the maps on preprocessedMetadataWrapper for `condesed_type : X`; and having `node_type: X` maps are logical property. But, this felt unclean as now we:
1. couple two different classes together i.e. GraphMetadata dataclass wrapper and PreprocessedMetadata dataclass wrapper. 
2. have client code be verbose and client knowing the weird logic to compose these maps:
```python
# with your proposal and my initial thoughts:
gbml_config_pb_wrapper.preprocessed_metadata_pb_wrapper.compute_node_type_to_feature_dim_map(
     gbml_config_pb_wrapper.graph_metadata_pb_wrapper
)

# with the change introduced:
gbml_config_pb_wrapper.node_type_to_feature_dim_map
```

The goal is to make client code less verbose vs. more and to do that through these convenience getters.

Btrw, I am not sure if I agree w/ :
> would make them reachable anywhere the preprocessed metadata is available, and would let callers like the linked function simplify their internal condensed-typed translation 

- We would need to pass around two things instead of one just to compute the values making the code more verbose.
- The example linked doesnt really operate on node_type or edge_type - most likely all our internal code can just operate on condensed types and more of a requirement for client code vs library code.

gigl/src/common/types/pb_wrappers/gbml_config.py
svij-sc: ty, addressed.

gigl/src/common/types/pb_wrappers/gbml_config.py
mkolodner-sc: Discussed offline, alternatives would involve over-coupling GraphMetadata, PreprocessedMetadata, and GBMLConfigPbWrapper or large proto/code changes that would require large refactors. I still have some belief that these fields would be better accessed as `preprocessed_metadata_pb_wrapper.node_type_to_feature_dim_map`, for example, but since there isn't an easy world for us to achieve this, the current solution makes sense. Thanks for the work/discussion!

mkolodner-sc (approved):

kmontemayor2-sc (approved):
Thanks Shubham! I'd prefer if we can have some follow up to update our examples, or at least some tracker that we should do so :)

PR #644, opened 2026-05-18
Fix (seeding) Randomness utility

svij-sc:
Updating random --> determinisitc utility:
1. Remove TF as we are not using it anymore
2. Simplify defaults and update what is being called according to recent dependency updates.

### DIFF
--- a/gigl/common/utils/compute/random.py
+++ b/gigl/common/utils/compute/random.py
@@ -39,7 +39,7 @@ def seed_everything(
     When ``should_enable_expensive_deterministic_compute=True`` (opt-in; default False because it costs
     throughput and should not be enabled for training or for production inference - can be used for debugging purposes.
 
-    - Important: Graph Sampling currently do not follow determism outlined here.
+    - Important: Graph Sampling currently do not follow determinism outlined here.
     Example:
         >>> seed_everything(42)
         42
--- a/gigl/common/utils/compute/random.py
+++ b/gigl/common/utils/compute/random.py
@@ -60,7 +60,7 @@ def seed_everything(
         torch.backends.cudnn.benchmark = False
         torch.backends.cudnn.deterministic = True
         torch.use_deterministic_algorithms(True)
-        logger.info(
+        logger.warning(
             f"seed_everything: seeded python/numpy/torch with seed={seed}; "
             f"expensive deterministic algorithms ON; "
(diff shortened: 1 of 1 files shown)


gigl/common/utils/compute/random.py
```diff
@@ -1,49 +1,69 @@
...
+      (``torch.manual_seed`` calls ``torch.cuda.manual_seed_all`` internally.
+      Also covers PyTorch Geometric.
 
-    if should_consider_numpy:
-        import numpy as np
+    When ``should_enable_expensive_deterministic_compute=True`` (opt-in; default False because it costs
+    throughput and should not be enabled for training or for production inference - can be used for debugging purposes.
 
-        np.random.seed(seed)
+    - Important: Graph Sampling currently do not follow determism outlined here.
```
mkolodner-sc: nit: determism

gigl/common/utils/compute/random.py
```diff
@@ -1,49 +1,69 @@
...
+        deterministic: If True, also enforces bitwise-deterministic torch
+            ops (cudnn flags, ``use_deterministic_algorithms``,
+            ``CUBLAS_WORKSPACE_CONFIG``). Default False — most training
+            pipelines want seeded RNGs without paying the throughput cost.
 
-        torch.manual_seed(seed)
+    """
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
```
mkolodner-sc, line 57: Should we also do `torch.cuda.manual_seed_all(seed)` as is done [here](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9/graphlearn_torch/python/utils/common.py#L31)?

mkolodner-sc (approved):
Thanks!

gigl/common/utils/compute/random.py
svij-sc: not needed, see: https://docs.pytorch.org/docs/2.12/notes/randomness.html#pytorch-random-number-generator

> You can use [torch.manual_seed()](https://docs.pytorch.org/docs/2.12/generated/torch.manual_seed.html#torch.manual_seed) to seed the RNG for all devices (both CPU and CUDA):

yliu2-sc (approved):

PR #645, opened 2026-05-19
Fix PPR sampler memory and labeled homogeneous ABLP

mkolodner-sc:
## Summary

Makes the PPR sampler cheaper and fixes the labeled-homogeneous ABLP edge case that surfaced while exercising PPR through Graph Store.

Changes include:
- Precompute PPR total-degree tensors by node type through `DistDataset.degree_tensor`.
- Store degree tensors as `int32` and share them across sampling workers instead of rebuilding/copying per worker.
- Update `DistPPRNeighborSampler` to consume precomputed degree tensors directly.
- Fix labeled homogeneous ABLP PPR sampling by only passing `etype=None` for true homogeneous graphs.
- Fix PPR for GraphStore mode so that memory is shared for the degree tensor
- Attach single-edge-type PPR outputs directly to homogeneous `Data` batches.
- Expand PPR and degree unit coverage.


gigl/distributed/graph_store/shared_dist_sampling_producer.py
```diff
@@ -871,7 +872,23 @@ def __init__(
...
+                logger.info(
+                    f"Pre-computed degree tensors for PPR sampling across "
+                    f"{len(degree_tensors)} node types."
+                )
+            else:
+                logger.info(
+                    f"Pre-computed degree tensor for PPR sampling with "
+                    f"{degree_tensors.size(0)} nodes."
+                )
+        share_memory(self._degree_tensors)
```
kmontemayor2-sc: qq why don't we have the dataset share it's tensor?

gigl/distributed/utils/degree.py
```diff
@@ -5,8 +5,10 @@
 and aggregate them across distributed machines. Degrees are computed from the
 CSR (Compressed Sparse Row) topology stored in GraphLearn-Torch Graph objects.
 
-Note: Degree tensors are not moved to shared memory and may be duplicated across
-processes on the same machine.
+For homogeneous graphs, callers receive a single ``torch.Tensor``. For
+heterogeneous graphs, degrees are accumulated per anchor node type (summing
+across all edge types incident to that node type) before the distributed
+all-reduce, so callers receive ``dict[NodeType, torch.Tensor]``.
```
kmontemayor2-sc, line 11: Do we lose anything from switching from edge_type to node_type here?

This means we lose fidelity on how we chose to re-queue nodes right? I suppose in the future we can add support for edge_type?

The benefit here is that we can save on memory right? Or is this a correctness issue?

gigl/distributed/utils/degree.py
```diff
@@ -69,52 +86,59 @@ def compute_and_broadcast_degree_tensor(
...
+            degrees = torch.empty(0, dtype=torch.int32)
+        else:
+            degrees = _compute_degrees_from_indptr(topo.indptr)
+
+        if anchor_type in local_dict:
+            existing = local_dict[anchor_type]
+            max_len = max(len(existing), len(degrees))
+            summed = _pad_to_size(existing, max_len).to(torch.int64)
+            summed[: len(degrees)] += degrees.to(torch.int64)
+            local_dict[anchor_type] = summed.to(torch.int32)
```
kmontemayor2-sc: qq why do we use int32 for degree tensor and not like, int16?

kmontemayor2-sc: Also should we clamp here before casting down?

gigl/distributed/utils/degree.py
```diff
@@ -131,33 +155,73 @@ def _pad_to_size(tensor: torch.Tensor, target_size: int) -> torch.Tensor:
...
+    torch.distributed.all_reduce(local_size, op=torch.distributed.ReduceOp.MAX)
+    max_size = int(local_size.item())
+
+    # Pad, convert to int64 for all_reduce, and move to the process-group device.
+    padded = _pad_to_size(tensor, max_size).to(torch.int64).to(device)
+    torch.distributed.all_reduce(padded, op=torch.distributed.ReduceOp.SUM)
+
+    # Correct for over-counting and move back to CPU. We keep int32 so high-degree
+    # nodes do not saturate at int16.
+    return (padded // local_world_size).to(torch.int32).cpu()
```
kmontemayor2-sc: ditto on clamping?

gigl/distributed/utils/neighborloader.py
```diff
@@ -357,6 +357,18 @@ def attach_ppr_outputs(
...
+        if len(ppr_edge_indices) > 1:
+            raise ValueError(
+                "Expected at most one PPR edge type for homogeneous Data output, "
+                f"got {set(ppr_edge_indices.keys())}"
+            )
+        if ppr_edge_indices:
+            edge_type = next(iter(ppr_edge_indices))
+            data.edge_index = ppr_edge_indices[edge_type]
+            data.edge_attr = ppr_weights[edge_type]
+        return
```
kmontemayor2-sc, line 370: why do we early return here?

Also to clarify we did end up on treating ppr edges as real edges right?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -191,57 +182,26 @@ def __init__(
             for nt in all_node_types
         ]
 
-    def _build_total_degree_tensors(
+    def _normalize_degree_tensors(
```
kmontemayor2-sc: should we say this is more like "convert_degree_to_dict` or something? "Normalize" implies to me it's dividing the values by something (e.g. z normalization)

tests/unit/distributed/utils/degree_test.py
```diff
@@ -25,16 +27,58 @@
...
+
+        max_len = max(expected[anchor_node_type].numel(), degrees.numel())
+        summed_degrees = torch.zeros(max_len, dtype=torch.int64)
+        summed_degrees[: expected[anchor_node_type].numel()] += expected[
+            anchor_node_type
+        ].to(torch.int64)
+        summed_degrees[: degrees.numel()] += degrees.to(torch.int64)
+        expected[anchor_node_type] = summed_degrees.to(torch.int32)
+
+    return expected
```
kmontemayor2-sc: instead of this (which looks the same as the business logic) can we just hard-code the expected values?

kmontemayor2-sc (commented):
Thanks Matt! Did a first pass here, fwiw I feel like this could have been multiple PRs for the different fixes / etc but this pr is fine as-is.

gigl/distributed/utils/degree.py
mkolodner-sc: This is to save memory. Previously what we were doing was computing the edge type degree in the dataset and then in the PPR sampler we were computing a node-type indexed total degree tensor. This was creating a new giant degree tensor inside of each sampler process, which was increasing the memory substantially. 

We've aimed to simplify for now by just computing degrees based on node type earlier s.t. it isn't creating duplicate tensors. There is no loss or gain in functionality from what we had previously for the current PPR sampler logic -- it is strictly a memory improvement.

In the future we can aim to have edge type degree breakdowns, but should cross that bridge when that is needed for sampling.

gigl/distributed/utils/neighborloader.py
mkolodner-sc: Yes, ppr edges are treated as real edges with a relation "ppr". For returning early, we do this since otherwise it would go into the heterogeneous logic

gigl/distributed/dist_sampling_producer.py
kmontemayor2-sc, line 1: Robot review, should we also be sharing memory here? (though it's probably best to do it on the dataset?)

>   - Colocated PPR producer doesn't share degree memory, contradicting the docstring (dist_sampling_producer.py:189 vs
  dist_ppr_sampler.py:158-162) — the "workers share a single allocation" guarantee only holds in graph-store mode; colocated workers each get a
  private pickled copy. Either add share_memory() for symmetry or scope the docstring to graph-store mode. (Important)

gigl/distributed/utils/degree.py
mkolodner-sc: We chose int32 for degree tensors because int16 has caused compatibility issues in this path. In particular, torch.distributed.all_reduce on an int16 tensor produced RuntimeError: Invalid scalar type while we were implementing the C++ PPR sampler. Because of that, the C++ PPR logic treats total-degree tensors as int32. This PR (and using int32 here) keeps the Python-side degree tensor dtype aligned with that requirement.

gigl/distributed/graph_store/shared_dist_sampling_producer.py
mkolodner-sc: Good call, moved the share memory to right after its construction in the dataset

gigl/distributed/utils/neighborloader.py
```diff
@@ -357,6 +357,19 @@ def attach_ppr_outputs(
         f"PPR edge index and weight edge types must match, "
         f"got {set(ppr_edge_indices.keys())} vs {set(ppr_weights.keys())}"
     )
+    if isinstance(data, Data):
+        if len(ppr_edge_indices) > 1:
```
kmontemayor2-sc: should we also throw an error if there are no ppr edges?

gigl/distributed/dist_dataset.py
```diff
@@ -341,7 +346,10 @@ def degree_tensor(
             if self.graph is None:
                 raise ValueError("Dataset graph is None. Cannot compute degrees.")
 
-            self._degree_tensor = compute_and_broadcast_degree_tensor(self.graph)
+            self._degree_tensor = compute_and_broadcast_degree_tensor(
+                self.graph, self._edge_dir
+            )
+        share_memory(entity=self._degree_tensor)
```
kmontemayor2-sc: ugh sorry to go back and forth on this, it may be weird to do this here instead of in `share_ipc`? Or do we not call `degree_tensor` until the subprocess launch?

In fact we do already share memory in `share_ipc` [1], is the problem here that we don't call `degree_tensor` until *after* we're in the subprocesses already?  

Can you remind me where we first call `degree_tensor` and where that's located in the process tree?

[1]: https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/dist_dataset.py#L981

mkolodner-sc: > Can you remind me where we first call degree_tensor and where that's located in the process tree?

Currently, the degree tensor is not built when we call `build_dataset`. Instead, it is first built after we are inside one of the data loaders and know we are doing PPR sampling (specified by the SamplingOptions). 


> In fact we do already share memory in share_ipc [1](https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/dist_dataset.py#L981), is the problem here that we don't call degree_tensor until after we're in the subprocesses already?

Yes exactly, this was calling failures specifically on large-scale graph store cases since it was creating copies of the tensor before we had an opportunity to call share_ipc (and thereby share the memory of the degree tensor). The original fix I had here handled this by explicitly calling share memory before handing the degree tensor to the sampling workers for the GS setting, where otherwise it was creating many copies of this degree tensor. 

I think the current solution to call share_memory immediately after building the degree tensor makes the most sense, since it doesn't need to rely on share_ipc and provides the most confidence for putting the degree tensor onto shared memory.

gigl/distributed/utils/degree.py
kmontemayor2-sc: sg! Can we document that here?

mkolodner-sc: Done :)

gigl/distributed/dist_dataset.py
kmontemayor2-sc: We'd still have `local world size` copies of this object in shared memory then right? Unless we want to create the degree tensor in the top-process?

IIRC our goal here was to abstract this away from the user s.t. they don't need to pass in this implementation detail into the loaders right?

And in that vein it's probably weird if we ask them to call `dataset.degree_tensor` naked in the top-level process? I think ideally we can poke at `/dev/shm` directly to setup the shared memory properly across the local nodes here. 

My concern with calling `share_memory` *inside* the property is that it may be surprising to users share memory tensors do have additional restrictions on them iirc - I guess the distdataset already does that under the hood so maybe it's not too bad.

I guess I have a slight preference here for calling `share_memory` in our loaders after we get this tensor, but that's up to you. WDYT?

mkolodner-sc: > We'd still have local world size copies of this object in shared memory then right? Unless we want to create the degree tensor in the top-process?
IIRC our goal here was to abstract this away from the user s.t. they don't need to pass in this implementation detail into the loaders right?
And in that vein it's probably weird if we ask them to call dataset.degree_tensor naked in the top-level process? I think ideally we can poke at /dev/shm directly to setup the shared memory properly across the local nodes here.


Correct, this was a consideration made in the initial design. The tradeoff with the current approach is that there may still be `local_world_size` copies of the degree tensor, but since the degree tensor isn't impacting the memory bottleneck, we proceeded with this approach. Since this is the intention of the design and isn't blocking training or inference with PPR, I'd prefer if we save any further optimization as a follow-up if that becomes a blocker. One note is that the graph_store case doesn't have this constraint I believe, since it makes the share_memory call from the `DistServer`, so there are not `local world size` copies in that setting.

> My concern with calling share_memory inside the property is that it may be surprising to users share memory tensors do have additional restrictions on them iirc - I guess the distdataset already does that under the hood so maybe it's not too bad.
I guess I have a slight preference here for calling share_memory in our loaders after we get this tensor, but that's up to you. WDYT?

That makes sense too, I can move the `share_memory` call back to the loaders in that case.

kmontemayor2-sc (approved):

yliu2-sc (approved):

PR #646, opened 2026-05-19
Fix latent bugs surfaced by ty migration

kmontemayor2-sc:
Resolves 4 `# ty: ignore` comments from PR https://github.com/Snapchat/GiGL/pull/585 by fixing the underlying logic (not suppressing). Includes None-division risk in loss.py and masked None flow in optimizer code paths.


gigl/src/common/models/layers/loss.py
```diff
@@ -100,18 +100,27 @@ def forward(
 
 
 class SoftmaxLoss(nn.Module):
```
svij-sc, line 102: Side node, this isnt really Softmax loss but we have been calling it that.
cc @yliu2-sc  - isnt this just a flavor of InfoNCE ?

tests/unit/experimental/torchrec_utils_test.py
```diff
@@ -0,0 +1,83 @@
+"""Regression tests for ``apply_sparse_optimizer`` in ``gigl.experimental.knowledge_graph_embedding.common.torchrec.utils``.
```
svij-sc: do we need test for this experimental code?
Are these auto generated?

tests/unit/src/common/models/layers/legacy_loss_test.py
```diff
@@ -0,0 +1,78 @@
...
+    to ``Tensor``. Until then, ty rejects the call, so we ignore the
+    specialization mismatch at the construction site.
+    """
+    pos_scores = torch.tensor([[0.8, 0.6]])
+    hard_neg_scores = torch.tensor([[0.1, 0.2]])
+    random_neg_scores = torch.tensor([[0.05, 0.15]])
+    return {
+        CondensedEdgeType(0): BatchScores(
+            pos_scores=pos_scores,  # ty: ignore[invalid-argument-type] TODO(ty-torch-tensor-specialization): cleared by PR 3 relaxing BatchScores to Tensor.
+            hard_neg_scores=hard_neg_scores,  # ty: ignore[invalid-argument-type] TODO(ty-torch-tensor-specialization): cleared by PR 3 relaxing BatchScores to Tensor.
```
svij-sc: can we take a manual pass at this test first?

tests/unit/experimental/torchrec_utils_test.py
kmontemayor2-sc: they were auto-generated, would you rather remove them?

tests/unit/src/common/models/layers/legacy_loss_test.py
kmontemayor2-sc: Hmmm do you mean why are we adding the new type ignores? I have a follow-up pr which fixes these.

gigl/src/common/models/layers/loss.py
yliu2-sc: this is not even used in any runs or configs

svij-sc (approved):

yliu2-sc (approved):

PR #648, opened 2026-05-20
Migrate RESOURCE_CONFIG_PATH env var to GIGL_RESOURCE_CONFIG_URI

kmontemayor2-sc:
Aligns the resource-config env var with the GIGL_* convention introduced in #642 and the existing GIGL_RESOURCE_CONFIG_URI_ENV_KEY constant. All writers (get_resource_config, launch_graph_store_local.py, Makefile's notebooks_test target) now set the new name; readers fall back to the deprecated RESOURCE_CONFIG_PATH with a one-time warning so externally-set values keep working for one release cycle.

**Scope of work done**

### DIFF
--- a/gigl/env/constants.py
+++ b/gigl/env/constants.py
@@ -1,11 +1,21 @@
-"""Environment-variable keys exported by ``launch_custom``.
+"""Environment-variable keys used across GiGL.
 
-These keys are set on the subprocess env (never on the parent
-``os.environ``) by ``gigl.src.common.custom_launcher.launch_custom`` so
-that receiving CLIs can ``os.environ.get(...)`` their runtime context.
+Most of these keys are set on subprocess env (never on the parent
+``os.environ``) by ``gigl.src.common.custom_launcher.launch_custom`` so that
+receiving CLIs can ``os.environ.get(...)`` their runtime context.
+
+``GIGL_RESOURCE_CONFIG_URI`` is also written to the parent ``os.environ`` by
+``gigl.env.pipelines_config.get_resource_config`` so that downstream readers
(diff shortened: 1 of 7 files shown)


gigl/env/constants.py
```diff
@@ -1,15 +1,54 @@
...
 GIGL_CPU_DOCKER_URI_ENV_KEY: Final[str] = "GIGL_CPU_DOCKER_URI"
 GIGL_CUDA_DOCKER_URI_ENV_KEY: Final[str] = "GIGL_CUDA_DOCKER_URI"
 GIGL_COMPONENT_ENV_KEY: Final[str] = "GIGL_COMPONENT"
+
+_LEGACY_RESOURCE_CONFIG_ENV_KEY: Final[str] = "RESOURCE_CONFIG_PATH"
+_legacy_resource_config_env_warned: bool = False
+_logger = Logger()
+
+
+def read_resource_config_uri_from_env() -> Optional[str]:
```
svij-sc, line 32: Is the plan to remove this in the next release?
Side note: Btw, I remember we did this prior with some other deprecations - have the prior changes been migrated? i.e. logs / migration is compelte?

svij-sc (approved):

mkolodner-sc (approved):

gigl/env/constants.py
kmontemayor2-sc: Yeah we can remove eventually, this should only impact gigl devs so I can announce internally.

> Side note: Btw, I remember we did this prior with some other deprecations - have the prior changes been migrated? i.e. logs / migration is compelte?

I'm not sure what you're referring to here? We can resolve as a follow up if urgent :)

PR #653, opened 2026-05-27
Expose GiGL env vars in Vertex AI launcher

kmontemayor2-sc:
As a follow up to https://github.com/Snapchat/GiGL/pull/642 we should expose the same vars on vai.

**Scope of work done**

### DIFF
--- a/gigl/src/common/vertex_ai_launcher.py
+++ b/gigl/src/common/vertex_ai_launcher.py
@@ -51,7 +51,6 @@ _VALID_RESERVATION_AFFINITY_TYPES: Final[frozenset[str]] = frozenset(
 def launch_single_pool_job(
     vertex_ai_resource_config: VertexAiResourceConfig,
     job_name: str,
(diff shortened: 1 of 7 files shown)


gigl/src/common/vertex_ai_launcher.py
```diff
@@ -43,6 +51,7 @@
 def launch_single_pool_job(
     vertex_ai_resource_config: VertexAiResourceConfig,
     job_name: str,
+    applied_task_identifier: str,
```
mkolodner-sc: Do we need to add applied_task_identifier here in the first place? IIRC we use job_name typically as the applied task identifier.

kmontemayor2-sc: hmmm good callout, simplified to base the job name based off component and removed the new arg.

Nice catch!

mkolodner-sc (approved):
Thanks!

yliu2-sc (approved):

PR #655, opened 2026-05-29
Add graph-store PPR E2E wiring

mkolodner-sc:
## Summary

Stacked on #645.

Enables PPR sampling in the existing homogeneous Graph Store training and inference entrypoints, then adds a short E2E test that exercises that path.

Changes include:
- Add `parse_sampler_options` for `sampler_type: ppr` task args.
- Thread optional `SamplerOptions` through homogeneous Graph Store training and inference loaders.
- Keep k-hop as the default behavior when no sampler type is provided.
- Default Graph Store `local_world_size` from `GraphStoreInfo` and fail fast when it disagrees with the cluster topology.
- Add the homogeneous Cora Graph Store PPR E2E config, E2E test registration, and Makefile target.


examples/link_prediction/graph_store/homogeneous_inference.py
```diff
@@ -110,16 +111,10 @@
 from gigl.src.common.utils.bq import BqUtils
 from gigl.src.common.utils.model import load_state_dict_from_uri
 from gigl.src.inference.lib.assets import InferenceAssets
-from gigl.utils.sampling import parse_fanout
+from gigl.utils.sampling import parse_fanout, parse_sampler_options
 
 logger = Logger()
 
-# Default number of inference processes per machine incase one isnt provided in inference args
```
mkolodner-sc, line 117: In Graph Store mode, the source of truth should be cluster_info.num_processes_per_compute, not a local CPU/GPU heuristic. The previous fallback could make inference spawn a different number of compute processes than storage expected, causing storage rendezvous failures like “only N/M clients joined.”

gigl/utils/sampling.py
```diff
@@ -88,6 +90,45 @@ def parse_fanout(fanout_str: str) -> Union[list[int], dict[EdgeType, list[int]]]
...
+    if num_neighbors_per_hop is None:
+        num_neighbors_per_hop = args.get("ppr_num_neighbors_per_hop", "1000")
+
+    return PPRSamplerOptions(
+        alpha=float(args.get("ppr_alpha", "0.5")),
+        eps=float(args.get("ppr_eps", "0.0001")),
+        max_ppr_nodes=int(max_ppr_nodes),
+        num_neighbors_per_hop=int(num_neighbors_per_hop),
+        max_fetch_iterations=_parse_optional_int(args.get("ppr_max_fetch_iterations")),
+    )
```
kmontemayor2-sc: instead of this can we encode `PPRSamplerOptions` as a json dict in the config? 

Also IMO it's a bit weird to have the sampling parameterized like this since I don't think the model can / should be the same for PPR vs khop sampling right?

gigl/utils/sampling.py
```diff
@@ -88,6 +90,45 @@ def parse_fanout(fanout_str: str) -> Union[list[int], dict[EdgeType, list[int]]]
...
+    return int(value)
+
+
+def parse_sampler_options(args: Mapping[str, str]) -> Optional[SamplerOptions]:
+    sampler_type = args.get("sampler_type", "khop").strip().lower().replace("-", "_")
+    if sampler_type == "":
+        sampler_type = "khop"
+
+    if sampler_type in {"khop", "k_hop", "neighbor", "neighbor_sampler"}:
+        return None
```
kmontemayor2-sc: should we also valdiate fanout / etc here if khop?

gigl/utils/sampling.py
```diff
@@ -88,6 +90,45 @@ def parse_fanout(fanout_str: str) -> Union[list[int], dict[EdgeType, list[int]]]
...
+def _parse_optional_int(value: Optional[str]) -> Optional[int]:
+    if value is None:
+        return None
+    normalized = value.strip().lower()
+    if normalized in {"", "none", "null"}:
+        return None
+    return int(value)
+
+
+def parse_sampler_options(args: Mapping[str, str]) -> Optional[SamplerOptions]:
```
kmontemayor2-sc: This function is good to add but I think we should more rigorously define how it should be used / etc.

mkolodner-sc: Hmm, thanks for drawing attention to this. I thought about this some more and have some thoughts that it might be too opinionated to have the parsing done this way, since it's not very rigorous at the moment and will need to be updated if any of our args change for the PPRSamplerOptions in the future, for example.

Opted instead to have it be parsed directly by the trainer/inferencer logic with a json.loads().

examples/link_prediction/graph_store/configs/e2e_hom_cora_sup_gs_ppr_task_config.yaml
```diff
@@ -0,0 +1,79 @@
...
+    log_every_n_batch: "1"
+    num_neighbors: "[10, 10]"
+    ppr_sampler_options: >-
+      {
+        "alpha": 0.5,
+        "eps": 0.0001,
+        "max_ppr_nodes": 20,
+        "num_neighbors_per_hop": 100,
+        "max_fetch_iterations": 2
+      }
```
kmontemayor2-sc, line 67: can we add comments here about what there are" (e.g. pointing to the sampler options class)

kmontemayor2-sc: Also, it's a bit odd that we sometimes use `ast.literal_eval` and sometimes use `json`, since we need `literal_eval` should we also use that here?

gigl/utils/sampling.py
kmontemayor2-sc: > Also IMO it's a bit weird to have the sampling parameterized like this since I don't think the model can / should be the same for PPR vs khop sampling right?

Ping on this :)

kmontemayor2-sc (commented):
Actually, is there a reason we want a full e2e test here? Why not just example the graph store integration test [1]?

We don't really need the "full" e2e test suite here right? And I feel like doing it this way makes our examples more confusing.

[1]: https://github.com/Snapchat/GiGL/blob/main/tests/integration/distributed/graph_store/graph_store_integration_test.py

PR #661, opened 2026-06-01
Add utils for exporting shared gigl variables

kmontemayor2-sc:
Similar to https://github.com/Snapchat/GiGL/pull/653 and https://github.com/Snapchat/GiGL/pull/642 we should expose these env vars in other jobs (e.x preprocessor user code). 

This just adds some utils so it's easier to do in the future, namely `initialize_gigl_runtime` which we will call instead of `initialize_metrics` so that the env vars will be exported everywhere. 

Will actually make the above change as a follow up pr.


gigl/src/common/utils/gigl_runtime.py
```diff
@@ -0,0 +1,43 @@
...
+        get_gigl_runtime_env_vars(
+            applied_task_identifier=applied_task_identifier,
+            task_config_uri=task_config_uri,
+            resource_config_uri=resource_config_uri,
+            component=component,
+            cpu_docker_uri=cpu_docker_uri,
+            cuda_docker_uri=cuda_docker_uri,
+        )
+    )
+    initialize_metrics(task_config_uri=task_config_uri, service_name=service_name)
```
mkolodner-sc, line 42: Hmm, should we be coupling the metric initialization with the runtime environment? Is there a reason for including it here?

mkolodner-sc (approved):
Thanks Kyle! Generally LGTM provided one question about using `initialize_metrics` with this utility is addressed.

gigl/src/common/utils/gigl_runtime.py
kmontemayor2-sc: This was something I was chatting with @svij-sc about a while ago. 

The idea here is that the metrics are part of the "Gigl environment" as as these env vars. We still have the `initialize_metrics` you can use here too. But in the future if there was something similar to `metrics` (say idk email alerter?) then I thnk it'd make sense to also do it in the env function.

gigl/src/common/utils/gigl_runtime.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Optional
+
+from gigl.common import Uri
+from gigl.src.common.constants.components import GiGLComponents
+from gigl.src.common.utils.gigl_env import get_gigl_runtime_env_vars
+from gigl.src.common.utils.metrics_service_provider import initialize_metrics
+
+
+def initialize_gigl_runtime(
+    *,
```
svij-sc: robot syntax `*`  - we do not adopt this a pattern in codebase. Lets update.

gigl/src/common/utils/gigl_runtime.py
svij-sc: @kmontemayor2-sc  - I am guessing in a future PR the plan is to replace the `initialize_metrics` calls with `initialize_gigl_runtime` ?

svij-sc (approved):
Approving, but lets address the above comments please.

gigl/src/common/utils/gigl_runtime.py
kmontemayor2-sc: we actually do this some places and I think it's a good pattern in general. I can revert if this is blocking but I did read and intend for this

gigl/src/common/utils/gigl_runtime.py
kmontemayor2-sc: yes it will be a follow up pr

gigl/src/common/utils/gigl_runtime.py
kmontemayor2-sc: Reverted.

PR #663, opened 2026-06-02
Enable `initialize_gigl_runtime` in all components 

kmontemayor2-sc:
**Scope of work done**

As a follow up to https://github.com/Snapchat/GiGL/pull/661 we should expose these env vars everywhere (this let's user code be consistent across the code base). 


gigl/orchestration/local/runner.py
```diff
@@ -107,6 +103,28 @@ def run(
...
+    def _initialize_component_runtime(
+        pipeline_config: PipelineConfig,
+        component: GiGLComponents,
+    ) -> None:
+        if component in {GiGLComponents.SubgraphSampler, GiGLComponents.SplitGenerator}:
+            initialize_metrics(
+                task_config_uri=pipeline_config.task_config_uri,
+                service_name=pipeline_config.applied_task_identifier,
+            )
+            return
```
mkolodner-sc: Could we merge this logic into initialize_gigl_runtime so that we don't need an additional static_method to call here?

gigl/orchestration/kubeflow/components/data_preprocessor/component.yaml
```diff
@@ -16,4 +18,6 @@ implementation:
       --task_config_uri, {inputValue: task_config_uri},
       --resource_config_uri, {inputValue: resource_config_uri},
       --custom_worker_image_uri, {inputValue: custom_worker_image_uri},
+      --cpu_docker_uri, {inputValue: cpu_docker_uri},
+      --cuda_docker_uri, {inputValue: cuda_docker_uri},
```
mkolodner-sc, line 22: Why do we need to add this to the data preprocessor component, seeing as there are still components like  config populator which don't have these fields?

gigl/orchestration/kubeflow/components/post_processor/component.yaml
```diff
@@ -14,4 +16,6 @@ implementation:
       --job_name, {inputValue: job_name},
       --task_config_uri, {inputValue: task_config_uri},
       --resource_config_uri, {inputValue: resource_config_uri},
+      --cpu_docker_uri, {inputValue: cpu_docker_uri},
+      --cuda_docker_uri, {inputValue: cuda_docker_uri},
```
mkolodner-sc, line 20: ditto question for post processor

gigl/src/training/v1/trainer.py
```diff
@@ -48,11 +49,24 @@ def run(
...
+                *[
+                    env_var.EnvVar(name=name, value=value)
+                    for name, value in get_gigl_runtime_env_vars(
+                        applied_task_identifier=applied_task_identifier,
+                        task_config_uri=task_config_uri,
+                        resource_config_uri=resource_config_uri,
+                        component=GiGLComponents.Trainer,
+                        cpu_docker_uri=cpu_docker_uri,
+                        cuda_docker_uri=cuda_docker_uri,
+                    ).items()
```
mkolodner-sc, line 61: Were we not setting any env vars for V1 prior to this change?

mkolodner-sc (commented):
Thanks Kyle! A few comments/questions

gigl/orchestration/kubeflow/components/data_preprocessor/component.yaml
kmontemayor2-sc: I want to have a consistent "GiGL env" for all potential user-defined code, e.g. a user can rely on the variables being always available.

Potentially, we should update the config populator too, (And if you want I can add that in this PR, or as a follow up?).

gigl/orchestration/kubeflow/components/post_processor/component.yaml
kmontemayor2-sc: let's continue the discussion in the other thread https://github.com/Snapchat/GiGL/pull/663/changes#r3382941360 :)

gigl/orchestration/local/runner.py
kmontemayor2-sc: good call, done :)

gigl/src/training/v1/trainer.py
kmontemayor2-sc: I guess not lmao

gigl/orchestration/kubeflow/components/data_preprocessor/component.yaml
mkolodner-sc: Gotcha, in that case could we make sure it is added for all components then as part of this PR?

mkolodner-sc: Also, dumb question -- is there a dataflow_docker_uri that could exist (analogous to the 3 kinds of images we produce) is it just cpu and cuda docker uris?

kmontemayor2-sc: > Gotcha, in that case could we make sure it is added for all components then as part of this PR?

Done

> Also, dumb question -- is there a dataflow_docker_uri that could exist (analogous to the 3 kinds of images we produce) is it just cpu and cuda docker uris?

I suppose so - we don't have a use for it right now though so should we can add in the future (adding it would make this PR quite a bit bigger I'd prefer to not do it here)

mkolodner-sc: > I suppose so - we don't have a use for it right now though so should we can add in the future (adding it would make this PR quite a bit bigger I'd prefer to not do it here)

Thanks! Can we add a TODO for this in that case?

mkolodner-sc (approved):
Thanks Kyle!

yliu2-sc (approved):

PR #666, opened 2026-06-02
Pass resource config URI to graph store networking integration test workers

kmontemayor2-sc:
get_graph_store_info() reads get_resource_config() to build the readiness URI (added in #533), but the launched Vertex AI workers had no GIGL_RESOURCE_CONFIG_URI in their env, so they failed with "No resource config provided". Inject it via the compute pool's environment_variables, which launch_graph_store_job propagates to both compute and storage pools.

**Scope of work done**

Where is the documentation for this feature?: N/A

Did you add automated tests or write a test plan?

***Updated Changelog.md?*** NO

***Ready for code review?:*** NO

PR #668, opened 2026-06-04
Run integration Vertex AI tests against a freshly-built source image

kmontemayor2-sc:
Follow up https://github.com/Snapchat/GiGL/pull/666 - would have caught this error much earlier.

  Fixes a source/image skew: the integration tests that launch Vertex AI jobs used to run their
  workers on the **pinned release image** (`DEFAULT_GIGL_RELEASE_SRC_IMAGE_CPU` = `src-cpu:0.2.0`),
  so worker-side source changes (e.g. `get_graph_store_info`) weren't validated until a release.

  Now `make integration_test` **always builds a fresh `src-cpu` image from the current source** and
  runs the suite against it (the Vertex-AI-launching tests pick it up via `GIGL_CPU_DOCKER_URI`):

  - **Makefile**: `integration_test` builds+pushes `${INTEGRATION_TEST_CPU_IMAGE}` then runs
    `tests.integration.main` with `GIGL_CPU_DOCKER_URI` exported. New
    `INTEGRATION_TEST_CPU_IMAGE_TAG`/`INTEGRATION_TEST_CPU_IMAGE` vars (tag defaults to `${DATE}` locally).
  - **Tests** (`tests/integration/{distributed/utils/networking_test.py, common/services/vertex_ai_test.py}`):
    read `GIGL_CPU_DOCKER_URI` fail-fast in `setUp`; workers run real functions on the fresh image —
    `_assert_graph_store_info(...)` and `_assert_machine_cpu_count(...)` (verifies provisioned vCPU
    count per pool); explicit `timeout_s` on all CustomJob configs.
  - **CI**: `on-pr-merge` `ci-integration-test` and the `/integration_test` comment job pass an
    immutable per-run tag `${{ github.run_id }}.${{ github.run_attempt }}`.
  - **Docs**: CLAUDE.md + README.md updated.

  Net diff is 7 files (the branch history first added a `tests/smoke/` suite, then folded it back into
  `tests/integration` — no `tests/smoke/` remains).

  ## Test Plan
  - [x] `ty`, `ruff`, `mdformat --check` pass on changed files
  - [x] unittest discovery collects the integration suite (72 cases); worker fns import
  - [x] `make -n integration_test` dry-run: build image tag == exported `GIGL_CPU_DOCKER_URI`
  - [ ] Merge-queue CI runs `make integration_test` (builds image + launches Vertex AI jobs) green


.github/workflows/on-pr-comment.yml
```diff
@@ -122,7 +122,8 @@ jobs:
         workload_identity_provider: ${{ secrets.WORKLOAD_IDENTITY_PROVIDER }}
         gcp_service_account_email: ${{ secrets.GCP_SERVICE_ACCOUNT_EMAIL }}
         command: |
-            make integration_test
+            # Immutable per-run image tag so concurrent runs can't overwrite each other's tag.
```
mkolodner-sc, line 125: Does this work? This puts a shell comment inside the `command:` block. That command is later passed as _CMD and executed via $_CMD in .github/cloud_builder/run_command_on_active_checkout.yaml:50, where expanded # is not parsed as a comment. This may fail before make integration_test with #: command not found

tests/integration/common/services/vertex_ai_test.py
```diff
@@ -66,19 +93,24 @@ def setUp(self):
...
-    def tearDown(self):
+    def tearDown(self) -> None:
         super().tearDown()
 
     def test_launch_job(self):
         job_name = f"GiGL-Integration-Test-{uuid.uuid4()}"
-        container_uri = "condaforge/miniforge3:25.3.0-1"
-        command = ["python", "-c", "import logging; logging.info('Hello, World!')"]
-
+        command = [
```
mkolodner-sc, line 101: Robot review:

Vertex AI worker commands import test-only dependencies missing from src-cpu
vertex_ai_test.py:104/140/153 and networking_test.py:135 import helper functions from test modules inside the freshly built runtime image. Those modules import parameterized at top level (vertex_ai_test.py:8, networking_test.py:6), but parameterized is only in the test dependency group (pyproject.toml:148), while the CPU image uses non-dev install and Dockerfile.src only runs uv pip install .. The remote workers will likely fail on import before reaching the assertions.
Fix: put worker entrypoints in a minimal module with only runtime deps, or inline the tiny commands so they do not import test modules.

mkolodner-sc (commented):
Thanks Kyle! Mostly LGTM, a few comments

.github/workflows/on-pr-comment.yml
kmontemayor2-sc: Good call, moved comment up :)

tests/integration/common/services/vertex_ai_test.py
kmontemayor2-sc: This is wrong - we do include [1] `tests/` in our docker images (no wether or not we should a different question...)


[1] https://github.com/Snapchat/GiGL/blob/8c75c4d243a86ad00bbe2020bac30fbf34a11db1/containers/Dockerfile.src#L15

yliu2-sc (approved):

svij-sc (approved):

PR #676, opened 2026-06-22
Remove positive/negative-edge zero-fanout workaround in k-hop sampler

kmontemayor2-sc:
DistNeighborSampler now skips label (positive/negative supervision) edges directly in its k-hop loop, mirroring DistPPRNeighborSampler. As a result patch_fanout_for_sampling no longer injects zero fanout for label edges; it excludes them from num_neighbors instead, resolving the long-standing TODO at neighborloader.py. Behavior-preserving (zero fanout and skipping produce identical subgraphs) and covers both colocated and graph-store modes.

- dist_neighbor_sampler.py: skip is_label_edge_type edges before indexing num_neighbors
- neighborloader.py: patch_fanout_for_sampling excludes label edges; add empty-edges guard
- base_dist_loader.py: fix stale create_sampling_config docstring
- neighborloader_test.py: update expectations (labels excluded, not zeroed)

**Scope of work done**


gigl/distributed/utils/neighborloader.py
```diff
@@ -79,32 +80,43 @@ def patch_fanout_for_sampling(
         if not all(hop >= 0 for hop in num_neighbors):
             raise ValueError(f"Hops provided must be non-negative, got {num_neighbors}")
         return num_neighbors
+
+    message_passing_edge_types = [
+        edge_type for edge_type in edge_types if not is_label_edge_type(edge_type)
+    ]
```
mkolodner-sc, line 91: Is this just here in case a user still provides a fanout with the labeled edge type?

mkolodner-sc (approved):
Thanks Kyle!

gigl/distributed/utils/neighborloader.py
kmontemayor2-sc: Hmmm that's a good point actually - do you think we should allow passing in label edge types? Maybe `user, like, content` becomes a label edge type, but we also want to use it for some other task?

WDYT? I guess we can allow fanout here (with a warning?)

kmontemayor2-sc: Synced offline, we now throw if: 

1. there's any label edge types
2. There's no edge types.

yliu2-sc (approved):

PR #681, opened 2026-06-25
  DistABLPLoader: vectorized label remap + optional AnchorLabels edge-list output

kmontemayor2-sc:
## Summary

Reworks how `DistABLPLoader` remaps Anchor-Based Link Prediction (ABLP) labels from global node IDs to sampled-subgraph-local indices and introduces an optional PyG-style `[2, E]` label edge-index output.

At production scale, we observed approximately 60% lower ABLP next-batch latency in colocated mode and 30% lower latency in Graph Store mode (roughly 40s → 15s and 60s → 40s, respectively).

### Vectorized label remapping

Label remapping now uses a vectorized sorted-membership lookup (`torch.searchsorted`) instead of a Python loop over anchors.

The remapper assumes that sampling fans out over every non-padding label. If a label is unexpectedly absent from the sampled subgraph, the loader fails fast rather than silently omitting supervision.

Pair order within an anchor is unspecified. The ABLP contrastive loss is order-invariant, so the output remains loss-equivalent to the previous representation.

### Label edge-index output

`DistABLPLoader(..., use_label_edge_index_output=True)` returns labels as `[2, E]` tensors:

- Row 0 contains local anchor indices.
- Row 1 contains local label-node indices in the sampled supervision node store.

This matches PyG’s conventional edge-index representation and lets training code consume all label pairs without per-anchor Python iteration.

The default remains `use_label_edge_index_output=False` for backward compatibility. This returns the existing ragged `dict[int, torch.Tensor]` representation, although that path is deprecated and emits a `FutureWarning`.

### Utilities and documentation

The remapping and legacy dictionary conversion logic now live in `gigl/distributed/utils/ablp.py`.

The module documents:

- Tensor dimensions and shape transitions.
- The `[2, E]` row semantics.
- Concrete input/output examples.
- The equivalent legacy dictionary representation.
- A Graphviz representation of the example label graph.

### Examples

The homogeneous and heterogeneous link-prediction examples for both colocated and Graph Store modes now consume the label edge-index output directly.

## Performance

In a synthetic CPU microbenchmark with 8,192 sampled nodes, 512 anchors, and 8 labels per anchor, median remapping times were:

- Previous per-anchor dictionary loop: 106.2 ms
- Direct label edge-index output: 1.45 ms
- Vectorized remapping plus legacy dictionary conversion: 2.53 ms

The compatibility path was approximately 42× faster than the previous implementation in this benchmark, while direct tensor output was approximately 73× faster.

## Tests

Coverage includes:

- Constructed expected outputs for padded, duplicate-label, homogeneous, heterogeneous, positive/negative, and multi-edge-type inputs.
- Equivalent labels from the tensor and legacy dictionary output paths.
- Exact `[2, E]` output semantics.
- Empty and fully padded anchor batches.
- Unsorted local-to-global node maps.
- CUDA device-placement equivalence with CPU output.
- A `ValueError` for non-unique local-to-global node maps.
- A `ValueError` when a non-padding label is unexpectedly absent from the sampled subgraph.

Focused formatting, linting, type checking, utility tests, loader equivalence tests, and the CUDA regression test pass.


gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+        ...     num_anchors=3,
+        ... )
+        >>> labels.to_dict()
+        {0: tensor([3]), 1: tensor([5, 7]), 2: tensor([], dtype=torch.int64)}
+
+    Args:
+        anchor_index (torch.Tensor): ``[E]`` long tensor of local anchor rows.
+        label_index (torch.Tensor): ``[E]`` long tensor of local label node ids.
+        num_anchors (int): Total number of anchors ``N_anchors`` (rows of the
+            source padded label tensor), including anchors with no labels.
```
mkolodner-sc: Do we need the num_anchors data_class? This should already be available in Data/HeteroData object right

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+        ...     anchor_index=torch.tensor([0, 1, 1], dtype=torch.long),
+        ...     label_index=torch.tensor([3, 5, 7], dtype=torch.long),
+        ...     num_anchors=3,
+        ... )
+        >>> labels.to_dict()
+        {0: tensor([3]), 1: tensor([5, 7]), 2: tensor([], dtype=torch.int64)}
+
+    Args:
+        anchor_index (torch.Tensor): ``[E]`` long tensor of local anchor rows.
+        label_index (torch.Tensor): ``[E]`` long tensor of local label node ids.
```
mkolodner-sc: I'm wondering if we can represent this instead as a [2, E] edge tensor, since that is essentially what it seems to be used for here. This would make it so we'd only need one .to(device) call. If we can make this just one tensor instead of two and we can remove the num_anchors field as well (which already exists in the HeteroData), I'm wondering if we can remove the need for a dataclass here all together. 

The benefit is that the resulting labeled object would just be an edge tensor and would feel more PyG-esque and natural to use, rather than a GiGL-specific dataclass for labeled objects. A [2, E] (labeled) edge index tensor also feels similar to what we use for other objects such as the message passing edge index tensor or PPR edges, which we don't have custom dataclasses for.

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
     select_label_edge_types,
 )
-from gigl.utils.data_splitters import get_labels_for_anchor_nodes
+from gigl.utils.data_splitters import PADDING_NODE, get_labels_for_anchor_nodes
 from gigl.utils.sampling import ABLPInputNodes
 
 logger = Logger()
 
 
+@dataclass(frozen=True)
```
mkolodner-sc: Could we move the newly introduced remapping logic to a separate utility file, maybe creating a new ablp_loader utility file? Might help make our code a bit easier to read/follow-through s.t. if users are clicking through and trying to understand what/how to use ABLP loader at a high level, they don't have to be exposed to this internal label mapping logic as the first thing they see.

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+
+    anchor_index: torch.Tensor
+    label_index: torch.Tensor
+    num_anchors: int
+
+    def to_dict(self) -> dict[int, torch.Tensor]:
+        """Expand to the legacy ragged ``dict[int, torch.Tensor]`` form.
+
+        Every anchor ``0..num_anchors-1`` receives a key; anchors with no labels
+        map to an empty ``long`` tensor on the same device as ``label_index``.
```
mkolodner-sc: Can we add a note here that this we may eventually want to  remove this path in the future?

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -213,11 +524,20 @@ def __init__(
...
+                ``y.label_index`` and ``query_idx[y.anchor_index]`` (both ``[E]``)
+                directly; see :class:`AnchorLabels` for the shape and the ``[E]``
+                vocabulary. Defaults to ``False`` (the backward-compatible ragged
+                dict).
         """
 
         # Set self._shutdowned right away, that way if we throw here, and __del__ is called,
         # then we can properly clean up and don't get extraneous error messages.
         self._shutdowned = True
+        self._use_list_output = use_list_output
```
mkolodner-sc: Can we add a warning here that this will be deprecated in a future release? Unless you see some benefit to us keeping this around.

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -775,48 +1116,25 @@ def _set_labels(
...
+        # (AnchorLabels.to_dict), expanded here when the caller wants the dict. Both
+        # forms feed an order-invariant contrastive loss, so the choice is purely
+        # about the consumer's preferred shape.
+        output_positive_labels, output_negative_labels = edge_list_set_labels(
+            node_local_to_global_by_type=node_type_to_local_node_to_global_node,
+            positive_labels_by_edge_type=positive_labels_by_label_edge_type,
+            negative_labels_by_edge_type=negative_labels_by_label_edge_type,
+            to_device=self.to_device,
         )
-        output_negative_labels: dict[EdgeType, dict[int, torch.Tensor]] = defaultdict(
```
mkolodner-sc, line 808: Glad we can remove a lot of this more complex logic now, thanks :)

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -775,48 +1116,25 @@ def _set_labels(
...
-                )  # shape [N, M], where N is the number of nodes and M is the number of negative labels for the current anchor node
-
-                # Gets the indexes of the items in local_node_to_global_node which match any of the negative labels for the current anchor node
-                output_negative_labels[
-                    label_edge_type_to_message_passing_edge_type(edge_type)
-                ][local_anchor_node_id] = torch.nonzero(negative_mask)[:, 0].to(
-                    self.to_device
-                )
-                # Shape [X], where X is the number of indexes in the original local_node_to_global_node which match a node in the negative labels for the current anchor node
+        if not self._use_list_output:
```
mkolodner-sc: For those still on the legacy path, do we have comparisons for how the code that previous path that has now been removed compares to this new "legacy-supported" path? I see we've validated that the results are the same, but could customers who don't migrate to the new approach expect slowdown?

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+
+def _membership_remap(
+    label_tensor: torch.Tensor,
+    sorted_node: torch.Tensor,
+    sort_perm: torch.Tensor,
+    to_device: torch.device,
+) -> tuple[torch.Tensor, torch.Tensor, int]:
+    """Resolve one padded label tensor to a flat ``(anchor, label)`` pair stream.
+
+    This is the actual global-id-to-local-index lookup; :class:`AnchorLabels` and
```
mkolodner-sc: To be honest, this docstring/example is a bit hard to read and wrap my head around. 

It'd be nice if we could clean this up a bit to make it a more readible/concise and easier to follow -- perhaps we can consolidate the detailed descriptions of what's happening with this new logic to one singular docstring at the top of some new "ablp_utilities" file where this might be better placed, and make each of these function-level docstrings can be a bit smaller and less duplicitous as a result.

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+    """
+    num_anchors = int(label_tensor.size(0))
+    num_nodes = int(sorted_node.size(0))
+    empty = torch.empty(0, dtype=torch.long, device=to_device)
+    if num_anchors == 0:
+        return empty, empty, num_anchors
+
+    num_labels = int(label_tensor.size(1))
+    flat = label_tensor.reshape(-1)  # [N_anchors * M] before the pad mask
+    # step 0 (see docstring example): tag each flattened entry with its anchor row.
```
mkolodner-sc: Similarly, in-line comments are useful, but these feel a bit too long and hard to follow for me (we have step 0, step 0 continued, precondition for step 1, step 1 ...)

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+        to_device (torch.device): Device for every output tensor.
+
+    Returns:
+        Tuple ``(y_positive, y_negative)``, each a
+        ``dict[message_passing_edge_type, AnchorLabels]`` with no entry for an
+        edge type that had no anchors this batch.
+    """
+    sorted_cache: dict[NodeType, tuple[torch.Tensor, torch.Tensor]] = {}
+
+    def _sorted_for(node_type: NodeType) -> tuple[torch.Tensor, torch.Tensor]:
```
mkolodner-sc: sorted_for is a bit ambiguous of a name, can we make this clearer? Ditto for the other functions here (i.e. "remap" is also vague.

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+        raise ValueError(
+            "vectorized label remap requires a unique node local->global map; "
+            "duplicate global ids break the searchsorted membership lookup."
+        )
+
+    # step 1 (see docstring example): position of each candidate id in sorted_node.
+    sorted_positions = torch.searchsorted(sorted_node, flat)  # [K]
+    # searchsorted returns N_nodes for an id larger than every entry, which would
+    # gather out of bounds at step 2; clamp it back into range.
+    sorted_positions = sorted_positions.clamp_(max=num_nodes - 1)
```
mkolodner-sc: In what scenarios do we expect this to happen?

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+    # via sort_perm (becomes AnchorLabels.label_index).
+    local_index = sort_perm[sorted_positions][is_exact_match]  # [E]
+    anchor_of_matched = anchor_of_entry[is_exact_match]  # [E]
+
+    # Result rows stay grouped by anchor (step 0 tagging was row-major and the masks
+    # preserve order); within-anchor order is unspecified -- the ABLP loss is
+    # order-invariant (see AnchorLabels).
+    return (
+        anchor_of_matched.to(to_device).to(torch.long),
+        local_index.to(to_device).to(torch.long),
```
mkolodner-sc: Do we need .to(to_device).to(torch.long)? Asking since we also call to_device in the training example loops

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -49,12 +50,298 @@
...
+        num_anchors,
+    )
+
+
+def edge_list_set_labels(
```
mkolodner-sc: nit: I think some of the new function and var names introduced in this PR are a bit non-intuitive for me. A few examples
- `edge_list_set_labels` might be easier to make sense of if we put the action first i.e. set_edge_list_labels
- node_local_to_global_by_type could be better and more aligned with other vars as local_id_to_global_id_by_node_type
- sorted_node and sorted_perm could have easier names to follow as well

mkolodner-sc (commented):
Thanks a lot for looking into this Kyle! This is a super exciting change, and looking forward to realizing the throughput and general complexity improvements from this work. 

Did a first pass here through the production logic and left some comments.

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: And we even get some perf gains!

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Removed, thanks for the feedback :)

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Good callout, done!

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Created new util file

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Done - though in practice I think we'll keep this around for a while

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: done.

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: It should be a speedup regardless, we're still vectorizing here.

The robots did some microbenchmark and found this was 42x faster even with the dict output.

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Updated a bit, with examples. WDYT/?

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Ditto.

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: Sure, updated, what d you think?

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: not needed, good call.

gigl/distributed/dist_ablp_neighborloader.py
kmontemayor2-sc: updated, wdyt?

examples/link_prediction/graph_store/heterogeneous_training.py
```diff
@@ -211,6 +211,7 @@ def _setup_dataloaders(
         channel_size=sampling_worker_shared_channel_size,
         process_start_gap_seconds=process_start_gap_seconds,
         shuffle=shuffle,
+        use_edge_index_output=True,
```
mkolodner-sc: nit: could we make this var a bit more clear? Reading this, there's no immediate association that we are referencing labeled edge index, edge index could be referring to message passing edges otherwise

Ditto elsewhere we use `edge_index_output`

mkolodner-sc (approved):
Thanks Kyle! Could we update the PR description as well which still references `AnchorLabels`? Otherwise this LGTM :)

yliu2-sc (approved):

examples/link_prediction/graph_store/heterogeneous_training.py
kmontemayor2-sc: done, thanks.

PR #682, opened 2026-06-29
WIP: Feature Quantization Scaffolding

jchmura-sc:
TODO
- unit / integration test updates
- get E2E numbers (compare with [scatter quantization pr](https://github.com/Snapchat/GiGL/pull/687)
- split into smaller PRs


gigl/common/data/dataloaders.py
jchmura-sc, line 396: Double check why we inject `tf.int64`

jchmura-sc: https://www.tensorflow.org/tutorials/load_data/tfrecord#data_types_for_tftrainexample

gigl/distributed/utils/neighborloader.py
```diff
@@ -324,6 +341,67 @@ def set_missing_features(
...
+        x = getattr(node_store, "x", None)
+        if x is None:
+            node_store.x = dequantized
+            return
+        if x.size(0) != dequantized.size(0):
+            raise ValueError(
+                "Cannot materialize quantized features with "
+                f"{dequantized.size(0)} rows into existing x with {x.size(0)} rows."
+            )
+        node_store.x = torch.cat([x, dequantized], dim=1)
```
jchmura-sc: Here is where we'd need to scatter write raw _and_ quantized node features if preserving original feature order is a requirement.

gigl/common/utils/feature_quantization/tensorflow_ops.py
```diff
@@ -0,0 +1,94 @@
...
+        pos = tf.cast(x > 0, tf.float32)
+        q.neg_mean = tf.math.divide_no_nan(tft.sum(x * neg), tft.sum(neg))
+        q.pos_mean = tf.math.divide_no_nan(tft.sum(x * pos), tft.sum(pos))
+        tf.debugging.check_numerics(q.neg_mean, "non-finite neg_mean.")
+        tf.debugging.check_numerics(q.pos_mean, "non-finite pos_mean.")
+        logger.info(
+            "Computed 1-bit quantization stats: "
+            f"neg_mean={q.neg_mean}, pos_mean={q.pos_mean}"
+        )
+    else:
```
jchmura-sc: Alternatively we can simplify proto model and compute all stats regardless of `bits`, but I suspect avoiding `tft.quantiles` in the 1-bit case is worth doing.

proto/snapchat/research/gbml/preprocessed_metadata.proto
```diff
@@ -3,6 +3,36 @@ syntax = "proto3";
...
+    float neg_mean = 1;
+    // Mean value for positive features, produced by packed bit/code 1.
+    float pos_mean = 2;
+  }
+
+  message FeatureQuantizationMetadata{
+    // Field in output TFRecords that stores packed uint8 features.
+    string quantized_feature_key = 1;
+    // Model-visible feature names produced after dequantizing and appending.
+    repeated string dequantized_feature_keys = 2;
```
jchmura-sc, line 24: Needed in order to preserve model schema semantics. If we change to scatter-semantics, this will no longer be necessary (since the logical feature schema would be unchanged regardless of quantization). Though, we would need a new bit mask field which tells us which indices of the original node feature tensor are quantized.

gigl/src/data_preprocessor/data_preprocessor.py
```diff
@@ -273,19 +278,42 @@ def __get_feature_dimension_for_single_data_reference(
                         feature_dimension += feature_shape[0]
             return feature_dimension
 
+        feature_outputs = preprocessing_spec.features_outputs
+        q_spec = (
+            preprocessing_spec.feature_quantization_spec
+            if isinstance(preprocessing_spec, NodeDataPreprocessingSpec)
+            else None
+        )
+        if q_spec is not None:
```
jchmura-sc: We have a couple options here
1. Override the schema as we have here. A bit messy but ensures that schema artifacts are logically consistent
2. Create separate artifact which is written for the trainer. Update the `preprocessed_metadata.schema_uri` to point to this new schema uri.
3. Ignore this and have the trainer just read from quantized schema in case a feature appears in both quantized and non-quantized schema.

On first glance, I prefer what we have here, ie. (1)

gigl/distributed/distributed_neighborloader.py
```diff
@@ -408,6 +409,7 @@ def _setup_for_graph_store(
                 is_homogeneous_with_labeled_edge_type=is_homogeneous_with_labeled_edge_type,
                 edge_types=edge_types,
                 node_feature_info=node_feature_info,
+                node_quantization_metadata=None,
```
jchmura-sc: Quantization support for Graph store mode can be a follow up

gigl/types/graph.py
```diff
@@ -107,6 +113,34 @@ class FeatureInfo:
     dtype: torch.dtype
 
 
+@dataclass
+class FeatureQuantizationMetadata:
```
jchmura-sc, line 117: can be made frozen, as in the scatter pr

gigl/distributed/base_sampler.py
```diff
@@ -322,6 +344,30 @@ async def _collate_fn(
                     for ntype, fut in nfeat_fut_dict.items():
                         nfeats = await wrap_torch_future(fut)
                         result_map[f"{as_str(ntype)}.nfeats"] = nfeats
+            if self.dist_node_quantized_feature is not None:
```
jchmura-sc, line 347: Why don't we fire off the `async_get` concurrently for _all_ feature stores simulatenously (when not using synchronous path)...

jchmura-sc: Also, it might be nice to add debug logs here that report how many bytes of transfer occurred for each feature (at the least, so that we can confirm the expected reduction when quantization is enabled)

PR #683, opened 2026-07-01
Fix Logger.__getattr__ infinite recursion

kmontemayor2-sc:
Logger(logging.LoggerAdapter) stores the wrapped logger as self.logger, but __getattr__ delegated to self._logger (nonexistent). Since __getattr__ only fires on failed lookups, resolving self._logger re-entered __getattr__("_logger") and recursed forever -> RecursionError.

This was dormant for the info/debug/... methods (defined directly on LoggerAdapter) but fired on any other attribute access, making Logger impossible to copy/deepcopy/pickle and turning undefined-attribute access into a crash instead of a clean AttributeError.

Read logger straight from __dict__ (cannot re-enter __getattr__) and raise AttributeError when it is not yet set. Add a regression test covering undefined-attr, delegation, deepcopy, and pickle.

**Scope of work done**


tests/unit/common/test_logger.py
```diff
@@ -0,0 +1,38 @@
...
+        logger = Logger(name="test_undefined_attr")
+        with self.assertRaises(AttributeError):
+            logger.this_attribute_does_not_exist
+
+    def test_delegates_to_wrapped_logger(self) -> None:
+        logger = Logger(name="test_delegation")
+        # ``level`` is defined on the wrapped logging.Logger, reached via __getattr__.
+        self.assertEqual(logger.level, logger.logger.level)
+
+    def test_deepcopy_round_trips(self) -> None:
```
svij-sc, line 24: why do we want to do this anyways? Just seems odd to want to pickle a logger.

svij-sc (approved):
Confused why we need this at all, but approving pre-emptively.

mkolodner-sc (approved):
stamp

tests/unit/common/test_logger.py
kmontemayor2-sc: during experimentation with ray the module gets picked and since we have `logger = Logger()` as a module-level variable we need to address this.

PR #684, opened 2026-07-02
Fail node id enumeration fast on NULL node ids

### LINKED ISSUE #288: Enumerator should not take `null` node_ids
cc @mkolodner-sc

Montana:
Hi Snapchat,

This Fixes #288.

The problem from what I can surmise is there's A NULL node id in the source node table survives the `SELECT DISTINCT` in `UNIQUE_NODE_ENUMERATION_QUERY`, gets silently assigned its own enumerated `int_id`, and pollutes the graph. The existing row-count assertion cannot catch this (a single NULL keeps input/output counts equal), and downstream edge enumeration joins on the original node id never match NULL, producing NULL enumerated src/dst ids in the enumerated edge tables.

**Fix.** Add a NULL guard directly in the enumeration query using BigQuery's documented `IF(x IS NULL, ERROR(...), x)` pattern, so the enumeration job fails immediately with a clear message naming the offending column and table. Doing this in SQL avoids the extra table scan a separate validation query would cost on billion-row node tables, and covers both the full-table and sharded read paths, which share this query.

**Test.** Adds `test_enumeration_fails_on_null_node_ids` to the enumerator integration test: uploads a node table containing a NULL node id and asserts `Enumerator.run` fails (via `SystemExit`) with the new error message. The message assertion also guards against a false pass from unrelated failures.

**Verification performed locally:**
- Rendered query parses as valid BigQuery SQL (sqlglot, `bigquery` dialect)
- `ruff check` and `ruff format --check` clean with the repo-pinned `ruff==0.15.10`
- CHANGELOG formatted with repo-pinned `mdformat==0.7.22` (`wrap=120`)

I don't have GCP credentials to run the BQ-backed integration test myself; would appreciate a maintainer kicking off the integration CI. Happy to adjust if you'd prefer filtering NULLs with a warning instead of failing the job.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Fail node id enumeration fast when the source node table contains NULL node ids, instead of silently enumerating NULL
+  as its own node (https://github.com/Snapchat/GiGL/issues/288)
+
 ## [0.3.1] - Jun 4, 2026
 
 ### Fixed
--- a/gigl/src/data_preprocessor/lib/enumerate/queries.py
+++ b/gigl/src/data_preprocessor/lib/enumerate/queries.py
@@ -3,8 +3,20 @@ DEFAULT_ENUMERATED_NODE_ID_FIELD = "int_id"
 
 UNIQUE_NODE_ENUMERATION_QUERY = """
 WITH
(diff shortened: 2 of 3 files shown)


RobbieBizzle:
really good catch @Montana, @snapchat should hire you by now as many hours you put in ahaha - Robbie

kmontemayor2-sc:
Thanks for the work here on the long-standing tech debt @Montana!

Apologies for the delays - we have been quite busy with OOO 😅 

We all really appreciate the work here but we need to do some checks internally for this change

PR #686, opened 2026-07-07
Remove dead generate-hashed-requirements workflow

kmontemayor2-sc:
Description:
  The `generate-hashed-requirements` workflow is a leftover from the pre-`uv` world and no longer functions.

  - It invokes `make generate_*_hashed_requirements` targets that **no longer exist** in the Makefile (the Makefile only `include`s `gigl/dep_vars.env`; none of those targets are defined). Dispatching it today fails with "No rule to make target".
  - It produces `requirements/*_requirements_unified.txt` files that **no longer exist** in the repo, and nothing else references them.
  - Dependency installation now uses `uv sync --locked` against `uv.lock` (see `requirements/install_py_deps.sh`), which is the source of truth for pinned deps.

  This PR deletes the workflow and removes the stale `Makefile` comment that pointed at the removed hashed-requirements targets and the `/make_cuda_hashed_req` comment command.

  No behavior change: the workflow was non-functional and unreferenced.


mkolodner-sc (approved):
Thanks Kyle!

svij-sc (approved):

PR #688, opened 2026-07-07
  Make example scripts self-consistent in naming and distributed behavior

kmontemayor2-sc:
## Summary

  A consistency pass over the 11 example train/infer scripts (plus `tests/api_test`), making every script say and do the same thing the same way, followed by adopting the new `GbmlConfigPbWrapper` typed feature maps (#643). Changes fall into six buckets:

  ### 1. Renames — one canonical name per concept

  | Concept | Canonical | Replaced |
  |---|---|---|
  | Processes per machine (config key / field / CLI) | `local_world_size` | `num_inference_processes_per_machine`, KDD `--process_count` |
  | Spawned process index | `local_rank` | `process_number` (KDD) |
  | Sampling workers (config key / field) | `sampling_workers_per_process` | `sampling_workers_per_inference_process` |
  | Model artifact field | `model_uri` | `model_state_dict_uri`, `saved_model_uri` / `saved_model_path` |
  | Output shard filename | `machine_{...}_local_process_{local_rank}` | `..._local_process_number_{...}` |

  Also fixed the `spillter_kwargs` → `splitter_kwargs` docstring typo. The KDD_2025 tutorial got renames only — its deliberately simple logic is untouched.

  ### 2. `local_world_size` resolution — one rule everywhere

  - **Colocated scripts** share one identical block: explicit config value wins → `torch.cuda.device_count()` when CUDA → else a per-file `DEFAULT_CPU_BASED_LOCAL_WORLD_SIZE` constant (4 for inference, 1 for training).
  - **Always raise `ValueError` on GPU oversubscription** — the homogeneous inference scripts previously only warned.
  - **Graph-store scripts** now take `local_world_size` directly from `cluster_info.num_processes_per_compute` (the launcher-set `COMPUTE_CLUSTER_LOCAL_WORLD_SIZE`) instead of a config arg — the old CPU default of 2 silently disagreed with the launcher's 1, desyncing spawned processes from the compute
  process group.
  - ⚠️ Behavior note: colocated LP training previously defaulted to 1 process even on GPU machines; it now defaults to GPU count, so e2e loss curves / step counts will shift on multi-GPU machines.

  ### 3. Distributed init — one canonical sequence

  Every colocated spawned process now does: get device → `cuda.set_device` → `init_process_group` (rank/world_size computed inline via the machine formula, the only place it appears) → `rank = torch.distributed.get_rank()` / `world_size = get_world_size()` locals used everywhere downstream. Lead-process
  guards are uniformly `if rank == 0:`. Graph-store heterogeneous inference now also reads rank from the process group after `init_compute_process` instead of a manual formula (its redundant `machine_rank`/`machine_world_size` fields are gone).

  ### 4. Cleanup — one canonical teardown

  - Colocated inference: `barrier → loader.shutdown() → gc.collect() → destroy_process_group()` (LP files were missing the destroy).
  - Graph-store: `loader.shutdown() → shutdown_compute_process() → gc.collect()` (GS homogeneous inference was missing `shutdown_compute_process`).
  - GS homogeneous inference's BigQuery load moved out of the spawned process into orchestration, guarded by `cluster_info.compute_node_rank == 0`, matching its hetero sibling.
  - LP heterogeneous inference orchestration gained the missing temporary-gloo-group destroy.
  - All five non-KDD inference orchestrations now call `mp.set_start_method("spawn")` like the training scripts always have.

  ### 5. Dead/deprecated API usage removed

  Dropped the deprecated `local_process_rank` / `local_process_world_size` loader kwargs from every example call site — they are silently ignored unless a (also-deprecated) `context` is passed, so they only taught users a dead parameter.

  ### 6. Adopt `GbmlConfigPbWrapper` typed feature maps (#643)

  All hand-built `{node,edge}_type_to_feature_dim` dict comprehensions and condensed-key lookups (8 scripts, 2 api_test files, `cora.ipynb`, `dblp.ipynb`) replaced with `gbml_config_pb_wrapper.node_type_to_feature_dim_map` / `.edge_type_to_feature_dim_map`.

  ### Investigated, not a bug

  `graph_store/heterogeneous_training.py` derives `query_node_type = supervision_edge_type.dst_node_type` — reversed vs. standard mode. This is deliberate compensation for the reversed storage-side `splitter_kwargs.supervision_edge_types` in the GS DBLP config (traced through
  `DistNodeAnchorLinkSplitter`'s direction handling), and is now documented with comments instead of "fixed".

  ### Verification

  - `make type_check` and `make format_py` clean; repo-wide greps show zero stale occurrences of any old name.
  - `make unit_test_py PY_TEST_FILES="config_validator_test.py"` (imports the example configs): 4/4 pass.
  - Dual final review — an adversarial whole-branch review (MERGE-READY) and a Codex review; every confirmed finding was fixed on-branch.

  ### Known follow-ups (out of scope)

  - Optional hardening: guard `mp.set_start_method` against pre-initialized contexts (equally applies to the pre-existing training pattern).
  - Library-level: finish deprecating/removing the loader `local_process_rank`/`local_process_world_size` params.
  - Pre-existing on main: a `dblp.ipynb` cell references `node_feature_dims[node_type]` where `node_type` may be undefined.


examples/link_prediction/graph_store/heterogeneous_training.py
```diff
@@ -179,6 +179,21 @@ def _setup_dataloaders(
     """
     rank = torch.distributed.get_rank()
 
+    # We anchor on the *destination* of the supervision edge here, which is the reverse of
```
mkolodner-sc: There's no code change here, but I wonder if there was a bug here at some point which has been made more clear thanks to this comment. My understanding in GiGL was that the supervision edge type would always be specified in the outward. In our task metadata we have:

```
taskMetadata:
  nodeAnchorBasedLinkPredictionTaskMetadata:
    # We aim to predict paper -> author links in the graph.
    supervisionEdgeTypes:
    - dstNodeType: author
      relation: to
      srcNodeType: paper
```

mkolodner-sc (commented):
Thanks Kyle! Generally LGTM, just one comment about a comment that might have exposed a bug with our example loops?

examples/link_prediction/graph_store/heterogeneous_training.py
kmontemayor2-sc: oops you're correct! We should fix this as a separate change (the root cause is that `get_ablp_input` is not aware of the edge dir. https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/graph_store/dist_server.py#L542C9-L542C23

Let's get that change in separately? And have this comment reflect the way the world *should* be?


Thanks for the catch, Matt!

mkolodner-sc: Ya makes sense to do separately! can we remove this comment for now then or add a TODO to fix in the follow-up?

examples/link_prediction/graph_store/heterogeneous_inference.py
```diff
@@ -396,6 +392,8 @@ def _run_example_inference(
     # - the total number of machines (world size)
 
     program_start_time = time.time()
+    mp.set_start_method("spawn")
```
mkolodner-sc, line 395: qq -- were we not doing this before which was leading to some issues? Or why did we add this in specifically here

examples/link_prediction/heterogeneous_inference.py
```diff
@@ -48,6 +48,10 @@
 
 logger = Logger()
 
+# Default number of inference processes per machine when one isn't provided via
+# `local_world_size` in inferencer args and there are no GPUs available.
+DEFAULT_CPU_BASED_LOCAL_WORLD_SIZE = 4
```
mkolodner-sc, line 53: Is it confusing that we sometimes set this to 1 but sometimes to 4? Should we be more consistent here?

mkolodner-sc (approved):
Thanks Kyle! LGTM provided comments are addressed and tests pass

svij-sc (approved):
stamping - leaning on @mkolodner-sc for review

examples/link_prediction/graph_store/heterogeneous_inference.py
kmontemayor2-sc: Adding this is in to be consistent. IIRC we saw some issues  (or perf degredation?) when we *didn't* set this before right?

examples/link_prediction/heterogeneous_inference.py
kmontemayor2-sc: We have 4 for inference since it's more parallizeable, but since we use DDP for training (and we end up sharing the same cores anyways) there may not be benefit for more parallel cpu training.

PR #690, opened 2026-07-08
Concurrently Issue All `dist_feature.async_get()`Requests in Base Sampler `_collate_fn()`

### LINKED ISSUE #689: Concurrently Issue All `dist_feature.async_get()`Requests in Base Sampler `_collate_fn()`
**TLDR**
In our samplers, we should consider firing off all `async_get()` requests to all distributed feature stores in parallel, and awaiting them all at once when building the `SampleMessage`. Currently, we await them serially, potentially limiting concurrency.

**Description**
In the [base sampler _collate_fn](https://github.com/Snapchat/GiGL/blob/fea2af34860245469abffd76cde9423f1b24d3f4/gigl/distributed/base_sampler.py#L223), we issue `async_get()` calls on each distributed feature store to retrieve the tensors required for constructing a `SampleMessage`. There is a synchronous version (`self.use_all2all=True`) which we don't use so is not relevant here.

For each feature store (e.g. node features, edge features, node labels, ...) we essentially do something like: 

```python
# In base_sampler._collate_fn()
for feat in self.distributed_features: # e.g. node_feats, edge_feats, ...
    fut_dict = {}
    for k_type, k in output.items(): # looping over node types / edge types
         fut_dict[k_type] = feat.async_get(k, k_type)
    for k_type, fut in fut_dict.items():
        result[...] = await wrap_torch_tensor(fut)
```

I suspect that we might be able to get slightly better feature fetch throughput from more concurrency by issuing all `async_get()` requests in parallel, across feature stores. Conceptually:

```python
# In base_sampler._collate_fn()
fut_dict = {}
for feat in self.distributed_features: # e.g. node_feats, edge_feats, ...
    for k_type, k in output.items(): # looping over node types / edge types
         fut_dict[(feat_type, k_type)] = feat.async_get(k, k_type)

for (feat_type, k_type), fut in fut_dict.items():
    result[...] = await wrap_torch_tensor(fut)
```

**Scope**
Change is minimal. Testing on a workload with multiple large feature stores (i.e. edge features) would help. [Quantization work](https://github.com/Snapchat/GiGL/pull/682) might need this to [hide overhead of managing a new quantized node feature store](https://github.com/Snapchat/GiGL/blob/66ae0eee93b428cf4a7c1b351214caf4b065c299/gigl/distributed/base_sampler.py#L360-L370)

jchmura-sc:
#### Base Sampler Collate Time (s)
| Experiment | RPC | Mean (s) |  Median (s) |
| :--- | :--- | :--- | :--- |
| No Edge Feats (Colocated) | Baseline | 1.33  | 0.77 |
| | **Parallel** | **1.30** *(-2.0%)* | **0.77** *(-0.0%)* | 
| No Edge Feats (GraphStore) | Baseline | 6.33  | 5.81 | 
| | **Parallel** | **4.26** *(-32.6%)* |  **3.27** *(-43.6%)* |
| Edge Feats(Colocated) | Baseline | 1.68 | 1.67 | 
| | **Parallel** | **1.54** *(-7.6%)* | **1.56** *(-6.4%)* | 

> **Note:** Summary statistics computed over `mean_collate_s`, which averaged across **500 collate calls**.

Close #689


gigl/distributed/base_sampler.py
```diff
@@ -281,23 +282,18 @@ async def _collate_fn(
                         )
             input_type = output.input_type
             assert input_type is not None
+            futs = {}
+            label_keys = set()
```
kmontemayor2-sc: nit. can we type these? e.g. `futs: dict[str, torch.futures.Future[Tensor] = {}`

gigl/distributed/base_sampler.py
```diff
@@ -360,33 +360,39 @@ async def _collate_fn(
                 )
             if self.with_edge:
                 result_map["eids"] = output.edge
+            futs = {}
+            label_keys = set()
```
kmontemayor2-sc: why do two sets of rpc send / gather here?

would it be easiest if we only sent out one batch of RPCs?

kmontemayor2-sc (commented):
Thanks for the work Jacob! I would appreciate if we can test this at scale, is it possible to do so and follow up offline?

gigl/distributed/base_sampler.py
jchmura-sc: Agreed, the heterogenous / homogenous are independent but we can simplify this!

gigl/distributed/base_sampler.py
```diff
@@ -333,16 +329,13 @@ async def _collate_fn(
                         eids = result_map.get(f"{as_str(etype)}.eids", None)
                     if eids is not None:
                         eids = eids.to(torch.long)
-                        efeat_fut_dict[etype] = self.dist_edge_feature.async_get(
-                            eids, etype
+                        result_key = (
+                            f"{as_str(etype)}.efeats"
+                            if self.edge_dir == "out"
+                            else f"{as_str(reverse_edge_type(etype))}.efeats"
```
kmontemayor2-sc: nit. I prefer theis style

```
if ..
   val = foo
else:
   val = bar
```

I find the ternaries a bit harder to read personally.

gigl/distributed/base_sampler.py
```diff
@@ -412,34 +407,51 @@ async def _collate_fn(
...
+                # (e.g. CrossEntropyLoss) expects for data.y. Multi-label
+                # (K>1) keeps the full 2-D matrix.
+                # TODO (mkolodner-sc): Consider investigating always returning
+                # 2-D — this may be a breaking change for single-label
+                # training pipelines (e.g. CrossEntropyLoss expects 1-D data.y).
+                value = value if value.shape[1] > 1 else value.T[0]
+            result_map[result_key] = value
+
+        collate_time = time.perf_counter() - collate_start
+        tensor_sizes = {}
```
kmontemayor2-sc: nit. add type?

gigl/distributed/base_sampler.py
```diff
@@ -412,34 +407,51 @@ async def _collate_fn(
...
+        total_bytes = 0
+        for k in futs:
+            nbytes = result_map[k].nbytes
+            total_bytes += nbytes
+            tensor_sizes[k] = f"{nbytes / 1e6:.2f}MB"
+
+        logger.debug(
+            f"Collected remote tensors: {tensor_sizes} | Total: {total_bytes / 1e6:.2f}MB | Time: {collate_time:.3f}s"
+        )
```
kmontemayor2-sc: Hmmm,  I have soem minor concern about adding this loop in the hot path, but I supposed it's probably not too critical? Do you think we'll need these metrics in the future?

jchmura-sc: I think that relative to the rest of the function, the time spent here should be negligible (collate_time on the order of 1s in colocated, 4s graph store).

But you are right that this is directly in the hot path. I will be adding several such debug logs when I profile the quantization branch - we can make a call on which to keep over there.

I did appreciate the ability to see how large the features are in each `SampleMessage`, but I will remove it from this PR.

jchmura-sc: Also, since we only export [INFO logs to GCP](https://github.com/Snapchat/GiGL/blob/928c60389dffc04c88145ef95798972d3f86b9b6/gigl/common/logger.py#L45), this would not be readily available anyway.

kmontemayor2-sc: > collate_time on the order of 1s in colocated, 4s graph store

Wait really? That seems much higher than I'd expect, I wonder if there are other optimizations we can take there.

anyways thanks for removing, I suppose it's not the end of the world if this fn is already so slow.

kmontemayor2-sc (approved):

yliu2-sc (approved):

PR #692, opened 2026-07-08
Make graph-store ABLP supervision-edge direction edge_dir-aware

kmontemayor2-sc:
Callers specify supervision edge types in outward form (anchor -> supervision), matching taskMetadata. The colocated ABLP path already reverses these internally for edge_dir="in" to match how label edges are stored; the graph-store path did not, so the DBLP example compensated with a pre-reversed config and query = dst_node_type. That double reversal canceled out but trained the mirror of the declared task (author anchors -> paper candidates instead of paper -> author).

Make the graph-store path edge_dir-aware to match colocated mode:
- DistServer.get_ablp_input reverses the supervision edge for edge_dir="in" before resolving stored label edge types.
- DistABLPLoader._setup_for_graph_store asserts the outward convention and applies the same reversal, consistently across label-type derivation and the per-server loop.
- Correct the GS DBLP config to outward ("paper","to","author").
- Switch the GS training example to query = src / labeled = dst.
- Add an edge_dir="in" regression test for get_ablp_input.

Behavior note: GS DBLP training now anchors on papers (the declared task), so e2e loss curves will shift.

**Scope of work done**


mkolodner-sc (approved):
Thanks Kyle for the quick fix!

svij-sc (approved):

PR #693, opened 2026-07-08
Surface swallowed sampler-coroutine exceptions + warn on partial feature coverage

kmontemayor2-sc:
## What & why

  GLT's `ConcurrentEventLoop` catches every exception raised inside a sampling
  coroutine and only logs `coroutine task failed: <str(e)>` — no traceback, no
  message sent to the channel. When that happens the loader blocks forever on
  `channel.recv()`, so a single sampler failure turns into a silent, indefinite
  hang.

  This PR makes those failures loud and immediate. It does **not** fix the
  underlying feature-coverage bug that triggered the incident (that's a
  separate PR) — it's the general guardrail that converts *any* swallowed sampler
  exception into a fast, actionable error.

  ## Changes

  - **`gigl/distributed/utils/sampling_errors.py`** (new): shared wire codec —
    reserved key `#SAMPLING_ERROR`, `encode_sampling_error()` (traceback → uint8
    tensor), `raise_if_sampling_error()` (decodes and raises `RuntimeError`). One
    source of truth so producer and consumer can't drift.
  - **`base_sampler.py`**: `BaseDistNeighborSampler` gains an `__init__` that
    initializes `_sampling_error_sent`; `_send_adapter` is wrapped in try/except —
    on failure it logs the full traceback via `logger.exception` and forwards a
    one-time poison-pill `SampleMessage` in channel mode (re-raises in
    channel-less mode). One pill per sampler instance, so a failure storm can't
    block a bounded channel.
  - **`distributed_neighborloader.py` / `dist_ablp_neighborloader.py`**:
    `raise_if_sampling_error(msg)` at the top of each `_collate_fn`, so the
    training process raises on the next `next(loader)` with the worker's real
    traceback.
  - **`base_dist_loader.py`**: `_warn_on_partial_feature_coverage` (called from
    `__init__`) — non-fatal warning when a heterogeneous dataset has features on
    only a subset of its node/edge types. Warning, never a raise: reachability of
    a featureless type is a runtime property that can't be proven at construction.

  ## Testing

  - `sampling_errors_test.py` (new): 5/5 — codec round-trip incl. non-ASCII and
    empty-string, writable-tensor, no-op-when-absent.
  - `remote_channel_test.py`: 6/6 — poison pill survives the graph-store receive
    path and raises loader-side.
  - `distributed_neighborloader_test.py`: 25/25 (1 pre-existing GPU skip) — incl.
    two loader-level tests under bounded `proc.join(timeout=...)`.
  - `make type_check`: clean.
  - Confirmed end-to-end that a reachable featureless edge type now raises
    `RuntimeError` ("A sampling worker failed... KeyError: ...") in ~30s instead
    of hanging.

### DIFF
--- a/tests/unit/distributed/graph_store/remote_channel_test.py
+++ b/tests/unit/distributed/graph_store/remote_channel_test.py
@@ -166,7 +166,7 @@ class RemoteReceivingChannelTest(TestCase):
         channel.reset()
 
         pill = {SAMPLING_ERROR_KEY: encode_sampling_error("boom")}
-        # Responses are mocked returned messages from the server 
+        # Responses are mocked returned messages from the server
         responses = iter([(pill, False), (None, True)])
 
         def _mock_async_request_server(server_rank, func, channel_id):
--- a/tests/unit/distributed/graph_store/remote_channel_test.py
+++ b/tests/unit/distributed/graph_store/remote_channel_test.py
@@ -166,6 +166,7 @@ class RemoteReceivingChannelTest(TestCase):
         channel.reset()
 
         pill = {SAMPLING_ERROR_KEY: encode_sampling_error("boom")}
+        # Responses are mocked returned messages from the server 
         responses = iter([(pill, False), (None, True)])
 
         def _mock_async_request_server(server_rank, func, channel_id):
--- a/tests/unit/distributed/graph_store/remote_channel_test.py
+++ b/tests/unit/distributed/graph_store/remote_channel_test.py
@@ -3,6 +3,11 @@ from unittest.mock import patch
 import torch
 
 from gigl.distributed.graph_store.remote_channel import RemoteReceivingChannel
+from gigl.distributed.utils.sampling_errors import (
+    SAMPLING_ERROR_KEY,
+    encode_sampling_error,
+    raise_if_sampling_error,
+)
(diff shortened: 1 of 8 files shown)


tests/unit/distributed/graph_store/remote_channel_test.py
```diff
@@ -144,3 +149,34 @@ def _mock_async_request_server(server_rank, func, channel_id):
...
+        channel = RemoteReceivingChannel(
+            server_rank=[0],
+            channel_id=[10],
+            prefetch_size=1,
+            active_mask=[True],
+        )
+        channel.reset()
+
+        pill = {SAMPLING_ERROR_KEY: encode_sampling_error("boom")}
+        responses = iter([(pill, False), (None, True)])
```
mkolodner-sc, line 170: What does this iter map to? Might be good to clarify with a comment

mkolodner-sc (approved):
Thanks Kyle! LGTM with one nit comment

tests/unit/distributed/graph_store/remote_channel_test.py
kmontemayor2-sc: done, thanks

yliu2-sc (approved):

PR #694, opened 2026-07-09
Derive with_edge from edge feature presence (perf)

kmontemayor2-sc:
## What

  `BaseDistLoader.create_sampling_config` hardcoded `with_edge=True` for every
  loader. This derives it from whether the dataset actually has edge features:

      with_edge = dataset_schema.edge_feature_info is not None

  `DatasetSchema.edge_feature_info` is already populated in both colocated and
  graph-store modes (fetched from storage rank 0 in graph-store mode), so the
  derivation is correct on compute nodes as well.

  ## Why

  For datasets with no edge features anywhere, `with_edge=True` forces GLT to
  collect sampled edge ids per hop and transport them in every `SampleMessage`.
  Those eids are pure overhead when there are no edge features to gather: the
  sampler writes an `int64` eid tensor of one element per sampled edge (gated on
  `self.with_edge`). With `with_edge=False`, GLT takes the `sample` path instead
  of `sample_with_edge`, `output.edge` is `None`, no eids are collected or
  serialized, and edge-feature-store construction is skipped.

  ## Behavior change (changelog)

  Batches produced from featureless datasets no longer carry GLT's `data.edge`
  (homogeneous) or `data[etype].edge` (heterogeneous) sampled-edge-id attribute.
  A repo-wide grep found no consumer that reads this attribute off a produced
  batch — the only in-repo references are the sampler's own write sites. External
  code that reads `batch.edge` on a featureless dataset would now find it
  absent/`None`. Datasets that do have edge features are unaffected.

  ## Tests

  - Direct unit tests for `create_sampling_config`: `with_edge` is `False` when
    `edge_feature_info` is `None`, `True` for homogeneous and heterogeneous edge
    features.
  - End-to-end: a featureless homogeneous dataset yields batches with no sampled
    edge ids (`data.edge is None`).
  - Weighted sampling without edge features still avoids zero-weight edges.
  - PPR unaffected (outputs come from metadata, not sampled eids).
  - Graph-store provenance: compute-side `edge_feature_info` is populated from the
    storage cluster.

  All affected files pass; `make type_check` passes.

### DIFF
--- a/tests/unit/distributed/distributed_neighborloader_test.py
+++ b/tests/unit/distributed/distributed_neighborloader_test.py
@@ -890,7 +890,11 @@ class DistributedNeighborLoaderTest(TestCase):
 
 
 class WithEdgeDerivationTest(TestCase):
-    """Regression coverage for deriving ``with_edge`` from edge-feature presence."""
+    """Covers ``create_sampling_config`` deriving ``with_edge`` from edge-feature presence.
+
+    Exercises both paths: ``with_edge=True`` when the dataset has edge features
+    (homogeneous or per-edge-type), and ``with_edge=False`` when it has none.
+    """
 
     def test_config_with_edge_false_when_no_edge_features(self) -> None:
         schema = DatasetSchema(
--- a/tests/unit/distributed/distributed_neighborloader_test.py
+++ b/tests/unit/distributed/distributed_neighborloader_test.py
@@ -8,10 +8,12 @@ from graphlearn_torch.distributed import shutdown_rpc
 from parameterized import param, parameterized
 from torch_geometric.data import Data, HeteroData
 
+from gigl.distributed.base_dist_loader import BaseDistLoader
 from gigl.distributed.dataset_factory import build_dataset
 from gigl.distributed.dist_dataset import DistDataset
 from gigl.distributed.distributed_neighborloader import DistNeighborLoader
 from gigl.distributed.utils import get_free_port
+from gigl.distributed.utils.neighborloader import DatasetSchema
 from gigl.distributed.utils.serialized_graph_metadata_translator import (
     convert_pb_to_serialized_graph_metadata,
 )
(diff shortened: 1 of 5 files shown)


tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -861,5 +889,82 @@ def test_distributed_neighbor_loader_invalid_inputs_colocated(
             DistNeighborLoader(**kwargs)
 
 
+class WithEdgeDerivationTest(TestCase):
+    """Regression coverage for deriving ``with_edge`` from edge-feature presence."""
```
mkolodner-sc: nit: I'd prefer us not to track this as regression coverage -- otherwise our codebase may become flooded with regression coverages as we add more changes (starts asking question of regression from what?). This docstring should focus more on coverage of both with_edge paths instead (which this test does).

tests/unit/distributed/graph_store/remote_dist_dataset_test.py
```diff
@@ -200,6 +200,36 @@ def test_graph_metadata_getters_homogeneous(self, mock_request):
...
         self.assertIsNone(remote_dataset.fetch_node_types())
 
+    def test_fetch_edge_feature_info_populated_from_server(self, mock_request):
+        """Edge feature info is fetched from server rank 0 for the compute-side schema.
+
+        The ``with_edge`` sampling flag is derived on the compute side as
+        ``dataset_schema.edge_feature_info is not None`` (see
+        ``BaseDistLoader.create_sampling_config``). This test confirms the field
+        is populated from the storage cluster when the graph has edge features,
+        so that derivation yields ``with_edge=True`` off-node.
```
mkolodner-sc, line 210: Since we've added a test here for with_edge=True, do you think it's worth also adding one for with_edge=false for the remote dataset tests?

mkolodner-sc (approved):
Thanks Kyle for this optimization! Generally LGTM with some minor comments

tests/unit/distributed/distributed_neighborloader_test.py
kmontemayor2-sc: sg updated!

tests/unit/distributed/graph_store/remote_dist_dataset_test.py
kmontemayor2-sc: done!

yliu2-sc (approved):

PR #695, opened 2026-07-09
Use metadata-free GCS bucket references

kmontemayor2-sc:
## Summary

  Replace object-only `storage.Client.get_bucket(...)` calls in `GcsUtils` with `storage.Client.bucket(...)`.

  These paths only construct blob references, list object prefixes, or delete objects. They do not read bucket metadata, so `Client.bucket()` is
  behavior-preserving and avoids requiring `storage.buckets.get` for callers that only have object-scoped IAM roles such as `roles/
  storage.objectUser`.

  ## Changes

  - Updated `does_gcs_file_exist`, `delete_gcs_file_if_exist`, `count_blobs_in_gcs_path`, `delete_files`, and `_delete_files_in_bucket_dir` to
  use metadata-free bucket references.
  - Added unit coverage that fails if `get_bucket()` is used for these object-only helpers.
  - Added integration coverage for real GCS object existence, counting, single-object deletion, batch deletion, and directory deletion.

  ## Tests

  - `make unit_test_py PY_TEST_FILES="gcs_test.py"`: passed
  - `make integration_test PY_TEST_FILES="gcs_test.py"`: passed
  - `make type_check`: passed
  - `uv run ruff check --config pyproject.toml gigl/common/utils/gcs.py tests/unit/utils/gcs_test.py tests/integration/common/gcs_test.py`:
  passed
  - `uv run ruff format --check --config pyproject.toml gigl/common/utils/gcs.py tests/unit/utils/gcs_test.py tests/integration/common/
  gcs_test.py`: passed

  Note: full `make check_format` passed Python/C++/Scala phases but failed on unrelated pre-existing Markdown formatting issues outside this
  change.


mkolodner-sc (approved):
Thanks Kyle!

tests/unit/utils/gcs_test.py
```diff
@@ -54,6 +68,76 @@ def test_init_succeeds_when_client_has_no_project(self):
...
+        gcs_path = GcsUri("gs://test-bucket/test-path/test-file.txt")
+
+        gcs_utils.delete_files(gcs_files=[gcs_path])
+
+        mock_client.bucket.assert_called_once_with("test-bucket")
+        mock_client.get_bucket.assert_not_called()
+        mock_bucket.blob.assert_called_once_with("test-path/test-file.txt")
+        mock_client.batch.assert_called_once_with()
+        mock_blob.delete.assert_called_once_with()
+
```
svij-sc: It really feels like we are over testing here :\ 
With testing internal calls, et al.
If we really care about this behaviour - prolly one test is sufficient ?

I am worried a lot at reducing the ability for humans to read the test code here, especially now when we should be more carefully reviewing tests when most code is derived from bots.

svij-sc (approved):

tests/unit/utils/gcs_test.py
mkolodner-sc: This is a fair callout -- perhaps we can rely on some of the functionality of the tests that already exist since this is more of a small refactor than a new knob here.

kmontemayor2-sc: good callout, done :)

PR #697, opened 2026-07-10
Add typed channel quotas to distributed PPR sampler

mkolodner-sc:
## Summary

Adds generic typed/channel-aware PPR support to `DistPPRNeighborSampler`, stacked on top of the residual top-up extraction API from #699.

A typed PPR channel can be either a single canonical edge type or a tuple of canonical edge types. The sampler runs one PPR state per channel with channel-specific traversal restrictions and quotas, then merges the channel outputs into one final PPR sequence per anchor.

Because this PR is stacked on #699, the merge can also use residual top-up candidates from completed PPR states. That lets the sampler fill remaining sequence slots with already-discovered residual candidates without issuing extra neighbor fetches.

## Changes

- Adds `PPRSamplerOptions.typed_channel_quotas`.
- Supports channel keys as:
  - one canonical edge type, e.g. `(src_type, relation, dst_type)`
  - a tuple of canonical edge types, for grouped traversal channels
- Validates typed PPR is used only for heterogeneous sampling.
- Builds per-channel traversal allowlists over edge types.
- Runs channel-specific PPR states and merges outputs with per-channel quotas.
- Deduplicates nodes across channels.
- Uses residual top-up candidates from the same PPR states when the quota merge emits fewer than `max_ppr_nodes`.
- Scores residual top-up candidates on the same scale as finalized PPR candidates via `ppr_score + residual`.
- Emits multi-column PPR edge attributes:
  - best calibrated score
  - per-channel calibrated scores
  - per-channel presence bits
- Shares frontier fetches across channels by unioning nodes per edge type before issuing one-hop fetches.
- Adds unit tests for channel validation, merge behavior, residual top-up behavior, metadata shape, and shared fetch behavior.

### DIFF
--- a/gigl/distributed/dist_ppr_sampler.py
+++ b/gigl/distributed/dist_ppr_sampler.py
@@ -1,6 +1,7 @@
 import asyncio
+import math
 from collections import defaultdict
-from typing import Optional, Union
+from typing import Iterator, Optional, Sequence, Union, cast
 
 import torch
 
@@ -9,7 +10,6 @@ import torch
 from gigl_core import PPRForwardPush
 from graphlearn_torch.sampler import (
     HeteroSamplerOutput,
-    NeighborOutput,
     NodeSamplerInput,
     SamplerOutput,
 )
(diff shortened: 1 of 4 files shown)


mkolodner-sc:
Closing in favor of #701, which uses the renamed mkolodner/typed-channel-ppr-sampler branch.

PR #699, opened 2026-07-10
Add residual top-up extraction to PPR forward push

mkolodner-sc:
## Summary

Adds an opt-in residual top-up extraction API to `PPRForwardPush`.

The existing `extractTopK` behavior is unchanged. The new API returns the normal top-k finalized PPR nodes first, then appends additional candidates from residual mass that was discovered during forward push but did not necessarily cross the requeue threshold. This gives downstream samplers a way to fill more sequence slots without issuing extra neighbor fetches.

## Changes

- Adds `extractTopKWithResidualTopUp(maxPprNodes, maxResidualNodes)` in C++.
- Exposes the new method through pybind and the Python stub.
- Scores residual top-up candidates on the same scale as finalized candidates via `ppr_score + residual`.
- Improves neighbor cache behavior so previously fetched `(node, edge type)` neighbor lists are reused within a PPR state.
- Adds C++ tests for neighbor cache reuse and residual top-up extraction.

### DIFF
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -146,11 +146,11 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
     return result;
 }
 
-void PPRForwardPush::pushResiduals(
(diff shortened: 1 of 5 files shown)


mkolodner-sc:
Closing in favor of #702, which uses the renamed mkolodner/ppr-residual-topup branch.

PR #702, opened 2026-07-10
Add residual top-up extraction to PPR forward push

mkolodner-sc:
## Summary

# Reasoning for Change

Previously, the PPR forward push would return all neighbors re-enqueued during forward push. This is how the base PPR forward push algorithm works, but results in small sequences volumes that could be more meaningful if they were larger. Decreasing epsilon previously provided a knob to get these larger sequence lengths, but came at heavy throughput costs. Instead of relying on epsilon for this behavior, this change exposes a way to allow us to "top-off" the existing PPR sequence with nodes that were found during forward-push but didn't meet the threshold for re-enqueue. The logic here is that these are the nodes that would have been re-enqueued next by decreasing epsilon, so this gives us a clean way to get these nodes and their ppr scores without having to pay heavy throughput costs.

On internal benchmarking, we have seen sequences toped off this way perform as good as the sequences with lower epsilon at a much faster runtime.

# Changes Summary

Adds an configurable but opt-in residual top-up extraction API to `PPRForwardPush`.

The existing `extractTopK` behavior is unchanged. The new API returns the normal top-k finalized PPR nodes first, then appends additional candidates from residual mass that was discovered during forward push but did not necessarily cross the requeue threshold. This gives downstream samplers a way to fill more sequence slots without issuing extra neighbor fetches.

## Changes

- Adds `extractTopKWithResidualTopUp(maxPprNodes, maxResidualNodes)` in C++.
- Exposes the new method through pybind and the Python stub.
- Scores residual top-up candidates on the same scale as finalized candidates via `ppr_score + residual`.

## Validation

- Ran focused ruff check on the updated Python stub.
- Full `uv run` validation in the temp worktree was blocked by local environment setup because `gigl-core` attempted to rebuild and could not find Torch CMake config.

### DIFF
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -287,7 +287,7 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
             const int32_t topK = std::min(maxPPRNodes, static_cast<int32_t>(scores.size()));
             const int32_t residualTopUpBudget = enableResidualTopUp ? maxPPRNodes - topK : 0;
             std::vector<std::pair<int32_t, double>> selectedPairs;
-            selectedPairs.reserve(static_cast<size_t>(topK + residualTopUpBudget));
+            selectedPairs.reserve(static_cast<size_t>(topK) + static_cast<size_t>(residualTopUpBudget));
             std::unordered_set<int32_t> selectedPPRNodeIds;
             if (topK > 0) {
                 if (residualTopUpBudget > 0) {
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -146,8 +146,7 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
     return result;
 }
 
-void PPRForwardPush::pushResiduals(
-    const std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>>& fetchedByEtypeId) {
+void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
     // Step 1: Persist fetched neighbor lists in the per-state cache. drainQueue()
     // consults this cache before requesting future lookups, so storing every
     // fetched row here avoids re-fetching a (node, edge type) pair if it re-enters
@@ -265,9 +264,15 @@ void PPRForwardPush::pushResiduals(
     }
 }
 
-std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPush::extractTopK(
-    int32_t maxPprNodes) {
-    std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> result;
+PPRExtractResult PPRForwardPush::extractTopK(int32_t maxPprNodes) {
+    return extractTopKWithResidualTopUp(maxPprNodes, /*maxResidualNodes=*/0);
+}
+
+PPRExtractResult PPRForwardPush::extractTopKWithResidualTopUp(int32_t maxPprNodes, int32_t maxResidualNodes) {
+    TORCH_CHECK(maxPprNodes >= 0, "maxPprNodes must be non-negative, got ", maxPprNodes, ".");
+    TORCH_CHECK(maxResidualNodes >= 0, "maxResidualNodes must be non-negative, got ", maxResidualNodes, ".");
+
+    PPRExtractResult result;
     // Emit an entry for every node type, even if unreachable in this batch (empty tensors,
     // all-zero valid_counts).  This keeps the output shape consistent across batches so
     // downstream model architectures see a fixed set of PPR edge types every iteration.
@@ -277,9 +282,13 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
         std::vector<int64_t> validCounts;
 
         for (int32_t seedIdx = 0; seedIdx < _batchSize; ++seedIdx) {
-            const auto& scores = _state[seedIdx][nodeTypeId].pprScores;
+            const auto& nodeTypeState = _state[seedIdx][nodeTypeId];
+            const auto& scores = nodeTypeState.pprScores;
             int32_t topK = std::min(maxPprNodes, static_cast<int32_t>(scores.size()));
+            int32_t residualTopK = 0;
+            std::unordered_set<int32_t> selectedNodeIds;
             if (topK > 0) {
+                selectedNodeIds.reserve(static_cast<size_t>(topK));
                 std::vector<std::pair<int32_t, double>> scorePairs(scores.begin(), scores.end());
                 std::partial_sort(scorePairs.begin(),
                                   scorePairs.begin() + topK,
@@ -287,11 +296,44 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
                                   [](const auto& a, const auto& b) { return a.second > b.second; });
 
                 for (int32_t rankIdx = 0; rankIdx < topK; ++rankIdx) {
-                    flatIds.push_back(static_cast<int64_t>(scorePairs[rankIdx].first));
+                    int32_t nodeId = scorePairs[rankIdx].first;
+                    flatIds.push_back(static_cast<int64_t>(nodeId));
                     flatWeights.push_back(scorePairs[rankIdx].second);
+                    selectedNodeIds.insert(nodeId);
                 }
             }
-            validCounts.push_back(static_cast<int64_t>(topK));
+
+            if (maxResidualNodes > 0) {
+                std::vector<std::pair<int32_t, double>> residualPairs;
+                residualPairs.reserve(nodeTypeState.residuals.size());
+                for (const auto& [nodeId, residual] : nodeTypeState.residuals) {
+                    if (residual <= 0.0 || selectedNodeIds.find(nodeId) != selectedNodeIds.end()) {
+                        continue;
+                    }
+
+                    auto scoreIter = scores.find(nodeId);
+                    double pprScore = (scoreIter != scores.end()) ? scoreIter->second : 0.0;
+                    double calibratedScore = pprScore + residual;
+                    if (calibratedScore <= 0.0) {
+                        continue;
+                    }
+                    residualPairs.emplace_back(nodeId, calibratedScore);
+                }
+
+                residualTopK = std::min(maxResidualNodes, static_cast<int32_t>(residualPairs.size()));
+                if (residualTopK > 0) {
+                    std::partial_sort(residualPairs.begin(),
+                                      residualPairs.begin() + residualTopK,
+                                      residualPairs.end(),
+                                      [](const auto& a, const auto& b) { return a.second > b.second; });
+
+                    for (int32_t rankIdx = 0; rankIdx < residualTopK; ++rankIdx) {
+                        flatIds.push_back(static_cast<int64_t>(residualPairs[rankIdx].first));
+                        flatWeights.push_back(residualPairs[rankIdx].second);
+                    }
+                }
+            }
+            validCounts.push_back(static_cast<int64_t>(topK + residualTopK));
         }
 
         result[nodeTypeId] = {torch::tensor(flatIds, torch::kLong),
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -267,10 +267,16 @@ void PPRForwardPush::pushResiduals(
 }
 
 std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPush::
-    extractTopKWithResidualTopUp(int32_t maxPprNodes, int32_t maxResidualNodes, int32_t maxTotalNodes) {
-    TORCH_CHECK(maxPprNodes >= 0, "maxPprNodes must be non-negative, got ", maxPprNodes, ".");
+    extractTopKWithResidualTopUp(int32_t maxPPRNodes, int32_t maxResidualNodes, int32_t maxTotalNodes) {
+    TORCH_CHECK(maxPPRNodes >= 0, "maxPPRNodes must be non-negative, got ", maxPPRNodes, ".");
     TORCH_CHECK(maxResidualNodes >= 0, "maxResidualNodes must be non-negative, got ", maxResidualNodes, ".");
     TORCH_CHECK(maxTotalNodes >= -1, "maxTotalNodes must be >= -1 (-1 means unbounded), got ", maxTotalNodes, ".");
+    TORCH_CHECK(maxTotalNodes == -1 || maxTotalNodes >= maxPPRNodes,
+                "maxTotalNodes must be -1 or >= maxPPRNodes, got maxTotalNodes=",
+                maxTotalNodes,
+                " and maxPPRNodes=",
+                maxPPRNodes,
+                ".");
 
     std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> result;
     // Emit an entry for every node type, even if unreachable in this batch (empty tensors,
@@ -284,10 +290,8 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
         for (int32_t seedIdx = 0; seedIdx < _batchSize; ++seedIdx) {
             const auto& nodeTypeState = _state[seedIdx][nodeTypeId];
             const auto& scores = nodeTypeState.pprScores;
-            int32_t pprNodeLimit = maxPprNodes;
-            int32_t totalNodeLimit = maxPprNodes + maxResidualNodes;
+            int32_t totalNodeLimit = maxPPRNodes + maxResidualNodes;
             if (maxTotalNodes >= 0) {
-                pprNodeLimit = std::min(maxPprNodes, maxTotalNodes);
                 totalNodeLimit = maxTotalNodes;
             }
             // With no residual top-up budget, emit finalized PPR scores as-is.
@@ -297,15 +301,15 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
             // sorted consistently.
             bool includeResidualMassInOutputScores = maxResidualNodes > 0;
 
-            int32_t topK = std::min(pprNodeLimit, static_cast<int32_t>(scores.size()));
+            int32_t topK = std::min(maxPPRNodes, static_cast<int32_t>(scores.size()));
             int32_t residualBudget = std::min(maxResidualNodes, std::max<int32_t>(0, totalNodeLimit - topK));
             int32_t residualTopK = 0;
             std::vector<std::pair<int32_t, double>> selectedPairs;
             selectedPairs.reserve(static_cast<size_t>(topK + residualBudget));
-            std::unordered_set<int32_t> selectedNodeIds;
+            std::unordered_set<int32_t> selectedPPRNodeIds;
             if (topK > 0) {
                 if (residualBudget > 0) {
-                    selectedNodeIds.reserve(static_cast<size_t>(topK));
+                    selectedPPRNodeIds.reserve(static_cast<size_t>(topK));
                 }
                 std::vector<std::pair<int32_t, double>> scorePairs(scores.begin(), scores.end());
                 // Selection is intentionally two-phase: finalized nodes are selected
@@ -328,7 +332,7 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
                     }
                     selectedPairs.emplace_back(nodeId, outputScore);
                     if (residualBudget > 0) {
-                        selectedNodeIds.insert(nodeId);
+                        selectedPPRNodeIds.insert(nodeId);
                     }
                 }
             }
@@ -337,7 +341,7 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
                 std::vector<std::pair<int32_t, double>> residualPairs;
                 residualPairs.reserve(nodeTypeState.residuals.size());
                 for (const auto& [nodeId, residual] : nodeTypeState.residuals) {
-                    if (residual <= 0.0 || selectedNodeIds.find(nodeId) != selectedNodeIds.end()) {
+                    if (residual <= 0.0 || selectedPPRNodeIds.find(nodeId) != selectedPPRNodeIds.end()) {
                         continue;
                     }
 
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -294,6 +294,7 @@ PPRExtractResult PPRForwardPush::extractTopKWithResidualTopUp(int32_t maxPprNode
                 pprNodeLimit = std::min(maxPprNodes, maxTotalNodes);
                 totalNodeLimit = maxTotalNodes;
             }
+            bool emitCalibratedScores = maxResidualNodes > 0;
 
             int32_t topK = std::min(pprNodeLimit, static_cast<int32_t>(scores.size()));
             int32_t residualBudget = std::min(maxResidualNodes, std::max<int32_t>(0, totalNodeLimit - topK));
@@ -313,7 +314,14 @@ PPRExtractResult PPRForwardPush::extractTopKWithResidualTopUp(int32_t maxPprNode
 
                 for (int32_t rankIdx = 0; rankIdx < topK; ++rankIdx) {
                     int32_t nodeId = scorePairs[rankIdx].first;
-                    selectedPairs.emplace_back(nodeId, scorePairs[rankIdx].second);
+                    double calibratedScore = scorePairs[rankIdx].second;
+                    if (emitCalibratedScores) {
+                        auto residualIter = nodeTypeState.residuals.find(nodeId);
+                        if (residualIter != nodeTypeState.residuals.end()) {
+                            calibratedScore += residualIter->second;
+                        }
+                    }
+                    selectedPairs.emplace_back(nodeId, calibratedScore);
                     if (residualBudget > 0) {
                         selectedNodeIds.insert(nodeId);
                     }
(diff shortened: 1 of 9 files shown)


gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -146,8 +147,7 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
     return result;
 }
 
-void PPRForwardPush::pushResiduals(
-    const std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>>& fetchedByEtypeId) {
+void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
```
kmontemayor2-sc: btw I'd prefer if these changes could be made in a separate PR :) makes it easier to distinguish the business logic changes vs refactors

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -265,9 +265,18 @@ void PPRForwardPush::pushResiduals(
     }
 }
 
-std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPush::extractTopK(
-    int32_t maxPprNodes) {
-    std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> result;
+PPRExtractResult PPRForwardPush::extractTopK(int32_t maxPprNodes) {
+    return extractTopKWithResidualTopUp(maxPprNodes, /*maxResidualNodes=*/0);
```
kmontemayor2-sc: Why do we always call with `0` here?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -265,9 +265,18 @@ void PPRForwardPush::pushResiduals(
...
+}
+
+PPRExtractResult PPRForwardPush::extractTopKWithResidualTopUp(int32_t maxPprNodes,
+                                                              int32_t maxResidualNodes,
+                                                              int32_t maxTotalNodes) {
+    TORCH_CHECK(maxPprNodes >= 0, "maxPprNodes must be non-negative, got ", maxPprNodes, ".");
+    TORCH_CHECK(maxResidualNodes >= 0, "maxResidualNodes must be non-negative, got ", maxResidualNodes, ".");
+    TORCH_CHECK(maxTotalNodes >= -1, "maxTotalNodes must be -1 or non-negative, got ", maxTotalNodes, ".");
```
kmontemayor2-sc: is it easier to say "Must be >= -1 (-1 means unbounded)"?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +286,88 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+                pprNodeLimit = std::min(maxPprNodes, maxTotalNodes);
+                totalNodeLimit = maxTotalNodes;
+            }
+            bool emitCalibratedScores = maxResidualNodes > 0;
+
+            int32_t topK = std::min(pprNodeLimit, static_cast<int32_t>(scores.size()));
+            int32_t residualBudget = std::min(maxResidualNodes, std::max<int32_t>(0, totalNodeLimit - topK));
+            int32_t residualTopK = 0;
+            std::vector<std::pair<int32_t, double>> selectedPairs;
+            selectedPairs.reserve(static_cast<size_t>(topK + residualBudget));
```
kmontemayor2-sc: any reason we do it this way vs just give it the right size at init?

I found https://stackoverflow.com/questions/8928547/vector-initialization-or-reserve which implies it doesn't really matter but it does read "cleaner" to me to do the sizing at init?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +286,88 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
                                   scorePairs.begin() + topK,
                                   scorePairs.end(),
                                   [](const auto& a, const auto& b) { return a.second > b.second; });
 
                 for (int32_t rankIdx = 0; rankIdx < topK; ++rankIdx) {
-                    flatIds.push_back(static_cast<int64_t>(scorePairs[rankIdx].first));
-                    flatWeights.push_back(scorePairs[rankIdx].second);
+                    int32_t nodeId = scorePairs[rankIdx].first;
+                    double calibratedScore = scorePairs[rankIdx].second;
+                    if (emitCalibratedScores) {
```
kmontemayor2-sc: BTW can we name this var better? Is it more like "should top up"?

Maybe I'm a bit lost, it'd be nice if we can have more documentation in the source code?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -273,6 +282,79 @@ async def _batch_fetch_neighbors(
             for eid, output in zip(eids, outputs)
         }
 
+    def _extract_ppr_state_top_k(
+        self,
+        ppr_state,
+        device: torch.device,
+        max_ppr_nodes: Optional[int] = None,
```
kmontemayor2-sc: why is this an arg here? It's only every called internally right?

Should we either remove it or make it required and always pass it in?

kmontemayor2-sc: Ditto generally.

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -273,6 +282,79 @@ async def _batch_fetch_neighbors(
...
+        ntype_to_valid_counts: dict[NodeType, torch.Tensor] = {}
+
+        ppr_node_quota = (
+            max_ppr_nodes if max_ppr_nodes is not None else self._max_ppr_nodes
+        )
+        if residual_topup_nodes > 0:
+            max_total_nodes_for_extraction = (
+                max_total_nodes if max_total_nodes is not None else -1
+            )
+            extracted_results = ppr_state.extract_top_k_with_residual_top_up(
```
kmontemayor2-sc: qq why not always call this? it still works with the other path right?

tests/unit/distributed/dist_ppr_sampler_test.py
```diff
@@ -815,6 +818,132 @@ def test_ppr_sampler_homogeneous_ablp(self) -> None:
...
+        sampler._max_ppr_nodes = max_ppr_nodes
+        sampler._enable_residual_topup = enable_residual_topup
+        sampler._max_fetch_iterations = None
+        sampler._is_homogeneous = True
+        sampler._node_type_to_id = {DEFAULT_HOMOGENEOUS_NODE_TYPE: 0}
+        sampler._ntype_id_to_ntype = [DEFAULT_HOMOGENEOUS_NODE_TYPE]
+        sampler._node_type_id_to_edge_type_ids = [[]]
+        sampler._edge_type_id_to_dst_ntype_id = []
+        sampler._degree_tensors_for_cpp = [torch.zeros(0, dtype=torch.int32)]
+        return sampler
```
kmontemayor2-sc: Can we avoid monkey patching like this?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +286,88 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+                if (residualBudget > 0) {
+                    selectedNodeIds.reserve(static_cast<size_t>(topK));
+                }
                 std::vector<std::pair<int32_t, double>> scorePairs(scores.begin(), scores.end());
                 std::partial_sort(scorePairs.begin(),
                                   scorePairs.begin() + topK,
                                   scorePairs.end(),
                                   [](const auto& a, const auto& b) { return a.second > b.second; });
 
                 for (int32_t rankIdx = 0; rankIdx < topK; ++rankIdx) {
```
kmontemayor2-sc, line 311: Robot review:

```
  ▎ Selection metric and emitted metric disagree — the result isn't top-k under the score it's sorted by.
  ▎
  ▎ The primary top-k is selected by raw pprScore (the partial_sort on scorePairs at line 310), but each selected node is then emitted with the
  ▎ calibrated score pprScore + residual (lines 318-322) and the combined set is re-sorted by that calibrated score later (lines 331+). So the
  ▎ returned set is top-k by raw ppr, merely re-ordered by calibrated score — it is not the top-k under the emitted metric.
  ▎
  ▎ Concretely, a residual candidate whose calibrated score would outrank a selected finalized node can still be excluded, because it competes
  ▎ only for residualBudget = totalNodeLimit - topK slots rather than against the finalized set. Reproduced against the built extension:
  ▎ topup(2,2,2):  ids=[0,1], weights=[0.5, 0.0625]     # node 3 (calibrated 0.1875) excluded
  ▎ topup(2,2,-1): ids=[0,3,1], weights=[0.5, 0.1875, 0.0625]
  ▎ This matches the "do not displace finalized-PPR nodes" note in ppr_forward_push.h:309-320, so it may be intended. If so, please add a
  ▎ one-line note here that the emitted set is top-k by raw ppr score and only re-sorted by the calibrated score. If displacement is actually
  ▎ desired, build a single candidate pool scored by pprScore + residual and take top-maxTotalNodes instead of the two-phase select.
```

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -374,36 +456,16 @@ async def _compute_ppr_scores(
...
+        residual_topup_nodes = (
+            self._max_ppr_nodes if self._enable_residual_topup else 0
+        )
+        return self._extract_ppr_state_top_k(
+            ppr_state,
+            device,
+            max_ppr_nodes=self._max_ppr_nodes,
+            residual_topup_nodes=residual_topup_nodes,
+            max_total_nodes=self._max_ppr_nodes,
+        )
```
kmontemayor2-sc: Robot review:

```
  ▎ Because the regular path passes residual_topup_nodes = max_ppr_nodes and max_total_nodes = max_ppr_nodes, the C++ side computes
  ▎ residualBudget = max_ppr_nodes - topK. Whenever a seed already has >= max_ppr_nodes finalized PPR scores, residualBudget == 0 and residual
  ▎ top-up never fires — which is precisely the well-explored/hub-seed case in the truncated (max_fetch_iterations) regime that motivated top-up,
  ▎ where large sub-threshold residuals are most likely to exist. Worth confirming this is the intended envelope (see Comment A), and
  ▎ documenting it.
```

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +286,88 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+                    if (residual <= 0.0 || selectedNodeIds.find(nodeId) != selectedNodeIds.end()) {
+                        continue;
+                    }
+
+                    auto scoreIter = scores.find(nodeId);
+                    double pprScore = (scoreIter != scores.end()) ? scoreIter->second : 0.0;
+                    double calibratedScore = pprScore + residual;
+                    if (calibratedScore <= 0.0) {
+                        continue;
+                    }
```

```
  ▎ if (calibratedScore <= 0.0) { continue; } is dead. This branch is only reached when residual > 0.0 (guaranteed by the guard at line 335) and
  ▎ pprScore >= 0 (all pprScores entries are absorbed mass, always non-negative), so calibratedScore = pprScore + residual is always strictly
  ▎ positive. Either remove it, or keep it as an explicit defensive guard with a comment saying so — right now it reads like a live filter. Note
  ▎ the primary-selection path has no equivalent check, so the asymmetry is also slightly confusing.
```

kmontemayor2-sc (commented):
Thanks Matt! In generally I think we have a good amount of algorithmic changes here and it'd be nice if we could comment what / why the new code is doing.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Makes sense, I reverted the type-alias/refactor-only changes from this PR and left a TODO to move the repeated tensor tuple/map types into a follow-up refactor-only PR, so this PR stays focused on the residual top-up logic

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: We've made the residual top-off a configurable knob (which defaults to True) and this is the previous path, so 0 means “do not include residual top-up nodes” and preserves the previous finalized-PPR-only behavior. The nonzero residual budget is only used by extractTopKWithResidualTopUp; this wrapper delegates to the shared implementation to avoid duplicating extraction code.

I've gone ahead and removed this since we can just use extractTopKWithResidualTopUp in both cases later on, thanks!

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: topK + residualBudget is only an upper bound, not the final size. We always add topK finalized PPR nodes, but we may find fewer residual candidates than residualBudget. If we initialized the vector with that size, it would contain placeholder pairs and we’d need extra indexing/trimming logic. So reserve gives us the allocation benefit without changing the logical size.

tests/unit/distributed/dist_ppr_sampler_test.py
mkolodner-sc: pulled the C++ extraction-method selection into a small helper and updated the tests to call it directly with fake PPR states, so we no longer need object.__new__ or manually assigned private sampler fields.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Done

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: This is intended. Residual top-up is a fill-only mechanism: finalized PPR nodes are selected first by raw PPR score, and residual candidates only compete for the remaining budget. After that, the selected rows are emitted/sorted by ppr_score + residual when top-up is enabled. I added source comments to make explicit that this is a two-phase policy and not a global top-k over the emitted score when maxTotalNodes is tight.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Yep, that is the intended envelope for the regular sampler right now. We cap total output at max_ppr_nodes, so residual candidates only fill slots when finalized PPR produces fewer than that cap; they do not displace finalized PPR nodes. I added documentation at the sampler call site and in the constructor docs to make this clearer.

mkolodner-sc:
> Thanks Matt! In generally I think we have a good amount of algorithmic changes here and it'd be nice if we could comment what / why the new code is doing.

@kmontemayor2-sc Done, added more context in the PR description and in the PR for the reasoning/change made here

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -37,6 +37,20 @@
...
+    max_ppr_nodes: int,
+    residual_topup_nodes: int,
+    max_total_nodes: int,
+) -> dict[int, tuple[torch.Tensor, torch.Tensor, torch.Tensor]]:
+    """Call the shared C++ extraction path for the configured quotas."""
+    return ppr_state.extract_top_k_with_residual_top_up(
+        max_ppr_nodes,
+        residual_topup_nodes,
+        max_total_nodes,
+    )
```
kmontemayor2-sc: is this function needed?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <torch/torch.h>
 
+#include <algorithm>
```
kmontemayor2-sc, line 5: btw why do we include this?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +282,94 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
 
         for (int32_t seedIdx = 0; seedIdx < _batchSize; ++seedIdx) {
-            const auto& scores = _state[seedIdx][nodeTypeId].pprScores;
-            int32_t topK = std::min(maxPprNodes, static_cast<int32_t>(scores.size()));
+            const auto& nodeTypeState = _state[seedIdx][nodeTypeId];
+            const auto& scores = nodeTypeState.pprScores;
+            int32_t pprNodeLimit = maxPprNodes;
+            int32_t totalNodeLimit = maxPprNodes + maxResidualNodes;
+            if (maxTotalNodes >= 0) {
+                pprNodeLimit = std::min(maxPprNodes, maxTotalNodes);
```
kmontemayor2-sc: Hmmm, should we have a condition that `maxTotalNodes >= maxPprNodes`? Does it make sense to have more PPR nodes than total nodes?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -265,8 +266,12 @@ void PPRForwardPush::pushResiduals(
     }
 }
 
-std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPush::extractTopK(
-    int32_t maxPprNodes) {
+std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPush::
+    extractTopKWithResidualTopUp(int32_t maxPprNodes, int32_t maxResidualNodes, int32_t maxTotalNodes) {
```
kmontemayor2-sc: nit. Can we do `maxPPRNodes`? PPR is an initialism and it reads better to me all caps (we should do this generally in our code base, see the `PPRForwardPush` class name.

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +282,94 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+            // same output score scale so the merged list is comparable and
+            // sorted consistently.
+            bool includeResidualMassInOutputScores = maxResidualNodes > 0;
+
+            int32_t topK = std::min(pprNodeLimit, static_cast<int32_t>(scores.size()));
+            int32_t residualBudget = std::min(maxResidualNodes, std::max<int32_t>(0, totalNodeLimit - topK));
+            int32_t residualTopK = 0;
+            std::vector<std::pair<int32_t, double>> selectedPairs;
+            selectedPairs.reserve(static_cast<size_t>(topK + residualBudget));
+            std::unordered_set<int32_t> selectedNodeIds;
```
kmontemayor2-sc: Is this more like `residualNodeCandidateIds`?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -273,6 +304,71 @@ async def _batch_fetch_neighbors(
...
+        ppr_state,
+        device: torch.device,
+        max_ppr_nodes: int,
+        residual_topup_nodes: int,
+        max_total_nodes: int,
+    ) -> tuple[
+        Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+        Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+        Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+    ]:
```
kmontemayor2-sc, line 302: can we add a docstring for what this is returning?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -374,36 +470,17 @@ async def _compute_ppr_scores(
...
+        # Regular sampling uses residual top-up as fill-only under the
+        # max_ppr_nodes cap.  If finalized PPR scores already fill that cap,
+        # max_total_nodes leaves no budget for residual candidates.
+        residual_topup_nodes = self._max_ppr_nodes if self._enable_residual_topup else 0
+        return self._extract_ppr_state_top_k(
+            ppr_state,
+            device,
+            max_ppr_nodes=self._max_ppr_nodes,
+            residual_topup_nodes=residual_topup_nodes,
+            max_total_nodes=self._max_ppr_nodes,
```
kmontemayor2-sc: why do we add `max_total_nodes` if it's always `max_ppr_nodes`?

gigl/distributed/sampler_options.py
```diff
@@ -44,6 +44,12 @@ class PPRSamplerOptions:
 
     For homogeneous graphs these live directly on ``data.edge_index`` / ``data.edge_attr``.
```
kmontemayor2-sc, line 46: Can we try to make this comment more "why" and not "what" e.g. "Enable residual top-up to get longer sequence lengths ... "

tests/unit/distributed/dist_ppr_sampler_test.py
kmontemayor2-sc: Hmm, the tests still seem very mock heavy - what are they testing here? 

It seems like we coverage of the new behavior in the cpp tests right? What are the python tests adding?

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Removed, good catch, thanks!

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: std::min, std::sort, and std::partial_sort require this. These functions worked before because we had the include in the .h file, but it's better for algorithm imports to live in the .cpp file, so we moved it from .h to here

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Removed

tests/unit/distributed/dist_ppr_sampler_test.py
mkolodner-sc: Hmm, that's fair. I'll remove these tests since they aren't adding much over the current cpp coverage

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -265,8 +266,18 @@ void PPRForwardPush::pushResiduals(
     }
 }
 
-std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPush::extractTopK(
-    int32_t maxPprNodes) {
+std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> PPRForwardPush::
+    extractTopKWithResidualTopUp(int32_t maxPPRNodes, int32_t maxResidualNodes, int32_t maxTotalNodes) {
```
kmontemayor2-sc: We never pass in `maxTotalNodes` via python right? 

I guess what I was getting at earlier is why do we have this API separate from `maxNodes: int, enableTopUp: bool`? 

Since we do just pass in `maxPPRNodes` for both if top up is enabled?

mkolodner-sc: Done. I simplified the API to max_ppr_nodes: int and enable_residual_topup: bool -- max_ppr_nodes is now the final per-seed output cap, and C++ derives the residual top-up budget from the remaining slots when top-up is enabled.

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +280,88 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
-            const auto& scores = _state[seedIdx][nodeTypeId].pprScores;
-            int32_t topK = std::min(maxPprNodes, static_cast<int32_t>(scores.size()));
+            const auto& nodeTypeState = _state[seedIdx][nodeTypeId];
+            const auto& scores = nodeTypeState.pprScores;
+            // With no residual top-up budget, emit finalized PPR scores as-is.
+            // When residual top-up is enabled, residual candidates are scored as
+            // ppr_score + residual; selected finalized PPR nodes must use that
+            // same output score scale so the merged list is comparable and
+            // sorted consistently.
+            bool includeResidualMassInOutputScores = enableResidualTopUp;
```
kmontemayor2-sc: nit. Can we remove this var all together now?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +280,88 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+
+            int32_t topK = std::min(maxPPRNodes, static_cast<int32_t>(scores.size()));
+            int32_t residualBudget = enableResidualTopUp ? std::max<int32_t>(0, maxPPRNodes - topK) : 0;
```
kmontemayor2-sc: nit. maybe having this as `optional<int>` is more explicit? Up to you

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -277,21 +280,88 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+                    for (int32_t rankIdx = 0; rankIdx < residualTopK; ++rankIdx) {
+                        selectedPairs.emplace_back(residualPairs[rankIdx].first, residualPairs[rankIdx].second);
+                    }
+                }
+            }
+
+            if (includeResidualMassInOutputScores && selectedPairs.size() > 1) {
+                std::sort(selectedPairs.begin(), selectedPairs.end(), [](const auto& a, const auto& b) {
+                    return a.second > b.second;
+                });
```
kmontemayor2-sc: hmmm do we need to have these two different sorts? If we added all of `residual_pairs` to `selectedPairs` what would happen?

Furthermore, in this world, should we still sort on line 307? 

https://github.com/Snapchat/GiGL/pull/702/changes#diff-ab52b8cce401b72029c9c54e647bdb31032956d64ab90574c33c6529b89633a7R307-R311

kmontemayor2-sc (commented):
Hi Matt, once we resolve the sorting questions this lgtm! thanks for the iterations here :)

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: I think using this as an explicit int and having the check be > 0 makes sense.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Good call. I updated this so the top-up path no longer does multiple ordering sorts over the candidate pools.

We still need to keep selection two-phase so residual candidates cannot displace finalized PPR nodes. So instead of adding all residual pairs into selectedPairs and taking a single global top-k, we first select finalized PPR nodes by raw PPR score, then use only the remaining budget for residual candidates.

The optimization is that the intermediate steps now use std::nth_element, not std::partial_sort, because we only need to select the top K candidates from each pool, not order them yet. Then we do one final std::sort over the selected output rows, bounded by maxPPRNodes, so the emitted sequence is ordered by ppr_score + residual.

So in the top-up case this is now:
- nth_element to select raw-PPR top K finalized nodes
- nth_element to select residual top-up candidates for the remaining slots
- one final small sort over the emitted result

This preserves the no-displacement top-up semantics while avoiding unnecessary full ordering work on the larger candidate pools.

kmontemayor2-sc: out of curiosity, is it possible the top-ups would replace the top-k nodes? They would have been skipped right so they have lower values in general?

I guess I'm wondering if we've done any analysis on the nodes we're adding / the scores they have here. It's possible the total nodes would be the same just the order is different?

mkolodner-sc: It is possible. This is because the re-enqueue criteria for PPR forward push is based not only on the ppr score that is emitted, but also the degree of that node. For example, if we have a node with 1_000_000 edges, the re-enqueue threshold is based not only on the residual of that node, but also how much residual they would be distributing to their neighbors. So a node could have a high residual but might not valuable enough to expand further under the degree-scaled threshold.

kmontemayor2-sc (approved):
Thanks Matt!

yliu2-sc (approved):

PR #703, opened 2026-07-13
Reuse fetched PPR neighbor cache entries

mkolodner-sc:
Moves fetched PPR neighbor rows into the persistent per-state neighbor cache immediately.

Previously, pushResiduals() kept newly fetched neighbor lists in a temporary fetched map for the current push, while older neighbor lists lived in _neighborCache. That meant the push loop had to check both places, and a node that re-entered the frontier later could still trigger another lookup if its neighbor row had not been promoted into _neighborCache.
Now, every fetched (node_id, edge_type_id) row is written to _neighborCache before residuals are pushed. After that, both drainQueue() and the residual-push loop use _neighborCache as the single source of truth for known adjacency.

Example: Example: suppose node `u` is fetched and pushed in iteration 1. Later, residual flows back to `u` through another path and `u` re-enters the frontier. Before this change, `u`’s adjacency may not have been retained unless it matched the old promotion path. With this change, `drainQueue()` sees `u`’s neighbor list in `_neighborCache` and avoids re-fetching it.

The PPR math is unchanged; this only avoids redundant neighbor fetches and simplifies the push path by removing the split between “newly fetched” and “previously cached” neighbor rows.

### DIFF
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -148,9 +148,10 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
 
 void PPRForwardPush::pushResiduals(
     const std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>>& fetchedByEtypeId) {
-    // Step 1: Unpack the input map into a C++ map keyed by packKey(nodeId, edgeTypeId)
-    // for fast lookup during the residual-push loop below.
-    std::unordered_map<uint64_t, std::vector<int32_t>> fetched;
+    // Step 1: Persist fetched neighbor lists in the per-state cache. drainQueue()
+    // consults this cache before requesting future lookups, so storing every
+    // fetched row here avoids re-fetching a (node, edge type) pair if it re-enters
+    // the frontier later in the same PPR channel.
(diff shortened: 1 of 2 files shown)


gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -172,7 +173,10 @@ void PPRForwardPush::pushResiduals(
             for (int64_t neighborIdx = 0; neighborIdx < count; ++neighborIdx) {
                 neighborIds[neighborIdx] = static_cast<int32_t>(flatNeighborIdsAccessor[offset + neighborIdx]);
             }
-            fetched[packKey(nodeId, edgeTypeId)] = std::move(neighborIds);
+            uint64_t cacheKey = packKey(nodeId, edgeTypeId);
+            if (_neighborCache.find(cacheKey) == _neighborCache.end()) {
+                _neighborCache.emplace(cacheKey, std::move(neighborIds));
```
kmontemayor2-sc, line 178: btw does this increase our memory usage? 

we're now storing all neighbors for a ppr batch generation right?

kmontemayor2-sc (approved):
Thanks Matt!

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes this does come with this tradeoff, but since training bottleneck are usually on the throughput instead of memory, this is an acceptable tradeoff.

yliu2-sc (approved):

PR #706, opened 2026-07-15
BqUtils: support a BigQuery quota project override + consolidate client construction

kmontemayor2-sc:
What

  - BqUtils.__init__ now accepts an optional quota_project_id and resolves the BigQuery client's quota/billing project as explicit arg → GIGL_BIGQUERY_QUOTA_PROJECT env var → none (unset ⇒ unchanged default behavior). When set, it's applied via ClientOptions(quota_project_id=...).
  - Adds BqUtils.load_files_to_bq(source_uris, bq_path, job_config, should_run_async=False) — loads files (e.g. Avro) from GCS into a BigQuery table (fulfilling the long-standing TODO to centralize this in BqUtils).
  - Routes the two remaining direct bigquery.Client(...) sites through BqUtils: the Avro export in gigl/common/data/export.py (load_embeddings_to_bigquery / load_predictions_to_bigquery) and the shard-key check in gigl/common/beam/sharded_read.py. After this change, the only bigquery.Client(...) construction in gigl/ lives inside BqUtils.

  Why

  The BigQuery client's quota project (used for API-enablement and per-API quota checks) defaults to the project associated with the active credentials. When a job runs under credentials whose default project differs from the project that owns the BigQuery resources — e.g. a workload that creates load jobs in a data project while its credentials belong to a different infrastructure project — the enablement/quota check runs against that default project, which may not have the BigQuery API enabled.

  A per-client quota_project_id lets callers direct BigQuery's quota/billing to the appropriate project. It is intentionally BigQuery-scoped (set only on the BqUtils client), so it does not affect other Google Cloud clients (e.g. GCS) — unlike the process-global GOOGLE_CLOUD_QUOTA_PROJECT. Consolidating client construction in BqUtils means this (and future client config) applies uniformly to all callers without per-call-site plumbing.

  Setting the quota project requires the caller to have serviceusage.services.use on that project (the roles/serviceusage.serviceUsageConsumer role).

  Configuration

  - GIGL_BIGQUERY_QUOTA_PROJECT (new, optional) — default quota project for all BqUtils BigQuery clients. Unset ⇒ no override (current behavior).
  - Per-call override: pass quota_project_id=... (takes precedence over the env var). export.py's public functions forward an optional quota_project_id for this.


gigl/src/common/utils/bq.py
```diff
@@ -58,9 +62,44 @@ def _load_file_to_bq_with_retry(
...
+    def __init__(
+        self,
+        project: Optional[str] = None,
+        quota_project_id: Optional[str] = None,
+    ) -> None:
+        """Initializes a BqUtils instance wrapping a single BigQuery client.
+
+        The quota / billing project for the underlying client is resolved as:
+        the explicit ``quota_project_id`` argument if provided, else the
+        ``GIGL_BIGQUERY_QUOTA_PROJECT`` environment variable if set to a
```
mkolodner-sc, line 74: qq -- how would `GIGL_BIGQUERY_QUOTA_PROJECT` get set in practice? Other than knowing to look in `gigl/env/constants.py`, is there somewhere we are documenting all of the supported env variables for users to consume? Otherwise, it might get hard to keep track.

Realistically, if we don't expect users to use the environment-supported path here, should we only support explicitly provided quota_project_ids for now, and have the fallback directly be the no override?

tests/integration/common/beam/sharded_read_test.py
```diff
@@ -0,0 +1,121 @@
...
+    ShardedExportRead,
+)
+from gigl.env.pipelines_config import get_resource_config
+from gigl.src.common.utils.bq import BqUtils
+from tests.test_assets.test_case import TestCase
+
+NUM_ROWS = 20
+
+
+class ShardedReadIntegrationTest(TestCase):
```
mkolodner-sc, line 20: Hmm, does adding this test belong in this PR? This seems a bit large of a test provided the change we've made to the sharded read file is to just use `fetch_bq_table_schema`, and should be covered by a test for that, rather than the whole sharded read logic.

gigl/common/beam/sharded_read.py
```diff
@@ -31,11 +31,11 @@ def _assert_shard_key_in_table(table_name: str, shard_key: str) -> None:
...
-    client = bigquery.Client()
-    table_ref = bigquery.TableReference.from_string(table_name)
-
-    table = client.get_table(table_ref)
-    column_names = [field.name for field in table.schema]
+    # BqUtils.fetch_bq_table_schema resolves the table via BigQuery's own path
+    # parsing, so it accepts the same `project.dataset.table` strings (and, as a
+    # superset, `project:dataset.table`) as the previous TableReference lookup.
+    bq_utils = BqUtils()
+    column_names = list(bq_utils.fetch_bq_table_schema(bq_table=table_name).keys())
```
mkolodner-sc, line 38: What's the reason we need this change in this PR? It seems independent to the quota project override goal of the other changes here.

tests/integration/common/data/export_test.py
```diff
@@ -105,3 +110,58 @@ def test_embedding_export(self, _, should_run_async: bool):
             bq_client.count_number_of_rows_in_bq_table(bq_export_table_path),
             num_nodes * 2,
         )
+
+
+class PredictionExportIntegrationTest(TestCase):
```
mkolodner-sc, line 115: Similarly, this new test and the test deletion deltas feels a bit large in scope for the main production logic change of this PR to have quota project override enabled.

mkolodner-sc (commented):
Thanks Kyle! This mostly LGTM, though I think it'd be nice if we can break the changes the sharded_read and export logic into their own PRs. While their business logic deltas are small, I feel they are incurring a fairly large testing delta that would warrant them moving to their own PR, rather than muddling the scope of the quota project id changes, which is the main business logic contribution of this PR IIUC.

tests/integration/common/data/export_test.py
kmontemayor2-sc: Sure, broke up into three prs:

https://github.com/Snapchat/GiGL/pull/715
https://github.com/Snapchat/GiGL/pull/716
https://github.com/Snapchat/GiGL/pull/717

PR #707, opened 2026-07-15
Clean up PPR sampler naming

mkolodner-sc:
## Summary

Splits out behavior-neutral PPR sampler cleanup ahead of the typed-PPR changes.

This PR intentionally does not change PPR sampling behavior. It renames local variables and tightens nearby comments/docstrings in `DistPPRNeighborSampler` so the follow-up PRs can focus on event-loop performance and typed-channel behavior.

## Stack

- Base: #702 (`mkolodner/ppr-residual-topup`)
- Next: #708

## Reasoning for Change

The typed-PPR work touches a lot of code in the distributed PPR sampler. Pulling readability cleanup into its own PR makes the business-logic diffs easier to review and keeps later review comments focused on the functional changes.

## Changes

- Uses explicit names for node-type and edge-type variables in the PPR sampler.
- Cleans up PPR extraction/fetch helper wording to match the current integer-ID flow.
- Leaves the sampler algorithm, output shape, and C++ PPR behavior unchanged.

PR #708, opened 2026-07-15
Move existing PPR sampler work off event loop

mkolodner-sc:
## Summary

Moves the existing distributed PPR sampler's CPU-heavy C++ work off the asyncio event-loop path.

This PR is limited to the existing/non-typed PPR path. Typed-channel PPR enablement is left for the next PR in the stack.

## Stack

- Base: #707
- Next: #709

## Reasoning for Change

PPR sampling runs several CPU-heavy C++ steps while async graph-store/RPC work is also in flight. When those steps run on the event-loop thread or hold the GIL longer than necessary, they can delay RPC completion callbacks and reduce concurrency across sampler coroutines.

This change keeps the PPR math and returned sequences unchanged while moving construction, queue draining, residual pushes, and extraction through executor/GIL-release boundaries where possible.

## Changes

- Releases the GIL around `PPRForwardPush` construction, queue draining, residual push, and extraction in the pybind layer.
- Runs PPR state construction, `drain_queue`, `push_residuals`, and top-k extraction through `run_in_executor` from the Python sampler.
- Removes a redundant frontier `.to(device)` before one-hop fetches; extracted tensors are still moved to the sampler device at the existing output boundary.

### DIFF
--- a/gigl-core/core/sampling/python_ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/python_ppr_forward_push.cpp
@@ -18,9 +18,8 @@ namespace py = pybind11;
 
 namespace gigl {
 
-// pushResiduals: a wrapper is needed solely to release the GIL during the C++ push.
-// pybind11/stl.h handles all type conversions automatically; the other methods use
-// direct member function pointers for the same reason.
+// pushResiduals receives Python-owned containers, so convert them while the GIL
+// is held and release only around the C++ state update.
 static void pushResidualsWrapper(PPRForwardPush& state, const py::dict& fetchedByEtypeId) {
     std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> neighborTensorsByEtypeId;
     // Dict iteration touches Python objects — GIL must be held here.
@@ -45,8 +44,9 @@ static void pushResidualsWrapper(PPRForwardPush& state, const py::dict& fetchedB
 
 static std::optional<std::unordered_map<int32_t, torch::Tensor>> drainQueueWrapper(PPRForwardPush& state) {
     std::optional<std::unordered_map<int32_t, torch::Tensor>> drained;
-    // The drain mutates only this PPRForwardPush instance and returns torch tensors.
-    // No Python objects are touched until pybind converts the result after return.
+    // drainQueue mutates only this PPRForwardPush instance and materializes CPU
+    // tensors for frontier node IDs. pybind converts those tensor handles back
+    // to Python tensors after return without copying the underlying storage.
     {
         py::gil_scoped_release release;
         drained = state.drainQueue();
@@ -57,8 +57,8 @@ static std::optional<std::unordered_map<int32_t, torch::Tensor>> drainQueueWrapp
 static std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>>
 extractTopKWithResidualTopUpWrapper(PPRForwardPush& state, int32_t maxPPRNodes, bool enableResidualTopUp) {
     std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> result;
-    // Extraction walks C++ state and builds torch tensors; Python only sees the
-    // result after pybind converts it on return.
+    // Extraction walks C++ state and builds torch tensors. Returning through
+    // pybind creates Python container/wrapper objects, not tensor data copies.
     {
         py::gil_scoped_release release;
         result = state.extractTopKWithResidualTopUp(maxPPRNodes, enableResidualTopUp);
@@ -79,6 +79,8 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
                       std::vector<std::vector<int32_t>>,
                       std::vector<int32_t>,
                       std::vector<torch::Tensor>>(),
+             // Constructor argument conversion happens before the C++ body; the
+             // body only initializes PPR state and can run without the GIL.
              py::call_guard<py::gil_scoped_release>())
         .def("drain_queue", gigl::drainQueueWrapper)
         .def("push_residuals", gigl::pushResidualsWrapper)
--- a/gigl-core/core/sampling/python_ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/python_ppr_forward_push.cpp
@@ -8,6 +8,7 @@
 #include <torch/extension.h>
 
 #include <cstdint>
+#include <optional>
 #include <tuple>
 #include <unordered_map>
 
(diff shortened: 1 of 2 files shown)


gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -42,6 +43,29 @@ static void pushResidualsWrapper(PPRForwardPush& state, const py::dict& fetchedB
     }
 }
 
+static std::optional<std::unordered_map<int32_t, torch::Tensor>> drainQueueWrapper(PPRForwardPush& state) {
+    std::optional<std::unordered_map<int32_t, torch::Tensor>> drained;
+    // The drain mutates only this PPRForwardPush instance and returns torch tensors.
+    // No Python objects are touched until pybind converts the result after return.
```
kmontemayor2-sc: btw do we know if pybind tensor conversions are cheap? Something else to look into? If we're copying the underlying memory that could be very expensive.

gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -54,11 +78,12 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
                       double,
                       std::vector<std::vector<int32_t>>,
                       std::vector<int32_t>,
-                      std::vector<torch::Tensor>>())
-        .def("drain_queue", &gigl::PPRForwardPush::drainQueue)
+                      std::vector<torch::Tensor>>(),
+             py::call_guard<py::gil_scoped_release>())
```
kmontemayor2-sc, line 84: qq does init need the gil release? Or we we just doing this for completion?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -285,7 +283,7 @@ async def _batch_fetch_neighbors(
             edge_type_ids.append(edge_type_id)
             sample_tasks.append(
                 self._sample_one_hop(
-                    srcs=nodes_by_edge_type_id[edge_type_id].to(device),
```
kmontemayor2-sc, line 288: Why do we remove this? Is it just not needed?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -445,7 +446,9 @@ async def _compute_ppr_scores(
             # means all drained nodes either had cached neighbors or no outgoing
             # edges — we still call push_residuals to flush their residuals into
             # ppr_scores_.
-            nodes_by_edge_type_id = ppr_state.drain_queue()
+            nodes_by_edge_type_id = await loop.run_in_executor(
+                None, ppr_state.drain_queue
+            )
```
kmontemayor2-sc, line 459: Sorry for my own understanding, why do we need to run this in the executor? Shouldn't just using the gil release be sufficient here? Have we tested this without `run_in_executor` but with GIL release? (and also preferable `run_in_executor` *without* GIL release?

I guess I'm confused how calling this method and then immediately awaiting gets us a speedup (vs just the gil release?)

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -455,25 +458,27 @@ async def _compute_ppr_scores(
...
         # there is no remaining budget for residual candidates.
-        return self._extract_ppr_state_top_k(
+        return await loop.run_in_executor(
+            None,
+            self._extract_ppr_state_top_k,
             ppr_state,
             device,
-            max_ppr_nodes=self._max_ppr_nodes,
+            self._max_ppr_nodes,
         )
```
kmontemayor2-sc, line 490: Something else to look into (not required now) is that it's possible that all of this going into and out of the python / c++ boundaries is expensive. Have we considered the following schema?

1. Create some new `compute_ppr_scores` cpp *loop* that mirrors the logic here
2. Since we have this logic because of `_batch_fetch_neighbors` (or specifically `sample_one_hop` right?) We can pass some object into cpp and we just call `sample_one_hop` in the cpp code.

kmontemayor2-sc (commented):
Thanks Matt!

gigl-core/core/sampling/python_ppr_forward_push.cpp
mkolodner-sc: The pybind conversion for `torch::Tensor` should not copy tensor storage. It creates Python tensor container objects around the existing `torch::Tensor` handles. The data copy here is earlier, inside drainQueue(), where we materialize the frontier node IDs from C++ sets into CPU tensors. I added comments to make that distinction explicit. In practice, we've observed the memory of these CPU tensors to not be significant on internal benchmarking.

gigl-core/core/sampling/python_ppr_forward_push.cpp
mkolodner-sc: It can still be useful to guarantee that the initialization of a PPRState object on a fresh batch isn't blocking the event loop for a different concurrent worker.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Yes it is not needed. These come directly from drainQueue(), which currently creates CPU torch::kLong tensors. _sample_one_hop can consume those CPU ID tensors directly, so the old .to(device) was an avoidable CPU→sampler-device transfer before the fetch.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: `run_in_executor` and GIL release help at different layers. `run_in_executor` keeps the asyncio event-loop thread from sitting inside a synchronous C++ call, so other RPC completions/coroutines can be scheduled while PPR work runs in a worker thread. py::gil_scoped_release makes sure that worker thread does not hold the Python GIL while it is doing C++ work, which lets the event-loop thread actually execute Python callbacks/coroutines while the worker is busy.

For example, if push_residuals takes 50ms and we call it directly, the event loop is stuck inside C++ for that full 50ms, even if the GIL is released. If we only use run_in_executor, the event-loop thread is free, but the worker may still hold the GIL and delay Python callbacks. With both, the 50ms C++ push runs in a worker thread without holding the GIL, so the event loop can keep processing RPC completions and scheduling the next PPR work.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: > it's possible that all of this going into and out of the python / c++ boundaries is expensive

I've done some internal benchmarking on this serialization cost explicitly and the latency is not significant from this. Agreed that this can be investigated in the future though if it becomes an issue -- I've added a TODO here

gigl-core/core/sampling/python_ppr_forward_push.cpp
kmontemayor2-sc: is the creation of the object expensive? It's just setting up some empty collections right?

I don't think this is blocking but I am a bit curious if we suspect this is a crucial fix

gigl/distributed/dist_ppr_sampler.py
kmontemayor2-sc: oh good catch, we do the `to(device)` in collate_fn right?

gigl/distributed/dist_ppr_sampler.py
kmontemayor2-sc: hmmm, so the `await loop.run_in_executor` blocks *this* thread, but we may have other threads (in the same python process) that would get freed up?

I suspect this pattern may be useful elsewhere in graph store but I'll need to think about that :)

kmontemayor2-sc (approved):
LGTM! Nice work Matt :)

Do you think there's similar low-hanging (but hard to find!) fruit here for the khop sampler? 

E.g. is it possible to put this bit in the background? https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/distributed/dist_neighbor_sampler.py#L297

Or do we need to also write a CPP version of that `_sample_from_nodes` so we can force-detach the gil? May not be difficult? WDYT?

yliu2-sc (approved):

gigl-core/core/sampling/python_ppr_forward_push.cpp
mkolodner-sc: I don't have the exact numbers for the savings from this change in particular, but I don't believe this is a crucial step relative to moving other items off the event loop in this PR (such as drain_queue), as it isn't part of the hot loop that is being called multiple times per batch. Still, it'd be good to ensure that this creation doesn't block event loop for other concurrent threads that might be trying to do fetches and whatnot at the same time this is being called.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Yes I believe so based on [1](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9/graphlearn_torch/python/distributed/dist_loader.py#L359)

PR #709, opened 2026-07-15
Add typed-channel PPR sampler

mkolodner-sc:
## Summary

Adds the Python/API layer for typed-channel PPR sampling.

This PR is stacked on #724, which contains the C++ typed PPR drain and extraction implementation. This layer exposes typed-channel PPR through sampler options and wires it into `DistPPRNeighborSampler`.

## Motivation

A collapsed heterogeneous PPR traversal can identify relevant neighbors, but it loses which relation paths contributed to that relevance. Typed-channel PPR keeps that information by running PPR over configured edge-type channels and emitting channel-aware edge attributes.

Downstream consumers can use:

- the best calibrated PPR score
- per-channel calibrated scores
- per-channel presence indicators

## What Changed

- Keeps `max_ppr_nodes` as the per-seed PPR sequence cap.
  - For untyped PPR, this preserves the existing behavior.
  - For typed PPR, this remains the total per-seed cap across all channels.

- Adds `typed_channel_ratios` to `PPRSamplerOptions`.
  - If omitted, the sampler uses the regular untyped PPR path.
  - Keys can be a single canonical edge type, e.g. `("user", "views", "item")`.
  - Keys can also be a tuple of canonical edge types to define one grouped traversal channel.
  - Values are ratios in `(0, 1]` and must sum to `1.0`.

- Converts typed-channel ratios to integer channel target counts during sampler construction.
  - For example, `max_ppr_nodes=200` and ratios `{views: 0.6, likes/shares: 0.4}` targets 120 and 80 nodes.
  - Fractional counts are rounded deterministically while preserving the total `max_ppr_nodes` cap.

- Adds typed-channel parsing, validation, and traversal-map construction.
  - Restricts typed PPR to heterogeneous sampling.
  - Rejects invalid channel keys and invalid ratio values.
  - Builds per-channel edge traversal allowlists for the C++ PPR states.

- Wires typed PPR execution into the distributed sampler.
  - Builds one `PPRForwardPush` state per typed channel.
  - Each state traverses only the edge types configured for that channel.
  - Calls the typed drain and extraction APIs from #724.
  - Keeps typed PPR work off the asyncio event-loop thread where applicable.

- Emits typed PPR edge attributes with shape `[N, 1 + 2 * num_channels]`:
  - column 0: best score across channels
  - next `num_channels` columns: per-channel scores
  - final `num_channels` columns: per-channel presence bits

## Tests

Added focused Python coverage for:

- typed-channel ratio parsing and validation
- ratio-to-target-count conversion
- traversal-map construction
- typed sampler option wiring
- disabling residual top-up for typed PPR
- end-to-end typed sampler output shape

### DIFF
--- a/gigl/distributed/dist_ppr_sampler.py
+++ b/gigl/distributed/dist_ppr_sampler.py
@@ -90,7 +90,7 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
     exposes the next frontier, Python performs the distributed neighbor fetch,
     ``push_residuals`` updates the state, and C++ extraction emits the final
     top-k plus residual top-up output. Typed PPR runs one ``PPRForwardPush``
-    state per traversal channel. Its typed drain step unions the channel
+    traversal per configured channel. Its typed drain step unions the channel
     frontiers before the same distributed fetch, pushes results back into the
     active channel states, and then uses one typed C++ extraction step to apply
     channel target counts, deduplicate shared candidates, and emit the final
@@ -111,11 +111,11 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
         - ``data[(seed_type, "ppr", neighbor_type)].edge_index``: same format as above.
         - ``data[(seed_type, "ppr", neighbor_type)].edge_attr``: scalar PPR
           score for regular PPR. For typed PPR, edge attrs are multi-column:
-          ``[best_calibrated_score, calibrated_channel_scores..., channel_presence_bits...]``.
-          Typed-PPR scores are calibrated within each channel/seed pool and
-          globally ranked by the best calibrated score. Channel columns follow
-          the insertion order of ``typed_channel_ratios``. Column 0 is the
-          scalar best score for consumers that need a single PPR weight.
+          ``[best_score, channel_scores..., channel_presence_bits...]``.
+          Scores use the same PPR mass scale as regular scalar PPR output.
+          Channel columns follow the insertion order of
+          ``typed_channel_ratios``. Column 0 is the scalar best score for
+          consumers that need a single PPR weight.
 
     Args:
         alpha: Restart probability (teleport probability back to seed). Higher values
@@ -133,22 +133,23 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
             scores are available.
         num_neighbors_per_hop: Maximum number of neighbors to fetch per hop.
         typed_channel_ratios: Optional target proportions for typed PPR
-            traversal channels. If not provided, PPR uses the regular untyped
-            path: each state may traverse all eligible edge types for the
-            current node type and emits scalar PPR scores without channel
-            attribution.
+            traversal channels. If not provided, PPR treats all eligible edge
+            types as one shared traversal space and emits a single scalar PPR
+            score per output row.
             Keys may be either a single canonical edge type
             ``(src_type, relation, dst_type)`` or a tuple of canonical edge
-            types. Each key defines one traversal channel whose PPR state may
-            traverse only those exact edge types.
+            types. Each key defines one traversal channel that may use only
+            those exact edge types. Edge types may appear in multiple channels
+            when those channels intentionally overlap.
             Values are positive ratios that must sum to ``1.0``. The sampler
             converts ratios to per-channel target counts from ``max_ppr_nodes``.
             Finalized PPR candidates and residual top-up candidates both obey
             these target counts. If the same node appears in multiple channels,
-            it is attributed to the channel where it has the highest calibrated
-            score. If sparse channels or duplicate nodes leave unused target
-            slots, the remaining slots are redistributed globally by score so
-            the returned sequence can still fill up to ``max_ppr_nodes``.
+            it is attributed to the channel where it has the highest emitted
+            PPR score for that seed. If sparse channels or duplicate nodes
+            leave unused target slots, the remaining slots are redistributed
+            globally by score so the returned sequence can still fill up to,
+            but never exceed, ``max_ppr_nodes``.
             Example::
 
                 typed_channel_ratios = {
@@ -163,12 +164,12 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
             attributed to the views channel and 80 nodes attributed to the
             grouped likes/shares channel. The views channel traverses only
             ``("user", "views", "item")`` edges. The grouped likes/shares
-            channel traverses either likes or shares edges in one PPR state.
+            channel traverses either likes or shares edges as one channel.
             Both finalized PPR rows and residual top-up rows fill those same
             targets. The targets are best-effort rather than strict per-seed
             guarantees: if a channel cannot provide enough unique candidates,
             unused slots are filled by the remaining highest-scoring candidates
-            from any channel.
+            from any channel, up to ``max_ppr_nodes`` total rows.
 
         degree_tensors: Pre-computed total-degree tensors (int32). Homogeneous
             graphs use a single tensor; heterogeneous graphs use tensors keyed
@@ -639,7 +640,7 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
         dict[NodeType, torch.Tensor],
         dict[NodeType, torch.Tensor],
     ]:
-        """Run one PPR state per typed channel and extract the merged result.
+        """Run one PPR traversal per typed channel and extract the merged result.
 
         Each channel receives the same seed nodes but a different edge-type
         traversal allowlist. Fetch frontiers are unioned across active channels
--- a/gigl/distributed/dist_ppr_sampler.py
+++ b/gigl/distributed/dist_ppr_sampler.py
@@ -21,6 +21,13 @@ from graphlearn_torch.typing import EdgeType, NodeType
 from graphlearn_torch.utils import merge_dict
 
 from gigl.distributed.base_sampler import BaseDistNeighborSampler
+from gigl.distributed.utils.dist_typed_sampler import (
+    TypedPPRChannelKey,
+    TypedPPRChannelTraversalMaps,
+    build_edge_type_channel_group_edge_type_ids,
+    compute_typed_channel_target_counts,
+    parse_typed_channel_ratio_groups,
+)
 from gigl.types.graph import DEFAULT_HOMOGENEOUS_NODE_TYPE, is_label_edge_type
 
 # Trailing "." is an intentional separator.  These constants are used both to
@@ -39,6 +46,21 @@ _PPR_HOMOGENEOUS_EDGE_TYPE = (
     "to",
     DEFAULT_HOMOGENEOUS_NODE_TYPE,
 )
+
+# C++ PPR extraction output: flat node IDs, flat weights, and per-seed valid
+# counts. Homogeneous extraction uses tensors directly; heterogeneous extraction
+# uses dictionaries keyed by node type.
+PPRResult = tuple[
+    Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+    Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+    Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+]
+# Heterogeneous-only view of PPRResult after typed PPR extraction.
+HeteroPPRResult = tuple[
+    dict[NodeType, torch.Tensor],
+    dict[NodeType, torch.Tensor],
+    dict[NodeType, torch.Tensor],
+]
 
 
 class DistPPRNeighborSampler(BaseDistNeighborSampler):
@@ -87,7 +109,13 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
     **Heterogeneous (HeteroData)** — one PPR edge type per
     ``(seed_type, neighbor_type)`` pair, with ``"ppr"`` as the relation:
         - ``data[(seed_type, "ppr", neighbor_type)].edge_index``: same format as above.
-        - ``data[(seed_type, "ppr", neighbor_type)].edge_attr``: same format as above.
+        - ``data[(seed_type, "ppr", neighbor_type)].edge_attr``: scalar PPR
+          score for regular PPR. For typed PPR, edge attrs are multi-column:
+          ``[best_calibrated_score, calibrated_channel_scores..., channel_presence_bits...]``.
+          Typed-PPR scores are calibrated within each channel/seed pool and
+          globally ranked by the best calibrated score. Channel columns follow
+          the insertion order of ``typed_channel_ratios``. Column 0 is the
+          scalar best score for consumers that need a single PPR weight.
 
     Args:
         alpha: Restart probability (teleport probability back to seed). Higher values
@@ -104,6 +132,43 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
             during Forward Push when fewer than ``max_ppr_nodes`` finalized PPR
             scores are available.
         num_neighbors_per_hop: Maximum number of neighbors to fetch per hop.
+        typed_channel_ratios: Optional target proportions for typed PPR
+            traversal channels. If not provided, PPR uses the regular untyped path: each
+            state may traverse all eligible edge types for the current node
+            type and emits scalar PPR scores without channel attribution.
+            Keys may be either a single canonical edge type
+            ``(src_type, relation, dst_type)`` or a tuple of canonical edge
+            types. Each key defines one traversal channel whose PPR state may
+            traverse only those exact edge types.
+            Values are positive ratios that must sum to ``1.0``. The sampler
+            converts ratios to per-channel target counts from ``max_ppr_nodes``.
+            Finalized PPR candidates and residual top-up candidates both obey
+            these target counts. If the same node appears in multiple channels,
+            it is attributed to the channel where it has the highest calibrated
+            score. If sparse channels or duplicate nodes leave unused target
+            slots, the remaining slots are redistributed globally by score so
+            the returned sequence can still fill up to ``max_ppr_nodes``.
+            Example::
+
+                typed_channel_ratios = {
+                    ("user", "views", "item"): 0.6,
+                    (
+                        ("user", "likes", "item"),
+                        ("user", "shares", "item"),
+                    ): 0.4,
+                }
+
+            With ``max_ppr_nodes=200``, this example targets 120 nodes
+            attributed to the views channel and 80 nodes attributed to the
+            grouped likes/shares channel. The views channel traverses only
+            ``("user", "views", "item")`` edges. The grouped likes/shares
+            channel traverses either likes or shares edges in one PPR state.
+            Both finalized PPR rows and residual top-up rows fill those same
+            targets. The targets are best-effort rather than strict per-seed
+            guarantees: if a channel cannot provide enough unique candidates,
+            unused slots are filled by the remaining highest-scoring candidates
+            from any channel.
+
         degree_tensors: Pre-computed total-degree tensors (int32). Homogeneous
             graphs use a single tensor; heterogeneous graphs use tensors keyed
             by NodeType. The colocated and graph-store loader paths retrieve
@@ -121,6 +186,7 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
         num_neighbors_per_hop: int = 100_000,
         degree_tensors: Union[torch.Tensor, dict[NodeType, torch.Tensor]],
         max_fetch_iterations: Optional[int] = None,
+        typed_channel_ratios: Optional[dict[TypedPPRChannelKey, float]] = None,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
@@ -162,7 +228,22 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
             ]
             self._is_homogeneous = True
 
-        self._typed_ppr_channel_target_counts: Optional[list[int]] = None
+        typed_channel_groups, typed_channel_ratio_list = (
+            parse_typed_channel_ratio_groups(typed_channel_ratios)
+        )
+        self._typed_ppr_channel_target_counts = (
+            compute_typed_channel_target_counts(
+                typed_channel_ratio_list,
+                max_ppr_nodes,
+            )
+            if typed_channel_ratio_list is not None
+            else None
+        )
+        if self._typed_ppr_channel_target_counts is not None:
+            if self._is_homogeneous:
+                raise ValueError(
+                    "Typed PPR channel ratios are only supported for heterogeneous PPR sampling."
+                )
 
         # Convert the public homogeneous/heterogeneous degree-tensor shape to
         # the node-type keyed form used internally by PPR.
@@ -226,9 +307,16 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
             for node_type in all_node_types
         ]
 
-        self._typed_ppr_channel_to_node_type_id_to_edge_type_ids: list[
-            list[list[int]]
-        ] = []
+        self._typed_ppr_channel_to_node_type_id_to_edge_type_ids: TypedPPRChannelTraversalMaps = []
+        if typed_channel_groups is not None:
+            self._typed_ppr_channel_to_node_type_id_to_edge_type_ids = (
+                build_edge_type_channel_group_edge_type_ids(
+                    edge_type_groups=typed_channel_groups,
+                    edge_type_to_edge_type_id=self._etype_to_etype_id,
+                    node_type_to_edge_types=self._node_type_to_edge_types,
+                    node_types=self._ntype_id_to_ntype,
+                )
+            )
 
     def _convert_degree_tensors_to_dict(
         self,
--- a/gigl/distributed/dist_ppr_sampler.py
+++ b/gigl/distributed/dist_ppr_sampler.py
@@ -22,7 +22,7 @@ from graphlearn_torch.utils import merge_dict
 
 from gigl.distributed.base_sampler import BaseDistNeighborSampler
 from gigl.distributed.utils.dist_typed_sampler import (
-    PPRSequenceLength,
+    MaxPPRNodes,
     TypedPPRChannelTraversalMaps,
     build_edge_type_channel_group_edge_type_ids,
     parse_typed_channel_target_groups,
@@ -113,22 +113,21 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
           ``[best_calibrated_score, calibrated_channel_scores..., channel_presence_bits...]``.
           Typed-PPR scores are calibrated within each channel/seed pool and
           globally ranked by the best calibrated score. Channel columns follow
-          the insertion order of the typed sequence mapping. Column 0 is the
-          scalar best score for consumers that need a single PPR weight.
+          the insertion order of the typed ``max_ppr_nodes`` mapping. Column 0
+          is the scalar best score for consumers that need a single PPR weight.
 
     Args:
         alpha: Restart probability (teleport probability back to seed). Higher values
                keep samples closer to seeds. Typical values: 0.15-0.25.
         eps: Convergence threshold. Smaller values give more accurate PPR scores
              but require more computation. Typical values: 1e-4 to 1e-6.
-        ppr_sequence_length: PPR output sequence specification. Pass an integer
-            to run regular untyped PPR and return up to that many nodes per
-            seed. Pass a mapping from typed channel key to integer target count
-            to run typed PPR; the sum of target counts is the total sequence
-            length.
+        max_ppr_nodes: PPR output sequence specification. Pass an integer to
+            run regular untyped PPR and return up to that many nodes per seed.
+            Pass a mapping from typed channel key to integer target count to
+            run typed PPR; the sum of target counts is the per-seed output cap.
         enable_residual_topup: Whether to include residual candidates discovered
-            during Forward Push when fewer than the requested sequence length
-            finalized PPR scores are available.
+            during Forward Push when fewer than the requested ``max_ppr_nodes``
+            output slots have finalized PPR scores.
         num_neighbors_per_hop: Maximum number of neighbors to fetch per hop.
         degree_tensors: Pre-computed total-degree tensors (int32). Homogeneous
             graphs use a single tensor; heterogeneous graphs use tensors keyed
@@ -142,7 +141,7 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
         *args,
         alpha: float = 0.5,
         eps: float = 1e-4,
-        ppr_sequence_length: PPRSequenceLength = 50,
+        max_ppr_nodes: MaxPPRNodes = 50,
         enable_residual_topup: bool = True,
         num_neighbors_per_hop: int = 100_000,
         degree_tensors: Union[torch.Tensor, dict[NodeType, torch.Tensor]],
@@ -189,26 +188,26 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
 
         typed_channel_groups: Optional[list[tuple[EdgeType, ...]]] = None
         self._typed_ppr_channel_target_counts: Optional[list[int]] = None
-        if isinstance(ppr_sequence_length, bool):
+        if isinstance(max_ppr_nodes, bool):
             raise ValueError(
-                "ppr_sequence_length must be an integer or typed channel mapping."
+                "max_ppr_nodes must be an integer or typed channel mapping."
             )
-        if isinstance(ppr_sequence_length, int):
-            if ppr_sequence_length < 0:
+        if isinstance(max_ppr_nodes, int):
+            if max_ppr_nodes < 0:
                 raise ValueError(
-                    f"ppr_sequence_length must be non-negative, got {ppr_sequence_length}."
+                    f"max_ppr_nodes must be non-negative, got {max_ppr_nodes}."
                 )
-            self._max_ppr_nodes = ppr_sequence_length
-        elif isinstance(ppr_sequence_length, dict):
+            self._max_ppr_nodes = max_ppr_nodes
+        elif isinstance(max_ppr_nodes, dict):
             typed_channel_groups, typed_channel_target_counts = (
-                parse_typed_channel_target_groups(ppr_sequence_length)
+                parse_typed_channel_target_groups(max_ppr_nodes)
             )
             self._typed_ppr_channel_target_counts = typed_channel_target_counts
             self._max_ppr_nodes = sum(typed_channel_target_counts)
         else:
             raise ValueError(
-                "ppr_sequence_length must be an integer or typed channel mapping, "
-                f"got {ppr_sequence_length!r}."
+                "max_ppr_nodes must be an integer or typed channel mapping, "
+                f"got {max_ppr_nodes!r}."
             )
         if self._typed_ppr_channel_target_counts is not None:
             if self._is_homogeneous:
--- a/gigl/distributed/dist_ppr_sampler.py
+++ b/gigl/distributed/dist_ppr_sampler.py
@@ -22,10 +22,11 @@ from graphlearn_torch.utils import merge_dict
 
 from gigl.distributed.base_sampler import BaseDistNeighborSampler
 from gigl.distributed.utils.dist_typed_sampler import (
-    MaxPPRNodes,
+    TypedPPRChannelKey,
     TypedPPRChannelTraversalMaps,
     build_edge_type_channel_group_edge_type_ids,
-    parse_typed_channel_target_groups,
+    compute_typed_channel_target_counts,
+    parse_typed_channel_ratio_groups,
 )
 from gigl.types.graph import DEFAULT_HOMOGENEOUS_NODE_TYPE, is_label_edge_type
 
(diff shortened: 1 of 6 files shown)


gigl/distributed/sampler_options.py
```diff
@@ -75,6 +80,33 @@ class PPRSamplerOptions:
             The algorithm still runs to convergence — re-enqueued nodes propagate
             through cached neighbors at negligible cost. ``None`` (default) means
             no fetch limit.
+        typed_channel_quotas: Optional top-k quotas for typed PPR
+            traversal channels defined by canonical edge-type allowlists. Keys
+            may be either a single canonical edge type
+            ``(src_type, relation, dst_type)`` or a tuple of canonical edge
+            types. Each key defines one traversal channel whose PPR state may
```
kmontemayor2-sc: BTW any reason we chose the "typed channel" moniker vs "meta path" which I think I've seen to describe this "sequence of edge types" elsewhere?

gigl/distributed/sampler_options.py
```diff
@@ -75,6 +80,33 @@ class PPRSamplerOptions:
...
+            traverse only those exact edge types.
+            Channel order follows the insertion order of this mapping. Each
+            channel may contribute up to its quota to the candidate pool, while
+            the final returned sequence remains capped by ``max_ppr_nodes``.
+            Quotas may sum above ``max_ppr_nodes`` to give sparse or
+            overlapping channels room to fill the sequence.
```
kmontemayor2-sc: Sorry what does this mean? Can we give examples?

gigl/distributed/sampler_options.py
```diff
@@ -75,6 +80,33 @@ class PPRSamplerOptions:
...
+            Example::
+
+                typed_channel_quotas = {
+                    ("user", "views", "item"): 64,
+                    (
+                        ("user", "likes", "item"),
+                        ("user", "shares", "item"),
+                    ): 32,
+                }
```
kmontemayor2-sc, line 114: Can we likewise explain what this means?

gigl/distributed/utils/distributed_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+# channel containing multiple edge types.
+TypedPPRChannelKey = Union[EdgeType, tuple[EdgeType, ...]]
+"""A typed-PPR traversal channel key.
+
+A single canonical edge type creates one channel restricted to that edge type.
+A tuple of canonical edge types creates one channel whose forward-push state may
+traverse any edge type in the group. When typed PPR emits multi-column
+``edge_attr`` tensors, channel columns follow the insertion order of the
+``typed_channel_quotas`` mapping.
+"""
```
kmontemayor2-sc: Does this comment below to `TypedPPRChannelKey`? Can we add a new line here so it's "more obvious" what the comment is describing?

gigl/distributed/utils/distributed_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
+"""Construction-time typed-PPR option parsing for distributed samplers.
+
+The helpers in this module validate ``typed_channel_quotas`` and convert public
+edge-type keys into the compact integer traversal maps consumed by the C++
+forward-push kernel. They run once during ``DistPPRNeighborSampler``
+initialization, not in the per-batch PPR sampling hot loop.
+"""
+
+from typing import Optional, Sequence, Union, cast
```
kmontemayor2-sc: nit. Any reason we can't `from collections.abc import Sequence`?

gigl/distributed/sampler_options.py
```diff
@@ -75,6 +80,33 @@ class PPRSamplerOptions:
             The algorithm still runs to convergence — re-enqueued nodes propagate
             through cached neighbors at negligible cost. ``None`` (default) means
             no fetch limit.
+        typed_channel_quotas: Optional top-k quotas for typed PPR
```
kmontemayor2-sc: Should we also add some note saying if not provided will collapse the graph to be "homogeneous" or something?

gigl/distributed/utils/distributed_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    if not typed_channel_quotas:
+        return None, None
+
+    typed_channel_groups: TypedPPRChannelEdgeTypeGroups = []
+    typed_channel_quota_list: list[int] = []
+    invalid_quotas: dict[TypedPPRChannelKey, object] = {}
+
+    def is_canonical_edge_type(value: object) -> bool:
+        """Return whether ``value`` has GraphLearn's canonical EdgeType shape."""
```
kmontemayor2-sc: more like PyG edge type right?

gigl/distributed/utils/distributed_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+        return (
+            isinstance(value, tuple)
+            and len(value) == 3
+            and all(isinstance(part, str) for part in value)
+        )
+
+    for edge_type_key, quota in typed_channel_quotas.items():
+        if not isinstance(quota, int) or isinstance(quota, bool) or quota <= 0:
+            invalid_quotas[edge_type_key] = quota
```
kmontemayor2-sc: we shouldn't need this check should we? Type check should catch this?

gigl/distributed/utils/distributed_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+    return typed_channel_groups, typed_channel_quota_list
+
+
+def build_edge_type_channel_group_edge_type_ids(
+    edge_type_groups: TypedPPRChannelEdgeTypeGroups,
+    edge_type_to_edge_type_id: dict[EdgeType, int],
+    node_type_to_edge_types: dict[NodeType, list[EdgeType]],
+    node_types: Sequence[NodeType],
+) -> list[list[list[int]]]:
```
kmontemayor2-sc: Can we explain what this triple list is better?

gigl/distributed/utils/dist_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+    known_edge_types = set(edge_type_to_edge_type_id.keys())
+    channel_edge_type_ids_by_node_type: list[list[list[int]]] = []
+    for channel_edge_types in edge_type_groups:
+        unknown_edge_types = set(channel_edge_types) - known_edge_types
+        if unknown_edge_types:
+            raise ValueError(
+                "typed_channel_quotas includes non-traversable edge types "
+                f"{sorted(unknown_edge_types)!r}. Edge types must exist in the "
+                "graph and must not be label edge types."
+            )
```
kmontemayor2-sc, line 199: shall we also add the known edge types here?

kmontemayor2-sc: Also we don't enforce the label edge type do we?

gigl/distributed/utils/distributed_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+
+        channel_edge_type_set = set(channel_edge_types)
+        node_type_id_to_channel_edge_type_ids = [
+            [
+                edge_type_to_edge_type_id[edge_type]
+                for edge_type in node_type_to_edge_types.get(node_type, [])
+                if edge_type in channel_edge_type_set
+            ]
+            for node_type in node_types
+        ]
```
kmontemayor2-sc: style preference: Can we not have the double nested comprehension like this? I find them a bit hard to read?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -473,20 +556,125 @@ async def _compute_ppr_scores(
...
+                    self._alpha,
+                    self._requeue_threshold_factor,
+                    node_type_id_to_edge_type_ids,
+                    self._edge_type_id_to_dst_ntype_id,
+                    self._degree_tensors_for_cpp,
+                )
+                for node_type_id_to_edge_type_ids in (
+                    self._typed_ppr_channel_to_node_type_id_to_edge_type_ids
+                )
+            ]
```
kmontemayor2-sc, line 687: Hmmm, do you think it'd be better to build each state in it's own executor? e.g. return some list of asyncio futures?

kmontemayor2-sc: Also can we add a return type?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -473,20 +556,125 @@ async def _compute_ppr_scores(
...
+            ] = [dict() for _ in ppr_states]
+
+            if unioned_node_ids_by_edge_type_id:
+                union_fetched_by_edge_type_id = await self._batch_fetch_neighbors(
+                    unioned_node_ids_by_edge_type_id
+                )
+                for channel_index, edge_type_ids in zip(
+                    fetch_channel_indices,
+                    edge_type_ids_by_fetch_channel,
+                ):
```
kmontemayor2-sc, line 723: Are we concerned that these ever can have different lengths?

```
>>> for z in zip([1, 2, 3], ["a", "b"]): print(z)
...
(1, 'a')
(2, 'b')
```

kmontemayor2-sc: There is `strict=True` if we want to enforce https://docs.python.org/3/library/functions.html#zip

gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -66,6 +66,56 @@ extractTopKWithResidualTopUpWrapper(PPRForwardPush& state, int32_t maxPPRNodes,
...
+    // channel in a single sequenced loop iteration.
+    TypedPPRQueueDrainResult queueDrainResult;
+    {
+        py::gil_scoped_release release;
+        queueDrainResult = drainTypedPPRChannelQueues(statePtrs, fetchIterationCounts, maxFetchIterations);
+    }
+    return py::make_tuple(std::move(queueDrainResult.drainedChannelIndices),
+                          std::move(queueDrainResult.fetchChannelIndices),
+                          std::move(queueDrainResult.edgeTypeIdsByFetchChannel),
+                          std::move(queueDrainResult.unionedNodeIdsByEdgeTypeId));
```
kmontemayor2-sc: For my own learning? What does `std::move` do here? What does it mean to "move" something out of the cpp boundary?

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -25,6 +25,33 @@ struct SeedNodeTypeState {
...
 
+// Batched drain result for typed-PPR channels.
+//
+// Typed PPR keeps one PPRForwardPush state per channel.  During an iteration,
+// each channel drains its own queue, but Python should issue at most one shared
+// neighbor fetch per edge type. This struct carries both pieces of information:
+// which channel states still need pushResiduals(), and the unioned frontier to
+// fetch once for all channels that requested it.
+struct TypedPPRQueueDrainResult {
+    // Channels whose drainQueue() returned a value this iteration.
```
kmontemayor2-sc: QQ how do we identify these channel ids?

kmontemayor2-sc: Does this mean the ordering of the channels needs to be maintained?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -147,6 +147,60 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
...
+    TypedPPRQueueDrainResult queueDrainResult;
+    std::unordered_map<int32_t, std::unordered_set<int64_t>> unionedFrontierNodeIdsByEdgeTypeId;
+
+    // TODO: If typed queue draining becomes a measured blocker, evaluate
+    // parallelizing this channel loop with per-thread frontier maps and a
+    // deterministic merge into queueDrainResult.
+    for (size_t channelIndex = 0; channelIndex < states.size(); ++channelIndex) {
+        PPRForwardPush* state = states[channelIndex];
+
+        auto channelFrontierByEdgeTypeId = state->drainQueue();
```
kmontemayor2-sc: drainQ is expensive right? It does seem like this would have value in being parallelized?

Have we done any exploration here? I think it could be worth it.

kmontemayor2-sc: Another way to look at this is "how much slower is using typed channels vs the old approach, for heterogeneous graphs"?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -147,6 +147,60 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
...
+                                                    int32_t maxFetchIterations) {
+    TORCH_CHECK(states.size() == fetchIterationCounts.size(),
+                "Expected one fetch iteration count per PPR state, got ",
+                fetchIterationCounts.size(),
+                " counts for ",
+                states.size(),
+                " states.");
+
+    TypedPPRQueueDrainResult queueDrainResult;
+    std::unordered_map<int32_t, std::unordered_set<int64_t>> unionedFrontierNodeIdsByEdgeTypeId;
```
kmontemayor2-sc: Any reason we can't just have this be `map<NodeTypeId, set<NextNodesToFetch>>`?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -368,6 +422,211 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+//   maxScore: largest score in the channel, used to calibrate scores to [0, 1].
+//
+// Expected output: channelSelectionCandidates contains this channel's calibrated
+// candidates for later quota selection.
+static void addTypedPPRChannelCandidates(std::vector<std::pair<int32_t, double>>& channelSelectionCandidates,
+                                         const std::vector<std::pair<int32_t, double>>& nodesAndScores,
+                                         double maxScore) {
+    for (const auto& [nodeId, score] : nodesAndScores) {
+        double calibratedScore = maxScore > 0.0 ? score / maxScore : 0.0;
+        channelSelectionCandidates.emplace_back(nodeId, calibratedScore);
```
kmontemayor2-sc: when is maxScore 0?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -368,6 +422,211 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+//
+// Inputs:
+//   outputScoresByNodeId: mutable map from node ID to emitted edge_attr features.
+//   channelOutputCandidates: mutable sortable candidates for the fill pass.
+//   nodesAndScores: extracted (node_id, score) candidates for one channel.
+//   maxScore: largest score in the channel, used to calibrate scores to [0, 1].
+//   channelIndex: index of the typed PPR channel being added.
+//   numChannels: total typed PPR channels, used to derive feature width.
+//
+// Expected output: outputScoresByNodeId has this channel's score/presence
```
kmontemayor2-sc, line 473: Shall we have a standard of making output args the last in the arg list?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -368,6 +422,211 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+// calibrated candidates.
+static void addTypedPPRSeedFeaturesAndCandidates(std::unordered_map<int32_t, std::vector<double>>& outputScoresByNodeId,
+                                                 std::vector<std::pair<int32_t, double>>& channelOutputCandidates,
+                                                 const std::vector<std::pair<int32_t, double>>& nodesAndScores,
+                                                 double maxScore,
+                                                 int32_t channelIndex,
+                                                 int32_t numChannels) {
+    // Typed edge_attr rows store one best score across channels, followed by
+    // one score and one presence bit per channel.
+    int32_t numEdgeAttrFeatures = 1 + (2 * numChannels);
```
kmontemayor2-sc, line 491: qq why 1+?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -368,6 +422,211 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+
+    int32_t selectedReserveSize = std::min(maxPPRNodes, static_cast<int32_t>(globalCandidates.size()));
+    if (selectedReserveSize < static_cast<int32_t>(globalCandidates.size())) {
+        std::partial_sort(globalCandidates.begin(),
+                          globalCandidates.begin() + selectedReserveSize,
+                          globalCandidates.end(),
+                          higherGlobalCandidate);
+    } else if (globalCandidates.size() > 1) {
+        std::sort(globalCandidates.begin(), globalCandidates.end(), higherGlobalCandidate);
+    }
```
kmontemayor2-sc: qq why this is / else?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -368,6 +422,211 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+    // Move unique candidates into a vector so the final top-k ranking can use
+    // partial_sort. This avoids sorting the full candidate set when
+    // maxPPRNodes is smaller than the number of unique channel candidates.
+    std::vector<GlobalCandidate> globalCandidates;
+    globalCandidates.reserve(bestCandidateByNodeId.size());
+    for (const auto& candidateEntry : bestCandidateByNodeId) {
+        globalCandidates.push_back(candidateEntry.second);
+    }
+
+    int32_t selectedReserveSize = std::min(maxPPRNodes, static_cast<int32_t>(globalCandidates.size()));
```
kmontemayor2-sc: Side note: Could we get rid of some of these casts if we treated `maxPPRNodes` as `size_t` or whatever instead?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -411,6 +670,143 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
 }
 
+std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> extractTypedTopKWithResidualTopUp(
+    const std::vector<PPRForwardPush*>& states,
+    const std::vector<int32_t>& channelQuotas,
+    int32_t maxPPRNodes,
+    bool enableResidualTopUp) {
+    const auto* firstState = states.front();
+    int32_t batchSize = firstState->_batchSize;
+    int32_t numNodeTypes = firstState->_numNodeTypes;
```
kmontemayor2-sc, line 727: Does this assume that all states have the same num node types?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -411,6 +670,143 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
     return extractedPPRByNodeTypeId;
 }
 
+std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>> extractTypedTopKWithResidualTopUp(
```
kmontemayor2-sc: What is this return type?

kmontemayor2-sc: Also IIRC at some point you had some PR to create type aliases for collections, it may be useful here. Is it possible for us to get that in?

kmontemayor2-sc (commented):
Thanks Matt! Took a first pass here :)

gigl/distributed/sampler_options.py
mkolodner-sc: A meta path implies an ordered sequence of edge types usually to my understanding. In this case, the channels don't necessarily have an ordering -- they are just edges that we are permitted to traverse across for that forward push operation. As a result I've named them differently here.

gigl/distributed/sampler_options.py
mkolodner-sc: Added examples, this is saying quotas are candidate budgets, not guaranteed output counts. A channel may find fewer candidates than its quota, and multiple channels may find the same node, so the merged/deduped result can be smaller than the sum of quotas. We still cap final output at max_ppr_nodes, but allowing quotas to sum above that cap gives each channel room to contribute enough candidates before the global merge.

gigl/distributed/utils/distributed_typed_sampler.py
mkolodner-sc: Added named type aliases and expanded the docstring. The shape is now documented as channel_traversal_maps[channel_id][node_type_id] -> list[edge_type_id], i.e. for each typed channel, for each node type, which integer edge-type IDs that channel may traverse.

gigl/distributed/utils/dist_typed_sampler.py
mkolodner-sc: Yeah the label edge types have already been filtered out when building node_type_to_edge_types / edge_type_to_edge_type_id

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Added the return type. For the executor question, I’d keep building the list in a single executor call rather than scheduling one executor task per channel. This construction should be fairly lightweight, and per-channel tasks are only likely to help if state construction becomes a measured bottleneck and there are enough channels to amortize threadpool scheduling overhead. The current approach still gets the important benefit of moving construction off the event-loop thread, while keeping ordering and error handling simpler.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Added strict=True to enforce this, but they are produced together and should be aligned

gigl-core/core/sampling/python_ppr_forward_push.cpp
mkolodner-sc: `std::move` casts those temporary C++ containers to rvalues so pybind can consume/move from them instead of copying the vectors/maps. The moved-from C++ containers should not be used afterward. This does not copy tensor storage across the Python/C++ boundary; tensors remain reference-counted tensor objects. I added a comment there.

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: Channel IDs are positional indices into the typed channel state vector, aligned with channelQuotas and the Python typed_channel_quotas insertion order. I updated the struct comment to make that explicit. The C++ drain appends indices in ascending state-vector order, so ordering is stable.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: This makes sense, although I'd prefer we do this as a follow-up investigation if possible so we can land the core typed PPR changes first in its current state. I've added a TODO to look into this.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: I kept the unioned frontier keyed by edge type rather than node type because the distributed neighbor fetch is edge-type scoped. A node type can have multiple outgoing relations with different adjacency tables / destination types, so grouping only by node type would lose required fetch information.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: The representation is [best_score, channel_1 score, channel_2 score, ... channel_1 presence bit, channel 2 presence bit], so the 1 is for the best score.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Empty candidate lists never enter calibration, so a non-empty candidate list with maxScore <= 0 would indicate invalid PPR state rather than an expected edge case. I changed this to TORCH_CHECK before calibration and then divide directly by maxScore, so we fail loudly instead of silently emitting zeroed calibrated scores.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: partial_sort helps when the candidate set is larger than maxPPRNodes: we only need the sorted top-k prefix, and the discarded suffix can remain unordered. For example, if there are 1,000 candidates and maxPPRNodes=50, we only rank the first 50 that we will emit. If the candidate set is already within the cap, every candidate will be emitted, so there is no discarded suffix; in that case a regular sort is clearer and gives the full ordered output directly.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: I left maxPPRNodes as int32_t because that matches the existing API / pybind boundary and avoids signed/unsigned churn. Using size_t would remove a few casts but introduce more signedness friction around limits and negative checks.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes, typed extraction assumes all channel states share batch size and node-type schema. They are constructed from the same seed batch and graph schema; only the traversal allowlist differs. I added a comment at the firstState usage.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: The return type is now named PPRExtractResult: a map from integer node type ID to (ids, weights_or_edge_attr, valid_counts) tensors.

gigl/distributed/utils/dist_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+TypedPPRChannelEdgeTypeGroups = list[tuple[EdgeType, ...]]
+# One channel's traversal map. The outer list is indexed by integer node-type
+# ID; each inner list contains the integer edge-type IDs that channel may
+# traverse from that node type.
+TypedPPRChannelTraversalMap = list[list[int]]
+# All typed-channel traversal maps, ordered to match quota order.
+TypedPPRChannelTraversalMaps = list[TypedPPRChannelTraversalMap]
+
+
+def parse_typed_channel_quota_groups(
```
kmontemayor2-sc: btw should we add tets for this?

gigl/distributed/utils/dist_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+                "(src_type, relation, dst_type) or a non-empty tuple of "
+                f"canonical edge types, got {edge_type_key!r}."
+            )
+        typed_channel_groups.append(edge_types)
+        typed_channel_quota_list.append(quota)
+
+    return typed_channel_groups, typed_channel_quota_list
+
+
+def build_edge_type_channel_group_edge_type_ids(
```
kmontemayor2-sc, line 156: Ditto on the tests

tests/unit/distributed/dist_ppr_sampler_test.py
```diff
@@ -814,6 +935,10 @@ def test_ppr_sampler_destination_only_node_type(self) -> None:
         """Verify PPR output includes destination-only node types."""
         mp.spawn(fn=_run_ppr_destination_only_node_type, args=())
 
+    def test_typed_ppr_sampler_loader_outputs_channel_attrs(self) -> None:
+        """Verify typed PPR runs end-to-end through the loader."""
+        mp.spawn(fn=_run_typed_ppr_loader_shape_check, args=())
+
```
kmontemayor2-sc, line 937: Should this test be in this PR?

gigl/distributed/utils/dist_typed_sampler.py
kmontemayor2-sc: Oh are they in the `dist_ppr_sampler_test`? Should we move them to some `distributed_typed_sampler_test`?

gigl/distributed/utils/dist_typed_sampler.py
kmontemayor2-sc, line 1: Hmmm, should we rename to `dist` to be more in line with the other files?

gigl/distributed/utils/distributed_typed_sampler.py
```diff
@@ -0,0 +1,176 @@
...
+``edge_attr`` tensors, channel columns follow the insertion order of the
+``typed_channel_quotas`` mapping.
+"""
+# Parsed typed-channel edge-type allowlists, ordered to match quota order.
+TypedPPRChannelEdgeTypeGroups = list[tuple[EdgeType, ...]]
+# One channel's traversal map. The outer list is indexed by integer node-type
+# ID; each inner list contains the integer edge-type IDs that channel may
+# traverse from that node type.
+TypedPPRChannelTraversalMap = list[list[int]]
+# All typed-channel traversal maps, ordered to match quota order.
```
kmontemayor2-sc: What does "quota order" mean here?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -72,7 +95,13 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
     ``(seed_type, neighbor_type)`` pair, with ``"ppr"`` as the relation:
         - ``data[(seed_type, "ppr", neighbor_type)].edge_index``: same format as above.
-        - ``data[(seed_type, "ppr", neighbor_type)].edge_attr``: same format as above.
+        - ``data[(seed_type, "ppr", neighbor_type)].edge_attr``: scalar PPR
+          score for regular PPR. For typed PPR, edge attrs are multi-column:
+          ``[best_calibrated_score, calibrated_channel_scores..., channel_presence_bits...]``.
+          Typed-PPR scores are calibrated within each channel/seed pool and
+          globally ranked by the best calibrated score. Channel columns follow
+          the insertion order of ``typed_channel_quotas``. Graph Transformer
+          PPR sequence construction uses column 0 as the scalar ``ppr_weight``.
```
kmontemayor2-sc: BTW should we mention "Graph Transformer" here?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -89,6 +118,34 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
+                    ): 32,
+                }
+
+            This example creates two traversal channels. The views channel can
+            contribute up to 64 ranked candidates into the cross-channel merge.
+            The grouped likes/shares channel traverses either likes or shares
+            edges in one PPR state and can contribute up to 32 ranked candidates
+            from that combined pool. After channel candidates are merged and
+            deduplicated, at most ``max_ppr_nodes`` rows are emitted for each
+            seed.
```
kmontemayor2-sc: So is it possible here that the 33rd item in the second bucket `((user, likes, item), (user, shares, item))` would have a higher score than the first item in the first bucket `(user, views, item)`?

I suppose that's intended behavior?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -89,6 +118,34 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
+            produce duplicate nodes that are deduplicated before output.
+            Example::
+
+                typed_channel_quotas = {
+                    ("user", "views", "item"): 64,
+                    (
+                        ("user", "likes", "item"),
+                        ("user", "shares", "item"),
+                    ): 32,
+                }
```
kmontemayor2-sc: Hmmm, imagine we have some edge types like:

```
a -> to -> b
b -> to -> c
d -> to -> b
```

Then if we have:

```
typed_channel_quotas: {
    ((b -> to -> c), (d -> to -> b): 10
    (a -> to -> b): 1
}
```

With anchor nodes of type A, then we only get to the first quota `(b -> to -> c)` via the second? E.g. we do need to do `a -> b -> c`?

I'm not sure why this sticks out to me, but it seems like there's some interplay between the like "available paths" and "quota" that may be relavent here?

I'm sorry if that doesn't quite make sense, do you see what I'm getting at here?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -209,6 +276,21 @@ def __init__(
             for node_type in all_node_types
         ]
 
+        self._typed_ppr_channel_to_node_type_id_to_edge_type_ids: (
+            TypedPPRChannelTraversalMaps
+        )
```
kmontemayor2-sc: nit. set default = [] and then remove else brnach?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -473,20 +573,126 @@ async def _compute_ppr_scores(
...
+        Args:
+            seed_nodes: Global node IDs for the seed batch.
+            seed_node_type: Heterogeneous node type for ``seed_nodes``.
+            typed_ppr_channel_quotas: Per-channel candidate quotas, aligned
+                with ``self._typed_ppr_channel_to_node_type_id_to_edge_type_ids``.
+
+        Returns:
+            Heterogeneous PPR extraction output with typed edge-attribute
+            feature vectors.
+        """
```
kmontemayor2-sc, line 666: btw I was sort of hoping these changes could be in a separate, non-api focussed PR.

(e.g. we'd throw if the typed_channel gets provided when we have the API but not logic behind it).

gigl-core/core/sampling/python_ppr_forward_push.cpp
kmontemayor2-sc: I got the robots to do some digging, they say it's overall neutral, I have a minor preference of removing std::move if it's a no-op here

> std::move makes these expressions rvalues, but pybind11’s STL casters still materialize Python containers by iterating the C++ containers: see the [vector/list caster](https://github.com/pybind/pybind11/blob/v3.0.4/include/pybind11/stl.h#L342-L357) and [map/dict caster](https://github.com/pybind/pybind11/blob/v3.0.4/include/pybind11/stl.h#L270-L289). Rvalue forwarding can matter for movable custom element types, but not for these integer containers. PyTorch’s at::Tensor caster accepts [const at::Tensor&](https://github.com/pytorch/pytorch/blob/v2.8.0/torch/csrc/utils/pybind.h#L32-L47) and its [implementation wraps that reference](https://github.com/pytorch/pytorch/blob/v2.8.0/torch/csrc/utils.cpp#L329-L345), so std::move does not move the Tensor or copy its underlying storage. Therefore these four std::moves do not avoid copying or improve the generated conversion code here.

gigl-core/core/sampling/ppr_forward_push.h
kmontemayor2-sc: > typed channel state vector, aligned with channelQuotas and the Python typed_channel_quotas insertion order

Hmmm, do we ever need to worry about cross process serialization here? We had some issues with mangled order right? Was that just with proto?

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc: I do think this is important work btw - especially as our graphs get more heterogeneous.

Can you work with @svij-sc on trying to setup an experiment there?

kmontemayor2-sc: I don't think we need experiments to be done to merge this PR in, but I would prefer if we can start launching them :)

kmontemayor2-sc (commented):
Hi Matt, thanks for breaking up the PRs. 

Does this comment make any sense to you?  https://github.com/Snapchat/GiGL/pull/709/changes#r3677995032

I wonder if we want to think about the semantics here a bit more

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Sure, I'll migrate this to the other code which focuses on the C++/sampling logic changes.

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: I don’t think we need to worry about cross-process serialization here. TypedPPRQueueDrainResult is returned from C++ to Python through pybind in the same sampler process; it is not serialized through proto/RPC. The order-sensitive fields are std::vectors (drainedChannelIndices, fetchChannelIndices, edgeTypeIdsByFetchChannel), and pybind preserves vector/list order. Channel IDs are positional indices into the ppr_states / quota vectors, and C++ appends them in ascending channel index after the parallel drains complete.

The unordered fields (unionedNodeIdsByEdgeTypeId, and the node IDs inside each union) intentionally do not carry ordering semantics; they are keyed fetch inputs. The previous “mangled order” concern was with serialization/map ordering, not this in-process pybind return path.

tests/unit/distributed/dist_ppr_sampler_test.py
mkolodner-sc: I'd prefer to keep this here, since this requires E2E (including API) changes to work and this branches out of the CPP PR. I plan to squash this in to the CPP PR and merge them together anyway.

gigl/distributed/utils/distributed_typed_sampler.py
mkolodner-sc: Discussed offline, we've reframed quota to be a decimal between 0 and 1 per channel for how much of the final sequence length they should be responsible for populating

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Discussed offline, we've reframed quota to be a decimal between 0 and 1 per channel for how much of the final sequence length they should be responsible for populating

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -97,13 +125,51 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
             PPR nodes when finalized scores already fill the cap.
         enable_residual_topup: Whether to include residual candidates discovered
             during Forward Push when fewer than ``max_ppr_nodes`` finalized PPR
             scores are available.
         num_neighbors_per_hop: Maximum number of neighbors to fetch per hop.
+        typed_channel_ratios: Optional target proportions for typed PPR
+            traversal channels. If not provided, PPR uses the regular untyped
+            path: each state may traverse all eligible edge types for the
+            current node type and emits scalar PPR scores without channel
+            attribution.
```
kmontemayor2-sc: nit. A user doesn't know what "untyped" means here, or what a "state" is.

Should we rephrase as "If not set, then we will treat all edge types as the same and do PPR amongst then equally"?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -97,13 +125,51 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
+            Keys may be either a single canonical edge type
+            ``(src_type, relation, dst_type)`` or a tuple of canonical edge
+            types. Each key defines one traversal channel whose PPR state may
+            traverse only those exact edge types.
+            Values are positive ratios that must sum to ``1.0``. The sampler
+            converts ratios to per-channel target counts from ``max_ppr_nodes``.
+            Finalized PPR candidates and residual top-up candidates both obey
+            these target counts. If the same node appears in multiple channels,
+            it is attributed to the channel where it has the highest calibrated
+            score. If sparse channels or duplicate nodes leave unused target
```
kmontemayor2-sc: ditto on what a "calibrated score" is.

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -97,13 +125,51 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
+            the returned sequence can still fill up to ``max_ppr_nodes``.
+            Example::
+
+                typed_channel_ratios = {
+                    ("user", "views", "item"): 0.6,
+                    (
+                        ("user", "likes", "item"),
+                        ("user", "shares", "item"),
+                    ): 0.4,
+                }
```
kmontemayor2-sc, line 165: btw what if the same edge type exists in multiple channels? Is this behavior we want to support?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -97,13 +125,51 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
+
+            With ``max_ppr_nodes=200``, this example targets 120 nodes
+            attributed to the views channel and 80 nodes attributed to the
+            grouped likes/shares channel. The views channel traverses only
+            ``("user", "views", "item")`` edges. The grouped likes/shares
+            channel traverses either likes or shares edges in one PPR state.
```
kmontemayor2-sc: ditto on "state"

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -97,13 +125,51 @@ class DistPPRNeighborSampler(BaseDistNeighborSampler):
...
+            Both finalized PPR rows and residual top-up rows fill those same
+            targets. The targets are best-effort rather than strict per-seed
+            guarantees: if a channel cannot provide enough unique candidates,
+            unused slots are filled by the remaining highest-scoring candidates
```
kmontemayor2-sc, line 175: should we specify we never go over `max_ppr_nodes`?

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: I rephrased the public docs to avoid internal terms. It now says that when typed_channel_ratios is not set, PPR treats all eligible edge types as one shared traversal space and emits a scalar PPR score per output row. I avoided saying “equally” because PPR still follows the graph/degrees rather than assigning equal weight per edge type.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: I think overlapping channels are valid and should be supported. A repeated edge type can be intentional if the user wants overlapping traversal views. If the same output node appears through multiple channels, we deduplicate it and attribute it to the channel where it has the highest emitted PPR score.

For example, if we have channel 1 with A -> likes -> A, channel 2 with A -> follows -> A, we could have a channel 3 that traverses along both A -> likes A, A -> follows A, for a more homogeneous/theoretically grounded forward push signal.

kmontemayor2-sc (approved):
Nice work Matt, thanks :)

yliu2-sc (approved):

PR #711, opened 2026-07-20
retry: fail loud when deadline_s is used off the main thread

kmontemayor2-sc:
retry(deadline_s=...) enforces the deadline with a SIGALRM timeout, which CPython only permits in the main thread of the main interpreter. Reached from a worker thread it raised the opaque "signal only works in main thread of the main interpreter" ValueError from signal.signal.

Guard the deadline path: off the main thread, raise a RuntimeError that names the decorated function and points the caller at enforcing the deadline on the operation itself (e.g. a client-native timeout). Main-thread behavior is unchanged. Adds a test for the off-main-thread path.

This keeps the same behavior (failing if we use this var wrong) just provides better error mesasges :)

**Scope of work done**

### DIFF
--- a/gigl/common/utils/retry.py
+++ b/gigl/common/utils/retry.py
@@ -113,9 +113,9 @@ def retry(
                         f"retry(deadline_s=...) uses a SIGALRM timeout that only "
                         f"works on the main thread; {f.__module__}:{f.__name__} was "
                         f"is called from thread:  "
-                        f"{threading.current_thread().name!r}. Enforce "
-                        f"off-main-thread deadlines on the operation itself (e.g. a "
-                        f"client-native timeout)."
+                        f"{threading.current_thread().name!r}. "
+                        f"If using deadline_s ensure you are calling from main "
+                        f"thread, or set the deadline in the target function."
                     )
                 return timed_f_retry(*args, **kwargs)
 
--- a/gigl/common/utils/retry.py
+++ b/gigl/common/utils/retry.py
@@ -112,7 +112,7 @@ def retry(
                     raise RuntimeError(
                         f"retry(deadline_s=...) uses a SIGALRM timeout that only "
                         f"works on the main thread; {f.__module__}:{f.__name__} was "
-                        f"called from thread "
+                        f"is called from thread:  "
                         f"{threading.current_thread().name!r}. Enforce "
                         f"off-main-thread deadlines on the operation itself (e.g. a "
                         f"client-native timeout)."
--- a/gigl/common/utils/retry.py
+++ b/gigl/common/utils/retry.py
@@ -107,12 +107,7 @@ def retry(
             timed_f_retry = timeout(seconds=deadline_s)(f_retry)
 
             @wraps(f)
-            def f_retry_with_deadline(*args, **kwargs) -> T:  # type: ignore[type-var]
-                # deadline_s is enforced with a SIGALRM timeout, which CPython
-                # only permits in the main thread of the main interpreter. Off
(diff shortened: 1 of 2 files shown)


gigl/common/utils/retry.py
```diff
@@ -103,8 +104,22 @@ def fn(*args, **kwargs) -> T:
...
+
+            @wraps(f)
+            def f_retry_with_deadline(*args, **kwargs) -> T:
+                if threading.current_thread() is not threading.main_thread():
+                    raise RuntimeError(
+                        f"retry(deadline_s=...) uses a SIGALRM timeout that only "
+                        f"works on the main thread; {f.__module__}:{f.__name__} was "
+                        f"called from thread "
+                        f"{threading.current_thread().name!r}. Enforce "
+                        f"off-main-thread deadlines on the operation itself (e.g. a "
```
yliu2-sc: more details on where/how to?

gigl/common/utils/retry.py
```diff
@@ -103,8 +104,22 @@ def fn(*args, **kwargs) -> T:
...
-            return global_retry_timeout_decorator(f_retry)
+            timed_f_retry = timeout(seconds=deadline_s)(f_retry)
+
+            @wraps(f)
+            def f_retry_with_deadline(*args, **kwargs) -> T:
+                if threading.current_thread() is not threading.main_thread():
+                    raise RuntimeError(
+                        f"retry(deadline_s=...) uses a SIGALRM timeout that only "
+                        f"works on the main thread; {f.__module__}:{f.__name__} was "
+                        f"called from thread "
```
yliu2-sc:
```suggestion
                        f"is called from thread:  "
```

yliu2-sc (approved):

gigl/common/utils/retry.py
kmontemayor2-sc: Sure, re-worded:)

gigl/common/utils/retry.py
```diff
@@ -103,8 +104,22 @@ def fn(*args, **kwargs) -> T:
...
 
         if deadline_s is not None:
-            global_retry_timeout_decorator = timeout(seconds=deadline_s)
-            return global_retry_timeout_decorator(f_retry)
+            timed_f_retry = timeout(seconds=deadline_s)(f_retry)
+
+            @wraps(f)
+            def f_retry_with_deadline(*args, **kwargs) -> T:
+                if threading.current_thread() is not threading.main_thread():
+                    raise RuntimeError(
```
svij-sc, line 113: this reads as an odd RuntimeError.
Is this what we really want vs. somehting more descriptive of what the issue was? i.e. we failed __ after __x tries, n time, in ___ thread/process.

kmontemayor2-sc: This change isn't about the timeout or retry failing. The code (as it stands) currently fails, and I'm adding a new error message that's hopefully more descriptive. the issue here is that we can't use `deadline_s` on non-main threads (python only lets you use sigalarm on main thread). 

I can re-word this to be more descriptive.

svij-sc (approved):
Thanks LGTM.

Btw, I wonder if there are ways to simplify this function - seems to be growing quite complex.

kmontemayor2-sc:
> Btw, I wonder if there are ways to simplify this function - seems to be growing quite complex.

Probably but I think this is in large part because we have `deadline_s` here too - which is kind of a separate change. 

Ideally we could drop `deadline_s` and make a separate `@with_deadline` decorator?

PR #713, opened 2026-07-21
Add agent skills (humanify, ace-it) + git guidance; repair skill frontmatter

kmontemayor2-sc:
## What

  Adds and repairs agent guidance under `.claude/`:

  - **`.claude/development.md`** — two additions to the `## Git` section: never
    amend an already-pushed commit (add a new commit instead), and propagate fixes
    up a stacked branch chain with `git merge` (ask first), never by rebasing or
    force-pushing.
  - **`.claude/skills/humanify/SKILL.md`** (new) — a skill for writing code
    comments, docstrings, and PR/commit descriptions: comment the *why*, write the
    standing state rather than the change, turn comma-runs into lists, and trim
    prose without collapsing scannable structure.
  - **`.claude/skills/ace-it/SKILL.md`** (new) — a skill for writing fewer,
    real-behavior tests: assert on real outputs instead of `mock.call_args`, give
    glue code e2e coverage rather than mock-plumbing unit tests, and match test
    count to risk.
  - **Frontmatter repair** — `mdformat` was corrupting the YAML frontmatter of
    every `SKILL.md` (folding `name`/`description`/`argument-hint` into a heading
    and escaping brackets). The `Makefile` now excludes `.claude/skills` from the
    markdown file set, and the frontmatter of all skills (`codex-verify`,
    `pragmatic`, `watch-action`, `humanify`, `ace-it`) is reconstructed as proper
    `---`-delimited YAML. Real docs under `.claude/` (`development.md`,
    `formatting.md`) are still formatted.

  ## Why

  Codifies recurring guidance so agents write comments and tests that survive
  refactors, and stops `mdformat` from silently breaking skill frontmatter (which
  the skill loader needs to parse).


.claude/skills/ace-it/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+name: ace-it
+description: Use when writing, adding, or porting any test — before creating a test file or test method, and before reaching for mock.patch or assert_called_with.
+---
+
+# ace-it — Writing Tests
+
+## Overview
+
+A test earns its place by catching a real behavior regression. If a test can only fail when someone rewrites the test
+itself (because it asserts how the code is wired, not what it produces), it is noise — delete it, don't write it.
```
svij-sc, line 11: 💌

.claude/skills/ace-it/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Core principle:** Prefer one test that exercises real behavior over many that mock collaborators. Match the number of
+tests to the risk of the change, not to the number of functions you touched.
+
+## When to Use
+
+Invoke this skill **before writing any test** — a new file, a new method, or porting tests from another branch. Also
+invoke it the moment you reach for `mock.patch` or `assert_called_with`.
```
svij-sc, line 19: do we need to reference this skill in claude.md too or is it automatic?

.claude/skills/ace-it/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+
+```python
+def test_from_yaml_loads_values(self) -> None:
+    hyperparams = HyperParams.from_yaml(UriFactory.create_uri(self.yaml_path))
+    self.assertEqual(hyperparams.graph_sampler_config.main_batch_size, 32)
+    self.assertEqual(hyperparams.optim_config.name, "AdamW")  # default backfilled
+    self.assertEqual(hyperparams.seed, 7)
+```
+
+One real behavior test like this beats five that patch collaborators.
```
svij-sc, line 40: This seems very verbose - curious if you have tried some simpler prompts in the past and how they worked out.

Btw, curious also in how you think these will evolve as we add more skills.
Should we explore custom sub agents: 
https://code.claude.com/docs/en/sub-agents

1. documentor
2. Test writer
3 ....

.claude/skills/ace-it/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+- The test body is mostly `@patch` decorators and `assert_called*`.
+- You're reading `mock.call_args` / `call_args.args[0]` to reach the object the function built.
+- The same multi-`patch` `with` block is copy-pasted across many methods.
+- You're mocking a collaborator you wrote (not an external service).
+- The test would still pass if the function's real logic were wrong, and only fail if someone changed which helper it
+  calls.
+- You're adding a unit test for a function whose only job is to submit/launch/forward.
+- You're keeping a ported test only because it was already there.
+
+**All of these mean: don't write it (or delete it). Reach for a real-behavior test, or push coverage to e2e.**
```
svij-sc, line 132: most important of all: "Do not modify existing tests without explicit approval from user. Prompt user whether or not change is okay. " 

thoughtS?

svij-sc (approved):
minor comments/discussion points - LGTM generally

.claude/skills/ace-it/SKILL.md
kmontemayor2-sc: It's been automatic for me so far.

.claude/skills/ace-it/SKILL.md
kmontemayor2-sc: Hmmm, we should explore looking at sub-agents! 

I've had good luck with this skill when testing locally so far - shall we update if we run into issues with it down the line?

.claude/skills/ace-it/SKILL.md
kmontemayor2-sc: I think it's fine to modify existing tests?

If we change behavior (which happens) we often change tests to - that's normal. 

I find it to be a better use of my time to review something and tell the robots to revert vs needing to grant them permission for all code changes.

yliu2-sc (approved):

PR #716, opened 2026-07-21
Use BqUtils for sharded read schema validation

kmontemayor2-sc:
## What

- Uses `BqUtils.fetch_bq_table_schema` for sharded-read shard-key validation.
- Removes the standalone BigQuery client and table-reference construction from `sharded_read.py`.
- Adds focused tests for valid shard keys, missing columns, and missing tables.

## Why

Schema lookup is already owned by `BqUtils`. Reusing that API removes duplicate BigQuery client construction while preserving the existing validation and error behavior.

## Testing

- `sharded_read_test.py`: 3 tests passed.
- Scoped `ty` checks passed for all changed Python files.
- Ruff check and formatting passed for all changed Python files.


mkolodner-sc (approved):
Thanks!

yliu2-sc (approved):

PR #717, opened 2026-07-21
Support a BigQuery quota project in BqUtils

kmontemayor2-sc:
## What

- Adds an optional `quota_project_id` argument to `BqUtils`.
- Supports `GIGL_BIGQUERY_QUOTA_PROJECT` as a process-level default, with an explicit argument taking precedence.
- Preserves the existing BigQuery client construction path when no non-empty override is set.
- Propagates a configured quota project into launched Python component environments.
- Documents configuration and the required Service Usage permission.

## Why

BigQuery quota and API-enablement checks can need a project distinct from both the active credential project and the project containing BigQuery resources. A BigQuery-scoped client option supports that deployment model without changing quota behavior for other Google Cloud clients.

## Testing

- `bq_test.py`: 11 tests passed.
- Quota environment propagation tests: 2 tests passed.
- Scoped `ty` checks passed for all changed Python files.
- Ruff and Markdown formatting passed for all changed files.


mkolodner-sc (approved):
LGTM, thanks!

yliu2-sc (approved):

PR #718, opened 2026-07-21
Implement MonitoredChannel

jchmura-sc:
**Purpose**
The purpose of this PR is to add telemetry to our `ShmChannel`.

We want to know how many `SampleMessage`'s are in each of our channels at any given time, since an empty channel implies a blocking `recv`, which implies a dataloader bottleneck. Conversely, if our channels stay comfortably above some low watermark then any changes _upstream_ of `dataloader.__next__` is unlikely to improve throughput.

**PR Changes**
1. New class `SizedShmChannel(ShmChannel)`: adds `qsize()` method to `ShmChannel` by attaching to the channels memory region and inspecting the C++ struct layout. It calculates queue size as the delta between the circular buffer's write and read sequence numbers.
2. New class `MonitoredShmChannel(SizedShmChannel)`: a subclass that integrates with the GiGL metrics service, pushing the `qsize` as a gauge on every `recv()` call.

**Usage (colocated)**
1. Call `gigl.src.common.utils.metrics_service_provider.initialize_metrics()`
2. Set `GIGL_ENABLE_PERF_MONITORING` env var to `"True"` or `"1"`

**Context (Colocated-Mode)**
Each of our `N` replicas launch `N_local` training workers. Each training worker initializes one or more data loaders (e.g. `DistABLPLoader`, `DistNeighborLoader`). Each dataloader is associated with a `ShmChannel`. For a given dataloader, multiple sampling workers (e.g. `DistNeighborSampler`, `DistPPRNeighborSampler`) push data into the queue.

The channel is roughly used like:

```python
channel = ShmChannel()

# In dataloader process (single consumer)
def __next__(self):
    msg: SampleMessage = channel.recv() # Blocks if empty
    ... # Convert SampleMessage to PyG data

# In each`SamplingRuntime` worker process(multiple producers)
async def _send_adapter(self, ...):
    ... # Sample subgraphs    
    msg: SampleMessage = self._collate_fn(...) # Hydrate features over RPC
    channel.send(msg) # Blocks if channel is at capacity
```

**Challenge**
There is no direct way to gauge the queue size since GLT's `ShmChannel` does not expose the size to python.


gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,71 @@
...
+            # Reference: https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9/graphlearn_torch/include/shm_queue.h#L65
+            # ==================================================================================
+            # Byte Offset | C++ Member Variable | Type     | Size    | Description
+            # ------------+---------------------+----------+---------+--------------------------
+            # 00 - 07     | max_block_num_      | size_t   | 8 bytes | Capacity (max block count)
+            # 08 - 15     | max_buf_size_       | size_t   | 8 bytes | Max buffer size in bytes
+            # 16 - 23     | block_meta_offset_  | size_t   | 8 bytes | Offset to BlockMeta array
+            # 24 - 31     | data_buf_offset_    | size_t   | 8 bytes | Offset to raw data buffer
+            # 32 - 39     | write_block_id_     | size_t   | 8 bytes | Total messages enqueued <-- READ HERE
+            # 40 - 47     | read_block_id_      | size_t   | 8 bytes | Total messages dequeued <-- READ HERE
```
jchmura-sc: These are monotonically increasing sequence numbers, _not_ wraparound array indices. Hence we can compute queue size as `write_block_id_ - read_block_id_`

gigl/src/common/utils/metrics_service_provider.py
```diff
@@ -61,11 +63,20 @@ def initialize_metrics(task_config_uri: Uri, service_name: str) -> bool:
         return False
 
 
-def get_metrics_service_instance() -> Optional[OpsMetricPublisher]:
+def get_metrics_service_instance() -> OpsMetricPublisher:
     if _metrics_instance is None:
-        logger.warning(
-            "initialize_metrics() was not called, using NopMetricsPulisher as default"
+        task_config_uri = os.environ.get(TASK_CONFIG_URI_ENV_KEY)
+        service_name = os.environ.get(JOB_NAME_GROUPING_ENV_KEY) or os.environ.get(
```
jchmura-sc: It's actually "GIGL_APPLIED_TASK_IDENTIFIER" this serves as the appropriate service name based on frozen config. Should double check if "GBML_JOB_NAME" is set and if so then this can be simplified

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,68 @@
...
+
+class SizedShmChannel(ShmChannel):
+    """Extends ShmChannel with queue size method `qsize()` by attaching to the channels memory region and inspecting the C++ struct layout."""
+
+    def qsize(self) -> int:
+        # Obtain the shmid from the underlying C++ SampleQueue instance
+        shmid = self._queue.__getstate__()
+
+        # Attach to the shared memory segment in the current process
+        ptr = _libc.shmat(shmid, None, 0)  # shmat returns (void *)(-1) on failure
```
jchmura-sc: Can we cache this?

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,96 @@
...
+    """Extends ShmChannel with queue size method `qsize()` by attaching to the channels memory region and inspecting the C++ struct layout."""
+
+    # Class-level handle cached across all instances in the current process
+    _libc: Optional[ctypes.CDLL] = None
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._finalizer: Optional[weakref.finalize] = None
+
+    def qsize(self) -> int:
```
jchmura-sc, line 37: Thoughts on aliasing this for `__len__`?

kmontemayor2-sc, line 37: BTW would it be easier to just copy over / re-implement our own CPP share memory channel?

We support cpp now - though I think there may be some fiddling required to setup GLT imports in our CPP code.

But it seems like a better long term solution to just add a `qsize` to the shm cpp directly, WDYT?

gigl/distributed/distributed_neighborloader.py
```diff
@@ -278,6 +278,7 @@ def __init__(
                 sampling_config=sampling_config,
                 worker_options=worker_options,
                 sampler_options=sampler_options,
+                channel_name=f"nbr_channel_{runtime.rank}_of_{runtime.world_size}",
```
kmontemayor2-sc, line 281: If we have multiple of the same type of neighborloader at the same time - would they step on each other?

I did something like https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/dist_ablp_neighborloader.py#L620 to differentiate between different "instances" of the same type of loader.

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,96 @@
...
+        # 16 - 23     | block_meta_offset_  | size_t   | 8 bytes | Offset to BlockMeta array
+        # 24 - 31     | data_buf_offset_    | size_t   | 8 bytes | Offset to raw data buffer
+        # 32 - 39     | write_block_id_     | size_t   | 8 bytes | Total messages enqueued <-- READ HERE
+        # 40 - 47     | read_block_id_      | size_t   | 8 bytes | Total messages dequeued <-- READ HERE
+        # 48 - 55     | alloc_offset_       | size_t   | 8 bytes | Internal ring write ptr
+        # 56 - ...    | released_offset_    | size_t   | 8 bytes | Internal ring release ptr
+        # ==================================================================================
+        ptr = self._shm_ptr
+        write_block_id = ctypes.c_size_t.from_address(ptr + 32).value
+        read_block_id = ctypes.c_size_t.from_address(ptr + 40).value
```
kmontemayor2-sc, line 55: Should we do `sizeof(size_t * member_num)`? 

I guess this is all so fragile maybe it doesn't matter too much?

gigl/distributed/utils/channel.py
kmontemayor2-sc: IMO `__len__` should more be like "item count" not "size of data"?

BTW if we do this anyways do you think introducing item count is useful?

gigl/src/common/utils/metrics_service_provider.py
kmontemayor2-sc: Why don't we call `initialize_metrics` first? I'm. a little confused as to why we're making this change

tests/unit/distributed/utils/channel.py
```diff
@@ -0,0 +1,82 @@
...
+        self.msg_2 = {"baz": torch.rand(3), "bat": torch.rand(7)}
+
+    def test_single_process_qsize(self):
+        # Initially empty channel
+        self.assertEqual(self.channel.qsize(), 0)
+        self.assertTrue(self.channel.empty())
+
+        # Send first message
+        self.channel.send(self.msg_1)
+        self.assertEqual(self.channel.qsize(), 1)
```
kmontemayor2-sc: oh wait is `qsize` the number of messages?


Hmmm maybe I got confused - should we clarify what `qsize` is in a doc string then?

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,96 @@
...
+
+class MonitoredShmChannel(SizedShmChannel):
+    """Monitored variant of SizedShmChannel that integrats with GiGL metrics_service and records queue size on recv() as a gauge."""
+
+    def __init__(self, channel_name: str, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._channel_name = channel_name
+
+    def recv(self, *args, **kwargs) -> SampleMessage:
+        publisher = get_metrics_service_instance()
```
kmontemayor2-sc: Maybe related to my other comments but this change requires service_metrics to be setup right? 

I think making this default is ok provided it doesn't hurt perf, but should we make it a no-op when no service metrics?

gigl/distributed/utils/channel.py
jchmura-sc: Yeah, `qsize()` _is_ item count. Added a docstring to clarify and added `__len__`

gigl/distributed/utils/channel.py
jchmura-sc: Yeah leaning towards it doesn't matter too much - e.g. Pytorch only supports 64-bit sytems

gigl/distributed/utils/channel.py
jchmura-sc: Good point, let me try to simplify and add the no-op fallback

gigl/distributed/distributed_neighborloader.py
jchmura-sc: Great point. Currently we have 1 channel per process but something like this is a good safety measure for the future.

gigl/distributed/utils/channel.py
jchmura-sc: I agree that implementing our own CPP shared memory channel is the most robust long term solution. But this would be a rather large change, requiring a few different moving parts copied over. I suspect we are not too worried about upstream features for these queues on the GLT side, but worth considering too.

I think the options are something like:
1. Copy everything over from GLT. Gives us full control but large change.
2. Build GLT from source with minimal changes to support `qsize`. Not worth the hassle in my opinion.
3. Try to extend GLT using our own bindings to support `qsize`.  This would be best but (a) [queue meta is private field of the channel](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9/graphlearn_torch/include/shm_queue.h#L238), and the fields we read to derive the `qsize` are [private within meta](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9/graphlearn_torch/include/shm_queue.h#L154-L165) (friendship would not be inherited to our subclasses).
4. Use the shared memory inspection as we currently have.

If it's okay, we can keep the implementation as is, leave a TODO, and consider full channel re-implementation later if we think it's worth it.

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,142 @@
...
+    # Counts instantiations of this class, per process.
+    # This is needed so we can generate unique channel names for each instance within the same process.
+    # NOTE: This is per-class, not per-instance.
+    _counter = count(0)
+
+    def __init__(self, channel_name: str, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._channel_name = f"{channel_name}_id{next(self._counter)}"
+
+        self._publisher: Optional[OpsMetricPublisher]
```
jchmura-sc: Is this fine to cache? There should not be subsequent calls to `initialize_metrics` which could invalidate this publisher.

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,142 @@
...
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+
+        self._publisher: Optional[OpsMetricPublisher]
+        try:
+            self._publisher = get_metrics_service_instance()
+        except RuntimeError:
+            self._publisher = None
```
jchmura-sc: In case `get_metrics_service_instance` semantics are updated to directly return None, the `try-catch` block can be simplified to

```python
self._publisher: Optional[OpsMetricPublisher] = get_metrics_service_instance()
```

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,142 @@
...
+            raise RuntimeError(f"shmat failed for shmid={shmid}: {error_msg}")
+
+        # Register automatic cleanup when this object is GC'd in this process
+        self._finalizer = weakref.finalize(self, libc.shmdt, ptr)
+        return ptr
+
+    @classmethod
+    def _get_libc(cls) -> ctypes.CDLL:
+        if cls._libc is None:
+            libc_path = ctypes.util.find_library("c") or "libc.so.6"
```
kmontemayor2-sc: QQ why do we do `or "libc.so.6"`? here?

gigl/distributed/utils/channel.py
kmontemayor2-sc: I think we're changing this to never be `None` right?

As for the `RuntimeError` I do think that we shouldn't try to catch it here - my understanding is that we do this since we require the metrics to be instantiated right? 

I think that's a bit odd - we're silently swallowing errors here - but maybe they should be real errors? We do this since we change the default channel to this one right?

Maybe the default channel shouldn't be updated? E.g. enabling these metrics is opt-in somehow?

gigl/distributed/utils/channel.py
kmontemayor2-sc: Probably fine to cache. Also shalle we default init to None? I feel like these only-def lines are dangerous.

jchmura-sc: Sounds good, I wasn't quite sure which style to adhere to in this instance.

gigl/distributed/utils/channel.py
jchmura-sc: Not strictly necessary, but it's possible that for some reason `ctypes.util.find_library("c")` will not find libc and return `None`. 

I can make this more clear by explicitly raising in that case. I suspect if it can't be found then hardcoding `libc.so` may not help.

gigl/distributed/utils/channel.py
jchmura-sc: Yeah... I suppose the only way around this would be to pass a flag through.

Then `BaseDistLoader.create_colocated_channel()` can return a `MonitoredShmChannel` if metrics were opted in. And in that case, we can call `get_metrics_service_instance` unconditionally.

It may still be useful in that case to catch the `RuntimeError` to help the user out? Something like:

```python
@staticmethod
def create_colocated_channel(
    worker_options: MpDistSamplingWorkerOptions,
    *
    enable_monitoring: bool = False,
) -> ShmChannel:
    if enable_monitoring:
        try:
            get_metrics_service_instance()
        except RuntimeError as e:
            raise RuntimeError(
                "Tried to create colocated channel with enable_monitoring=True, "
                "but metrics instance was not initialized. Either call "
                "initialize_metrics() or set enable_monitoring=False."
            ) from e
            channel = MonitoredShmChannel(...)
    else:
        channel = ShmChannel(...)

    if worker_options.pin_memory:
        channel.pin_memory()
    return channel
```

Thoughts? @kmontemayor2-sc

kmontemayor2-sc: Maybe something like a `GIGL_ENABLE_DEBUG` env var that we can use as a general debug tool?

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,121 @@
...
+    _counter = count(0)
+
+    def __init__(self, channel_name: str, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._channel_name = f"{channel_name}_id{next(self._counter)}"
+
+    def recv(self, *args, **kwargs) -> SampleMessage:
+        publisher: Optional[OpsMetricPublisher] = get_metrics_service_instance()
+        if publisher is not None:
+            publisher.add_gauge(f"{self._channel_name}_qsize", self.qsize())
```
kmontemayor2-sc: btw I think this is fine for now :)

kmontemayor2-sc: But can you add a TODO here (feel free to tag me on it) to make this some debug-enabled switch?

gigl/distributed/utils/channel.py
```diff
@@ -0,0 +1,121 @@
...
+    """Monitored variant of SizedShmChannel that integrates with GiGL metrics_service and records queue size on recv() as a gauge."""
+
+    # Counts instantiations of this class, per process.
+    # This is needed so we can generate unique channel names for each instance within the same process.
+    # NOTE: This is per-class, not per-instance.
+    _counter = count(0)
+
+    def __init__(self, channel_name: str, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._channel_name = f"{channel_name}_id{next(self._counter)}"
```
kmontemayor2-sc, line 127: btw this does require `channel_name` to be unique across processes right? 

Should we document that somewhere?

gigl/distributed/dist_ablp_neighborloader.py
```diff
@@ -91,6 +91,7 @@ def __init__(
         local_process_rank: Optional[int] = None,  # TODO: (svij) Deprecate this
         local_process_world_size: Optional[int] = None,  # TODO: (svij) Deprecate this
         non_blocking_transfers: bool = True,
+        enable_channel_monitoring: bool = False,
```
kmontemayor2-sc: this might be too specific actually :/ I'd rather us avoid having such a large API and instead do the env var.

Sorry for the back and forth.

gigl/distributed/utils/channel.py
jchmura-sc: Good point. Assuming the metrics instance does not add e.g. PID as metadata on publisher then these would alias.

There's a world in which a user may not really care (i.e. they don't want 10 to 100's of unique metrics but just an aggregated view) but best to communicate this. I'll add some documentation to make it clear.

gigl/distributed/utils/channel.py
jchmura-sc: Sorry, to clarify, which part would be toggled based on the switch?

I think we can switch to env var to enable channel monitoring but presumably once the `MonitoredShmChannel` is selected (based on the env var) there are no more flags?

kmontemayor2-sc: Adding the debug var would swap between MonitoredChannel and the assumed default SHMChannel.

If we do this then MonitoredChannel should throw if it can't get a metrics provider.

gigl/distributed/base_dist_loader.py
```diff
@@ -426,21 +429,49 @@ def create_sampling_config(
...
+
+        Raises:
+            RuntimeError: If "GIGL_ENABLE_PERF_MONITORING" is set but user did not previously call
+                `gigl.src.common.utils.metrics_service_provider.initialize_metrics()`.
         """
-        channel = ShmChannel(
-            worker_options.channel_capacity, worker_options.channel_size
-        )
+        enable_channel_monitoring = os.environ.get(
+            "GIGL_ENABLE_PERF_MONITORING", ""
```
jchmura-sc: Just confirming whether this env var should or should not be part of https://github.com/Snapchat/GiGL/blob/main/gigl/env/constants.py @kmontemayor2-sc

kmontemayor2-sc: Yup!

kmontemayor2-sc (approved):
Thanks Jacob!

mkolodner-sc (approved):
Thanks a lot :)

gigl/distributed/base_dist_loader.py
```diff
@@ -426,21 +430,51 @@ def create_sampling_config(
...
+        Raises:
+            RuntimeError: If GIGL_ENABLE_PERF_MONITORING is set but user did not previously call
+                `gigl.src.common.utils.metrics_service_provider.initialize_metrics()`.
         """
-        channel = ShmChannel(
-            worker_options.channel_capacity, worker_options.channel_size
-        )
+        enable_channel_monitoring = os.environ.get(
+            GIGL_ENABLE_PERF_MONITORING, ""
+        ).strip().lower() in ("1", "True")
```
jchmura-sc, line 457: typo, should check again "true"

PR #719, opened 2026-07-23
  [graph-store] Fix high-fan-out sampling deadlock with a per-channel in-flight cap

kmontemayor2-sc:
## What

  Fixes a scheduler deadlock in graph-store mode's shared sampling producer
  (`shared_dist_sampling_producer.py`) that triggers at high fan-out — whenever a
  storage server co-locates ≥2 compute-rank channels (~≥1.6 ranks/server).

  The single per-worker scheduler thread submits sampling batches round-robin
  across channels. Each `sample_*` call acquires a slot from the sampler's
  `BoundedSemaphore(worker_concurrency)`; a slot is held from submit until the
  result's `channel.send` completes. When a channel's consumer falls behind, its
  output channel saturates and its in-flight coroutines wedge in `send`, holding
  their slots. The scheduler then blocks on that channel's exhausted semaphore —
  and because it is the *single* thread serving all co-located channels, it can no
  longer drain the other channels or process commands. At high fan-out this closes
  a cross-rank cycle and the whole compute cluster stalls.

  ## Fix

  Add a per-channel in-flight submission cap. The scheduler **parks** a channel
  (skips its submit, does not re-enqueue) once it has `worker_concurrency` batches
  submitted-but-not-completed — mirroring the sampler semaphore exactly, so the
  thread never issues a submit that could block. A parked channel is **woken**
  (re-enqueued) only by the batch-completion callback (`_on_batch_done`) when an
  in-flight slot frees. The scheduler stays wait-free: it rotates past saturated
  channels and keeps draining commands, and a park-only pump cycle leaves
  `made_progress` unset so Phase 3 idles on its tick instead of busy-spinning.

  This changes only how work is *scheduled*, not per-rank data volume or sampling
  results.


tests/unit/distributed/graph_store/shared_dist_sampling_producer_test.py
```diff
@@ -424,3 +428,580 @@ def run_callback() -> None:
...
+            self.nowait_get_calls += 1
+        return self._queue.get_nowait()
+
+    def get(self, timeout: float | None = None) -> tuple[SharedMpCommand, object]:
+        with self._lock:
+            self.blocking_get_calls += 1
+        return self._queue.get(timeout=timeout)
+
+
+class StallFixWorkerLoopTest(TestCase):
```
mkolodner-sc, line 547: Roughly, how long do these tests take to run, out of curiosity. If it's very heavy, should we consider moving this to intregration testingi nstead?

mkolodner-sc (approved):
Thanks Kyle! The business logic fix looks good to me. Thanks for digging into it.

On the validation side, this does feel like a lot of complex testing. I understand why it's needed - my main concern is that the test setup now has enough moving parts that it may be hard to maintain or extend later.

I don’t have an obviously better drop-in alternative, but I wonder if part of this could be simplified maybe by extracting the parking/requeue state into a smaller helper and testing that we never submitting when `submitted_batches - completed_batches >= worker_concurrency`. 

SG to stamp either way to unblock landing this bug fix, but curious to hear your thoughts.

tests/unit/distributed/graph_store/shared_dist_sampling_producer_test.py
kmontemayor2-sc:
```
uv run python -m tests.unit.main \
	--env=test \
	--resource_config_uri=... \
	--test_file_pattern=shared_dist_sampling_producer_test.py \

...

----------------------------------------------------------------------
Ran 11 tests in 0.754s

OK
2026-07-29 14:29 [INFO] : Ran 11/11 test cases (test_utils.py:run_tests:103)
2026-07-29 14:29 [INFO] : It took  33.98 second(s) to run tests (test_utils.py:run_tests:105)
```

They're pretty fast :)

yliu2-sc (approved):

PR #721, opened 2026-07-27
Pin torchrec~=1.3.0 to match fbgemm-gpu~=1.3.0

kmontemayor2-sc:
gigl.nn depends on torchrec, which in turn depends on fbgemm-gpu for its kernels. fbgemm-gpu is already pinned to ~=1.3.0 in both the pyg27-torch28-cpu and pyg27-torch28-cu128 extras, but torchrec was left unconstrained. An unconstrained re-lock can pull a newer torchrec built against a newer fbgemm-gpu line (e.g. 1.5), whose symbols (such as KVZCHTBEConfig) are absent from the pinned 1.3.0 fbgemm-gpu, breaking `import gigl.nn` at runtime.

https://meta-pytorch.org/torchrec/setup-torchrec.html#version-compatability

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,8 @@ pyg27-torch28-cpu = [
   "torch_scatter; sys_platform != 'darwin'",
   "torch_sparse; sys_platform != 'darwin'",
   "torch_spline_conv; sys_platform != 'darwin'",
-  "torchrec ; sys_platform != 'darwin'",
+  # Must move in lockstep with fbgemm-gpu~=1.3.0 above; see the FBGEMM/torchrec compatibility matrix: https://docs.pytorch.org/FBGEMM/general/Releases.html
+  "torchrec~=1.3.0 ; sys_platform != 'darwin'",
 ]
 pyg27-torch28-cu128 = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -71,7 +71,8 @@ pyg27-torch28-cpu = [
   "torch_scatter; sys_platform != 'darwin'",
   "torch_sparse; sys_platform != 'darwin'",
   "torch_spline_conv; sys_platform != 'darwin'",
-  "torchrec ; sys_platform != 'darwin'",
+  # Must move in lockstep with fbgemm-gpu~=1.3.0 above; see https://meta-pytorch.org/torchrec/setup-torchrec.html#version-compatability
+  "torchrec~=1.3.0 ; sys_platform != 'darwin'",
```
svij-sc, line 75: nit: move the deps side by side as they are related to each other.

svij-sc (approved):

pyproject.toml
kmontemayor2-sc: We have these alphabetical for consistency - what if there's some combo package that's all tied up in the future?

(note that `pyg_lib` is above `torch` which is above the other pyg deps)

yliu2-sc (approved):

PR #722, opened 2026-07-27
Update Metrics Service Provider Singleton Semantics

jchmura-sc:
**Summary of Semantics**
- `get_metrics_service_instance` returns the value of the singleton metrics instance as is (could be None), but still raises if `initialize_metrics` was not called first.
- Failing to construct a custom metrics provider in `initiailize_metrics` does **not** fallback to `NoOpMetricsPublisher` - it overrides the metrics instance to `None`.

### DIFF
--- a/gigl/src/common/utils/metrics_service_provider.py
+++ b/gigl/src/common/utils/metrics_service_provider.py
@@ -12,6 +12,7 @@ from snapchat.research.gbml.gbml_config_pb2 import GbmlConfig
 logger = Logger()
 
 _metrics_instance: Optional[OpsMetricPublisher] = None
+_WERE_METRICS_INITIALIZED: bool = False
 JOB_NAME_GROUPING_ENV_KEY = "GBML_JOB_NAME"
 
 
@@ -31,7 +32,8 @@ def initialize_metrics(task_config_uri: Uri, service_name: str) -> bool:
         no-op default when no custom class is configured), ``False`` if a
         custom metrics class was specified but could not be loaded or instantiated.
     """
-    global _metrics_instance
+    global _metrics_instance, _WERE_METRICS_INITIALIZED
+    _WERE_METRICS_INITIALIZED = True
     os.environ[JOB_NAME_GROUPING_ENV_KEY] = service_name
     proto_utils = ProtoUtils()
     task_config: GbmlConfig = proto_utils.read_proto_from_yaml(
@@ -60,6 +62,10 @@ def initialize_metrics(task_config_uri: Uri, service_name: str) -> bool:
 
 
 def get_metrics_service_instance() -> Optional[OpsMetricPublisher]:
+    if not _WERE_METRICS_INITIALIZED:
+        raise RuntimeError(
+            "Metrics instance is not initialized. Call initialize_metrics() before getting the instance."
+        )
     return _metrics_instance
 
 
--- a/gigl/src/common/utils/metrics_service_provider.py
+++ b/gigl/src/common/utils/metrics_service_provider.py
@@ -54,22 +54,12 @@ def initialize_metrics(task_config_uri: Uri, service_name: str) -> bool:
         logger.info(f"Instantiated Custom Metrics Class from: {metrics_cls_path}")
         return True
     except Exception as e:
-        logger.error(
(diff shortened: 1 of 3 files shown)


gigl/src/common/utils/metrics_service_provider.py
```diff
@@ -54,22 +54,12 @@ def initialize_metrics(task_config_uri: Uri, service_name: str) -> bool:
...
-        logger.error(
-            f"Could not instantiate class {metrics_cls_path}: {e}. Falling back to No-op metrics."
-        )
-        _metrics_instance = NopMetricsPublisher()
+        logger.error(f"Could not instantiate class {metrics_cls_path}: {e}")
+        _metrics_instance = None
         return False
 
 
 def get_metrics_service_instance() -> Optional[OpsMetricPublisher]:
```
kmontemayor2-sc, line 64: IIRC we want to enforce that `get_metrics_service_instance` is called *after* `initialize_metrics`. E.g. we should add some separate `_WERE_METRICS_INITIALIZED` bool and check that in this fn (throwing if not true).

kmontemayor2-sc: cc: @mkolodner-sc @svij-sc to double check

gigl/src/common/utils/metrics_service_provider.py
```diff
@@ -54,22 +54,12 @@ def initialize_metrics(task_config_uri: Uri, service_name: str) -> bool:
...
-    if _metrics_instance is None:
-        logger.warning(
-            "initialize_metrics() was not called, using NopMetricsPulisher as default"
-        )
-        raise RuntimeError(
-            "Metrics instance is not initialized. Call initialize_metrics() before getting the instance."
-        )
```
mkolodner-sc, line 71: Hmm, I've noticed that with this change it seems like there's isn't a way to separately identify 
- metric initialization hasn't been called before get
- metric initialization **has** been called before get, but failed

since they both fallback to None now by the time `get_metrics_service_instance` is called.

From my understanding, we want the first case to properly raise an error, since this is a reproducible usage error by the user, but the second case should not raise an error, or else transient metric initialization failures may impact business logic.

cc @kmontemayor2-sc Since these two cases are not distinguishable anymore, should we go back and just have metric initialization failure fallback to `NopMetricsPublisher`, so that the two cases above can hold? Unless you had a different suggestion in mind. The alternative seems to be what the current PR is doing, but I'm not sure if we want to not be raising an error in the first case here.

gigl/src/common/utils/metrics_service_provider.py
mkolodner-sc: That makes sense to me, similar to what I've expressed in https://github.com/Snapchat/GiGL/pull/722#discussion_r3660875047, an initialization boolean so that we can distinguish between premature get calls and get calls after metric initialization has failed.

gigl/src/common/utils/metrics_service_provider.py
kmontemayor2-sc: The cases are distinguishable if we add some new `_WERE_METRICS_INITED` right?

kmontemayor2-sc (approved):
Thanks!

tests/unit/src/common/utils/metrics_service_provider_test.py
```diff
@@ -50,15 +51,27 @@ def test_valid_custom_class_returns_true(self) -> None:
...
 
         result = initialize_metrics(task_config_uri=uri, service_name="test_service")
 
         self.assertFalse(result)
+        self.assertIsNone(get_metrics_service_instance())
+
+    def test_custom_class_not_specified_returns_true_and_falls_back_to_nop(
+        self,
+    ) -> None:
+        """initialize_metrics returns False and falls back to NopMetricsPublisher when custom metrics class not provided."""
```
mkolodner-sc: Is this docstring correct for what we're testing? It says `initialize_metrics` returns False but it's checking for if it's true

mkolodner-sc (approved):
Thanks Jacob :)

tests/unit/src/common/utils/metrics_service_provider_test.py
jchmura-sc: Thanks for catching this, updating.

PR #723, opened 2026-07-28
Refactor PPR top-k selection helper

mkolodner-sc:
### Description

This PR refactors PPR extraction by moving the existing inline finalized-PPR selection and residual top-up logic into smaller C++ helpers.

The extraction behavior is intended to stay the same, but the implementation is now split into reusable pieces:

- `selectFinalizedPPRPairs`: selects finalized PPR candidates by raw PPR score
- `addResidualMassToPPRPairs`: adds residual mass to selected finalized PPR candidates when residual top-up is enabled
- `appendResidualTopUpPairs`: appends residual candidates into remaining sequence slots

These helpers replace logic that previously lived directly inside `extractTopKWithResidualTopUp`. The untyped extraction path still owns final output assembly and sorting, but the lower-level pieces can now be reused by the typed-channel PPR extraction work in the stacked PR.

No functional change is intended.

### Performance

This refactor should be mostly performance-neutral, if not slightly faster.

The main algorithmic work is the same as before: scan finalized PPR scores, select finalized top-k candidates, optionally scan residuals for top-up candidates, add residual mass to selected finalized candidates, append top-up rows, and sort the emitted rows. The PR mostly moves those existing steps into named helpers.

One small implementation detail changes: finalized PPR selection now uses `nth_element` followed by a final sort of the selected output rows, instead of relying on `partial_sort` in the non-top-up path. This avoids ordering intermediate candidates before we know the final emitted set. The asymptotic behavior is unchanged for the final output, but this can save some unnecessary ordering work, so any runtime difference should skew neutral-to-slightly-faster rather than slower.

### Testing

- `make -C gigl-core format_cpp`
- `git diff --check`
- `uv run ty check .github/scripts examples gigl tests snapchat scripts`

### DIFF
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -278,18 +278,20 @@ void PPRForwardPush::pushResiduals(
 static std::vector<std::pair<int32_t, double>> selectFinalizedPPRPairs(const SeedNodeTypeState& nodeTypeState,
                                                                        int32_t finalizedPPRNodeLimit) {
     const auto& scores = nodeTypeState.pprScores;
-    const auto higherScore = [](const auto& a, const auto& b) { return a.second > b.second; };
 
-    const int32_t pprTopK = std::min(finalizedPPRNodeLimit, static_cast<int32_t>(scores.size()));
+    const int32_t numReturnedPairs = std::min(finalizedPPRNodeLimit, static_cast<int32_t>(scores.size()));
     std::vector<std::pair<int32_t, double>> selectedPairs;
-    selectedPairs.reserve(static_cast<size_t>(pprTopK));
-    if (pprTopK > 0) {
+    selectedPairs.reserve(static_cast<size_t>(numReturnedPairs));
+    if (numReturnedPairs > 0) {
         std::vector<std::pair<int32_t, double>> scorePairs(scores.begin(), scores.end());
-        if (pprTopK < static_cast<int32_t>(scorePairs.size())) {
-            std::nth_element(scorePairs.begin(), scorePairs.begin() + pprTopK, scorePairs.end(), higherScore);
+        if (numReturnedPairs < static_cast<int32_t>(scorePairs.size())) {
+            std::nth_element(scorePairs.begin(),
+                             scorePairs.begin() + numReturnedPairs,
+                             scorePairs.end(),
+                             [](const auto& a, const auto& b) { return a.second > b.second; });
         }
 
-        for (int32_t rankIdx = 0; rankIdx < pprTopK; ++rankIdx) {
+        for (int32_t rankIdx = 0; rankIdx < numReturnedPairs; ++rankIdx) {
             selectedPairs.emplace_back(scorePairs[rankIdx].first, scorePairs[rankIdx].second);
         }
     }
@@ -314,8 +316,7 @@ static void appendResidualTopUpPairs(const SeedNodeTypeState& nodeTypeState,
     const int32_t residualTopUpBudget =
         std::max<int32_t>(0, sequenceLength - static_cast<int32_t>(selectedPairs.size()));
     if (residualTopUpBudget > 0) {
-        const auto& scores = nodeTypeState.pprScores;
-        const auto higherScore = [](const auto& a, const auto& b) { return a.second > b.second; };
+        const std::unordered_map<int32_t, double>& pprScoresByNodeId = nodeTypeState.pprScores;
         std::unordered_set<int32_t> selectedPPRNodeIds;
         selectedPPRNodeIds.reserve(selectedPairs.size());
         for (const auto& selectedPair : selectedPairs) {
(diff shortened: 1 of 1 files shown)


gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -266,11 +266,113 @@ void PPRForwardPush::pushResiduals(
...
+//   nodeTypeState: finalized PPR scores and residuals for one seed/node type.
+//   finalizedPPRNodeLimit: maximum finalized-PPR rows to select before top-up.
+//
+// Expected output: (node_id, raw_ppr_score) pairs selected by raw PPR score.
+// The order is unspecified; callers that emit these directly should sort the
+// returned vector before writing output tensors.
+static std::vector<std::pair<int32_t, double>> selectFinalizedPPRPairs(const SeedNodeTypeState& nodeTypeState,
+                                                                       int32_t finalizedPPRNodeLimit) {
+    const auto& scores = nodeTypeState.pprScores;
+    const auto higherScore = [](const auto& a, const auto& b) { return a.second > b.second; };
```
kmontemayor2-sc: qq do we just use this on line 289? Does it make more sense to define it around there (or inline it?)

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -266,11 +266,113 @@ void PPRForwardPush::pushResiduals(
...
+
+    const int32_t pprTopK = std::min(finalizedPPRNodeLimit, static_cast<int32_t>(scores.size()));
```
kmontemayor2-sc: nit. is this better named like `numReturnedPairs`?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -266,11 +266,113 @@ void PPRForwardPush::pushResiduals(
...
+        std::unordered_set<int32_t> selectedPPRNodeIds;
+        selectedPPRNodeIds.reserve(selectedPairs.size());
+        for (const auto& selectedPair : selectedPairs) {
+            selectedPPRNodeIds.insert(selectedPair.first);
+        }
+
+        std::vector<std::pair<int32_t, double>> residualPairs;
+        residualPairs.reserve(nodeTypeState.residuals.size());
+        for (const auto& [nodeId, residual] : nodeTypeState.residuals) {
+            if (residual <= 0.0 || selectedPPRNodeIds.find(nodeId) != selectedPPRNodeIds.end()) {
```
kmontemayor2-sc, line 332: QQ do we ever subtract from the residual value? If we do, then due to floating point weirdness a "valid(ish)" value may be negative.

Could be an edge case regardless - I'm not sure we really really care about residuals with such low value anyways?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -266,11 +266,113 @@ void PPRForwardPush::pushResiduals(
...
+// Expected output: selectedPairs has up to sequenceLength rows after appending
+// highest-scoring residual candidates that are not already selected. This helper
+// does not sort selectedPairs; callers sort only when their output needs it.
+static void appendResidualTopUpPairs(const SeedNodeTypeState& nodeTypeState,
+                                     std::vector<std::pair<int32_t, double>>& selectedPairs,
+                                     int32_t sequenceLength) {
+    const int32_t residualTopUpBudget =
+        std::max<int32_t>(0, sequenceLength - static_cast<int32_t>(selectedPairs.size()));
+    if (residualTopUpBudget > 0) {
+        const auto& scores = nodeTypeState.pprScores;
```
kmontemayor2-sc: btw minor style thing: I got confused later (on line 332 `auto scoreIter = scores.find(nodeId);`), I was like "why are we doing a list search here?"

Maybe using `auto` all the time isn't ideal? If I don't have intimate knowledge of `nodeTypeState` it's hard to know what types the inner fields are.

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -281,95 +383,32 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
-            }
-
-            if (enableResidualTopUp && selectedPairs.size() > 1) {
+            // Empty and singleton outputs are already ordered. Multi-row outputs can
+            // be unordered because the selection helpers use nth_element, so sort once
+            // to rank emitted rows by the score returned to callers.
+            if (selectedPairs.size() > 1) {
+                const auto higherScore = [](const auto& a, const auto& b) { return a.second > b.second; };
                 std::sort(selectedPairs.begin(), selectedPairs.end(), higherScore);
             }
```
kmontemayor2-sc: This is a logic change right?

IIRC we had some discussion before about wanting all the "top up" nodes at the end of the list, and. this change means they are interspersed throughout right?

kmontemayor2-sc (approved):
Thanks Matt! LGTM provided we are ok with this behavior change

https://github.com/Snapchat/GiGL/pull/723/changes#r3675350401

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: We do not intentionally subtract from residuals in forward push; pushed nodes are left in the residual map with 0.0, and neighbors receive positive residual mass. I kept the <= 0.0 filter because non-positive residuals cannot help top-up, and added a comment clarifying this is mostly filtering drained zero entries plus any unexpected non-positive values.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Made some renames to make things more clear here

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Before the helper extraction, the top-up path already did this:
```
// finalized PPR selected first
selectedPairs.emplace_back(nodeId, outputScore);

// residual top-up appended after
selectedPairs.emplace_back(residualNodeId, residualOutputScore);

// then all selected rows sorted together
if (enableResidualTopUp && selectedPairs.size() > 1) {
    std::sort(selectedPairs.begin(), selectedPairs.end(), higherScore);
}
```
So top-up nodes could already interleave in the emitted order.

I believe the important thing is that residual top-up nodes are included in the sequence only after finalized PPR selection has happened. Once the final set of nodes is selected, we expect all emitted rows, whether finalized PPR nodes or residual top-up nodes, to be ordered by descending emitted score.

yliu2-sc (approved):

PR #724, opened 2026-07-29
Add typed-channel PPR C++ extraction support

mkolodner-sc:
## Summary

Adds the C++ support needed for typed-channel PPR sampling.

This is the lower PR in the typed PPR stack. It adds the core typed-channel frontier drain and extraction logic used by the Python sampler layer in the follow-up PR.

## Motivation

For heterogeneous graphs, different edge types can represent meaningfully different relation paths. Typed-channel PPR lets us run separate PPR traversals over configured edge-type channels, then merge the results into one sequence while preserving channel-aware features.

## What Changed

- Adds typed-channel frontier draining across multiple `PPRForwardPush` states.
  - Each channel has its own traversal allowlist.
  - Active channel frontiers are unioned before neighbor fetch so shared one-hop fetches are issued once.

- Adds typed PPR extraction.
  - Accepts integer per-channel target counts from the caller.
  - Derives the final per-seed sequence cap from the sum of those target counts.
  - Deduplicates nodes across channels.
  - Attributes duplicate nodes to the channel where they have the strongest calibrated score.
  - Fills each channel toward its target, then globally overflows unused slots by best score.

- Handles residual top-up in the typed path.
  - Residual candidates participate in the same channel target and overflow policy.
  - Final emitted rows use calibrated scores when residual top-up is enabled.

- Emits typed edge attributes as:
  - `[best_score, channel_scores..., channel_presence_bits...]`

## Tests

Added focused C++ coverage for:

- typed frontier draining
- fetch-budget behavior
- typed extraction
- cross-channel deduplication
- residual top-up behavior
- channel target and overflow behavior


gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -374,11 +431,238 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+    for (const auto& [nodeId, score] : nodesAndScores) {
+        double calibratedScore = score / maxScore;
+        channelSelectionCandidates.emplace_back(nodeId, calibratedScore);
+    }
+}
+
+// Helper function for adding one channel's extracted PPR candidates into the
+// emitted typed feature table and a fill-pass candidate list.
+//
+// Unlike addTypedPPRChannelCandidates, this helper owns the output view: it
```
kmontemayor2-sc: Does it own the output view?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -374,11 +431,238 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+        if (a.bestCalibratedScore != b.bestCalibratedScore) {
+            return a.bestCalibratedScore > b.bestCalibratedScore;
+        }
+        if (a.channelCalibratedScore != b.channelCalibratedScore) {
+            return a.channelCalibratedScore > b.channelCalibratedScore;
+        }
+        if (a.channelIndex != b.channelIndex) {
+            return a.channelIndex < b.channelIndex;
+        }
+        return a.nodeId < b.nodeId;
```
kmontemayor2-sc, line 550: btw is there a case where the same node id can be returned from the same channel index twice?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -420,6 +704,148 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
 }
 
+PPRExtractResult extractTypedTopKWithResidualTopUp(const std::vector<PPRForwardPush*>& states,
+                                                   const std::vector<int32_t>& channelQuotas,
+                                                   int32_t maxPPRNodes,
+                                                   bool enableResidualTopUp) {
+    // Typed channels are constructed from the same seed batch and graph schema;
+    // only the edge-type traversal allowlist differs. Use the first state as
+    // the shared schema source for batch size and node-type count.
+    const auto* firstState = states.front();
```
kmontemayor2-sc, line 725: qq we do guarantee there's always at least one state rght?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -420,6 +704,148 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+    // Pre-size the per-seed feature map below. Output features may receive up
+    // to maxPPRNodes candidates per channel when residual top-up is on.
+    size_t outputCandidateReserveSize = 0;
+    for (int32_t channelIndex = 0; channelIndex < numChannels; ++channelIndex) {
+        int32_t channelPPRNodeBudget = std::min(channelQuotas[channelIndex], maxPPRNodes);
+        outputCandidateReserveSize += static_cast<size_t>(enableResidualTopUp ? maxPPRNodes : channelPPRNodeBudget);
+    }
+
+    PPRExtractResult extractedTypedPPRByNodeTypeId;
+    std::vector<int32_t> topUpChannelQuotas(static_cast<size_t>(numChannels), maxPPRNodes);
```
kmontemayor2-sc: If these are always the same value then does this even need to be a vector?

Or as we're discussing offline maybe we should ignore this and only use the per-channel quota?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -420,6 +704,148 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+
+                double baseMaxScore = 0.0;
+                for (const auto& nodeAndScore : baseNodesAndScores) {
+                    baseMaxScore = std::max(baseMaxScore, nodeAndScore.second);
+                }
+
+                double outputMaxScore = 0.0;
+                for (const auto& nodeAndScore : outputNodesAndScores) {
+                    outputMaxScore = std::max(outputMaxScore, nodeAndScore.second);
+                }
```
kmontemayor2-sc: nit. can we use `max_element`? https://en.cppreference.com/cpp/algorithm/max_element

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -420,6 +704,148 @@ std::unordered_map<int32_t, std::tuple<torch::Tensor, torch::Tensor, torch::Tens
...
+                    if (selectedNodeIds.find(nodeId) != selectedNodeIds.end()) {
+                        continue;
+                    }
+                    ++selectedNodeCount;
+                    selectedNodeIds.insert(nodeId);
+                    selectedNodes.push_back(nodeId);
+                }
+            }
+
+            if (enableResidualTopUp && selectedNodes.size() > 1) {
```
kmontemayor2-sc: qq why `> 1`?

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc, line 1: Sorry silly question but this code all uses the neighbor cache we have right?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -147,8 +147,65 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
...
+        PPRForwardPush* state = states[channelIndex];
+
+        auto channelFrontierByEdgeTypeId = state->drainQueue();
+        if (!channelFrontierByEdgeTypeId.has_value()) {
+            continue;
+        }
+
+        queueDrainResult.drainedChannelIndices.push_back(static_cast<int32_t>(channelIndex));
+
+        bool fetchBudgetRemaining = maxFetchIterations < 0 || fetchIterationCounts[channelIndex] < maxFetchIterations;
```
kmontemayor2-sc, line 197: should this be `maxFetchIterations <= 0`?

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: reworded this to say the helper writes the emitted edge_attr feature table.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: no, finalized PPR and residuals come from maps keyed by node id, and residual top-up excludes already-selected finalized nodes.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes. The sampler only calls typed extraction when typed channel targets are configured, and it constructs one state per configured channel.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: it's to skip no-op sorting for empty/singleton output

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes, both typed and untyped paths use the PPRForwardPush neighbor cache. Typed PPR has one state/cache per channel; the shared fetch result is pushed back into the active channel states, and each state stores fetched neighbors in its own cache.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: I would keep the current < 0 unbounded convention. Python passes -1 for unbounded; 0 means zero distributed fetch iterations, matching the untyped path’s fetch_iteration_count < max_fetch_iterations behavior.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: removed the separate top-up quota path; residual-aware candidates now go through the same channelTargetCounts selection as finalized PPR candidates

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -147,8 +149,77 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
...
+                states.size(),
+                " states.");
+
+    TypedPPRQueueDrainResult queueDrainResult;
+    // Fetch requests are edge-type scoped: each edge type has its own adjacency
+    // table and destination node type. Grouping by node type would merge
+    // relation-specific fetches that must remain separate.
+    std::unordered_map<int32_t, std::unordered_set<int64_t>> unionedSourceNodeIdsByEdgeTypeId;
+
+    std::vector<std::optional<std::unordered_map<int32_t, torch::Tensor>>> frontiersByChannel;
```
kmontemayor2-sc, line 171: What does each bit of this one mean?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -147,8 +149,77 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
...
+        for (size_t channelIndex = 0; channelIndex < frontierFutures.size(); ++channelIndex) {
+            frontiersByChannel[channelIndex] = frontierFutures[channelIndex].get();
+        }
+    }
+
+    // Merge per-channel drain results serially in channel order so the returned
+    // channel-index lists stay deterministic while the expensive independent
+    // drainQueue() calls above can run concurrently.
+    for (size_t channelIndex = 0; channelIndex < states.size(); ++channelIndex) {
+        auto& channelFrontierByEdgeTypeId = frontiersByChannel[channelIndex];
```
kmontemayor2-sc, line 190: I am still a bit worried we don't have stronger guarantees about channel ordering, I feel like this may cause very odd bugs down the line.

Do you think sorting the dict keys here would be a good idea?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -147,8 +149,77 @@ std::optional<std::unordered_map<int32_t, torch::Tensor>> PPRForwardPush::drainQ
...
+        }
+
+        queueDrainResult.drainedChannelIndices.push_back(static_cast<int32_t>(channelIndex));
+
+        bool fetchBudgetRemaining = maxFetchIterations < 0 || fetchIterationCounts[channelIndex] < maxFetchIterations;
+        if (!fetchBudgetRemaining) {
+            continue;
+        }
+
+        std::vector<int32_t> requestedEdgeTypeIds;
```
kmontemayor2-sc, line 203: should we pre-declare the size here? I wonder if push_back would be slower than using `[]` or what not?

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -116,4 +158,59 @@ class PPRForwardPush {
...
+//   enableResidualTopUp: Whether residual candidates may participate in target
+//                        filling alongside finalized PPR candidates.
+//
+// Expected output: per-node-type tensors. Tuple values match
+// extractTopKWithResidualTopUp:
+//   ids: int64 node IDs, flattened across seeds.
+//   weights: double feature matrix with columns
+//            [best_calibrated_score, per-channel scores..., presence bits...].
+//   valid_counts: int64 count of selected nodes per seed.
+PPRExtractResult extractTypedTopKWithResidualTopUp(const std::vector<PPRForwardPush*>& states,
```
kmontemayor2-sc, line 213: Robot review:

```
  **1. `extractTypedTopKWithResidualTopUp` has no precondition validation — three crash/UB paths reachable from Python.**
  It's exported at `python_ppr_forward_push.cpp:147-151` and re-exported in `gigl_core.__all__`, so all of these are
  user-reachable:
  - `ppr_forward_push.cpp:712` — `states.front()` on an empty vector is UB. `extract_typed_top_k_with_residual_top_up(states=[],
  channel_target_counts=[], enable_residual_topup=False)` reads element 0 of an empty vector.
  - `ppr_forward_push.cpp:593` — `channelTargetCounts[channelIndex]` is indexed for every channel in `[0, states.size())` with no length check.
  `states=[s0, s1], channel_target_counts=[100]` is an unchecked out-of-bounds read.
  - `ppr_forward_push.cpp:716` + `:577` — `std::accumulate` with no non-negativity check; a negative sum reaches
  `reserve(static_cast<size_t>(negative))` → `reserve(~1.8e19)` → `std::length_error`.

  This is asymmetric with the rest of the file: `drainTypedPPRChannelQueues` validates the same size relationship at
  `:155`, and `extractTopKWithResidualTopUp` validates `maxPPRNodes >= 0` at `:660`. The comment at `:707-711` states
  the invariant instead of enforcing it, which CLAUDE.md's fail-fast rule asks us not to do.

  ```cpp
  TORCH_CHECK(!states.empty(), "extractTypedTopKWithResidualTopUp requires at least one channel state.");
  TORCH_CHECK(states.size() == channelTargetCounts.size(), "Expected one channel target count per PPR state, got ",
              channelTargetCounts.size(), " counts for ", states.size(), " states.");
  for (int32_t targetCount : channelTargetCounts) {
      TORCH_CHECK(targetCount >= 0, "channelTargetCounts must be non-negative, got ", targetCount, ".");
  }
```

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -374,11 +445,221 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+                ", which indicates invalid PPR state.");
+
+    // Feature width is 1 + 2C:
+    //   column 0: best calibrated score across channels, used as the scalar
+    //             PPR weight by downstream ranking/sequence construction.
+    //   columns [1, C]: calibrated score for each channel.
+    //   columns [1 + C, 1 + 2C): presence bit for each channel.
+    int32_t numEdgeAttrFeatures = 1 + (2 * numChannels);
+    for (const auto& [nodeId, score] : nodesAndScores) {
+        double calibratedScore = score / maxScore;
```

I feel like `Global overflow (:605-635) ranks by the same scale-free score, so a node with tiny raw PPR mass in a sparse channel
  outranks a node with large raw mass in a dense channel. Overflow systematically prefers weak channels.` is important - is this the behavior we want?

```
Per-channel score calibration makes cross-channel comparison meaningless, so dedup resolves on channel index.
  :488 computes calibratedScore = score / maxScore where maxScore is the max within one (channel, seed, node type)
  — so the top candidate of every channel calibrates to exactly 1.0. :561 then builds
  AttributedCandidate{calibratedScore, calibratedScore, channelIndex, nodeId} (both score fields identical) and
  higherAttributedCandidate compares those normalized values across channels. Consequences:

  - A node that's the strongest candidate in channels 0 and 1 ties on both score fields and always falls through to
  :544 a.channelIndex < b.channelIndex → always attributed to channel 0. The doc claim in
  ppr_forward_push.h:200-203 ("attributed to the channel where it has the strongest calibrated score") is vacuous in
  exactly the case it's meant to resolve.
  - Global overflow (:605-635) ranks by the same scale-free score, so a node with tiny raw PPR mass in a sparse channel
  outranks a node with large raw mass in a dense channel. Overflow systematically prefers weak channels.

  Also, emitted column 0 is raw ppr_score in the untyped path (:687-690) but normalized [0,1] in the typed path —
  so edge_attr[:, 0] means a different quantity depending on config. Suggest keeping raw scores for attribution and
  ranking, calibrating only the emitted feature columns (or normalizing against a per-seed max across all channels).
  Whichever you pick, please state it in the doc comment — it determines what the ratios actually mean.
```

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Added comment, frontiersByChannel stores one drainQueue() result per channel, aligned by vector index with states; nullopt means that channel had no queued nodes. unionedSourceNodeIdsByEdgeTypeId is the deduplicated shared fetch request keyed by edge type, because fetches must remain edge-type scoped.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: I added reserve(channelFrontier.size()). That avoids reallocations while keeping the code simpler than pre-sizing and assigning by index. Since we are collecting map keys, reserve + push_back is the clearer approach and should be effectively equivalent here.

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: I added fail-fast TORCH_CHECKs for empty states, mismatched channelTargetCounts, and negative channel counts. I also added a focused C++ test for these invalid inputs.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: I changed typed extraction to use raw residual-aware PPR scores for attribution, overflow ranking, and column 0. That keeps typed column 0 on the same scale as regular PPR weight and avoids normalizing each channel’s top candidate to 1.0

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: There are two different ordering concerns here. Channel ordering itself does not come from map/dict iteration; it is the index in the states vector, which is built from the Python typed_channel_ratios insertion order and then carried through frontiersByChannel, drainedChannelIndices, and fetchChannelIndices. So sorting map keys would not affect channel ordering.

The place where unordered-map iteration could matter is the list of edge types requested within a given channel. I updated that path to collect and sort requestedEdgeTypeIds before returning them, so the per-channel edge-type list is deterministic too. The unioned fetch map remains a lookup map, and callers should not depend on its iteration order.

mkolodner-sc: Yep, I think sorting the map keys is useful for the edge-type list within each channel, and I updated that path to do so. That gives deterministic edgeTypeIdsByFetchChannel output instead of relying on unordered_map iteration.
I would not use sorted keys to define channel order itself. Channel order is part of the typed-channel API contract and follows the configured Python dict insertion order; the C++ side preserves that via the states vector index. If we changed channel order by sorting, the emitted channel-based PPR feature columns could appear in a different order than the user provided, which would be surprising. So the behavior is: channel order is explicit/stable by configured order, and edge-type IDs within each channel are sorted for determinism.

kmontemayor2-sc: > which is built from the Python typed_channel_ratios insertion order

Yes my concern is this may be non-deterministic (e.g., loading from protos is known to be non-deterministic right), so if we sort the keys on `typed_channel_ratios` then we can guarantee the ordering.

> If we changed channel order by sorting, the emitted channel-based PPR feature columns could appear in a different order than the user provided, which would be surprising

That's a good point though! Should we note that the outputs are tied to the input order and so we expect the channels to have the same (insertion) order across machines?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -374,11 +455,215 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
+                candidateIter->second = candidate;
+            }
+        }
+    }
+
+    std::vector<std::vector<AttributedCandidate>> candidatesByAttributedChannel(
+        static_cast<size_t>(candidatesByChannel.size()));
+    for (const auto& candidateEntry : bestCandidateByNodeId) {
+        const auto& candidate = candidateEntry.second;
+        candidatesByAttributedChannel[candidate.channelIndex].push_back(candidate);
```
kmontemayor2-sc, line 577: qq does this make a copy of each candidate?

kmontemayor2-sc (commented):
Thanks Matt!

kmontemayor2-sc (approved):
Thanks Matt!

yliu2-sc (approved):

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: added a note that typed channel output columns follow the insertion order of typed_channel_ratios, so callers should construct that mapping deterministically if it comes from an unordered source like a proto/map.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Yes, this copies the AttributedCandidate into the per-channel bucket. I think that’s okay here because AttributedCandidate is a small trivially-copyable value type: two doubles and two int32_ts. Avoiding the copy would mean storing pointers/iterators into the unordered_map or restructuring the later selection pass around indices, which would make this more fragile for very little benefit.
The copied candidates are the working set we mutate/reorder with nth_element, so keeping them as values also keeps the selection logic simple and avoids lifetime concerns around references into the map.

PR #725, opened 2026-07-29
Add `GIGL_DEBUG` environment variable to set log level DEBUG

jchmura-sc:
**Purpose of this PR**
Add `GIGL_DEBUG` env variable which, when set, uses `logging.DEBUG` log levels (GCP and locally).


gigl/common/logger.py
```diff
@@ -37,12 +39,17 @@ def _setup_logger(
     ) -> None:
         handler: logging.Handler
         if not logger.handlers:
+            if is_env_flag_enabled(GIGL_DEBUG):
+                log_level = logging.DEBUG
+            else:
+                log_level = logging.INFO
```
kmontemayor2-sc: I'm not sure this file is the best way to alert users when the var is set but I feel like we should at some point?

Maybe do it once per process (gated by some module `_DID_ALERT_FOR_DEBUG`

kmontemayor2-sc (approved):
Thanks!

yliu2-sc (approved):

PR #726, opened 2026-07-30
Add Hydra composition for GiGL configs

kmontemayor2-sc:
Why

  GiGL configs today are monolithic YAML files, so anything shared between pipelines has to be copy-pasted. Our own e2e resource configs carried three near-identical copies of the same infrastructure and preprocessing sections, and they drifted apart the way copies always do. Users maintaining families
  of related pipelines (per-region, per-experiment, per-environment) face the same problem at a larger scale.

  This PR lets task and resource configs use Hydra Defaults Lists so shared fragments live in one place and each pipeline's config declares only what makes it different. Dynamic values (dates, env vars, git hashes) resolve through OmegaConf resolvers instead of being hand-edited before every launch.

  Design decisions

  - Resolve once, at the pipeline entry point. ConfigValidator composes the source configs, then publishes fully resolved snapshots to GCS that every downstream component consumes. Composition can depend on sibling files, the environment, and time-based resolvers — resolving it once and freezing the
    result is what keeps a pipeline run reproducible and every component in agreement about its config. Retries reuse the frozen snapshots rather than re-resolving (the validator follows run-level KFP caching).
  - One reader API. ProtoUtils.read_proto_from_yaml now composes everything it reads. A config without a defaults list composes to itself, so there is no separate "composing" code path for users to know about, and a Hydra-authored config works anywhere a plain one did — not just through the KFP
    pipeline. Composition also snapshots and restores any active Hydra context, so GiGL reads keep working inside user applications that use Hydra themselves.
  - Snapshots carry provenance. The resolved snapshots are machine-written YAML, so source comments cannot survive; each one instead starts with a comment naming the source config it was composed from (including the docker image for container-local sources), so an artifact found in GCS can always be
    traced back.

  The e2e resource configs are migrated to shared fragments as the in-repo example. A new user guide (docs/user_guide/config_guides/hydra_composition.md) covers authoring, remote configs, and the pipeline behavior.

  Testing

  - Unit tests for composition, resolver handling, thread safety, and foreign-Hydra-context restoration, plus ConfigValidator resolution/materialization and compiled-pipeline wiring.
  - assert_yaml_configs_parse runs all 240 repo YAML configs through the composing reader.


deployment/configs/e2e_cicd_resource_config.yaml
```diff
@@ -1,38 +1,9 @@
...
-shared_resource_config:
-  # Resource labels are just compute labels that should be attached to all compute resources spun up by GiGL.
-  # So a practitioner can have a more fine grained understanding of resource utilization and cost of the resources in their GCP billing.
-  # Read more here: https://cloud.google.com/compute/docs/labeling-resources#what-are-labels
-  resource_labels:
-  # We have a 63 character limit for cost_resource_group_tag.
+defaults:
+  - e2e/shared@shared_resource_config
+  - e2e/preprocessor@preprocessor_config
+  - _self_
```
mkolodner-sc: Can provide an in-line break down of what this means here/how to interpret this? Reading this, I'm not sure what the _self_ refers to, ditto elsewhere

mkolodner-sc: Is there a reason we don't need to override other configs here also like trainer/inferencer?

docs/user_guide/config_guides/hydra_composition.md
```diff
@@ -0,0 +1,125 @@
...
+```yaml
+# shared/directed.yaml
+isGraphDirected: true
+```
+
+The package target after `@` places the selected group at the corresponding protobuf field. A fragment containing fields
+for the whole protobuf can use `@_global_`.
+
+Include `_self_` explicitly so it is clear whether values in the primary override group values or are overridden by
+them.
```
mkolodner-sc: This still isn't super clear to me what this means. What scenario would I do this? What if i have some parameters that fall into one group, but some in the other?

gigl/common/utils/hydra_config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    config_root = Path(uri.uri).absolute().parent
+
+    with _COMPOSE_LOCK:
+        # Hydra's compose API owns the process-global GlobalHydra singleton,
+        # so the swap below cannot protect user threads composing concurrently
+        # outside this lock.
+        gh_backup = get_gh_backup()
+        GlobalHydra.instance().clear()
+        try:
```
mkolodner-sc: why do we need a try finally which both have 
```
OmegaConf.register_new_resolver(
                    "now",
                    now_resolver,
                    replace=True,
                )
```

gigl/common/utils/hydra_config.py
```diff
@@ -0,0 +1,77 @@
...
+                replace=True,
+            )
+            restore_gh_from_backup(gh_backup)
+
+    if not isinstance(resolved, dict):
+        raise ValueError(
+            f"Hydra config {config_root / config_name} resolved to "
+            f"{type(resolved).__name__}, expected a mapping."
+        )
+    return cast(dict[str, Any], resolved)
```
mkolodner-sc: Could we avoid using cast here if possible and defer to type narrowing instead?

gigl/orchestration/kubeflow/components/config_validator/component.yaml
```diff
@@ -9,6 +9,9 @@ inputs:
 - {name: cpu_docker_uri, type: String, description: "Uri to dockerized source code compiled for cpu at runtime"}
 - {name: cuda_docker_uri, type: String, description: "Uri to dockerized source code compiled for gpu at runtime"}
 outputs:
+- {name: resolved_task_config_uri, type: String, description: 'Resolved task config URI'}
+- {name: resolved_resource_config_uri, type: String, description: 'Resolved resource config URI'}
+- {name: should_use_glt_backend, type: Boolean, description: 'Whether to use the GLT backend'}
```
mkolodner-sc: For my learning, why do we need to add these separate output fields and what does resolved entail? Is there a way we can avoid doing this?  This seems to be derivable from the input task_config_uri and resource_config_uri, if we could avoid this additional complication here that'd be preferred.

mkolodner-sc (commented):
Thanks Kyle! Left a few comments/questions

PR #730, opened 2026-08-01
Seed the neighbor-sampling RNG per rank instead of leaving it unset

kmontemayor2-sc:
`create_sampling_config` hardcoded `seed=None` into GLT's `SamplingConfig`, so `NeighborSampler` never called `RandomSeedManager::setSeed` -- the only `setSeed` call site in GLT -- and `is_set` stayed false for the whole run.

That is not just a determinism choice. `CPURandomSampler::UniformSample` reads `RandomSeedManager::getInstance().getSeed()` on every call, once per source row of a batch, and discards the result because the `std::mt19937` it feeds is `thread_local static` and already constructed. Unseeded, `getSeed()` constructs a `std::random_device` and draws from it, roughly 5 us of real work per source row for a value that is thrown away.

Measured on an unmodified GLT build, setting any seed took the sampler call from ~1450 us to ~50 us (~29x) with byte-identical neighbor output. In a multi-node GPU inference workload it took per-batch data loading from 47-184 ms to 21-25 ms and left recall unchanged.

The seed is derived from the global rank so ranks do not draw identical samples, mixed with a run-level base that defaults to wall-clock time so successive runs still vary as unseeded sampling did. The derived seed is logged with its base so a run stays reproducible after the fact. `crc32` is used rather than `hash()`, whose string hashing is randomized per process, and its range matches `setSeed(unsigned int)` exactly so no masking is needed.

Callers may still pass an explicit seed, which is used verbatim.


mkolodner-sc (approved):
Wow, nice find Kyle! Thanks :)

yliu2-sc (approved):

PR #731, opened 2026-08-03
Skip zero-fanout distributed sampling RPCs

kmontemayor2-sc:
We can skip RPC generation / etc here and provide a good speedup for sampling.


mkolodner-sc (approved):
Thanks!

yliu2-sc (approved):

PR #733, opened 2026-08-03
Feature Quantization [1/5]: Compile new protos

jchmura-sc:
**Purpose**
The purpose of this PR is to update `PreprocessedMetadata` proto with a new, optional `FeatureQuantizationMetadata` message. This stores the state needed to dequantize features during dataloading.

**Context**
- #733
- #734
- #735
- #736 
- #737 


proto/snapchat/research/gbml/preprocessed_metadata.proto
```diff
@@ -3,6 +3,34 @@ syntax = "proto3";
...
+    // Mean value for positive features, produced by packed bit/code 1.
+    float pos_mean = 2;
+  }
+
+  message FeatureQuantizationMetadata{
+    // Field in output TFRecords that stores packed uint8 features.
+    string packed_feature_key = 1;
+    // Original feature indices stored in packed_feature_key.
+    repeated uint32 quantized_feature_indices = 2;
+    // Stats required to dequantize features for the selected bit width.
```
kmontemayor2-sc, line 27: Somehow I remember there was some other field in this message before? 

```
    // Quantization level bit-width
    uint32 bits = 3;
```

Did we just move that to `MultiBitQuantizationState`?

kmontemayor2-sc (approved):

proto/snapchat/research/gbml/preprocessed_metadata.proto
jchmura-sc: Yep! As you hinted at offline, it could be more robust to embed bits directly within state to prevent misconfiguration (e.g. bits=1 but multibit state enabled).

proto/snapchat/research/gbml/preprocessed_metadata.proto
```diff
@@ -3,6 +3,34 @@ syntax = "proto3";
 package snapchat.research.gbml;
 
 message PreprocessedMetadata{
+  message MultiBitQuantizationState{
+    // Lower clipping bound; dequantized value for linear code 0.
+    float clip_min = 1;
+    // Upper clipping bound; dequantized value for max linear code.
+    float clip_max = 2;
```
kmontemayor2-sc, line 10: Came up in a different review but should we have these all be `double` (float64), to match downstream consumption?

jchmura-sc: Do you think this would make a meaningful difference? Our features are float32? Which part are you referring to downstream?

kmontemayor2-sc: Hmmm - probably not meaningful.

It's going from the stats from dataflow (in python, which is a double) -> float32 here -> loaded back as double in py right?

jchmura-sc: Yeah, I see. I think that at least for this scheme it won't have an impact.

yliu2-sc (approved):

PR #734, opened 2026-08-03
Feature Quantization [2/5]: Core Quantization Ops

jchmura-sc:
**Purpose**
The purpose of this PR is to add the core (standalone) quantize/dequantize functions.

Included is the `FeatureQuantizationMetadata` which is constructed from proto and used during dequantization.

**Context**
- #733
- #734
- #735
- #736 
- #737 


gigl/common/utils/feature_quantization/__init__.py
kmontemayor2-sc, line 1: WDYT about adding some README.md in this folder linking to the paper we're basing this off of - and some details about how we expect people to use it?

Or maybe just the former is good for this dir and then we can have something more user-facing in docs/ when you're done with the work?

gigl/common/utils/feature_quantization/numpy_ops.py
```diff
@@ -0,0 +1,46 @@
...
+arrays and torch is not available. Dequantization lives in torch_ops.py because
+the dataloader collate path operates on torch tensors that may already be on GPU.
+"""
+
+from collections.abc import Mapping
+
+import numpy as np
+
+
+def quantize_ndarray(
```
kmontemayor2-sc, line 11: I had some idea about tying the `quantize` and `dequantize` fns to the `FeatureQuantizationSpec` [1] (or similar). We'd serialize the FQDN and some args to the proto, and the build the object back based on those two.

That way, it's easier to plug in new quantization schemes in the future, IIRC the paper proposed two schemes and we found this one performed better. 

Though that could be be better done as a follow up? WDYT? Maybe we can add a new issue and some TODO?


[1]: https://github.com/Snapchat/GiGL/pull/735/changes#diff-205c4b2c8f6eddc834eac123327f40d8c92bc804b7ef59d1ed7c2edb3bea9216R51

gigl/common/utils/feature_quantization/numpy_ops.py
```diff
@@ -0,0 +1,46 @@
...
+        raise ValueError(f"bits must be one of {valid_bits}, got {bits}")
+    if features.ndim != 2:
+        raise ValueError(f"Expected a 2D feature array, got shape {features.shape}.")
+    if bits == 1:
+        # 1-bit quantization keeps only sign; values restore from neg/pos means.
+        codes = (features > 0).astype(np.uint8)
+    else:
+        # Min-max scale using clipped values and map to integer buckets.
+        levels = (1 << bits) - 1
+        lo, hi = stats["clip_min"], stats["clip_max"]
```
kmontemayor2-sc: QQ where do `clip_min` and `clip_max` come from? Should we document. they're needed in this docstring?

gigl/common/utils/feature_quantization/numpy_ops.py
```diff
@@ -0,0 +1,46 @@
...
+def _pack_codes(codes: np.ndarray, bits: int) -> np.ndarray:
+    """Pack low-bit feature codes high-bits-first along the final dimension."""
+    per_byte = 8 // bits
+    pad = (-codes.shape[-1]) % per_byte
+    if pad:
+        # Pad only the feature dimension of this 2D [row, feature] array.
+        codes = np.pad(codes, ((0, 0), (0, pad)), constant_values=0)
+    # Group the padded feature dimension into chunks that each form one byte.
+    codes = codes.reshape(codes.shape[0], -1, per_byte).astype(np.uint16)
+    shifts = bits * np.arange(per_byte - 1, -1, -1, dtype=np.uint16)
```
kmontemayor2-sc, line 57: is it always safe to recast to `unint16` here?

tests/unit/common/utils/feature_quantization/__init__.py
kmontemayor2-sc, line 1: Should we add a round trip test? Not that we need to check the values - but that what one quantizes the other can dequantize?

gigl/types/graph.py
```diff
@@ -107,6 +108,75 @@ class FeatureInfo:
     dtype: torch.dtype
 
 
+@dataclass(frozen=True)
+class FeatureQuantizationIndexTensors:
+    """Device-local indices used to scatter dequantized and raw features."""
+
+    quantized: torch.Tensor
+    raw: torch.Tensor
```
kmontemayor2-sc, line 116: QQ does this change need to be in this PR?

gigl/types/graph.py
```diff
@@ -107,6 +108,75 @@ class FeatureInfo:
...
+        """Logical feature positions that remain in raw form."""
+        quantized_indices = set(self.quantized_feature_indices)
+        return tuple(i for i in range(self.feature_dim) if i not in quantized_indices)
+
+    @property
+    def raw_feature_dim(self) -> int:
+        """Number of logical features that remain in raw form."""
+        return len(self.raw_feature_indices)
+
+    @lru_cache(maxsize=2)
```
kmontemayor2-sc: qq any reason to keep this cached? Do we expect to call it a bunch?

kmontemayor2-sc: Wait also, we should only ever be calling with one device right? Why have cache size 2?

gigl/common/utils/feature_quantization/numpy_ops.py
```diff
@@ -0,0 +1,46 @@
...
+    if features.ndim != 2:
+        raise ValueError(f"Expected a 2D feature array, got shape {features.shape}.")
+    if bits == 1:
+        # 1-bit quantization keeps only sign; values restore from neg/pos means.
+        codes = (features > 0).astype(np.uint8)
+    else:
+        # Min-max scale using clipped values and map to integer buckets.
+        levels = (1 << bits) - 1
+        lo, hi = stats["clip_min"], stats["clip_max"]
+        clipped = np.clip(features, lo, hi)
```
kmontemayor2-sc: Robot review:

```
NaN features are silently rewritten to clip_min here.

  np.clip passes NaN through, so np.rint(NaN).astype(np.uint8) is an undefined float→uint cast. On this platform it lands on 0, which dequantizes to clip_min:

  quantize_ndarray(np.array([[np.nan, 2.0]]), bits=4, stats={"clip_min": 0.0, "clip_max": 15.0})
  # dequantizes to [[0.0, 2.0]] — the NaN is now indistinguishable from a real 0.0

  The only signal is a RuntimeWarning: invalid value encountered in cast, which a Beam worker will swallow.

  Worth flagging that this is reachable, not theoretical:

  - NULL imputation doesn't catch it — transform/utils.py:79 branches on is not None, and float('nan') is None is False.
  - NaN survives the FloatList → TFRecord → parse round trip, and BigQuery FLOAT64 holds NaN fine.
  - preprocessing_fn is arbitrary user code (lib/types.py:70) with no output validation.
  - Nothing in gigl/ checks isfinite on the data path — the only isnan handling is in model layers, and that's the tell: simplehgn_conv.py:121-127 zeroes NaNs right after projecting node features ("correct NaNs"), same at :136. matmul only makes NaN from NaN, so this has been hit before.

  My concern is that quantization turns a loud failure into a silent one. Today a NaN feature either gets scrubbed by those layers or poisons the loss — either way you find out. Here it becomes a plausible-looking feature value before the model sees it, and those scrubs become dead code for quantized
  features. clip_min also isn't neutral: over a normalized [-3, 3] range, NaN becomes -3.0 — a strong negative signal rather than the 0.0 those layers deliberately picked.

  Suggest failing fast at entry:

  if not np.isfinite(features).all():
      raise ValueError("features must be finite; got NaN or Inf")

  If NaN has to be tolerated, could we give it an explicit reserved code rather than aliasing it onto a range endpoint? Mapping it to clip_min is the one option that's strictly worse than current behavior.
```

gigl/common/utils/feature_quantization/torch_ops.py
```diff
@@ -0,0 +1,49 @@
...
+
+from gigl.types.graph import FeatureQuantizationMetadata
+
+
+def dequantize_torch_tensor(
+    packed_features: torch.Tensor,
+    metadata: FeatureQuantizationMetadata,
+) -> torch.Tensor:
+    """Reconstruct approximate float features from packed uint8 codes."""
+    q = metadata
```
kmontemayor2-sc, line 18: any reason we add this?

gigl/common/utils/feature_quantization/torch_ops.py
```diff
@@ -0,0 +1,49 @@
...
+        return q.clip_min + (codes / levels) * (q.clip_max - q.clip_min)
+
+
+def _unpack_torch_tensor(
+    packed_features: torch.Tensor, *, dim: int, bits: int
+) -> torch.Tensor:
+    per_byte = 8 // bits
+    mask = (1 << bits) - 1
+    # Extract high-bits-first codes from each packed byte.
+    shifts = bits * torch.arange(per_byte - 1, -1, -1, device=packed_features.device)
```
kmontemayor2-sc: Robot review, nitty but do you think it's worth addressing if the fix is funcitonal?

```
Unpack materializes an int64 intermediate 64x the packed size, in the GPU collate path
- **Status**: `open-hardening`
- **Priority**: Medium
- **Location**: `gigl/common/utils/feature_quantization/torch_ops.py:47`
- **Description**: `torch.arange` defaults to int64, so `int16_tensor >> shifts` promotes the whole result to int64 before `bitwise_and` narrows it back. Verified at `bits=1`: a 128,000-byte packed tensor produces an 8,192,000-byte int64 intermediate — 64x expansion. This module's own docstring (lines 3-5) says it runs "in the dataloader collate path, where packed feature data is already represented as torch tensors" and may be on GPU, which is the worst place to take a 64x transient allocation.
- **Fix**: Keep the arithmetic narrow:
  ```python
  shifts = (bits * torch.arange(per_byte - 1, -1, -1, device=packed_features.device)).to(torch.uint8)
  codes = (packed_features.unsqueeze(-1) >> shifts).bitwise_and(mask)
  ```
  uint8 shifts work directly on uint8 input; drop the int16 hop entirely and confirm with a dtype assertion in the test.
```

gigl/common/utils/feature_quantization/__init__.py
jchmura-sc: Yeah that sounds like a good idea. I can do both. User-facing docs I'll add at the end.

gigl/common/utils/feature_quantization/numpy_ops.py
jchmura-sc: I like the idea, and I think it would be a good follow up. I'll add a TODO here as well.

I think it's best as a follow up if we want to e.g. research different quantization protocols.

gigl/common/utils/feature_quantization/numpy_ops.py
jchmura-sc: This is the metadata computed in the preprocessor: https://github.com/Snapchat/GiGL/blob/459b8cbd17143475bea4ebbb85bc013677949b48/gigl/src/data_preprocessor/lib/transform/feature_quantization.py#L82-L108

The `FeatureQuantizationMetadata` (and it's module) depends on torch which we don't have in the preprocessor.

But you bring up a good point - I think i'll unpack these in directly pass them as arguments since this quantization implementation is tied directly to the metadata.

gigl/types/graph.py
jchmura-sc: Note directly, no. It's ultimately used to streamline the scatter based feature write (in dataloading dequantize).

gigl/types/graph.py
jchmura-sc: Yeah this is called in our hot path: https://github.com/Snapchat/GiGL/blob/be4a9fef332e107b479543e0fd2ee1bbf0b3baeb/gigl/distributed/utils/neighborloader.py#L351-L369

But these are always identical for a given metadata instantiation - hence why I cache

jchmura-sc: Correct, we _should_ only be calling this on one device - I set cache size 2 in case of cpu? I can change to cache size 1 since I agree this should only be called on a single device. And I'll add some documentation to explain reasoning behind this

gigl/common/utils/feature_quantization/numpy_ops.py
jchmura-sc: Yes, assuming `bits` is at most 8, `codes` is in [0, (1 << bits) - 1] <= 255 (so technically uint8 is just barely enough), `uin16` will be fine. I'll add a comment to clarify the assumptions we rely on implicitly here.

gigl/common/utils/feature_quantization/torch_ops.py
jchmura-sc: Just to make the code below a little easier to read

gigl/common/utils/feature_quantization/numpy_ops.py
jchmura-sc: Will do, thanks

gigl/common/utils/feature_quantization/torch_ops.py
jchmura-sc: I will confirm and patch this, thank you

tests/unit/common/utils/feature_quantization/__init__.py
kmontemayor2-sc: Did we add this?

kmontemayor2-sc (approved):

tests/unit/common/utils/feature_quantization/__init__.py
kmontemayor2-sc: oh I see, we did, nvm :)

yliu2-sc (approved):

PR #735, opened 2026-08-03
Feature Quantization [3/5]: Preprocessor (write-side) Updates

jchmura-sc:
**Purpose**
The purpose of this PR is to make the (write-side) preprocessor changes for feature quantization:
- `FeatureQuantizationSpec` the API endpoint that the user can specific inside `NodeDataPreprocessingSpec`
- beam transform for computing quantization metadata and serializing to temporary json
- beam transform for quantizing node features based on that metadata (calls into `quantize_ndarray`)
- driver reads the json, constructs `FeatureQuantizationMetadata` proto, and adds to `preprocessed_metadata_pb`

**Context**
- #733
- #734
- #735
- #736 
- #737 


gigl/src/data_preprocessor/lib/transform/feature_quantization.py
```diff
@@ -0,0 +1,274 @@
...
+    packed_feature.value_count.max = 1
+    logger.info(
+        f"Updated transformed schema for feature quantization: dropped "
+        f"{len(spec.feature_keys)} features and added bytes feature "
+        f"{_NODE_PACKED_FEATURE_KEY}."
+    )
+    return quantized_schema
+
+
+def _build_feature_values(
```
jchmura-sc: TBD: Confirming we impact of switching to asymetric clip values. If we proceed then this can be removed

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
```diff
@@ -0,0 +1,281 @@
...
+from gigl.common.utils.feature_quantization.numpy_ops import quantize_ndarray
+from gigl.common.utils.tensorflow_schema import feature_spec_to_feature_index_map
+from gigl.src.data_preprocessor.lib.types import FeatureQuantizationSpec
+
+logger = Logger()
+_NODE_PACKED_FEATURE_KEY: Final[str] = "node_packed_features"
+_SignStats: TypeAlias = tuple[float, int, float, int]
+
+
+def apply_feature_quantization_transform(
```
kmontemayor2-sc, line 22: can we add docstrings here?

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
```diff
@@ -0,0 +1,281 @@
...
+    return quantized_schema
+
+
+def _flatten_feature_values(
+    batch: pa.RecordBatch, quantized_feature_keys: list[str]
+) -> list[float]:
+    return _build_feature_matrix(batch, quantized_feature_keys).ravel().tolist()
+
+
+def _build_feature_matrix(
```
kmontemayor2-sc, line 235: BTW do we have concerns going back and forth from pa to np?

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
```diff
@@ -0,0 +1,281 @@
...
+    batch: pa.RecordBatch, quantized_feature_keys: list[str]
+) -> np.ndarray:
+    key_to_idx: dict[str, int] = {name: i for i, name in enumerate(batch.schema.names)}
```
kmontemayor2-sc, line 238: Does this require is to have the same batch creation across all machines? does pyarrow / beam take care of this?

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+def _multi_bit_stats_from_quantiles(quantiles: list[float]) -> dict[str, float]:
+    if not quantiles:
+        raise ValueError("Cannot compute quantization stats from no values.")
+    quantile_count = len(quantiles) - 1
+    clip_min = float(quantiles[round(0.005 * quantile_count)])
+    clip_max = float(quantiles[round(0.995 * quantile_count)])
+    if clip_max <= clip_min:
+        clip_max = clip_min + 1e-5
```
kmontemayor2-sc, line 265: Is this case for when the data are all clustered tight enough together that you can't really distinguish the quantiles?

gigl/src/data_preprocessor/lib/transform/utils.py
```diff
@@ -362,6 +366,26 @@ def get_load_data_and_transform_pipeline_component(
...
+                apply_feature_quantization_transform(
+                    logical_features=transformed_features,
+                    logical_metadata=logical_metadata,
+                    logical_feature_keys=list(
+                        preprocessing_spec.features_outputs or []
+                    ),
+                    quantization_spec=quantization_spec,
+                    quantization_metadata_path=transformed_features_info.feature_quantization_metadata_path.uri,
+                )
+            )
```
kmontemayor2-sc, line 388: Does this always launch the quantization pipeline? Even if we don't want to quantize anything?

gigl/src/data_preprocessor/lib/types.py
```diff
@@ -48,6 +49,20 @@ class NodeOutputIdentifier(str):
...
+class FeatureQuantizationSpec:
+    """Selects logical feature fields to pack at a fixed bit width."""
+
+    feature_keys: list[str]
+    bits: int
+
+    def __post_init__(self) -> None:
+        if not self.feature_keys:
+            raise ValueError("Feature quantization expects at least one feature key.")
+        if self.bits not in (1, 2, 4, 8):
```
kmontemayor2-sc: nit. let's consider adding some SUPPORTED_QUANTIZATION_BITS in the utils folder so we can reference anywhere (and easier to update if needed)

kmontemayor2-sc (commented):
Thanks Jacob!

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
jchmura-sc: Yeah exactly - I can imagine this happening in case a feature happens to be fully zero-d out for instance. This prevents issues in the [min-max scaling](https://github.com/Snapchat/GiGL/blob/b9558cf3d4a10e22107ab36e832f475e68a6fcf4/gigl/common/utils/feature_quantization/numpy_ops.py#L40)

gigl/src/data_preprocessor/lib/transform/utils.py
jchmura-sc: If we don't want to quantize anything then `FeatureQuantizationSpec` will be `None` (which is automatically true for non-node preprocessing spec), and the quantization pipeline is skipped.

jchmura-sc: And if we happen to instantiate the spec with an empty list of feature to quantize, we fail loudly during construction of the `FeatureQuantizationSpec`

gigl/src/data_preprocessor/lib/types.py
jchmura-sc: Good idea, will do

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
jchmura-sc: Performance wise? I intentionally allow `zero_copy_only=False` which  can fallback to creating a copy in the presence of null values (though this triggers an exception later anyway). In case data is good this should be zero copy.

More broadly, I suspect that row-wise ops (i.e. bit-packing) which we do during quantization would not be possible to do in pyarrow compute

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
jchmura-sc: No, I don't think so.

The feature matrix is ultimately constructed in the `FeatureQuantizationSpec.feature_keys` order, regardless of per-worker batch schema order - they should all ultimately produces the exact same logical feature order.

But we do require that every batch contains every configured feature name in the quantizatino spec.

proto/snapchat/research/gbml/preprocessed_metadata.proto
kmontemayor2-sc, line 33: should we re-base this on main or something? I thought we added these already?

kmontemayor2-sc (commented):
Nice work, Jacob!

gigl/common/utils/feature_quantization/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Utilities for node feature quantization in GiGL."""
+
+from typing import Final
+
+SUPPORTED_QUANTIZATION_BITS: Final[tuple[int, ...]] = (1, 2, 4, 8)
```
mkolodner-sc: nit: I'd have a preference for moving constants-responsible to a different file rather than declaring them inside of `__init__`. if we'd want to import from `gigl.common.utils.feature_quantization` we can then add import from the constants file here

gigl/src/data_preprocessor/lib/transform/feature_quantization.py
```diff
@@ -0,0 +1,303 @@
...
+    Stores the packed bytes in ``node_packed_features`` and computes global
+    quantization statistics with Beam.
+
+    Side Effects:
+        Writes the quantization statistics JSON that ``data_preprocessor.py``
+        reads and serializes into the preprocessing metadata protobuf.
+
+    Args:
+        logical_features: RecordBatches containing the logical feature columns.
+        logical_metadata: Eager or deferred metadata for the logical schema.
```
mkolodner-sc: nit: can you elaborate what eager or deferred mean here

gigl/src/data_preprocessor/lib/types.py
```diff
@@ -48,6 +50,22 @@ class NodeOutputIdentifier(str):
...
 
+@dataclass(frozen=True)
+class FeatureQuantizationSpec:
+    """Selects logical feature fields to pack at a fixed bit width."""
+
+    feature_keys: list[str]
+    bits: int
+
+    def __post_init__(self) -> None:
+        if not self.feature_keys:
```
mkolodner-sc, line 61: Is there any edge case behavior if there are duplicate feature keys in this list?

tests/integration/pipeline/data_preprocessor/feature_quantization_transform_test.py
```diff
@@ -0,0 +1,118 @@
...
+            ),
+            (
+                "single_bit",
+                1,
+                [(-4.0, -2.0), (4.0, 8.0)],
+                {"neg_mean": -3.0, "pos_mean": 6.0},
+            ),
+        ]
+    )
+    def test_apply_feature_quantization_transform_writes_metadata(
```
mkolodner-sc, line 40: Should we add a test for the eager/deferred branches here as well?

mkolodner-sc (approved):
Thanks Jacob!

gigl/src/data_preprocessor/lib/types.py
jchmura-sc: We have a guard on the read side but I'll add add a check here to surface the error fast

tests/integration/pipeline/data_preprocessor/feature_quantization_transform_test.py
jchmura-sc: Good point, will add

yliu2-sc (approved):
Stamp

gigl/src/data_preprocessor/data_preprocessor.py
```diff
@@ -481,6 +482,31 @@ def generate_preprocessed_metadata_pb(
                 feature_dim=feature_dim_output,
                 transform_fn_assets_uri=node_transformed_features_info.transformed_features_transform_fn_assets_path.uri,
             )
+            metadata_path = (
+                node_transformed_features_info.feature_quantization_metadata_path.uri
+            )
+            if tf.io.gfile.exists(metadata_path):
```
kmontemayor2-sc: Hmmm, if the uri is set but doesn't exist, should that be an error?

tests/integration/pipeline/data_preprocessor/feature_quantization_transform_test.py
```diff
@@ -0,0 +1,144 @@
...
+                        | "Create RecordBatches" >> beam.Create(batches),
+                        logical_metadata=logical_metadata,
+                        logical_feature_keys=logical_feature_keys,
+                        quantization_spec=FeatureQuantizationSpec(
+                            feature_keys=logical_feature_keys, bits=bits
+                        ),
+                        quantization_metadata_path=metadata_path,
+                    )
+                )
+                if use_deferred_metadata:
```
kmontemayor2-sc, line 92: nit. I prefer to not have branching logic like this in tests, but I guess it's hard to do otherwise in this case?

kmontemayor2-sc (approved):

gigl/src/data_preprocessor/data_preprocessor.py
jchmura-sc: Agreed, this shouldn't happen, but currently would silently skip quantization. I'll update

tests/integration/pipeline/data_preprocessor/feature_quantization_transform_test.py
jchmura-sc: I don't think its straightforward to do without additional indirection or a duplicate test

PR #736, opened 2026-08-03
Feature Quantization [4/5]: Read updated TFRecords

jchmura-sc:
**Purpose**
The purpose of this PR is to make the read-side changes for feature quantization:
- graph metadata translator reads feature quantization proto and constructs `FeatureQuantizationMetadata`
- tfrecord data build reads quantized node feature tfrecord raw bytes and converts back to uint8

**Context**
- #733
- #734
- #735
- #736 
- #737 


gigl/distributed/utils/serialized_graph_metadata_translator.py
```diff
@@ -30,22 +32,94 @@ def _build_serialized_tfrecord_entity_info(
...
+        quantization_metadata (Optional[FeatureQuantizationMetadata]): Quantization
+            metadata for a node entity, when its features are quantized.
     Returns:
         SerializedTFRecordInfo: Stored metadata for current entity
     """
+    if quantization_metadata is not None:
+        if not isinstance(
+            preprocessed_metadata, PreprocessedMetadata.NodeMetadataOutput
+        ):
+            raise ValueError("Quantization is supported only for node entities.")
```
kmontemayor2-sc: Can we add a TODO to support edge feats here too?

gigl/distributed/utils/serialized_graph_metadata_translator.py
```diff
@@ -30,22 +32,94 @@ def _build_serialized_tfrecord_entity_info(
...
+        )
+
+        feature_keys_to_load: list[str] = []
+        # Identify non-quantized feature fields to load from float storage.
+        for key in preprocessed_metadata.feature_keys:
+            key_indices = set(range(*feature_index[key]))
+            quantized_key_indices = key_indices.intersection(quantized_indices)
+            if not quantized_key_indices:
+                feature_keys_to_load.append(key)
+            elif quantized_key_indices != key_indices:
```
kmontemayor2-sc: Hmmm, do we do some set comparison a bunch in this hot loop?

I guess I'm not sure how much it matters in practice

gigl/distributed/utils/serialized_graph_metadata_translator.py
```diff
@@ -30,22 +32,94 @@ def _build_serialized_tfrecord_entity_info(
...
     )
 
 
+def _build_feature_quantization_metadata(
+    quantized_metadata: PreprocessedMetadata.FeatureQuantizationMetadata,
+    feature_dim: int,
+) -> FeatureQuantizationMetadata:
+    state = quantized_metadata.WhichOneof("state")
+
+    neg_mean = pos_mean = clip_min = clip_max = None
```
kmontemayor2-sc: nit can we add types, I also have some style preference against these multi delcarations

kmontemayor2-sc (approved):
Thanks Jacob!

gigl/distributed/utils/serialized_graph_metadata_translator.py
jchmura-sc: Sounds good, added

gigl/distributed/utils/serialized_graph_metadata_translator.py
jchmura-sc: I don't think this is in the hot path? It only runs once during startup per node type?

But yeah this is defensive, and should not occur since we [only support quantizing scalar features](https://github.com/Snapchat/GiGL/blob/42676f5f8707243222ab9a93438995e73eba2db7/gigl/src/data_preprocessor/lib/transform/feature_quantization.py#L196).

Still, I'll simplify this to a `len` comparison which will be equivalent given that `quantized_key_indices` is a subset of `key_indices`

gigl/common/data/dataloaders.py
```diff
@@ -44,6 +45,10 @@ class SerializedTFRecordInfo:
     feature_dim: int
     # Entity ID Key for current entity. If this is a Node Entity, this must be a string. If this is an edge entity, this must be a Tuple[str, str] for the source and destination ids.
     entity_key: Union[str, Tuple[str, str]]
+    # Packed uint8 feature name to load for the current node entity.
+    packed_feature_key: Optional[str] = None
+    # Number of packed uint8 columns for the current node entity.
+    packed_feature_dim: int = 0
```
mkolodner-sc, line 51: should we have some __post_init__ which validates that the packed feature dim is valid?

gigl/distributed/utils/serialized_graph_metadata_translator.py
```diff
@@ -30,22 +32,100 @@ def _build_serialized_tfrecord_entity_info(
...
+    feature_dim: int,
+) -> FeatureQuantizationMetadata:
+    state = quantized_metadata.WhichOneof("state")
+
+    neg_mean: Optional[float] = None
+    pos_mean: Optional[float] = None
+    clip_min: Optional[float] = None
+    clip_max: Optional[float] = None
+    if state == "single_bit_state":
+        bits = 1
```
mkolodner-sc, line 108: qq -- do we actually end up using bits in any meaningful way?

mkolodner-sc (approved):
Thanks Jacob! Generally LGTM with some small comments :)

gigl/distributed/utils/serialized_graph_metadata_translator.py
jchmura-sc: It's ultimately needed to dequantize:

https://github.com/Snapchat/GiGL/blob/9aaa791d07b134768ea33d2e47b4c113e7e2b55c/gigl/common/utils/feature_quantization/torch_ops.py#L27-L51

gigl/common/data/dataloaders.py
jchmura-sc: Hmm, maybe not super necessary?

The packed feature dim is derived from QuantizationMetadata fields which are valided: https://github.com/Snapchat/GiGL/blob/9aaa791d07b134768ea33d2e47b4c113e7e2b55c/gigl/types/graph.py#L154

And we also catch unexpected packed dimension directly (albeit a bit later) https://github.com/Snapchat/GiGL/blob/9aaa791d07b134768ea33d2e47b4c113e7e2b55c/gigl/common/utils/feature_quantization/torch_ops.py#L20-L25

PR #737, opened 2026-08-04
Feature Quantization [5/5]: Distributed Integration

jchmura-sc:
**Purpose**
The purpose of this PR is to integrate feature quantization into our distributed data, loading, partioning, and sampling setup:
- graph partitioner partitions quantized node feature using node partion book
- neighbor samplers fetch quantized node features over rpc during subgraph hydration
- dataloaders dequantize/reconstruct node feature during collate_fn on gpu
- distributed dataset takes quantization metadata (to pass down to dataloaders) and new quantized node feature store

**Context**
- #733
- #734
- #735
- #736 
- #737 


gigl/distributed/graph_store/remote_dist_dataset.py
```diff
@@ -66,6 +67,29 @@ def fetch_node_feature_info(
...
+
+    @property
+    def node_quantization_metadata(
+        self,
+    ) -> Union[
+        FeatureQuantizationMetadata,
+        dict[NodeType, FeatureQuantizationMetadata],
+        None,
+    ]:
+        return self.fetch_node_quantization_metadata()
```
kmontemayor2-sc: we don't have the @property for the rest of these, specifically so that we can clarify we're making a network call.

Maybe for this small stuff it's not a big deal, but shall we stay in accordance with the existing pattern?

gigl/distributed/utils/neighborloader.py
```diff
@@ -324,6 +335,74 @@ def set_missing_features(
...
+    node_quantization_metadata: Optional[
+        Union[FeatureQuantizationMetadata, dict[NodeType, FeatureQuantizationMetadata]]
+    ],
+) -> tuple[_GraphType, dict[str, torch.Tensor]]:
+    """Materialize packed quantized node features into PyG node feature tensors."""
+    if node_quantization_metadata is None:
+        return data, metadata
+
+    def materialize(
+        store, packed_features: torch.Tensor, q: FeatureQuantizationMetadata
```
kmontemayor2-sc: nit. can we add a type to store?

gigl/distributed/utils/neighborloader.py
```diff
@@ -324,6 +335,74 @@ def set_missing_features(
...
+            if x.size(1) != q.raw_feature_dim:
+                raise ValueError(
+                    f"Expected {q.raw_feature_dim} raw node feature columns before "
+                    f"dequantization, got {x.size(1)}."
+                )
+            out[:, scatter_indices.raw] = x
+        store.x = out
+
+    if isinstance(data, Data):
+        if isinstance(node_quantization_metadata, dict):
```
kmontemayor2-sc, line 417: When is the metadata a dict but the data is a Data?

gigl/distributed/base_sampler.py
```diff
@@ -110,9 +111,28 @@ def __init__(self, *args, **kwargs) -> None:
...
 
+        self.dist_node_quantized_feature: Optional[DistFeature] = None
+        if (
+            self.collect_features
+            and data is not None
+            and getattr(data, "node_quantized_features", None) is not None
+        ):
+            # Mirrors GLT's dist_node_feature initialization:
+            # https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9/graphlearn_torch/python/distributed/dist_neighbor_sampler.py#L162-L167
+            self.dist_node_quantized_feature = DistFeature(
```
kmontemayor2-sc, line 126: Should we default set this to None?

gigl/distributed/dataset_factory.py
```diff
@@ -151,12 +152,37 @@ def _load_and_build_partitioned_dataset(
...
+    if packed_node_features is not None and node_quantization_metadata is not None:
+        packed_features_are_mapping = isinstance(packed_node_features, Mapping)
+        metadata_is_mapping = isinstance(node_quantization_metadata, Mapping)
+        if packed_features_are_mapping != metadata_is_mapping:
+            raise ValueError(
+                "Packed node features and node quantization metadata must both be "
+                "scalar or keyed by node type. "
+                f"Got packed node features of type {type(packed_node_features)} "
+                f"and node quantization metadata {node_quantization_metadata}."
+            )
```
kmontemayor2-sc: Do we store all of the features as the dict like this under the hood for labeled homogeneous graphs?

cc: @mkolodner-sc

gigl/distributed/dist_dataset.py
```diff
@@ -723,6 +760,62 @@ def _initialize_node_features(
...
+        if isinstance(node_quantized_features, Mapping):
+            assert isinstance(node_quantized_feature_id_to_index, Mapping)
+            self._node_quantized_features = {
+                node_type: Feature(
+                    feature_tensor=features_per_node_type,
+                    id2index=node_quantized_feature_id_to_index[node_type],  # ty: ignore[invalid-argument-type] TODO(ty-torch-keyed-access): fix ty false positives for torch-backed keyed container access.
+                    with_gpu=False,
+                    dtype=torch.uint8,
+                )
+                for node_type, features_per_node_type in node_quantized_features.items()
```
kmontemayor2-sc, line 802: Re: the "how do we store under the hood". My understanding is that "labeled homogeneous" we want to treat to users as "homogeneous" (aka no dicts), even if under the hood we store them as dict?

tests/unit/distributed/distributed_neighborloader_test.py
```diff
@@ -777,6 +798,44 @@ def test_isolated_homogeneous_neighbor_loader(
...
+            edge_partition_book=torch.zeros(2),
+            partitioned_edge_index=GraphPartitionData(
+                edge_index=torch.tensor([[0, 1], [1, 0]]), edge_ids=None
+            ),
+            partitioned_node_features=FeaturePartitionData(
+                feats=torch.tensor([[10.0, 20.0], [30.0, 40.0]]),
+                ids=torch.arange(2),
+            ),
+            partitioned_node_quantized_features=FeaturePartitionData(
+                feats=torch.tensor([[48], [144]], dtype=torch.uint8),
```
kmontemayor2-sc: Are these the backed data? Why do we have 2 under the hood? If we use 2 bits here shouldn't we be able to fit all the data in one uint8?

gigl/distributed/graph_store/remote_dist_dataset.py
jchmura-sc: Good point, I'll drop this.

gigl/distributed/base_sampler.py
jchmura-sc: We do, right above?

tests/unit/distributed/distributed_neighborloader_test.py
jchmura-sc: We have two nodes, and one quantized uint8 per node.

Node 0: packed code 48 = 0b110000 = 00 11 00 00 -> quantized values [0, 3]
Node 1: packed code 144 = 0b10010000 = 10 01 00 00 -> quantized values [2, 1]

Since we scatter into quantized feature indices (0, 2), the final node feature tensor will be: 

[[0, 10, 3, 20], [2, 30, 1, 40]]

gigl/distributed/utils/neighborloader.py
jchmura-sc: It's unnecessary - I was previously promoting metadata from scalar to dict in the case of labeled homogenous graphs, only to convert it back here. Thanks for flagging it

gigl/distributed/dist_dataset.py
jchmura-sc: Correct. 

- Internally, we promote to dict in [LoadedGraphTensors.treat_labels_as_edges](https://github.com/Snapchat/GiGL/blob/1cdeb88bacb0c70533e423887781717e69ed6fbb/gigl/types/graph.py#L337)
- `DistDataset` [keeps dictionary form](https://github.com/Snapchat/GiGL/blob/1cdeb88bacb0c70533e423887781717e69ed6fbb/gigl/distributed/dist_dataset.py#L749)
- `DistABLPLoader` will undo back to pure homogenous [during collate](https://github.com/Snapchat/GiGL/blob/1cad8e6dc8eb0778bfd97cab2af33d3eb05dba0d/gigl/distributed/dist_ablp_neighborloader.py#L961)

gigl/distributed/dataset_factory.py
```diff
@@ -217,7 +222,12 @@ def _load_and_build_partitioned_dataset(
     partition_output = partitioner.partition()
 
     logger.info(f"Initializing DistDataset instance with edge direction {edge_dir}")
-    dataset = DistDataset(rank=rank, world_size=world_size, edge_dir=edge_dir)
+    dataset = DistDataset(
+        rank=rank,
+        world_size=world_size,
+        edge_dir=edge_dir,
+        node_quantization_metadata=serialized_graph_metadata.node_quantization_metadata,
```
kmontemayor2-sc, line 229: BTW is there a reason to split node and edge feature quantization up like this?

Not blocking as we've already designed for this but I'm wondering if it could be simplified in the future.

tests/unit/distributed/distributed_neighborloader_test.py
kmontemayor2-sc: Thanks! Can we add some comment with essentially this info? We can also define expected_features in this test so all the data are together?

gigl/distributed/utils/neighborloader.py
```diff
@@ -324,6 +337,74 @@ def set_missing_features(
     return data
 
 
+def materialize_quantized_node_features(
```
kmontemayor2-sc, line 340: Robot review:

**[P1] Restrict lookups to types registered in each feature store**

Heterogeneous graphs may quantize only a subset of their node types. The metadata translator supports this by adding quantization metadata only for types that have it ([translator](https://github.com/Snapchat/GiGL/blob/40e1460f8c2921a75144ba3ea5f7ed0ebe8ba3e9/gigl/distributed/utils/serialized_graph_metadata_translator.py#L152-L182)), and `DistDataset` consequently builds a quantized-feature mapping containing only those types ([dataset initialization](https://github.com/Snapchat/GiGL/blob/40e1460f8c2921a75144ba3ea5f7ed0ebe8ba3e9/gigl/distributed/dist_dataset.py#L793-L803)).

However, the `DistFeature` is constructed with that sparse feature mapping and the complete node partition book ([construction](https://github.com/Snapchat/GiGL/blob/40e1460f8c2921a75144ba3ea5f7ed0ebe8ba3e9/gigl/distributed/base_sampler.py#L120-L134)), while this loop calls `async_get()` for every type in `output.node` ([lookup](https://github.com/Snapchat/GiGL/blob/40e1460f8c2921a75144ba3ea5f7ed0ebe8ba3e9/gigl/distributed/base_sampler.py#L405-L410)). GLT's lookup path directly indexes `local_feature[input_type]` ([`async_get` -> `_local_selecting_get`](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f/graphlearn_torch/python/distributed/dist_feature.py#L176-L184), [direct mapping access](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f/graphlearn_torch/python/distributed/dist_feature.py#L141-L145)), so an unquantized sampled type raises `KeyError`.

The all-to-all branch has the same issue because GLT's `get_all2all()` ultimately iterates every type in `sampler_result.node` ([GLT implementation](https://github.com/alibaba/graphlearn-for-pytorch/blob/88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f/graphlearn_torch/python/distributed/dist_feature.py#L159-L172)).

Please restrict raw and quantized lookups to the types registered in their respective feature mappings, and add a heterogeneous regression test with one quantized type and one raw-only type.

The concrete error path is:

`output.node["raw_only_type"]` -> `async_get(nodes, "raw_only_type")` -> `_local_selecting_get()` -> `_get_local_store()` -> `local_feature["raw_only_type"]` -> `KeyError`.

kmontemayor2-sc (commented):
Thanks Jacob! I the robot feedback seems valid but otherwise this LGTM :)

tests/unit/distributed/distributed_neighborloader_test.py
jchmura-sc: Sounds good, will do

gigl/distributed/dataset_factory.py
jchmura-sc: If by this you mean that eventually `DistDataset` would take a `edge_quantization_metadata`, then yeah we could group them into a single structure in the future.

Overall the main thing is just that the dataloaders need to get access to the metadata somehow.

gigl/distributed/utils/neighborloader.py
jchmura-sc: Thanks for surfacing.

kmontemayor2-sc (approved):
Nice work, thanks Jacob :)

gigl/distributed/utils/neighborloader.py
```diff
@@ -324,6 +337,74 @@ def set_missing_features(
...
+    ],
+) -> tuple[_GraphType, dict[str, torch.Tensor]]:
+    """Materialize packed quantized node features into PyG node feature tensors."""
+    if node_quantization_metadata is None:
+        return data, metadata
+
+    def materialize(
+        store: Union[Data, NodeStorage],
+        packed_features: torch.Tensor,
+        q: FeatureQuantizationMetadata,
```
mkolodner-sc: nit: can we use full names instead of q? Also, docstrings across these functions could be helpful

mkolodner-sc (approved):
Thanks Jacob!

PR #739, opened 2026-08-05
Support environment-based GraphStore discovery

kmontemayor2-sc:
**Scope of work done**


tests/unit/distributed/utils/networking_test.py
kmontemayor2-sc, line 1: A lot of mocks here but this is useful wr.t. testing the actual logic.

gigl/distributed/utils/networking.py
```diff
@@ -273,37 +282,119 @@ def wait_for_readiness_signal(
...
+        ]
+        if missing_environment_keys:
+            missing_keys = ", ".join(missing_environment_keys)
+            raise ValueError(
+                "GraphStore topology environment configuration is incomplete. "
+                f"Missing required environment variables: {missing_keys}."
+            )
+        return _get_graph_store_shape_and_readiness_from_environment()
+
+    if is_currently_running_in_vertex_ai_job():
```
mkolodner-sc, line 338: A few questions
- What settings are we expecting to use the environment keys?
- how does a user currently set them when launching a pipeline? 
- If we are in a VAI job, would the cluster spec be what determines the setup, or would it be the env? Moreso, if both happen to be true, currentl env keys take precedent here. Should this be the case? It might be useful to have a warning or error if they are both present but don't agree

mkolodner-sc (commented):
Thanks Kyle! Mostly LGTM, just some questions around how these new environment keys are expected to be used with the current VAI setup.

gigl/distributed/utils/networking.py
kmontemayor2-sc: 1. We use them for custom launcher jobs
2. We expect custom (e.g. ray) launchers to provide them
3. It would first be env and then the cluster spec, I think it makes sense to have this be an override. I think it's probably too complicated to do both and then throw if they don't match. In fact, if a user sets them then I think it's safe to assume they are going to differ from the VAI env.

mkolodner-sc: Gotcha, would it be possible to at least warn if they aren't matching, so we aren't completely silently selecting one over the other?

mkolodner-sc (approved):
Thanks Kyle!

yliu2-sc (approved):

PR #740, opened 2026-08-05
Add PPR min-hop edge attribute

mkolodner-sc:
## Summary

Adds hop-distance metadata to PPR sampler outputs so downstream models can distinguish how close each sampled node was to the anchor during forward push.

For untyped PPR, `edge_attr` is now `[ppr_score, min_hop]`.

For typed/channel PPR, `edge_attr` is now `[best_score, global_min_hop, channel_scores..., channel_min_hops..., channel_presence_bits...]`.

`min_hop` is emitted as the raw minimum discovered hop count, e.g. `1` for 1-hop, `2` for 2-hop, `4` for 4-hop.

## Why

This lets models use PPR rank/score together with structural proximity. In typed PPR, the global min-hop is always available in a fixed position, while per-channel min-hop values allow relation-specific proximity features when needed.

## Profiling

A local synthetic C++ PPR benchmark compared this branch against `origin/main`.

| Variant | Untyped mean | Typed mean |
|---|---:|---:|
| `origin/main` | `0.495s` | `0.251s` |
| This branch | `0.478s` | `0.253s` |

The optimized implementation is roughly baseline in this synthetic benchmark:

- Untyped: approximately baseline, slightly faster in this run.
- Typed/channel PPR: about `+1%` versus `origin/main`, effectively noise-level for this benchmark.

## Testing

- `git diff --check`
- `python3 -m py_compile gigl/distributed/dist_ppr_sampler.py gigl/distributed/sampler_options.py tests/unit/distributed/dist_ppr_sampler_test.py`
- C++ formatting check on modified PPR forward-push files

### DIFF
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -468,15 +468,15 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
     }
 }
 
-static int32_t getNodeMinHop(const SeedNodeTypeState& nodeTypeState, int32_t nodeId) {
+static double getNodeHopProximity(const SeedNodeTypeState& nodeTypeState, int32_t nodeId) {
     auto residualStateIter = nodeTypeState.residualStates.find(nodeId);
     TORCH_CHECK(residualStateIter != nodeTypeState.residualStates.end(),
                 "PPR extraction expected a discovered hop for emitted node ",
                 nodeId,
                 ".");
-    int32_t minHop = residualStateIter->second.minHop;
-    TORCH_CHECK(minHop >= 0, "PPR min-hop must be non-negative, got ", minHop, ".");
-    return minHop;
+    int32_t hop = residualStateIter->second.minHop;
+    TORCH_CHECK(hop >= 0, "PPR discovered hop must be non-negative, got ", hop, ".");
+    return 1.0 / (1.0 + static_cast<double>(hop));
 }
 
 // Owns the packed edge_attr layout used while merging typed PPR channels.
@@ -488,7 +488,7 @@ static int32_t getNodeMinHop(const SeedNodeTypeState& nodeTypeState, int32_t nod
 //
 // The width is 2 + (3 * numChannels):
 //   - 2 global columns:
-//       [best_score, min_hop]
+//       [best_score, hop_proximity]
 //     These preserve the regular PPR edge_attr contract at the front of the row.
 //   - 3 columns per typed channel:
 //       [channel_score, channel_hop_proximity, channel_presence]
@@ -496,7 +496,7 @@ static int32_t getNodeMinHop(const SeedNodeTypeState& nodeTypeState, int32_t nod
 //     hop proximity gives models a bounded closeness signal, and the presence bit
 //     is the explicit channel-reachability mask.
 //
-// Channel hop proximity is 1 / (1 + min_hop) when a channel reaches the node:
+// Hop proximity is 1 / (1 + hop) when a channel reaches the node:
 // anchor=1.0, 1-hop=0.5, 2-hop ~= 0.333, and so on. Missing channels remain 0,
 // which is finite, bounded, and never looks closer than a reached channel. For
 // present channels, callers can recover the original hop count as
@@ -512,39 +512,31 @@ public:
         return kNumGlobalFeatures + (kNumPerChannelFeatures * _numChannels);
     }
 
-    // New node feature vectors start empty. All columns can start at 0 because
-    // the first channel write initializes the global min-hop before any
-    // cross-channel min comparison is needed.
+    // New node feature vectors start empty. Scores, proximities, and presence
+    // bits all use 0 as their absent-channel value.
     [[nodiscard]] std::vector<double> makeFeatureVector() const {
         return std::vector<double>(numFeatures(), 0.0);
     }
 
-    // Merge one channel's emitted score/hop into an existing node feature row.
+    // Merge one channel's emitted score/proximity into an existing node feature row.
     //
     // A node can be seen in multiple typed channels. The scalar score column keeps
     // the best emitted PPR score for downstream consumers that expect one weight,
-    // while the global hop keeps the nearest discovered distance.
-    void updateScores(std::vector<double>& features,
-                      int32_t channelIndex,
-                      double score,
-                      int32_t minHop,
-                      bool isFirstChannelForNode) const {
-        // The extraction loop bounds channelIndex, and getNodeMinHop validates
-        // minHop before this call. Avoid duplicating those checks per candidate.
-        const double hopFeature = static_cast<double>(minHop);
+    // while the global proximity keeps the closest discovered distance.
+    void updateScores(std::vector<double>& features, int32_t channelIndex, double score, double hopProximity) const {
+        // The extraction loop bounds channelIndex, and getNodeHopProximity
+        // validates the underlying hop before this call.
         features[kBestScoreIndex] = std::max(features[kBestScoreIndex], score);
-        if (isFirstChannelForNode || hopFeature < features[kGlobalMinHopIndex]) {
-            features[kGlobalMinHopIndex] = hopFeature;
-        }
+        features[kGlobalHopProximityIndex] = std::max(features[kGlobalHopProximityIndex], hopProximity);
         features[channelScoreIndex(channelIndex)] = score;
-        features[channelHopProximityIndex(channelIndex)] = 1.0 / (1.0 + hopFeature);
+        features[channelHopProximityIndex(channelIndex)] = hopProximity;
         features[channelPresenceIndex(channelIndex)] = 1.0;
     }
 
 private:
     static constexpr int32_t kBestScoreIndex = 0;
-    static constexpr int32_t kGlobalMinHopIndex = 1;
-    static constexpr int32_t kNumGlobalFeatures = 2;     // best score + global min-hop
+    static constexpr int32_t kGlobalHopProximityIndex = 1;
+    static constexpr int32_t kNumGlobalFeatures = 2;     // best score + global hop proximity
     static constexpr int32_t kNumPerChannelFeatures = 3; // score + hop proximity + presence
 
     [[nodiscard]] int32_t channelScoreIndex(int32_t channelIndex) const {
@@ -599,8 +591,7 @@ static void addTypedPPRSeedFeaturesAndCandidates(const std::vector<std::pair<int
 
     for (const auto& [nodeId, score] : nodesAndScores) {
         auto scoreIter = outputScoresByNodeId.find(nodeId);
-        const bool isFirstChannelForNode = scoreIter == outputScoresByNodeId.end();
-        if (isFirstChannelForNode) {
+        if (scoreIter == outputScoresByNodeId.end()) {
             scoreIter = outputScoresByNodeId.emplace(nodeId, featureLayout.makeFeatureVector()).first;
         }
         auto& scoreFeatures = scoreIter->second;
@@ -608,8 +599,7 @@ static void addTypedPPRSeedFeaturesAndCandidates(const std::vector<std::pair<int
         // Record this node's score for the current channel and mark that the
         // channel reached the node. Current extraction emits one row per node
         // per channel, so no intra-channel merge is needed here.
-        featureLayout.updateScores(
-            scoreFeatures, channelIndex, score, getNodeMinHop(nodeTypeState, nodeId), isFirstChannelForNode);
+        featureLayout.updateScores(scoreFeatures, channelIndex, score, getNodeHopProximity(nodeTypeState, nodeId));
 
         // Keep a per-channel sortable candidate list so target counts can be
         // applied after cross-channel attribution.
@@ -797,7 +787,7 @@ PPRExtractResult PPRForwardPush::extractTopKWithResidualTopUp(int32_t maxPPRNode
             for (const auto& [nodeId, score] : selectedPairs) {
                 flatIds.push_back(static_cast<int64_t>(nodeId));
                 flatFeatureValues.push_back(score);
-                flatFeatureValues.push_back(static_cast<double>(getNodeMinHop(nodeTypeState, nodeId)));
+                flatFeatureValues.push_back(getNodeHopProximity(nodeTypeState, nodeId));
             }
             validCounts.push_back(static_cast<int64_t>(selectedPairs.size()));
         }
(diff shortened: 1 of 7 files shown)


gigl/distributed/sampler_options.py
```diff
@@ -41,11 +41,15 @@ class PPRSamplerOptions:
 
     - ``edge_index``: ``[2, N]`` int64 — row 0 is local seed indices, row 1 is local
       neighbor indices.
-    - ``edge_attr``: ``[N]`` float — PPR score for each (seed, neighbor) pair.
-      Typed PPR emits multi-column edge attrs:
-      ``[best_score, channel_scores..., channel_presence_bits...]``.
+    - ``edge_attr``: ``[N, 2]`` float — PPR score and minimum discovered-hop
+      count for each (seed, neighbor) pair: ``[ppr_score, min_hop]``.
+      ``min_hop`` is emitted as ``0`` for the anchor, ``1`` for 1-hop, ``2``
```
kmontemayor2-sc: Hmmm, for typed channels, the min_hop may be 0 if we never reach that node via that channel right? So now we have 0 mean two different things?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -727,12 +775,15 @@ PPRExtractResult extractTypedTopKWithResidualTopUp(const std::vector<PPRForwardP
     int32_t numNodeTypes = firstState->_numNodeTypes;
     auto numChannels = static_cast<int32_t>(states.size());
     int32_t maxPPRNodes = std::accumulate(channelTargetCounts.begin(), channelTargetCounts.end(), 0);
-    // Feature width is 1 + 2C:
+    // Feature width is 2 + 3C:
```
kmontemayor2-sc: This is getting a bit complciated - do is there some easier way to handle this you think?

kmontemayor2-sc: maybe have a class that has some `updateScores` which takes in the output vector?

And can likewise do this math internally?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -276,12 +276,18 @@ void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
...
-                auto residualIter = srcNodeTypeState.residuals.find(sourceNodeId);
-                double sourceResidual = (residualIter != srcNodeTypeState.residuals.end()) ? residualIter->second : 0.0;
+                auto sourceStateIter = srcNodeTypeState.residualStates.find(sourceNodeId);
+                TORCH_CHECK(sourceStateIter != srcNodeTypeState.residualStates.end(),
+                            "PPR push expected residual state for queued node ",
+                            sourceNodeId,
+                            ".");
+                auto& sourceState = sourceStateIter->second;
+                double sourceResidual = sourceState.residual;
+                int32_t sourceHop = sourceState.minHop;
```
kmontemayor2-sc: do we only use this var once?

Do you think it's clearer if we only keep `sourceState` around?

kmontemayor2-sc: Oh do we want to save the value, then set the state to 0?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -448,13 +462,27 @@ static void appendResidualTopUpPairs(const SeedNodeTypeState& nodeTypeState,
...
+                "PPR extraction expected a discovered hop for emitted node ",
+                nodeId,
+                ".");
+    return residualStateIter->second.minHop;
+}
+
+static double getHopFeature(int32_t minHop) {
+    TORCH_CHECK(minHop >= 0, "PPR min-hop must be non-negative, got ", minHop, ".");
+    return static_cast<double>(minHop);
+}
```
kmontemayor2-sc: Do you think it's better / quicker to convert a tensor vs convert each indivudually?

gigl/distributed/sampler_options.py
mkolodner-sc: Good catch. I changed typed per-channel min-hop to use -1 when that channel did not reach the node. Global min_hop remains non-negative (0 anchor, 1, 2, ...), so 0 no longer has two meanings.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Added a small TypedPPRFeatureLayout helper that owns the feature width, column indexing, default initialization, and updateScores.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Simplified, we now keep sourceState, snapshot only sourceResidual, zero the residual state, and read sourceState.minHop directly.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: I removed getHopFeature. Since extraction already builds one flat std::vector<double> before creating the tensor, converting the hop while appending avoids an extra tensor allocation/copy.

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -448,13 +461,105 @@ static void appendResidualTopUpPairs(const SeedNodeTypeState& nodeTypeState,
...
+//     hop proximity gives models a bounded closeness signal, and the presence bit
+//     is the explicit channel-reachability mask.
+//
+// Hop proximity is 1 / (1 + hop) when a channel reaches the node:
+// anchor=1.0, 1-hop=0.5, 2-hop ~= 0.333, and so on. Missing channels remain 0,
+// which is finite, bounded, and never looks closer than a reached channel. For
+// present channels, callers can recover the original hop count as
+// (1 - proximity) / proximity; use the presence bit before applying this
+// inverse because proximity 0 means the channel is missing.
+class TypedPPRFeatureLayout {
```
kmontemayor2-sc: Do you think we should have a new file for this? Can probably be header only

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -448,13 +461,105 @@ static void appendResidualTopUpPairs(const SeedNodeTypeState& nodeTypeState,
...
+    [[nodiscard]] std::vector<double> makeFeatureVector() const {
+        return std::vector<double>(numFeatures(), 0.0);
+    }
+
+    // Merge one channel's emitted score/proximity into an existing node feature row.
+    //
+    // A node can be seen in multiple typed channels. The scalar score column keeps
+    // the best emitted PPR score for downstream consumers that expect one weight,
+    // while the global proximity keeps the closest discovered distance.
+    void updateScores(std::vector<double>& features, int32_t channelIndex, double score, double hopProximity) const {
```
kmontemayor2-sc: Should we check the size is as expected here?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -465,48 +570,36 @@ static void addResidualMassToPPRPairs(const SeedNodeTypeState& nodeTypeState,
...
-        scoreFeatures[0] = std::max(scoreFeatures[0], score);
-        int32_t channelScoreIndex = 1 + channelIndex;
-        int32_t channelPresenceIndex = 1 + numChannels + channelIndex;
-
         // Record this node's score for the current channel and mark that the
         // channel reached the node. Current extraction emits one row per node
         // per channel, so no intra-channel merge is needed here.
-        scoreFeatures[channelScoreIndex] = score;
-        scoreFeatures[channelPresenceIndex] = 1.0;
+        featureLayout.updateScores(scoreFeatures, channelIndex, score, getNodeHopProximity(nodeTypeState, nodeId));
```
kmontemayor2-sc, line 521: doesn't `updateScores` update every time? Should we be doing that? If we saw a node already, then the only update may be that it has a better score?

Like the proxmitity should never get further away right?

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -21,18 +21,23 @@ using NeighborFetchMap = std::unordered_map<int32_t, NeighborFetchTensors>;
 using PPRExtractTensors = std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>;
 using PPRExtractResult = std::unordered_map<int32_t, PPRExtractTensors>;
 
+struct ResidualState {
+    double residual; // unabsorbed mass waiting to push
+    int32_t minHop;  // minimum discovered hop from seed
```
kmontemayor2-sc: If a seed node sees a different seed node in it's traversal, then will we increase the minHop?

IOW is this bit isolated per seed node? 
https://github.com/Snapchat/GiGL/pull/740/changes#diff-ab52b8cce401b72029c9c54e647bdb31032956d64ab90574c33c6529b89633a7R337-R338

gigl-core/tests/ppr_forward_push_test.cpp
```diff
@@ -240,20 +254,26 @@ TEST(PPRForwardPush, ExtractTypedTopKWithResidualTopUpUsesTargetsForResidualRows
...
-    EXPECT_NEAR(featureAccessor[1][2], 0.25, 1e-9);
-    EXPECT_NEAR(featureAccessor[1][3], 0.0, 1e-9);
-    EXPECT_NEAR(featureAccessor[1][4], 1.0, 1e-9);
+    EXPECT_NEAR(featureAccessor[1][1], 0.5, 1e-9);
+    EXPECT_NEAR(featureAccessor[1][2], 0.0, 1e-9);
+    EXPECT_NEAR(featureAccessor[1][3], 0.25, 1e-9);
+    EXPECT_NEAR(featureAccessor[1][4], 0.0, 1e-9);
+    EXPECT_NEAR(featureAccessor[1][5], 0.5, 1e-9);
+    EXPECT_NEAR(featureAccessor[1][6], 0.0, 1e-9);
+    EXPECT_NEAR(featureAccessor[1][7], 1.0, 1e-9);
```
kmontemayor2-sc: FYI I'm not terrible certain these tests are super readable. I think adding some helper like `EXPECT_FOO_NEAR` or what not could make this more understandable?

gigl/distributed/sampler_options.py
```diff
@@ -41,11 +41,20 @@ class PPRSamplerOptions:
...
-    - ``edge_attr``: ``[N]`` float — PPR score for each (seed, neighbor) pair.
-      Typed PPR emits multi-column edge attrs:
-      ``[best_score, channel_scores..., channel_presence_bits...]``.
+    - ``edge_attr``: ``[N, 2]`` float — PPR score and hop proximity for each
+      (seed, neighbor) pair: ``[ppr_score, hop_proximity]``.
+      ``hop_proximity`` is ``1 / (1 + hop)``: ``1.0`` for the anchor, ``0.5``
+      for 1-hop, and so on.
+      Typed PPR emits additional channel columns:
+      ``[best_score, hop_proximity, channel_scores..., channel_hop_proximities...,
+      channel_presence_bits...]``.
```
kmontemayor2-sc: qq is this more like `[best_score, hop_proximity, (channel_score, channel_prox, channel_precense), ...]`?

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Makes sense, I moved TypedPPRFeatureLayout into typed_ppr_feature_layout.h. It’s header-only since it’s just layout/index math plus validation.”

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: added validation for the expected feature row width, and also validate the channel index before writing.

gigl/distributed/sampler_options.py
mkolodner-sc: Sure, updated!

gigl-core/tests/ppr_forward_push_test.cpp
mkolodner-sc: I added expectTypedPPRFeatureRowNear(...) so the tests assert named feature groups instead of raw column offsets.”

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: Yes, this is isolated per seed. The residual/min-hop state is keyed by seed index, so seed 0 reaching seed 1 only affects seed 0’s traversal state. Seed 1’s own anchor stays hop 0 / proximity 1.0. I added a test for this.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Changed this to be monotonic: best score, channel score, global proximity, and channel proximity all use max. Since proximity is higher when closer, this prevents a later observation from making the feature look farther away. Presence is still just set to 1.0.

gigl-core/core/sampling/typed_ppr_feature_layout.h
```diff
@@ -0,0 +1,113 @@
...
+        features[channelHopProximityIndex(channelIndex)] =
+            std::max(features[channelHopProximityIndex(channelIndex)], hopProximity);
+        features[channelPresenceIndex(channelIndex)] = 1.0;
+    }
+
+private:
+    static constexpr int32_t kBestScoreIndex = 0;
+    static constexpr int32_t kGlobalHopProximityIndex = 1;
+    static constexpr int32_t kNumGlobalFeatures = 2;     // best score + global hop proximity
+    static constexpr int32_t kNumPerChannelFeatures = 3; // score + hop proximity + presence
```
kmontemayor2-sc: Should these be file-level CONSTANTS? Any reason we put them in the class?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -330,13 +336,20 @@ void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
                     // exceeded.
                     auto& dstNodeTypeState = _state[seedIdx][dstNodeTypeId];
                     for (int32_t neighborNodeId : neighborList->get()) {
-                        dstNodeTypeState.residuals[neighborNodeId] += residualPerNeighbor;
+                        const int32_t neighborHop = sourceState.minHop + 1;
+                        auto [residualStateIter, inserted] =
+                            dstNodeTypeState.residualStates.emplace(neighborNodeId, ResidualState{0.0, neighborHop});
+                        auto& residualState = residualStateIter->second;
+                        residualState.residual += residualPerNeighbor;
+                        if (!inserted && neighborHop < residualState.minHop) {
```
kmontemayor2-sc: QQ why only if if not inserted?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -693,14 +706,16 @@ PPRExtractResult PPRForwardPush::extractTopKWithResidualTopUp(int32_t maxPPRNode
 
             for (const auto& [nodeId, score] : selectedPairs) {
                 flatIds.push_back(static_cast<int64_t>(nodeId));
-                flatWeights.push_back(score);
+                flatFeatureValues.push_back(score);
+                flatFeatureValues.push_back(getNodeHopProximity(nodeTypeState, nodeId));
```
kmontemayor2-sc, line 709: nit. We can do some vectorized operation here by just setting the min hop, then adjusting the appropraite columns after we built the tensor right?

kmontemayor2-sc (approved):
Curious why we have this but otherwise LGTM :)

https://github.com/Snapchat/GiGL/pull/740/changes#r3738685303

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: The loop is already building a CPU vector row by row, and the scalar proximity conversion seems cheap compared with materializing the tensor. A post-build tensor transform would add another operation and make the intermediate tensor contain raw min-hop even though the external feature contract is proximity-only. I think the current version is clearer and should not introduce meaningful runtime overhead.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: The old inserted variable name was misleading -- it indicated whether the residual state was newly created or whether we found an existing state for that node. However, since emplace initializes new states with minHop = neighborHop, we don’t need to branch on that boolean. I simplified this to always update minHop with std::min(current_min_hop, neighborHop), which is a no-op for new states and preserves the shortest discovered hop for existing states.

yliu2-sc (approved):

PR #742, opened 2026-08-06
ppr preserve original edges

mkolodner-sc:


mkolodner-sc:
Superseded by #743, which is rebuilt from current main on mkolodner-sc/ppr-preserve-original-edges and excludes the old typed-PPR stack.

PR #743, opened 2026-08-06
Preserve fetched original edges for hetero PPR

mkolodner-sc:
## Summary

This PR adds opt-in preservation of original graph edges for heterogeneous PPR sampling.

When `include_original_edges_in_ppr_subgraph=True`, the PPR sampler now emits original graph edges whose endpoints are already present in the PPR-selected node set and whose adjacency rows were fetched during PPR traversal. The default remains `False`, which is the more faithful behavior relative to PyG’s `get_ppr` API: output only virtual seed-to-PPR-node edges.

## Changes

- Stores fetched PPR adjacency in the C++ `PPRForwardPush` cache with optional edge IDs.
- Adds C++ extraction for original edges from completed PPR cache state.
- Reuses the same extraction path for typed and untyped heterogeneous PPR.
- Deduplicates shared fetched rows across typed PPR channel states.
- Keeps Python throughput-friendly:
  - no extra `_sample_one_hop` pass
  - extraction runs via `run_in_executor`
  - pybind wrapper releases the GIL during C++ cache scanning
- Preserves edge IDs when available so existing edge-feature fetching can continue to work.
- Leaves homogeneous `Data` support explicitly unsupported for this flag.
- Does not include GraphTransformer/model consumption changes.

## Output Behavior

For heterogeneous PPR:
- Virtual PPR edges remain under `("seed_type", "ppr", "neighbor_type")`.
- Preserved original graph edges are emitted through the regular sampled-edge channel.
- Original edges follow the same `edge_dir` orientation convention as k-hop sampling.
- The preserved original-edge set is not guaranteed to be a full induced subgraph because it is limited to adjacency rows already fetched during PPR.

## Validation

- Focused distributed sampler tests:
  - `test_typed_ppr_sampler_loader_outputs_channel_attrs`
  - `test_ppr_sampler_can_preserve_original_edges_for_selected_nodes`
- `git diff --check`
- `make -C gigl-core check_format_cpp`
- `ruff check` on touched Python/stub files

## Local Benchmark

Synthetic no-edge-feature dataloader benchmark, flag off vs on:

| Degree | Mean Delta | Mean % Delta | Original Edges / Batch |
| --- | ---: | ---: | ---: |
| 4 | -0.014 ms/batch | -0.13% | ~10.5k |
| 8 | +1.555 ms/batch | +12.83% | ~33.3k |
| 16 | +2.458 ms/batch | +14.12% | ~77.8k |

Average across these cases: approximately `+1.33 ms/batch` mean overhead. While nonzero, this slowdown is relatively negligible.


gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -118,6 +123,38 @@ static PPRExtractResult extractTypedTopKWithResidualTopUpWrapper(const py::seque
...
+
+    py::dict pyResult;
+    for (const auto& [edgeTypeId, tensors] : result) {
+        const auto& [rows, cols, edgeIds] = tensors;
+        py::object edgeIdsObject = py::none();
+        if (edgeIds.has_value()) {
+            edgeIdsObject = py::cast(edgeIds.value());
+        }
+        pyResult[py::int_(edgeTypeId)] = py::make_tuple(rows, cols, edgeIdsObject);
+    }
```
kmontemayor2-sc, line 163: Any reason this whole block can be outside the gil?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -22,6 +22,58 @@ static uint64_t packKey(int32_t nodeId, int32_t edgeTypeId) {
...
+        combine(std::hash<bool>{}(key.hasEdgeId));
+        if (key.hasEdgeId) {
+            combine(std::hash<int64_t>{}(key.edgeId));
+        } else {
+            combine(std::hash<int32_t>{}(key.sourceNodeId));
+            combine(std::hash<int32_t>{}(key.destinationNodeId));
+        }
+        return seed;
+    }
+};
```
kmontemayor2-sc: QQ why do we have two structs here?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -22,6 +22,58 @@ static uint64_t packKey(int32_t nodeId, int32_t edgeTypeId) {
     return (static_cast<uint64_t>(static_cast<uint32_t>(nodeId)) << 32) | static_cast<uint32_t>(edgeTypeId);
 }
 
+static int32_t unpackNodeId(uint64_t key) {
+    return static_cast<int32_t>(static_cast<uint32_t>(key >> 32));
+}
+
+static int32_t unpackEdgeTypeId(uint64_t key) {
+    return static_cast<int32_t>(static_cast<uint32_t>(key));
+}
```
kmontemayor2-sc, line 34: why do we do this double cast?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -22,6 +22,58 @@ static uint64_t packKey(int32_t nodeId, int32_t edgeTypeId) {
...
+
+struct OriginalEdgeDedupeKey {
+    // Original-edge extraction merges cached adjacency from several PPR states
+    // in typed PPR. Those states can fetch overlapping rows, so dedupe at the
+    // emitted-edge level rather than skipping a whole cached row.
+    //
+    // When edge IDs are cached, the edge ID is the strongest identity: two
+    // parallel edges with the same (type, src, dst) but different edge IDs should
+    // both be emitted. When edge IDs are unavailable, fall back to the structural
```
kmontemayor2-sc: > two parallel edges with the same (type, src, dst) but different edge IDs

Can this happen?

It seems odd to have two paths here?

tests/unit/distributed/utils/neighborloader_test.py
```diff
@@ -248,6 +254,58 @@ def test_strip_non_ppr_edge_types(self):
         self.assertIn(_PPR_U2I, result.edge_types)
         self.assertIn(_PPR_U2U, result.edge_types)
 
+    def test_apply_ppr_outputs_with_original_edges_strips_empty_edge_stores(self):
```
kmontemayor2-sc: is this test useful?

gigl/distributed/sampler_options.py
```diff
@@ -137,6 +138,27 @@ class PPRSamplerOptions:
             same mass scale as finalized PPR scores: ``ppr_score + residual``.
             Residual candidates follow the same channel targets as finalized
             PPR candidates.
+        include_original_edges_in_ppr_subgraph: Whether heterogeneous PPR output
```
kmontemayor2-sc: nit. What do we think about like `include_sampled_edges` or something? This name is quite wordy

gigl/distributed/sampler_options.py
```diff
@@ -137,6 +138,27 @@ class PPRSamplerOptions:
...
+            batches should also include original graph edge types already fetched
+            during PPR traversal whose endpoints are both in the PPR-selected node
+            set. This gives downstream models a local typed-edge view over the
+            selected nodes while keeping the PPR budget as the node-selection
+            boundary and avoiding a second graph-store sampling pass. This is a
+            preserved-fetched-edge view, not a post-hoc induced subgraph over
+            every selected node: if a residual/top-up node was selected but never
```
kmontemayor2-sc: This sentence is quite hard for me to read, can we reword it somehow?

kmontemayor2-sc: Is all of this summed up as "Only edges which contributed to producing a PPR score and edge in the output object are included"?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -47,19 +48,28 @@
...
     Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+    PPRCacheState,
 ]
 # Heterogeneous-only view of PPRResult after typed PPR extraction.
 HeteroPPRResult = tuple[
     dict[NodeType, torch.Tensor],
     dict[NodeType, torch.Tensor],
     dict[NodeType, torch.Tensor],
+    PPRCacheState,
 ]
```
kmontemayor2-sc: Should we consider making these dataclasses? Or some CPP struct we expose in python?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -263,6 +294,11 @@ def __init__(
                 raise ValueError(
                     "Typed PPR channel ratios are only supported for heterogeneous PPR sampling."
                 )
+        if include_original_edges_in_ppr_subgraph and self._is_homogeneous:
```
kmontemayor2-sc: Is `self._is_homogeneous` true or false for the "homogeneous with label edge types"?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -750,13 +804,73 @@ def build_ppr_states() -> list[PPRForwardPush]:
...
+        self,
+        node_dict: dict[NodeType, torch.Tensor],
+        ppr_cache_states: list[PPRForwardPush],
+    ) -> tuple[
+        dict[EdgeType, torch.Tensor],
+        dict[EdgeType, torch.Tensor],
+        Optional[dict[EdgeType, torch.Tensor]],
+        dict[EdgeType, list[int]],
+    ]:
+        """Extract original edges over selected nodes from C++ PPR caches.
```
kmontemayor2-sc, line 804: can we add some documentation for what's returned here?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -750,13 +804,73 @@ def build_ppr_states() -> list[PPRForwardPush]:
...
+
+        num_sampled_edges = {
+            edge_type: [int(cols.size(0))] for edge_type, cols in cols_dict.items()
+        }
+        return (
+            rows_dict,
+            cols_dict,
+            edge_dict if self.with_edge else None,
+            num_sampled_edges,
+        )
```
kmontemayor2-sc, line 859: Is it valid to return 2, 3, or 4 items in the tuple here? 

Should we create a new dataclass to return here?

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -40,6 +45,11 @@ struct SeedNodeTypeState {
     std::unordered_set<int32_t> queuedNodes;                   // snapshot captured by drainQueue()
 };
 
+struct CachedNeighborList {
```
kmontemayor2-sc, line 54: Can we be consistent with if we put the struct in the h file, or the cpp file?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -359,6 +438,158 @@ void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
...
+    const auto* firstState = states.front();
+    int32_t numNodeTypes = firstState->_numNodeTypes;
+    const auto& edgeTypeToDestinationNodeTypeId = firstState->_edgeTypeToDstNtypeId;
+
+    for (const auto* state : states) {
+        TORCH_CHECK(state != nullptr, "extractOriginalEdgesFromPPRCaches received a null PPR state.");
+        TORCH_CHECK(state->_numNodeTypes == numNodeTypes,
+                    "All PPR states must share the same node type schema for original-edge extraction.");
+        TORCH_CHECK(state->_edgeTypeToDstNtypeId == edgeTypeToDestinationNodeTypeId,
+                    "All PPR states must share the same edge destination-type schema for original-edge extraction.");
```
kmontemayor2-sc: Can we add the expected / received values here?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -359,6 +438,158 @@ void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
...
+                        " has no source node type for original-edge extraction.");
+            const auto& selectedSourceLocalIds = selectedLocalIdsByNodeType[static_cast<size_t>(sourceNodeTypeId)];
+            const auto& selectedDestinationLocalIds =
+                selectedLocalIdsByNodeType[static_cast<size_t>(destinationNodeTypeId)];
+
+            auto sourceLocalIter = selectedSourceLocalIds.find(sourceNodeId);
+            if (sourceLocalIter == selectedSourceLocalIds.end() || selectedDestinationLocalIds.empty()) {
+                // Preserve only edges whose endpoints are both in the final
+                // PPR-selected node set. A cached row for an unselected source
+                // can still exist because it was expanded before top-k pruning.
```
kmontemayor2-sc, line 510: Let's say we have some graph like:

```
A -> {B, C} -> D
C -> E -> D
```

And for whatever reason A -> B does not end. up as a PPR node, then do we include the `A->B` edge in the output? Should we? That edge still contributes to the `A, PPR, D` edge right, so in some sense it's "used"?

gigl-core/core/sampling/python_ppr_forward_push.cpp
mkolodner-sc: The extraction can run outside the GIL, but the py::dict, py::tuple, py::cast, and Python tensor wrapper construction need the GIL. I added a comment

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: OriginalEdgeDedupeKey defines identity/equality; OriginalEdgeDedupeKeyHash lets std::unordered_set hash that same identity. This is standard C++ plumbing for a custom key.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: It mirrors packKey: first isolate the low 32 bits, then reinterpret them as the original signed int32_t. I added a comment. 

The reason the double cast is safer is that key is a uint64_t packed with the source node in the high 32 bits. A direct uint64_t -> int32_t conversion is implementation-defined when the 64-bit value is larger than INT32_MAX, which will be true for almost any nonzero source node. The intermediate uint32_t makes the truncation to the low 32 bits explicit and well-defined first.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: We needed both previously because edge IDs are optional in cases where we have no edge features. With IDs, dedupe by edge identity; without IDs, dedupe by (edge_type, src, dst). 

Sounds good to just have the path for deduping on (edge_type, src, dst) to simplify this.

tests/unit/distributed/utils/neighborloader_test.py
mkolodner-sc: Removed

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: We omit A -> B. Even if it contributed to mass later, the output is a message-passing graph over selected nodes, not path provenance. Including it would either create a dangling edge or force B into node_dict, which breaks the PPR node-budget boundary.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: It always returns 4, the ternary may have been confusing to read. Will make a dataclass for this.

gigl/distributed/sampler_options.py
mkolodner-sc: Reworded, let me know if there's still concerns here

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: It would be true, but we fail earlier in the BaseDistLoader. Can remove this check to make it more clear as a result

gigl-core/core/sampling/ppr_forward_push.h
mkolodner-sc: Removed the dedicated CachedNeighborList struct entirely instead of keeping another small helper type in the header. The neighbor cache now uses a standard std::pair<std::vector<int32_t>, std::optional<std::vector<int64_t>>>, and the remaining helper structs are implementation-only and stay in the cpp file. This keeps the header limited to public/shared API surface plus required class members.

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc: Oh I see, apparently you can also do:

```
template <>
struct std::hash<Point> {
    std::size_t operator()(const Point& p) const noexcept {
...

std::unordered_set<Point> points; // No hash passed in
```

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -359,6 +427,173 @@ void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
...
+
+    // All states are expected to come from the same sampler invocation. Typed
+    // PPR creates one state per traversal channel, so verify that those channels
+    // agree on the compact node/edge-type IDs before merging their caches.
+    const auto* firstState = states.front();
+    TORCH_CHECK(firstState != nullptr, "extractOriginalEdgesFromPPRCaches received a null first PPR state.");
+    int32_t numNodeTypes = firstState->_numNodeTypes;
+    const auto& edgeTypeToDestinationNodeTypeId = firstState->_edgeTypeToDstNtypeId;
+
+    for (size_t stateIndex = 0; stateIndex < states.size(); ++stateIndex) {
```
kmontemayor2-sc, line 451: qq any reason to not `const auto const * state : states`?

kmontemayor2-sc: We do this below, but I think we want to const right? One to say the pointer is const (e.g. we can't change the address) and the other to say the object that is being pointed to is const.

gigl-core/core/sampling/ppr_forward_push.cpp
kmontemayor2-sc: That makes sense! thanks for explaining :)

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -7,20 +7,26 @@
...
 using NeighborFetchMap = std::unordered_map<int32_t, NeighborFetchTensors>;
 
 // PPR extraction output, keyed by integer node type ID:
 //   ids, weights/edge_attr, valid_counts.
 using PPRExtractTensors = std::tuple<torch::Tensor, torch::Tensor, torch::Tensor>;
 using PPRExtractResult = std::unordered_map<int32_t, PPRExtractTensors>;
 
+// Original-edge extraction output, keyed by integer edge type ID:
+//   rows, cols, optional edge_ids.
+using OriginalEdgeExtractTensors = std::tuple<torch::Tensor, torch::Tensor, std::optional<torch::Tensor>>;
```
kmontemayor2-sc: QQ is there a reason we use tuples here instead of structs?

I thin it's probably more readable to swap to structs but let's do so as a follow up?

gigl-core/core/sampling/ppr_forward_push.h
```diff
@@ -163,7 +174,7 @@ class PPRForwardPush {
     // Neighbor lists keyed by packKey(nodeId, edgeTypeId).
     // Hash map: nodeId is a sparse graph ID from a large graph, so a dense array is
     // impractical (contrast with _state above).  Populated incrementally; avoids re-fetching.
-    std::unordered_map<uint64_t, std::vector<int32_t>> _neighborCache;
+    std::unordered_map<uint64_t, std::pair<std::vector<int32_t>, std::optional<std::vector<int64_t>>>> _neighborCache;
```
kmontemayor2-sc: Ditto on structs here?

gigl-core/core/sampling/python_ppr_forward_push.cpp
```diff
@@ -26,9 +26,14 @@ static void pushResidualsWrapper(PPRForwardPush& state, const py::dict& fetchedB
     for (auto item : fetchedByEtypeId) {
         auto edgeTypeId = item.first.cast<int32_t>();
         auto neighborTensors = item.second.cast<py::tuple>();
+        std::optional<torch::Tensor> edgeIds = std::nullopt;
+        if (py::len(neighborTensors) > 3 && !neighborTensors[3].is_none()) {
+            edgeIds = neighborTensors[3].cast<torch::Tensor>();
+        }
```
kmontemayor2-sc: WE can probably do the cpp equivalent here instead of `py::len` right?

gigl-core/core/sampling/python_ppr_forward_push.cpp
kmontemayor2-sc: Hmmm, so *creating* objects also requires us to take the GIL? I guess that makes sense so we don't double allocate?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -718,9 +753,9 @@ def build_ppr_states() -> list[PPRForwardPush]:
             if not drained_channel_indices:
                 break
 
-            fetched_by_channel: list[
-                dict[int, tuple[torch.Tensor, torch.Tensor, torch.Tensor]]
-            ] = [dict() for _ in ppr_states]
+            fetched_by_channel: list[PPRForwardPushFetchMap] = [
+                dict() for _ in ppr_states
```
kmontemayor2-sc: huh does this pass type check? It should be [PPRForwardPushFetchMap() for _ in ppr_states]` right?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -750,13 +785,81 @@ def build_ppr_states() -> list[PPRForwardPush]:
...
+        edge_dict: dict[EdgeType, torch.Tensor] = {}
+        for edge_type_id, (rows, cols, edge_ids) in extracted_edges.items():
+            edge_type = self._etype_id_to_etype[edge_type_id]
+            output_edge_type = (
+                reverse_edge_type(edge_type) if self.edge_dir == "in" else edge_type
+            )
+            rows_dict[output_edge_type] = rows.to(self.device)
+            cols_dict[output_edge_type] = cols.to(self.device)
+            if edge_ids is not None:
+                edge_dict[output_edge_type] = edge_ids.to(self.device)
```
kmontemayor2-sc: should we pin memory before this? See https://github.com/Snapchat/GiGL/pull/565 for some timing I found here before.

gigl/distributed/sampler_options.py
```diff
@@ -137,6 +138,23 @@ class PPRSamplerOptions:
             same mass scale as finalized PPR scores: ``ppr_score + residual``.
             Residual candidates follow the same channel targets as finalized
             PPR candidates.
+        include_sampled_edges: Whether heterogeneous PPR output batches should
```
kmontemayor2-sc, line 141: Can we add a TODO (somewhere) to consider supporting this for homogeneous graphs?

gigl/distributed/sampler_options.py
```diff
@@ -137,6 +138,23 @@ class PPRSamplerOptions:
...
+            also include original graph edge types alongside virtual PPR edges.
+            The sampler emits only original edges that were fetched during PPR
+            traversal and whose source and destination are both in the final
+            PPR-selected node set. Original edges are emitted through GLT's
```
kmontemayor2-sc: Do we require that the edge was travered? Let's say we have typed traversal `[[user, likes, content], [user, watches, content]]` and then we have the same like and watch edge to a specific piece of content, do we include both edges? Even if we don't use the like edge for PPR for whatever reason?

gigl-core/core/sampling/ppr_forward_push.cpp
```diff
@@ -298,7 +366,7 @@ void PPRForwardPush::pushResiduals(const NeighborFetchMap& fetchedByEtypeId) {
                 for (int32_t edgeTypeId : _nodeTypeToEdgeTypeIds[nodeTypeId]) {
                     auto cachedEntry = _neighborCache.find(packKey(sourceNodeId, edgeTypeId));
                     if (cachedEntry != _neighborCache.end()) {
-                        totalCachedNeighbors += static_cast<int32_t>(cachedEntry->second.size());
+                        totalCachedNeighbors += static_cast<int32_t>(cachedEntry->second.first.size());
```
kmontemayor2-sc: Why go from first to second here? I think this is another benefit of adding structs here.

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -958,19 +1088,18 @@ async def _sample_from_nodes(
...
+                # those edge types are part of the dataset schema.
+                row=sampled_edges.rows,
+                col=sampled_edges.cols,
+                edge=sampled_edges.edge_ids,
                 batch={input_type: input_seeds},
                 num_sampled_nodes={
                     node_type: [nodes.size(0)] for node_type, nodes in node_dict.items()
                 },
-                num_sampled_edges={},
+                num_sampled_edges=sampled_edges.num_sampled_edges,
```
kmontemayor2-sc: Robot comment

kmontemayor2-sc: `num_sampled_nodes` is inconsistent with the one-layer `num_sampled_edges` emitted when `include_sampled_edges=True`.

The root cause is GLT’s [`to_hetero_data`](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/loader/transform.py#L54-L107): it sets `num_hops` to the maximum length of `num_sampled_edges` ([lines 64–69](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/loader/transform.py#L64-L69)), then right-pads each `num_sampled_nodes` vector to `num_hops + 1` ([lines 85–97](https://github.com/alibaba/graphlearn-for-pytorch/blob/main/graphlearn_torch/python/loader/transform.py#L85-L97)).

Here, preserved edges are reported as one layer:

    num_sampled_edges = {
        edge_type: [edge_count],
    }

but every node type is reported as a one-element vector containing *all* output nodes:

    num_sampled_nodes = {
        node_type: [nodes.size(0)],
    }

For example, if a `user` seed PPR-selects two `story` nodes and preserved edges connect `user -> story`, the sampler initially produces:

    num_sampled_nodes = {
        "user": [1],
        "story": [2],
    }
    num_sampled_edges = {
        ("user", "viewed", "story"): [2],
    }

Since the edge metadata has one hop, GLT pads those node counts to:

    "user": [1, 0]
    "story": [2, 0]

This incorrectly classifies both `story` nodes as roots rather than nodes introduced by the represented edge layer. The metadata should instead be:

    num_sampled_nodes = {
        "user": [1, 0],
        "story": [0, 2],
    }

Could we retain `source_dict` and `new_nodes_dict` and emit their separate counts whenever preserved original edges are returned? A regression test with a cross-type PPR result, where the destination type has no input seeds, should assert both the `[0, 2]` value and alignment with the one-element `num_sampled_edges` vector.

gigl-core/core/sampling/ppr_forward_push.cpp
mkolodner-sc: Looked into this, opening namespace std for a .cpp-local helper makes the hash feel more global/reusable than it is; passing the hasher explicitly keeps the dedupe behavior attached to the one container that needs it.

gigl/distributed/sampler_options.py
mkolodner-sc: Clarified the docs. The rule is fetched adjacency row + both endpoints selected. It is not strict path provenance, so if both likes and watches rows were fetched and both endpoints are selected, both can be emitted.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: I added _move_extracted_tensor_to_device, which pins CPU tensors and uses non_blocking=True when the sampler target device is CUDA. I applied it to original-edge rows/cols/edge IDs and the other C++ PPR extraction tensors. CPU paths are unchanged.

tests/unit/distributed/dist_ppr_sampler_test.py
```diff
@@ -867,6 +868,70 @@ def _run_ppr_destination_only_node_type(_: int) -> None:
...
+        ),
+        pin_memory_device=torch.device("cpu"),
+        batch_size=1,
+    )
+
+    datum = next(iter(loader))
+    assert isinstance(datum, HeteroData)
+
+    ppr_edge_type = (USER, "ppr", STORY)
+    original_output_edge_type = reverse_edge_type(USER_TO_STORY)
```
kmontemayor2-sc, line 901: Do we expect users to reverse this edge type? Is that odd?

I think if they have some USER_TO_STORY var around generally, then we should probably let them use that var everywhere?

gigl/distributed/sampler_options.py
```diff
@@ -137,6 +138,25 @@ class PPRSamplerOptions:
...
+        include_sampled_edges: Whether heterogeneous PPR output batches should
+            also include original graph edge types alongside virtual PPR edges.
+            The sampler emits original edges from adjacency rows fetched while
+            running PPR and whose source and destination are both in the final
+            PPR-selected node set; an emitted edge does not have to be the
+            relation that uniquely caused the destination's PPR score. Original
+            edges are emitted through GLT's
+            regular sampled-edge channel, so their final HeteroData edge
+            orientation follows the same ``edge_dir`` convention as k-hop
+            sampling. Homogeneous PPR keeps the default PPR-only output because
```
kmontemayor2-sc: nit. I'm not sure this homogeneous comment is super useful? What information does it give to a user?

gigl-core/src/gigl_core/ppr_forward_push.pyi
```diff
@@ -16,7 +16,11 @@ class PPRForwardPush:
     def drain_queue(self) -> dict[int, torch.Tensor] | None: ...
     def push_residuals(
         self,
-        fetched_by_etype_id: dict[int, tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
+        fetched_by_etype_id: dict[
+            int,
+            tuple[torch.Tensor, torch.Tensor, torch.Tensor]
+            | tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor | None],
```
kmontemayor2-sc, line 22: does this take in three or four items?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -471,11 +502,15 @@ def _extract_ppr_state_top_k(
...
         ) in extracted_results.items():
             node_type = self._ntype_id_to_ntype[node_type_id]
-            # TODO: If these copies become a bottleneck, evaluate
-            # non_blocking=True together with pinned extraction output tensors.
-            node_type_to_flat_ids[node_type] = flat_ids.to(device)
-            node_type_to_flat_weights[node_type] = flat_weights.to(device)
-            node_type_to_valid_counts[node_type] = valid_counts.to(device)
+            node_type_to_flat_ids[node_type] = self._move_extracted_tensor_to_device(
+                flat_ids, device
+            )
```
kmontemayor2-sc: Should we do this for all of the items at once, and then sync cuda? like we do in https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/base_dist_loader.py#L1043-L1049?

kmontemayor2-sc: Note I mean both these three calls, and amongst all of `extracted_items`

kmontemayor2-sc: also in practice, do we set `device` to be cuda here? Or does that happen down stream?

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -937,6 +1071,51 @@ async def _sample_from_nodes(
...
+                            else 0
+                        ),
+                        (
+                            int(new_nodes_dict[node_type].size(0))
+                            if node_type in new_nodes_dict
+                            else 0
+                        ),
+                    ]
+                    for node_type in node_dict
+                }
```
kmontemayor2-sc, line 1087: nit. Can we add a comment for why we do this?

tests/unit/distributed/dist_ppr_sampler_test.py
kmontemayor2-sc: Synced offline, this is the existing behavior.

kmontemayor2-sc (approved):

yliu2-sc (approved):

gigl-core/src/gigl_core/ppr_forward_push.pyi
mkolodner-sc: Python normalizes neighbor fetches to a 4-tuple: (source_nodes, flat_neighbors, counts, optional_edge_ids). The C++ binding accepts either 3 or 4 for compatibility and treats the missing 4th item as None. I added a comment to make that explicit.

gigl/distributed/dist_ppr_sampler.py
mkolodner-sc: Good call, in practice, we do not set this to CUDA in the sampler path. Both colocated and graph-store worker options construct sampler workers with torch.device("cpu"), and create_dist_sampler passes that worker device into the sampler. CUDA placement happens downstream in BaseDistLoader._collate_fn, where the existing loader path moves tensors to self.to_device with non_blocking=True.

I updated the PR accordingly: PPR extraction no longer calls .to(device) at all. It now only translates C++ node/edge type IDs back to GiGL keys and leaves pinning/device transfer to the existing channel/loader pipeline.

PR #744, opened 2026-08-07
Move the Dataproc CI pools from n2d to n2

kmontemayor2-sc:
n2d stocks out in us-central1 often enough to fail SubgraphSampler and SplitGenerator cluster creation, and each stockout is expensive: create_cluster retries six times with a fixed 20 minute sleep, so a run can burn two hours before giving up.

n2 has a deeper pool at the same shape. Both stages keep 16 vCPU and the same memory, so Spark parallelism (shuffle partitions are derived from vCPU count) and executor sizing do not move.

n2 rather than a newer family because local SSD constrains the choice: e2 and n4 have none, and c3/c4 only offer it on machine types ending in `-lssd`, which NumCores.scala derives vCPUs from with a trailing-digit regex and cannot parse. examples/toy_visual_example already runs n2-standard-16 with num_local_ssds: 2.

Only the two CI configs change. examples/MAG240M keeps n2d at 240 replicas, where the shape was chosen for a 240 million node graph.

**Scope of work done**

### DIFF
--- a/deployment/configs/e2e_cicd_resource_config.yaml
+++ b/deployment/configs/e2e_cicd_resource_config.yaml
@@ -33,11 +33,6 @@ preprocessor_config:
     max_num_workers: 128
     machine_type: "n2d-highmem-64"
     disk_size_gb: 300
-# n2 rather than n2d for both Dataproc stages: n2d stocks out in us-central1 often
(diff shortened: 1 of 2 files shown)


mkolodner-sc (approved):
Thanks Kyle!

yliu2-sc (approved):

PR #748, opened 2026-08-14
Jaxtyping Shape Contracts

jchmura-sc:
**Purpose of this PR**

This PR adds targeted, runtime-checkable tensor contracts for: loader and sampler inputs, public model `forward` and `decode` methods, loss interfaces, and task-result containers.

Jaxtyping lets an annotation declare tensor dtype, rank, fixed dimensions, and relationships between named dimensions. For example, `Float[Tensor, "queries embedding_dim"]` and `Float[Tensor, "candidates embedding_dim"]` require matching `embedding_dim`; decoder output `Float[Tensor, "queries candidates"]` then documents both output axes. This makes malformed tensors fail close to the boundary.

The contracts are intentionally test-only. Unit, integration, and E2E launchers install a Jaxtyping hook before test discovery. Listed modules imported afterwards are instrumented; arguments are checked before execution and returns afterwards. An uncaught violation raises `jaxtyping.TypeCheckError`, which fails the test command. Production execution does not enable this mechanism, and this PR does not expose it as a user API.


tests/unit/distributed/sampler_test.py
jchmura-sc, line 1: These text fixtures were flagged as having invalid shapes w.r.t. to the prod path. Updates here pass shape checks.

gigl/distributed/dist_ppr_sampler.py
```diff
@@ -232,12 +233,15 @@ def __init__(
         max_ppr_nodes: int = 50,
         enable_residual_topup: bool = True,
         num_neighbors_per_hop: int = 100_000,
-        degree_tensors: Union[torch.Tensor, dict[NodeType, torch.Tensor]],
+        degree_tensors: Union[
```
jchmura-sc, line 236: An example of where the union type is actually encoding two different shapes without a requirement that they match: the homogenous type (first type) and heterogeneous (second type) need not be equal.

tests/unit/nn/models_test.py
```diff
@@ -59,7 +62,7 @@ def __init__(self):
     def forward(
         self, query_embeddings: torch.Tensor, candidate_embeddings: torch.Tensor
     ) -> torch.Tensor:
-        return query_embeddings + candidate_embeddings
+        return torch.mm(query_embeddings, candidate_embeddings.T)
```
jchmura-sc, line 65: This mirrors the production decoder: `[queries, embedding_dim] @ [embedding_dim, candidates]` returns pairwise `[queries, candidates]` scores. Addition returns embeddings, not decoder scores.

gigl/distributed/dist_ppr_sampler.py
kmontemayor2-sc: In this case, the shape of *all* nodes in the dict shall not be equal, e.x. `{"a": torch.ones(2), "b": torch.ones(5)}` is valid an expected. 

What's the jaxtyping behavior here?

gigl/distributed/sampler.py
```diff
@@ -37,11 +42,15 @@ def __init__(
         self._negative_label_by_edge_types = negative_label_by_edge_types
 
     @property
-    def positive_label_by_edge_types(self) -> dict[EdgeType, torch.Tensor]:
+    def positive_label_by_edge_types(
+        self,
+    ) -> dict[EdgeType, Int[torch.Tensor, "anchors positive_labels_per_anchor"]]:
         return self._positive_label_by_edge_types
```
kmontemayor2-sc, line 48: In this case, is `anchors` bound per instance of ABLPSamplerInput?

e.x.

```
in1 = Input(node: torch.ones(10), pos: {e: torch.zeroes(10, 2))
in2  = Input(node: torch.ones(10), pos: {e: torch.zeroes(10, 3))
```

Can we later distinguish between the label size for different objects? 

My question about dict key sizing from above still stands.

tests/integration/main.py
```diff
@@ -1,5 +1,9 @@
 import sys
 
+from tests.test_assets.runtime_type_checking import install_runtime_typechecking
+
+install_runtime_typechecking()
```
kmontemayor2-sc, line 5: Hmmm, in some of our tests we fork / spawn new processes, would we need to call this again in those?

https://github.com/Snapchat/GiGL/blob/main/tests/unit/distributed/distributed_weighted_sampling_test.py#L415

tests/test_assets/runtime_type_checking.py
```diff
@@ -0,0 +1,88 @@
...
+Production imports do not install this hook.
+"""
+
+from collections.abc import Callable
+from typing import Any, Final, Optional, get_args
+
+from beartype import beartype
+from jaxtyping import AbstractArray, install_import_hook
+
+_SHAPE_CONTRACT_MODULES: Final[tuple[str, ...]] = (
```
kmontemayor2-sc: Hmmm, so we'd need to add this for all source modules we want to test?

Do you think there's a way to enable this for all gigl/ and examples/ and then see if the calls seem as expected?

E.g. add some new `make check_tensor`?

kmontemayor2-sc (commented):
Neat! Thanks for exploring Jacob :) I left some comments :)

I guess for this we'd need to be careful that we only enable runtime shape checking for tests?

gigl/distributed/dist_ppr_sampler.py
jchmura-sc: As I understand it, typing the example like `dict[str, Float[Tensor, "nodes"]]` **would** force all values in the dictionary to have matching shape.

We specifically use `_nodes` (note the  underscore `_`) which effectively _skips_ any shape checks for the type. This allows us to communicate the expected semantics without actually enforcing a shape check (since they need not match).

gigl/distributed/sampler.py
jchmura-sc: This is a good question. Jaxtyping bindings are scoped to single call so your sample code would pass. This property in particular is essentially just validating a 2d integral output tensor.

Jaxtyping does not do any analysis across function calls and object lifetimes. It only validates between input and output tensors of the same function call. 

For something that can check across function calls, I think we'd need to look into static type checks like [pyrefly](https://pyrefly.org/en/docs/tensor-shapes/) which has an experimental feature for tensor shape checks.

tests/integration/main.py
jchmura-sc: I'd have to double check! We specifically install the runtime type check in test module main, so the spawned process should end up re-running it automaitcally since its at module scope

tests/test_assets/runtime_type_checking.py
jchmura-sc: We should be able to directly put package prefixes to make this more maintainable.

It's important to realize that this is a runtime type checker, so **only the code that runs is checked**. So having `make check_tensor` is nice, but in practice we need something to actually run (e.g. the tests) in order for it to do anything.

gigl/distributed/base_sampler.py
```diff
@@ -171,7 +172,7 @@ def _prepare_sample_loop_inputs(
     def _prepare_ablp_inputs(
         self,
         inputs: ABLPNodeSamplerInput,
```
kmontemayor2-sc, line 174: I wonder if we can extend `AbstractArray` so that we can also tie "anchors" into this class? https://docs.kidger.site/jaxtyping/api/advanced-features/#jaxtyping.AbstractArray

e.g. verify that this shape and the shape of `input_seeds` are compatible?

pyproject.toml
```diff
@@ -148,6 +149,7 @@ typing-stubs = [
   "types-tqdm==4.67.0.20250513",
 ]
 test = [
+  "beartype",
```
kmontemayor2-sc: BTW any reason we chose beartype over typeguard which the jax typing docs also recommend?

tests/unit/distributed/sampler_test.py
kmontemayor2-sc: Hmmm, I wonder if this is "wrong"? The code still works as expected right? I'm not sure if it's bad form to treat `[x, 1]` and `[x]` shaped tensors as equivalent, but maybe we could? WDYT?

kmontemayor2-sc: We could use `#num_labels` here right, to indicate the size can be 1? https://docs.kidger.site/jaxtyping/api/advanced-features/#jaxtyping.AbstractArray

tests/test_assets/runtime_type_checking.py
kmontemayor2-sc: [The docs](https://docs.kidger.site/jaxtyping/api/runtime-type-checking/#jaxtyping.install_import_hook) imply just `gigl` should be sufficient here? Did we test this?

tests/integration/main.py
jchmura-sc: Confirming that this auto-runs for spawned subprocesses, so no need to call again (when using `make unit_test_py`)

tests/test_assets/runtime_type_checking.py
jchmura-sc: Yep, updated

tests/unit/distributed/sampler_test.py
jchmura-sc: > I'm not sure if it's bad form to treat `[x, 1]` and `[x]` shaped tensors as equivalent, but maybe we could? WDYT?

I would say that they are broadly equivalent in that they are always broadcast-able to each other, but they could have different behaviors in some cases:
- `torch.cat(..., dim=1)`, or some other op where we specify `dim=1` would break under `[x]`
- `torch.softmax(..., dim=-1)` would produce different normalization for `[x, 1]` or `[x]`
- `torch.stack(...)` would produce 2d or 3d tensor depending on dimension of input.

From a docs perspective it could also help inform whether you need a `.squeeze()` or not. I've had situations where I had missed that, the code ran fine, but the computed loss was completely wrong.

jchmura-sc: > We could use `#num_labels` here right, to indicate the size can be 1? https://docs.kidger.site/jaxtyping/api/advanced-features/#jaxtyping.AbstractArray

Where exactly are you suggesting this? An annotation like `[foo, #num_labels]` allow shape `[x, y]` or `[x, 1]` but not `[x]`.

Also, if a function only has a single typed argument, adding the `#` wouldnt' change anything since `num_labels` (no `#`) can bind to any size.

gigl/distributed/base_sampler.py
jchmura-sc: We can actually do something slightly simpler, using symbol expressions `"{...}"`

Something like

```python
input_seeds: Int64[torch.Tensor, "{inputs.node.shape[0]}"],
```

seems to do what you suggested? This gets resolved and will raise an exception if they are incompatible.

Of course, this only runs when `jaxtyping` hooks is installed - i.e. testing.

gigl/common/data/export.py
```diff
@@ -308,10 +309,10 @@ def __init__(
 
     def add_prediction(
         self,
-        id_batch: torch.Tensor,
-        prediction_batch: torch.Tensor,
+        id_batch: Int64[torch.Tensor, "batch"],
+        prediction_batch: Shaped[torch.Tensor, "batch"],
         prediction_type: str,
-    ):
+    ) -> None:
```
kmontemayor2-sc, line 315: QQ does jaxtyping require the explicit `-> None` here?

gigl/distributed/base_sampler.py
kmontemayor2-sc: Neat!

tests/test_assets/runtime_type_checking.py
```diff
@@ -0,0 +1,125 @@
...
+    Repeated calls are safe. Runtime checking remains scoped to the current
+    process and modules imported after this function runs. A contract violation
+    raises ``jaxtyping.TypeCheckError`` at the call site, so an uncaught
+    violation fails the active test command.
+    """
+    global _import_hook
+    if _import_hook is None:
+        _import_hook = install_import_hook(
+            modules=_SHAPE_CONTRACT_PACKAGES,
+            #typechecker="tests.test_assets.runtime_type_checking.shape_contract_typechecker",
```
kmontemayor2-sc: nit. remove?

kmontemayor2-sc (approved):
Neat! I think I'm fine with these changes:)

BTW, do you know how much more coverage is untyped?

Can we get a review from @yliu2-sc for the GT shapes as well?

gigl/common/data/export.py
jchmura-sc: Nope, the bots added this but it's not required to pass shape checks.

tests/test_assets/runtime_type_checking.py
jchmura-sc: I was experimenting with how many things break when we just use the simpler `typechecker="typeguard.typechecked"` instead of the custom `shape_contract_typechecker`.

We can't really get away with just using the `typeguard.typechecked` since this type checks _everything_ (not just tensor shapes), which [fails in many places](https://console.cloud.google.com/logs/query;query=log_name%3D%22projects%2Fexternal-snap-ci-github-gigl%2Flogs%2Fcloudbuild%22%20resource.type%3D%22build%22%20resource.labels.build_id%3D%2227d826f2-4216-43fe-88ba-4d3452c3820b%22%20timestamp%3E%3D%222026-08-20T17:31:20.963652847Z%22;cursorTimestamp=2026-08-20T17:47:45.705745206Z;duration=PT5M?project=external-snap-ci-github-gigl), due to `__future__ annotations`. Will revert

yliu2-sc (approved):

jchmura-sc:
> BTW, do you know how much more coverage is untyped?

@kmontemayor2-sc  When running python unit tests, **71 functions** across the codebase are shape checked. An agent found **334** eligible functions inside `gigl/` which means our coverage is about 21%

PR #750, opened 2026-08-15
Edge Feature Quantization [2/5]: Preprocessor

jchmura-sc:
## Summary

- Quantizes selected main-edge features during preprocessing.
- Writes packed features and edge quantization metadata.
- Rejects supervision-edge quantization and quantized sampling weights.

## Stack

1. #749 - metadata proto
2. Preprocessor - this PR
3. TFRecord read path and partitioning
4. Distributed loading and materialization
5. User guide

PR #749, opened 2026-08-15
Edge Feature Quantization [1/5]: Metadata Proto

jchmura-sc:
## Summary

- Adds quantization metadata for main-edge features.
- Regenerates the Python and Scala protobuf bindings.

PR #751, opened 2026-08-15
Edge Feature Quantization [3/5]: Read and Partition

jchmura-sc:
## Summary

- Reads packed edge features and their quantization metadata from TFRecords.
- Carries quantized edge tensors through dataset construction and partitioning.
- Keeps this layer independent of distributed loader materialization.

## Stack

1. #749 - metadata proto
2. #750 - preprocessor
3. Read path and partitioning - this PR
4. Distributed loading and materialization
5. User guide


tests/test_assets/distributed/run_distributed_partitioner.py
```diff
@@ -119,6 +144,7 @@ def run_distributed_partitioner(
         (
             output_edge_index,
             output_edge_features,
+            _,
```
jchmura-sc: Use `output_edge_quantized_features` and assert None

gigl/common/data/load_torch_tensors.py
```diff
@@ -119,6 +119,134 @@ class SerializedGraphMetadata:
...
+def _validate_weight_edge_feature_name(
+    edge_entity_info: Union[
+        SerializedTFRecordInfo, dict[EdgeType, SerializedTFRecordInfo]
+    ],
+    weight_edge_feat_name: Optional[Union[str, dict[EdgeType, str]]],
+) -> None:
+    if weight_edge_feat_name is None:
+        return
+
+    configured_weights: list[tuple[EdgeType, str, SerializedTFRecordInfo]]
```
jchmura-sc: Remove this, directly run the validation loop 161-165 and inline it into each conditional branch, so that we don't need `configured_weights` at all

gigl/types/graph.py
kmontemayor2-sc, line 83: To jog my memory: edge weights will be stored here?

gigl/distributed/dataset_factory.py
kmontemayor2-sc, line 1: Robot review: 

```
**Anchor:** `gigl/distributed/dataset_factory.py:234` (or `gigl/distributed/dist_dataset.py`, `build()`)

The read path stops at partitioning. Following the data through:

1. `EdgeDataPreprocessingSpec.feature_quantization_spec` is now accepted, and the
   translator builds `edge_quantization_metadata` and sets the edge
   `feature_dim` to `raw_feature_dim` only (`serialized_graph_metadata_translator.py:60`).
2. The loader decodes the packed uint8 tensor, and `dataset_factory.py:197-199`
   registers it via `register_edge_quantized_features`, so it gets network-partitioned
   into `partition_output.partitioned_edge_quantized_features`.
3. Then it's dropped: `dist_dataset.py` has zero references to
   `partitioned_edge_quantized_features` — `build()` consumes and frees
   `partitioned_node_quantized_features` (lines 995-997) and
   `partitioned_edge_features` (lines 1009-1011) but never touches the edge
   quantized field, and `DistDataset` is constructed with only
   `node_quantization_metadata` (`dataset_factory.py:234`).
```

gigl/common/data/load_torch_tensors.py
```diff
@@ -119,6 +119,93 @@ class SerializedGraphMetadata:
...
+    else:
+        assert isinstance(quantization_metadata, dict)
+        edge_info_by_type: dict[EdgeType, SerializedTFRecordInfo] = (
+            serialized_graph_metadata.edge_entity_info
+        )
+        metadata_by_type: dict[EdgeType, FeatureQuantizationMetadata] = cast(
+            dict[EdgeType, FeatureQuantizationMetadata], quantization_metadata
+        )
+        if isinstance(weight_edge_feat_name, str):
+            edge_type = next(iter(edge_info_by_type))
```
kmontemayor2-sc, line 178: Robot review: 

```
## 3. [bug] `remove_sampling_weight_from_edge_quantization_metadata` binds a str weight name to an arbitrary edge type

**Anchor:** `gigl/common/data/load_torch_tensors.py:173`

In the heterogeneous branch, a `str` weight name is bound to
`next(iter(edge_info_by_type))` — whichever edge type happens to iterate first —
with no check that there is exactly one edge type:

if isinstance(weight_edge_feat_name, str):
    edge_type = next(iter(edge_info_by_type))
    weight_by_type: dict[EdgeType, str] = {edge_type: weight_edge_feat_name}

Concrete failure: `edge_entity_info = {A: infoA, B: infoB}` where only B has
quantization metadata, called with `weight_edge_feat_name="w"`.
`weight_by_type` becomes `{A: "w"}`, so the adjustment loop hits
`weight_by_type.get(B) is None` and returns B's metadata **unadjusted** — even
though loading strips the weight column from B's raw tensor. The metadata's
`feature_dim` and `quantized_feature_indices` are then off by one relative to
the trimmed tensor, and batch reconstruction scatters raw and dequantized
columns into the wrong positions. No error; just wrong features.

`load_torch_tensors_from_tf_record` rejects exactly this input
(`len(edge_entity_info) != 1` check at lines 493-495), but this is a public
standalone helper and enforces no such precondition. Since it exists precisely
so a follow-up can call it, please mirror that `len == 1` check here (or take
the already-normalized `dict[EdgeType, str]` and drop the str form entirely).****
```

gigl/distributed/dist_range_partitioner.py
kmontemayor2-sc: Robot review: 


```
## 5. [bug] The unpartitioned packed edge tensor is never freed — the new cleanup doesn't actually release memory

**Anchor:** `gigl/distributed/dist_range_partitioner.py:316`

del input_data, edge_index, target_indices, edge_feat, edge_weights_tensor

`edge_quantized_features` is missing from this `del` (every other optional
tensor appended to `input_parts` is there), so the full unpartitioned packed
tensor stays alive through the `torch.cat` of the partitioned chunks (lines
376-380) until the function returns. That's peak memory holding the source
tensor + received chunks + concatenated copy simultaneously, for what is often
the largest edge payload — e.g. 250M local edges × 128 packed dims ≈ 32 GB/rank.
OOM risk in exactly the memory-constrained deployments quantization targets.

Also worth noting for both partitioners: `input_parts` (line 285 here, line
1372 in `dist_partitioner.py`) still holds references to all the source
tensors, so even the `del`s that *are* present — including the new
`del edge_quantized_features` at `dist_partitioner.py:1419` — don't actually
free anything until scope exit. The retention via `input_parts` predates this
PR, but this PR's tensor makes it expensive; a `del input_parts` right after
`_partition_by_chunk` returns would make the cleanup real in both files.
```

gigl/distributed/dist_partitioner.py
```diff
@@ -1201,7 +1233,10 @@ def _partition_edge_index_and_edge_features(
         node_partition_book: dict[NodeType, PartitionBook],
         edge_type: EdgeType,
     ) -> Tuple[
-        GraphPartitionData, Optional[FeaturePartitionData], Optional[PartitionBook]
+        GraphPartitionData,
+        Optional[FeaturePartitionData],
+        Optional[FeaturePartitionData],
+        Optional[PartitionBook],
```
kmontemayor2-sc, line 1241: Can be a follow up but should we create some new output dataclass here? the four tuple is kind of big...


cc: @mkolodner-sc

kmontemayor2-sc (commented):
Thanks Jacob!

gigl/distributed/dataset_factory.py
jchmura-sc: Yeah, the edge quantized feature are actually integrated into the distributed setup in the follow up pr: https://github.com/Snapchat/GiGL/pull/752

gigl/types/graph.py
jchmura-sc: No, they are stored in `partitioned_edge_index.weights`, inside of GLT `GraphPartitionData`.

1. Before partitioning we [extract the weight col from the features](https://github.com/Snapchat/GiGL/blob/fbe034ec7c584d3e4fb407770d2ed10d98f3f67d/gigl/common/data/load_torch_tensors.py#L36) and [return the weights in a loaded graph tensor](https://github.com/Snapchat/GiGL/blob/fbe034ec7c584d3e4fb407770d2ed10d98f3f67d/gigl/common/data/load_torch_tensors.py#L557)
2. [register the weights for partitioning](https://github.com/Snapchat/GiGL/blob/fbe034ec7c584d3e4fb407770d2ed10d98f3f67d/gigl/distributed/dataset_factory.py#L189)
3. [partition edge weights along with edge index](https://github.com/Snapchat/GiGL/blob/fbe034ec7c584d3e4fb407770d2ed10d98f3f67d/gigl/distributed/dist_partitioner.py#L1407)

gigl/distributed/dist_partitioner.py
jchmura-sc: I added a todo for this (along with the analogous `_partition_node_features_and_labels`)

gigl/common/data/load_torch_tensors.py
jchmura-sc: Thanks for surfacing - I've added a guard and a test case to exercise this path

kmontemayor2-sc (approved):

gigl/common/data/dataloaders.py
```diff
@@ -433,6 +423,17 @@ def load_as_torch_tensors(
                     shape=[], dtype=tf.int64
                 )
 
+        if (
```
mkolodner-sc, line 426: qq -- was there a reason this was moved?

gigl/common/data/load_torch_tensors.py
```diff
@@ -119,6 +119,98 @@ class SerializedGraphMetadata:
...
+        edge_info = edge_info_by_type[edge_type]
+        raw_column_offset = 0
+        for feature_name in edge_info.feature_keys:
+            if feature_name == weight_feature_name:
+                break
+            feature_spec = edge_info.feature_spec[feature_name]
+            raw_column_offset += feature_spec.shape[-1] if feature_spec.shape else 1
+        weight_logical_index = metadata.raw_feature_indices[raw_column_offset]
+        adjusted_quantized_feature_indices = tuple(
+            quantized_feature_index - 1
```
mkolodner-sc, line 202: Is the adjustment here for the edge weight column? Can we note this somewhere in-line?

mkolodner-sc (approved):
Thanks a lot!

gigl/common/data/dataloaders.py
jchmura-sc: It was previously inside an additional block that checked `if serialized_tf_record_info.is_node_entity:`

Now, it applies to both nodes and edges, so can be shared.

gigl/common/data/load_torch_tensors.py
jchmura-sc: Yes, exactly! I've added a comment to explain

PR #745, opened 2026-08-12
Jchmura/edge feature quantization

jchmura-sc:

PR #752, opened 2026-08-15
Edge Feature Quantization [4/5]: Distributed Materialization

jchmura-sc:
## Summary

- Propagates quantized edge features through `DistDataset`, graph-store, and distributed loaders.
- Materializes reconstructed edge tensors before model consumption.
- Completes the preprocessing-to-loader integration coverage.

## Stack

1. #749 - metadata proto
2. #750 - preprocessor
3. #751 - read path and partitioning
4. Distributed loading and materialization - this PR
5. User guide

## Verification

- Ruff and formatting checks pass.
- Python compilation passes.
- The cumulative implementation matches #745, with its `base_dist_loader.py` hunks applied on current `main` to preserve newer PPR behavior.
- Full tests are blocked in this clean worktree because the local environment lacks Beam, Torch, and pytest dependencies.


gigl/distributed/base_sampler.py
```diff
@@ -451,6 +469,30 @@ async def _collate_fn(
...
                             self.dist_edge_feature.async_get(eids, etype)
                         )
+            if self.dist_edge_quantized_feature is not None and self.with_edge:
+                for etype in self.edge_types:
+                    # Like node features, an edge partition book covers every
+                    # edge type while a feature store may register only some.
+                    if etype not in self.dist_edge_quantized_feature.local_feature:
+                        continue
+                    result_edge_type = (
+                        reverse_edge_type(etype) if self.edge_dir == "in" else etype
```
kmontemayor2-sc, line 479: Does this in conjunction with https://github.com/Snapchat/GiGL/pull/752/changes#diff-26ddc5d5f6c11b1f6dba27e64f5466113df50438de5b6196c279bbb0939f4db0R536-R538 mean we're always flipping the edge type at lease once?

It's slightly confusing the result map has the opposite edge-type reversal rules as the metadata keys? But I assume you have verified this is the correct way to do it.

kmontemayor2-sc (approved):

gigl/distributed/base_sampler.py
jchmura-sc: Yeah it's a bit confusing frankly. 

It boils down to the fact that quantized edge features are transported from the sampler as `#META` entries. We remove `#META` entries before GLT collate since otherwise GLT would treat the packed uint8 tensor as ordinary `edge_attr`. So, GLT reversal is not applied to quantized features, and we key their transport by output_edge_type = reverse_edge_type(etype) if self.edge_dir == "out" else etype`.

During materialization, we use the same `output_edge_type` to pair the quantization metadata with the packed feature tensor and update the corresponding edge in the final `HeteroData` object.

For heterogeneous samples, GLT always reverses the edge type at least once:
- `edge_dir="in"`: result key is `reverse(etype)`, GLT reverses to `etype`
- `edge_dir="out"`: result key is `etype`, GLT reverses to `reverse(etype)`

Quantized features bypass GLT collate so GiGL applies the `output_edge_type` logic to it.

gigl/distributed/dist_dataset.py
```diff
@@ -899,6 +925,43 @@ def _initialize_edge_features(
...
+            partition_book=edge_partition_book,
+            partitioned_data=partitioned_edge_quantized_features,
+        )
+        if features is None or id_to_index is None:
+            logger.info("Found no packed quantized edge features to initialize")
+            return
+        if isinstance(features, Mapping):
+            assert isinstance(id_to_index, Mapping)
+            features = cast(dict[EdgeType, torch.Tensor], features)
+            id_to_index = cast(dict[EdgeType, torch.Tensor], id_to_index)
```
mkolodner-sc: Are we able to a void these casts with type narrowing instead?

mkolodner-sc (approved):
Thanks Jacob!

gigl/distributed/dist_dataset.py
jchmura-sc: Done!

PR #753, opened 2026-08-15
Edge Feature Quantization [5/5]: User Guide

jchmura-sc:
## Summary

- Documents main-edge feature quantization with a complete preprocessing code example.
- Explains edge restrictions, reconstruction, and expected performance effects.
- Makes the configuration scope explicit: each node or edge type independently selects feature keys and bit width, or remains unquantized.

## Stack

1. #749 - metadata proto
2. #750 - preprocessor
3. #751 - read path and partitioning
4. #752 - distributed loading and materialization
5. User guide - this PR

## Verification

- Markdown formatting passes.
- Pre-commit whitespace checks pass.
- The full documentation build is blocked in this clean worktree because the local build cannot locate the Torch CMake package.


docs/user_guide/config_guides/feature_quantization.md
```diff
@@ -48,6 +48,50 @@ def get_nodes_preprocessing_spec(
...
+            feature_quantization_spec=FeatureQuantizationSpec(
+                feature_keys=["match_score", "recency_days"],
+                bits=8,
+            ),
+        ),
+    }
+```
+
+Edge quantization is supported only for references with `EdgeUsageType.MAIN`. Positive and negative supervision edges
+cannot be quantized. A feature used as a sampling weight must also remain raw so the sampler can consume it directly.
```
kmontemayor2-sc, line 84: > A feature used as a sampling weight must also remain raw so the sampler can consume it directly.

Do we enforce this?

kmontemayor2-sc (approved):

docs/user_guide/config_guides/feature_quantization.md
jchmura-sc: Yes: https://github.com/Snapchat/GiGL/blob/116dd1341ba33bd1f9845ef30fc3c93f2e29efcd/gigl/common/data/load_torch_tensors.py#L490-L494


The edge weight column name is supplied at training time: https://github.com/Snapchat/GiGL/blob/116dd1341ba33bd1f9845ef30fc3c93f2e29efcd/gigl/distributed/dataset_factory.py#L368 so we don't have any checks in the preprocessor specifically.

mkolodner-sc (approved):

PR #754, opened 2026-08-17
Convert PPR sampling tuples/pairs to named structs

kmontemayor2-sc:
Replace the load-bearing std::tuple/std::pair types in the PPR sampling path with named structs and expose the C++ result structs to Python as pybind11 databags (attribute access) instead of positional tuples.

- Header: NeighborFetchTensors and PPRExtractTensors become structs (map using-aliases kept); internal std::pair<int32_t,double> becomes an anonymous-namespace NodeScore, with comparators reading named fields.
- Bindings: bind NeighborFetchTensors (with edge_ids=None ctor), PPRExtractTensors, OriginalEdgeExtractTensors, and TypedPPRQueueDrainResult; wrappers return the structs/maps directly. GIL-release strategy is unchanged (convert under GIL, compute released).
- Python consumer: construct NeighborFetchTensors at the fetch site and use attribute access at the unpack sites; drop the mirror tuple aliases.
- Update the .pyi stub and gigl_core exports to match.

No behavior change; existing C++ and Python correctness tests pass.

**Scope of work done**

### DIFF
--- a/gigl-core/core/sampling/ppr_forward_push.cpp
+++ b/gigl-core/core/sampling/ppr_forward_push.cpp
@@ -9,13 +9,22 @@
 #include <future>
 #include <numeric>
 #include <optional>
-#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
 #include <vector>
 
 namespace gigl {
+
+namespace {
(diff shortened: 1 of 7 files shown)


mkolodner-sc (approved):
Thanks Kyle!

yliu2-sc (approved):

PR #755, opened 2026-08-17
Materialize graph transformer attribute-name lists

dsaini2-sc:
## What

`GraphTransformerEncoder.__init__` now materializes its four attribute-name parameters as plain Python lists: `pe_attr_names`, `anchor_based_attention_bias_attr_names`, `anchor_based_input_attr_names`, and `pairwise_attention_bias_attr_names`.

## Why

The encoder stores these parameters on the instance, and the forward path reads them from there. A configuration loader can supply OmegaConf `ListConfig` objects, which TorchDynamo cannot trace — so an encoder constructed from a config could not be compiled with `torch.compile`, even though nothing about the model itself prevents it.

Materializing at construction keeps config objects out of the forward path. Note that converting only the local `anchor_bias_attr_names` / `anchor_input_attr_names` / `pairwise_bias_attr_names` aliases would not fix this: those are used for validation, while it is the parameters themselves that are stored and later read.

## Behaviour

No functional change. Iteration semantics are identical, including the downstream list concatenation at the `ppr_weight` validation.

One normalization: omitted attribute names are now stored as `[]` rather than `None`. Every consumer already handled this — `self._pe_attr_names` and `self._pairwise_attention_bias_attr_names` are read via truthiness checks (1527, 1434), and the remaining reads apply an `or []` guard (1424, 1566, 1569, 1570). `heterodata_to_graph_transformer_input`, which receives three of them, applies the same `or []` to its own arguments. The redundant guards are left in place to keep the diff minimal.

## Testing

`make unit_test_py PY_TEST_FILES="graph_transformer_test.py"` — 99 tests pass. `ty` reports no diagnostics.

Adds `test_config_supplied_attr_names_become_plain_lists`, which builds the encoder from an OmegaConf config and asserts all four stored attributes are plain lists, plus the empty-default case. The existing tests pass plain lists, so they would not have caught a regression here.

Verified that the new test discriminates: constructing the pre-change encoder from the same config retains `ListConfig` in all four attributes, while the patched encoder stores a plain `list` in all four.

### DIFF
--- a/gigl/nn/graph_transformer.py
+++ b/gigl/nn/graph_transformer.py
@@ -1305,9 +1305,19 @@ class GraphTransformerEncoder(nn.Module):
                 "{'none', 'edge_type_linear', 'edge_type_attention'}, "
                 f"got '{relation_message_mode}'"
             )
-        anchor_bias_attr_names = anchor_based_attention_bias_attr_names or []
-        anchor_input_attr_names = anchor_based_input_attr_names or []
-        pairwise_bias_attr_names = pairwise_attention_bias_attr_names or []
+        # Config loaders may pass OmegaConf ListConfig objects, which TorchDynamo cannot
+        # trace. Plain lists iterate identically and keep the forward path traceable.
+        pe_attr_names = list(pe_attr_names or ())
+        anchor_based_attention_bias_attr_names = list(
+            anchor_based_attention_bias_attr_names or ()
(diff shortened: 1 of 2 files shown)


kmontemayor2-sc (approved):
Thanks!

yliu2-sc (approved):

PR #756, opened 2026-08-18
Add host memory introspection utilities

dsaini2-sc:
Adds `gigl/utils/host_memory.py` and its tests. Two new files; no existing code path changes.

## Why

`psutil.virtual_memory().available` reads `/proc/meminfo`, which inside a container reports the **host's** memory rather than the container's limit. A memory budget check that trusts it alone can pass moments before the container is OOM-killed. This module reads the cgroup limit alongside meminfo and believes the smaller.

## What it handles

- Resolves the process's own cgroup rather than the hierarchy root, translated through the mount root, trying v2 then v1.
- Picks the cgroup with the least headroom. Every ancestor constrains the process, so the first limit found walking up is not necessarily the binding one.
- `cgroup_memory_breakdown` separates anonymous memory from reclaimable page cache, which is what decides whether sitting near the limit is survivable.
- `log_stage_memory` logs that split at pipeline phase boundaries.

## Testing

45 tests pass; `ty` and `ruff` clean. The tests build real cgroup hierarchies in temporary directories and read them through the real code path, mocking only cgroup discovery and `psutil`.

### DIFF
--- a/gigl/utils/host_memory.py
+++ b/gigl/utils/host_memory.py
@@ -7,11 +7,11 @@ the smaller.
 
 Most callers want :func:`available_memory_bytes`. The public API:
 
-- :func:`available_memory_bytes` — bytes this process can still allocate; the number to budget against.
-- :func:`cgroup_limit_and_usage` — the binding cgroup ``(limit, current)`` in bytes, or None when unlimited.
-- :func:`cgroup_memory_breakdown` — current usage split by reclaimability (anon, shmem, file, dirty, writeback).
-- :func:`log_stage_memory` — log the memory position after a pipeline stage.
-- :func:`describe_memory` — one line covering both views, for logs where a budget is decided.
+- :func:`available_memory_bytes` -- bytes this process can still allocate; the number to budget against.
+- :func:`cgroup_limit_and_usage` -- the binding cgroup ``(limit, current)`` in bytes, or None when unlimited.
+- :func:`cgroup_memory_breakdown` -- current usage split by reclaimability (anon, shmem, file, dirty, writeback).
+- :func:`log_stage_memory` -- log the memory position after a pipeline stage.
+- :func:`describe_memory` -- one line covering both views, for logs where a budget is decided.
 
 Everything else in the module is the ``/proc`` and cgroup plumbing behind those five.
 """
@@ -58,12 +58,14 @@ def _cgroup_paths() -> list[str]:
     try:
         with open("/proc/self/cgroup") as handle:
             for line in handle:
+                # Each line is `hierarchy-id:controllers:path`. v2 lines are `0::<path>`; v1 lines
+                # name their controllers, and only the one carrying `memory` is relevant here:
+                #   0::/kubepods/burstable/pod1234/5678
+                #   9:memory:/docker/5678
                 fields = line.strip().split(":", 2)
                 if len(fields) != 3:
                     continue
                 hierarchy_id, controllers, path = fields
-                # v2 lines are `0::<path>`; v1 lines name their controllers, and only the one
-                # carrying `memory` is relevant here.
                 if hierarchy_id == "0" or "memory" in controllers.split(","):
                     paths.append(path or "/")
     except OSError:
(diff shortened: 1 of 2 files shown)


tests/unit/utils/host_memory_test.py
```diff
@@ -0,0 +1,388 @@
...
+        Zero is a legitimate answer for a cgroup at its limit, and the upper bound is TOTAL rather
+        than a second reading of AVAILABLE, which moves between calls.
+        """
+        available = host_memory.available_memory_bytes()
+
+        self.assertGreaterEqual(available, 0)
+        self.assertLessEqual(available, host_memory.psutil.virtual_memory().total)
+
+
+class MemoryBreakdownTest(CgroupResolutionTest):
```
zfan3-sc, line 239: nit: inheriting from a concrete test class will inherits its test methods. consider refactor it to move test methods out of the base class

gigl/utils/host_memory.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    parts = [f"[mem] after {stage}:"]
+    limits = cgroup_limit_and_usage()
+    if limits is None:
+        parts.append("cgroup=unlimited")
+    else:
+        limit, current = limits
+        parts.append(
+            f"cgroup {current / 2**30:.1f}/{limit / 2**30:.1f} GiB "
+            f"({100.0 * current / limit:.0f}%) headroom {(limit - current) / 2**30:.1f} GiB"
```
zfan3-sc, line 241: nit: guard against division by 0?

zfan3-sc (approved):

gigl/utils/host_memory.py
```diff
@@ -0,0 +1,293 @@
+"""How much host memory this process may still allocate.
+
+``psutil.virtual_memory().available`` reads ``/proc/meminfo``, which inside a container reports the
+HOST's memory rather than the container's limit, so a budget check trusting it alone can pass
+immediately before the container is killed. The cgroup knows the real limit; read both and believe
+the smaller.
```
kmontemayor2-sc, line 6: nit. Can we update this docstring for recommended usage of the module? The public functions are all at the bottom which can make it hard for human readers to discover

gigl/utils/host_memory.py
```diff
@@ -0,0 +1,293 @@
...
+    Returns the paths to try, most specific first, each expanded up its ancestors, ending at the
+    root.
+    """
+    paths: list[str] = []
+    try:
+        with open("/proc/self/cgroup") as handle:
+            for line in handle:
+                fields = line.strip().split(":", 2)
+                if len(fields) != 3:
+                    continue
```
kmontemayor2-sc, line 53: nit, can we leave the expected output here?

gigl/utils/host_memory.py
```diff
@@ -0,0 +1,293 @@
...
+    except OSError:
+        pass
+    paths.append("/")
+    # A container may be limited at an ancestor rather than at its own leaf, so walk upward.
+    expanded: list[str] = []
+    for path in paths:
+        parts = [part for part in path.split("/") if part]
+        while True:
+            candidate = "/" + "/".join(parts)
+            if candidate not in expanded:
```
kmontemayor2-sc, line 68: BTW do we call this often? Do we suspect that using a set here and t hen converting to list later (if needed) would be faster?

kmontemayor2-sc: On the perf note (and if if this isn't expensive it's ok to ignore), can we do the following:

1. have `parts: tuple[str, ...] = tuple(part for part in path.split("/") if part)
2. `for i in range(len(parts)`
3. if `parts[:len(parts-i) in ... `
4. Add the sub-tuple to expanded (as a set)
5. Covert all to str at the end.

dsaini2-sc: Measured it: the full call is a few tens of microseconds, dominated by the `/proc` read, and it's only called at phase boundaries -- never in a loop. The list also keeps the documented most-specific-first ordering, so leaving as is.

tests/unit/utils/host_memory_test.py
dsaini2-sc: Good catch -- moved the fixture helpers to a non-test base class, so the resolution tests no longer run twice under MemoryBreakdownTest.

gigl/utils/host_memory.py
dsaini2-sc: thanks -- added

gigl/utils/host_memory.py
dsaini2-sc: added

gigl/utils/host_memory.py
dsaini2-sc: Done -- a cgroup can report a 0 limit, so the log line now skips the percentage instead of dividing, and added a test covering it.

kmontemayor2-sc (approved):
Thanks!

PR #757, opened 2026-08-19
Add opt-in tensor spill to disk for colocated datasets

dsaini2-sc:
Adds an opt-in tensor-spill layer to `gigl/utils/share_memory.py`, plus its tests. Off by default: with `GIGL_TENSOR_SPILL_DIR` unset, nothing changes.

## Why

In colocated mode each rank's features and topology live in POSIX shared memory. `/dev/shm` is tmpfs, i.e. RAM, and without swap those pages cannot be evicted; worse, `share_memory_()` on an anonymous tensor **copies** it into `/dev/shm`, so the largest tensors transiently exist twice -- GLT's `Graph.__init__` shares the CSR unconditionally, and at billion-node scale that transient 2x is what OOM-kills the graph build. Spilling large tensors to a local disk turns their bytes into reclaimable page cache, letting large graphs run colocated without a separate storage tier.

## What it adds

- `prepare_spill_dir()` -- run-start cleanup of the spill directory. Cleanup never runs at exit, because the process that spills exits before the consumers unmap; a spill directory belongs to one workload and is reused across its sequential runs.
- `spill_tensor_to_disk` / `allocate_disk_backed` / `allocate_preshared` -- placement. `None` always means "fall back to memory". Sequential-write destinations go to disk (~3.1x measured write cost, bytes never anonymous); scattered destinations prefer memory, because an uncached 8-byte file write is a 4 KiB read-modify-write and a CSR build revisits the same pages hundreds of times. Block reservation via `posix_fallocate` is mandatory, so a filesystem that fills or cannot reserve fails as `OSError` at allocation, never as SIGBUS on a later page write.
- `SpilledTensorHandle` + `share_memory_for_ipc` / `resolve_spilled_handles` -- crossing process boundaries by descriptor. Pickling an mmap-backed tensor copies every byte back into `/dev/shm`, silently undoing the spill, so handles travel instead.
- `release_page_cache` / `release_page_cache_by_path` -- release the cgroup charge of written-once data (msync, then `MADV_DONTNEED`, then `FADV_DONTNEED`; skipping the middle step makes the call a silent no-op).
- Introspection: `is_tensor_spilling_enabled`, `is_disk_backed`, `disk_backed_handle`, `has_live_mapping`.

Placement decisions read `available_memory_bytes()` from #756.

## Behaviour change

None when spilling is unset. `share_memory()` keeps its `-> None` signature; its one change is leaving an already disk-backed tensor alone rather than copying it into `/dev/shm`, reachable only with spilling on.

## Testing

32 tests pass; `ty` and `ruff` clean. The tests exercise the real filesystem: spill/re-map round trips, threshold and dtype refusals, handle pickling across processes, forced `posix_fallocate` refusal (both "no space" and "not supported") asserting memory fallback with no leftover file, preshared allocation asserted on `data_ptr` stability across `share_memory_()`, and page-cache release asserted on measured page residency (`mincore`) rather than return values.

## Open questions

- Module home: this extends `share_memory.py` because `share_memory()` itself must recognise disk-backed tensors; a separate module importing the registry check is workable if a ~1k-line utility file is unwanted.
- Env vars vs config plumbing: env vars are required at the `mp.spawn` boundary, but a config field mapped to them may be the friendlier surface.
- Defaults: the 2 GiB spill threshold and 16 GiB scattered-write reserve are the values validated at ~1B-node scale.
- Should the final direct-to-shared-memory allocation also require cgroup headroom, failing fast instead of risking a later OOM?

### DIFF
--- a/gigl/utils/share_memory.py
+++ b/gigl/utils/share_memory.py
@@ -1,10 +1,938 @@
+import errno
+import mmap
+import os
+import tempfile
+import weakref
 from collections import abc
-from typing import Optional, TypeVar, Union
+from dataclasses import dataclass
+from typing import Any, Optional, TypeVar, Union, cast
 
 import torch
 from graphlearn_torch.partition import PartitionBook, RangePartitionBook
 
+from gigl.common.logger import Logger
+from gigl.utils.host_memory import available_memory_bytes
+
+logger = Logger()
+
 _KeyType = TypeVar("_KeyType")  # Generic Key Type
+
+# Directory for spilling large tensors to disk instead of POSIX shared memory. When unset,
+# behaviour is exactly as before (everything goes to /dev/shm).
+#
+# Read from the ENVIRONMENT rather than passed as an argument on purpose: the dataset is
+# built inside `mp.spawn` (gigl/distributed/dataset_factory.py:468), which starts a fresh
+# interpreter. Module-level state set by a parent process does NOT survive that, but the
+# environment does -- it is inherited by the spawned child. An earlier attempt to control
+# this by monkey-patching from the trainer process was a silent no-op for exactly this
+# reason.
+_SPILL_DIR_ENV = "GIGL_TENSOR_SPILL_DIR"
+_SPILL_MIN_BYTES_ENV = "GIGL_TENSOR_SPILL_MIN_BYTES"
+# Set by whichever process cleans the spill directory, so its descendants know not to.
+_SPILL_PREPARED_ENV = "GIGL_TENSOR_SPILL_DIR_PREPARED"
+_DEFAULT_SPILL_MIN_BYTES = 2 * 2**30  # 2 GiB
+# Headroom left on the shared-memory mount when preallocating there. Other tensors, torch's own
+# queues and the sampling workers' channels all share it, and a shared storage that overruns the
+# mount takes SIGBUS on write rather than failing at allocation.
+_SHARED_MEMORY_RESERVE_BYTES = 2 * 2**30  # 2 GiB
+# Cgroup headroom to leave free when a scattered destination is placed in memory. Only what is
+# allocated AFTER this check runs belongs here -- at the measured worst case (the largest CSC of
+# a ~1B-node graph): the direct scatter's full-size cursor clone (indptr[:-1], 7.2 GiB anonymous)
+# and ~1 GiB of chunk/sort transients, ~8.2 GiB, so 16 leaves ~7.8 GiB of allocator/race margin.
+# The freshly written indptr's 7.2 GiB of dirty pages are NOT itemised: they are already charged
+# when this check reads the headroom. NOT covered: the next edge type; its own allocation re-runs
+# this check against whatever headroom is left, and falls back to the banded disk path if the
+# answer is no
+_DEFAULT_RANDOM_ACCESS_RESERVE_BYTES = 16 * 2**30  # 16 GiB
+
+_NUMPY_DTYPES = {
+    torch.float32: "float32",
+    torch.float16: "float16",
+    torch.int64: "int64",
+    torch.int32: "int32",
+    torch.uint8: "uint8",
+}
+
+
+def prepare_spill_dir() -> None:
+    """Clear leftover spill files, ONCE per run, before anything in the run can spill.
+
+    Cleanup is done at the START of a run, not with ``atexit``. The tensors are spilled inside the
+    ``mp.spawn`` dataset-building child (``dataset_factory._build_dataset_process``), and that
+    child **exits before the trainer uses the dataset** -- so unlinking on its exit would delete
+    files the trainer and its sampling workers still have mapped by path. Removing stale files up
+    front bounds disk use without that hazard.
+
+    Exactly one process may do this, and it must be a process that starts before any spilling
+    one. Per-process age-based cleanup is NOT a substitute and is actively wrong: with sequential
+    loading, the edge child spills and exits, then the node child starts, sees the edge files as
+    older than itself, and deletes files the parent has not mapped yet. The marker below is an
+    environment variable precisely because ``spawn`` children inherit the environment, so a child
+    can tell that its parent already did this.
+
+    Idempotent, and safe to call when spilling is disabled.
+    """
+    spill_dir = _spill_dir()
+    if spill_dir is None:
+        return
+    if os.environ.get(_SPILL_PREPARED_ENV) == "1":
+        logger.info(
+            f"share_memory: {spill_dir} was already prepared by an ancestor process; "
+            f"leaving its contents alone"
+        )
+        _cleared_spill_dirs.add(spill_dir)
+        return
+    _clear_spill_dir(spill_dir)
+    # Inherited by every process spawned from here on, which is how they know to skip cleanup.
+    os.environ[_SPILL_PREPARED_ENV] = "1"
+
+
+def _clear_spill_dir(spill_dir: str) -> None:
+    """Unlink every spill file this module could have written under ``spill_dir``."""
+    if spill_dir in _cleared_spill_dirs:
+        return
+    _cleared_spill_dirs.add(spill_dir)
+    removed = 0
+    freed = 0
+    try:
+        for entry in os.scandir(spill_dir):
+            if not (
+                entry.name.startswith(_SPILL_PREFIX) and entry.name.endswith(".bin")
+            ):
+                continue
+            try:
+                freed += entry.stat().st_size
+                os.unlink(entry.path)
+                removed += 1
+            except OSError:
+                pass
+    except FileNotFoundError:
+        return
+    if removed:
+        logger.info(
+            f"share_memory: removed {removed} stale spill file(s) from {spill_dir} "
+            f"({freed / 2**30:.1f} GiB)"
+        )
+
+
+def _ensure_spill_dir_prepared(spill_dir: str) -> None:
+    """Called before writing a spill file.
+
+    If some ancestor called :func:`prepare_spill_dir`, do nothing -- deleting anything here risks
+    removing a sibling's live file. Otherwise this process is the first in its run to spill, so it
+    takes on the cleanup itself; that covers standalone use and tests, where no orchestrator ran.
+    """
+    if os.environ.get(_SPILL_PREPARED_ENV) == "1":
+        return
+    _clear_spill_dir(spill_dir)
+    os.environ[_SPILL_PREPARED_ENV] = "1"
+
+
+_cleared_spill_dirs: set[str] = set()
+_SPILL_PREFIX = "spill_"
+
+# (start address, byte length, weak reference to the backing array, descriptor) for every live spill
+# mapping in this process. See _register_mapping for why this cannot be read back off the tensor.
+_spilled_mappings: list[tuple[int, int, "weakref.ref[Any]", "SpilledTensorHandle"]] = []
+
+
+def _spill_dir() -> Optional[str]:
+    d = os.environ.get(_SPILL_DIR_ENV)
+    return d or None
+
+
+def _env_bytes(name: str, default: int) -> int:
+    """A byte count from the environment, falling back to ``default`` when unset or unparseable."""
+    try:
+        return int(os.environ.get(name, default))
+    except ValueError:
+        return default
+
+
+def _spill_min_bytes() -> int:
+    return _env_bytes(_SPILL_MIN_BYTES_ENV, _DEFAULT_SPILL_MIN_BYTES)
+
+
+def is_tensor_spilling_enabled() -> bool:
+    """Whether ``GIGL_TENSOR_SPILL_DIR`` is set, i.e. large tensors should go to disk."""
+    return _spill_dir() is not None
+
+
+@dataclass(frozen=True)
+class SpilledTensorHandle:
+    """Where a spilled tensor lives, with no tensor attached, so it is cheap to pickle.
+
+    This is what crosses a process boundary in place of the tensor. Sending the tensor instead
+    silently undoes the spill: torch's multiprocessing reduction does not consider a numpy-owned
+    storage shareable, so pickling copies every byte into RAM-backed ``/dev/shm`` (measured: Shmem
+    +383 MiB on a 383 MiB tensor). Call :func:`load_spilled_tensor` on the far side.
+    """
+
+    path: str
+    dtype: torch.dtype
+    shape: tuple[int, ...]
+
+    def load(self) -> torch.Tensor:
+        """Re-map the tensor in the current process."""
+        return load_spilled_tensor(self.path, self.dtype, self.shape)
+
+
+@dataclass(frozen=True)
+class SpilledTensor:
+    """A tensor living in a file, plus everything needed to re-map it elsewhere.
+
+    ``tensor`` is an mmap view; ``handle`` is the part that is safe to send elsewhere.
+    """
+
+    tensor: torch.Tensor
+    path: str
+    dtype: torch.dtype
+    shape: tuple[int, ...]
+
+    @property
+    def handle(self) -> SpilledTensorHandle:
+        return SpilledTensorHandle(path=self.path, dtype=self.dtype, shape=self.shape)
+
+
+def load_spilled_tensor(
+    path: str, dtype: torch.dtype, shape: tuple[int, ...]
+) -> torch.Tensor:
+    """Re-map a spilled tensor in another process, without copying its bytes anywhere.
+
+    The counterpart to :func:`spill_tensor_to_disk`. Pass ``SpilledTensor``'s path, dtype and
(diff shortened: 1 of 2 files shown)


gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+    """Clear leftover spill files, ONCE per run, before anything in the run can spill.
+
+    Cleanup is done at the START of a run, not with ``atexit``. The tensors are spilled inside the
+    ``mp.spawn`` dataset-building child (``dataset_factory._build_dataset_process``), and that
+    child **exits before the trainer uses the dataset** -- so unlinking on its exit would delete
+    files the trainer and its sampling workers still have mapped by path. Removing stale files up
+    front bounds disk use without that hazard.
+
+    Exactly one process may do this, and it must be a process that starts before any spilling
+    one. Per-process age-based cleanup is NOT a substitute and is actively wrong: with sequential
```
kmontemayor2-sc: Why can't the spilling process do this?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+    def load(self) -> torch.Tensor:
+        """Re-map the tensor in the current process."""
+        return load_spilled_tensor(self.path, self.dtype, self.shape)
+
+
+@dataclass(frozen=True)
+class SpilledTensor:
+    """A tensor living in a file, plus everything needed to re-map it elsewhere.
+
+    ``tensor`` is an mmap view; ``handle`` is the part that is safe to send elsewhere.
```
kmontemayor2-sc: Should this be: "`SpilledTensor` is an mmap view, `SpilledTensorHandle` is ..."?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+    """Re-map a spilled tensor in another process, without copying its bytes anywhere.
+
+    The counterpart to :func:`spill_tensor_to_disk`. Pass ``SpilledTensor``'s path, dtype and
+    shape through whatever IPC channel is available and call this on the far side.
+
+    Raises:
+        ValueError: If ``dtype`` has no numpy equivalent, or the file is the wrong size for
+            ``shape`` -- which would otherwise be read as silently wrong data.
+    """
+    import numpy as np
```
kmontemayor2-sc: nit. should we put this import at the top of the file?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+        return None
+    _ensure_spill_dir_prepared(spill_dir)
+    return _spill_to_mmap(tensor, spill_dir)
+
+
+def _fadvise_dontneed(fd: int) -> None:
+    """``POSIX_FADV_DONTNEED`` over the whole file (length 0 means to end of file).
+
+    Raises ``OSError`` (ENOSYS) on platforms without ``posix_fadvise``, resolved with getattr
+    because the symbol does not exist off Linux, where the module must still import.
```
kmontemayor2-sc: Must the module import off linux?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+    it into ``/dev/shm`` -- measured at +16,128 kB for a 16 MiB tensor -- undoing the spill at a
+    point far from where it was made.
+
+    The reference to ``array`` is weak, so an entry disappears when the mapping does. Views of a
+    spilled tensor point inside the recorded range and are recognised too.
+    """
+    nbytes = tensor.numel() * tensor.element_size()
+    _spilled_mappings[:] = [
+        entry for entry in _spilled_mappings if entry[2]() is not None
+    ]
```
kmontemayor2-sc: Is this doing some "GC"?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+        _register_mapping(
+            array,
+            tensor,
+            SpilledTensorHandle(path=path, dtype=dtype, shape=tuple(shape)),
+        )
+        logger.info(
+            f"share_memory: allocated {n_bytes / 2**30:.1f} GiB buffer {tuple(shape)} {dtype} "
+            f"on disk at {path} instead of in anonymous memory"
+        )
+        return tensor
```
kmontemayor2-sc, line 379: Should all of this be in the same try block? Is this so we don't fail here if we can't create the tensor?

And is there a reason we'd return None here over just raising?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+
+    Why this helps: ``/dev/shm`` is tmpfs, i.e. RAM. Node features for a ~1B-node graph approach
+    a terabyte at fp32; even split across replicas that is the largest resident tensor, and
+    ``share_memory_()`` costs 1x in tmpfs plus a transient 2x in RSS. Backing the bytes with
+    a real file turns them into evictable page cache instead.
+
+    NOTE the mapping only stays file-backed within this process. Torch's multiprocessing reduction
+    copies a numpy-owned storage into shared memory when the tensor is pickled, so anything that
+    ships this tensor to another process undoes the spill. Ship ``SpilledTensor.path`` and call
+    :func:`load_spilled_tensor` on the far side instead.
```
kmontemayor2-sc: I wonder if there's some way around this? It'd be quite convenient if we could get torch to respect these file-backed tensors and not share them so we don't need to be careful later.

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+        writable = np.memmap(path, dtype=np_dtype, mode="r+", shape=tuple(tensor.shape))
+        # Copy in ~1 GiB slices so we never hold a second full-size buffer.
+        row_bytes = (
+            max(1, int(tensor[0].numel()) * tensor.element_size())
+            if tensor.dim() > 1
+            else tensor.element_size()
+        )
+        rows = max(1, int(2**30 // row_bytes))
+        for start in range(0, tensor.shape[0], rows):
+            writable[start : start + rows] = tensor[start : start + rows].numpy()
```
kmontemayor2-sc: QQ does `numpy()` copy memory?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+    Measured rather than inferred, because the obvious reading of the docs is wrong. Under
+    ``file_system`` torch calls ``_new_using_filename_cpu``, which sounds like an ordinary temporary
+    file and is not: the storage lands in ``/dev/shm`` (observed:
+    ``/dev/shm/torch_<pid>_<random>_0``, 16,000,064 bytes for a 16 MB tensor), and ``TMPDIR`` receives
+    only a 4 KiB ``torch-shm-dir-*`` directory holding the libshm manager's socket. Under
+    ``file_descriptor`` the segment is ``shm_open``ed and immediately unlinked, so nothing is visible
+    in the directory listing while it still consumes the mount's capacity. Returning ``TMPDIR``
+    for ``file_system`` would size the check against the wrong filesystem in precisely the case
+    the check exists for.
+    """
```
kmontemayor2-sc, line 477: This comment confuses me a bit - what's it's purpose?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+    return None
+
+
+def allocate_preshared(
+    shape: tuple[int, ...], dtype: torch.dtype, random_access: bool = False
+) -> torch.Tensor:
+    """Allocate a large tensor in its FINAL home, so ``share_memory_()`` cannot duplicate it.
+
+    GLT's ``Graph.__init__`` shares the topology unconditionally, and for an anonymous tensor that
+    copies every byte into ``/dev/shm`` -- the tensor exists twice during the copy, fatal for a
```
kmontemayor2-sc, line 551: We are planning on fixing this too, right?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+
+def _random_access_headroom_reserve_bytes() -> int:
+    """Cgroup headroom to leave free when choosing memory over disk for a scattered destination."""
+    return _env_bytes(
+        "GIGL_RANDOM_ACCESS_HEADROOM_RESERVE_BYTES",
+        _DEFAULT_RANDOM_ACCESS_RESERVE_BYTES,
+    )
+
+
+def _memory_is_the_better_home(n_bytes: int) -> Optional[str]:
```
kmontemayor2-sc, line 524: Is it simpler if this returns a bool?

gigl/utils/share_memory.py
```diff
@@ -1,11 +1,884 @@
...
+                    reserve_error.errno,
+                    f"{spill_dir} does not support reserving space ({reserve_error}); "
+                    f"refusing an unreserved {n_bytes / 2**30:.1f} GiB mapping",
+                ) from reserve_error
+            # ENOSPC, EDQUOT, EFBIG, EIO and anything else unexpected: the caller cleans up and
+            # uses memory instead
+            raise
+
+
+def allocate_disk_backed(
```
kmontemayor2-sc, line 324: Did we consider `torch.from_file(..., shared=True)` [1]?

It seems like if we use it we can simplify this code a lot (no need for `SpilledTensor`, or registry / weakrefs / etc). We can also override the ForklingPickler to have a special case for when `tensor.untyped_storage().filename is not None` (e.g. it's a memmapped file) and default to default torch behavior otherwise.




[1]: https://docs.pytorch.org/docs/2.13/generated/torch.from_file.html#torch-from-file

kmontemayor2-sc (commented):
Nice work Deepak! Thanks :)

Did we consider `torch.from_file`? See https://github.com/Snapchat/GiGL/pull/757/changes#r3814543741. I think that if that's feasible (and in my testing it seems like it should be, without adding additional mem overhead), then we should be able to simplify this a lot by using that API and not needing the registry or new dataclasses / handles / etc?

gigl/utils/share_memory.py
dsaini2-sc: There can be multiple spiller processes over a run, and a later one cannot tell whether existing `.bin` files are a sibling's live output from this run (the writer may have already exited, but its files get read later) or garbage from a previous run -- so cleanup has to happen once, before any spilling starts. Clarified in the docstring.

gigl/utils/share_memory.py
dsaini2-sc: These are the fields, not the classes -- `.tensor` (this dataclass's attribute) is the mmap view, and `.handle` is a property returning the `SpilledTensorHandle`. Added dots so they read as attributes.

gigl/utils/share_memory.py
dsaini2-sc: Yes -- moved to the top and dropped the function-local copies. Leftover habit from the standalone version; numpy is a hard dep here.

gigl/utils/share_memory.py
dsaini2-sc: Sort of -- but only of bookkeeping, not memory. The weakref means this registry never keeps a mapping alive; the mmap is freed by normal refcounting when the last tensor dies. This line just drops rows whose mapping is already gone, so the list doesn't grow forever and a recycled address range can't match a dead row.

gigl/utils/share_memory.py
dsaini2-sc: The utility has to support macOS as well(darwin markers in pyproject, `os.posix_fadvise` doesn't exist there); but but the utility still behaves correctly -- it refuses instead of crashing. Made explicit in the docstring.

gigl/utils/share_memory.py
dsaini2-sc: Measured this on torch 2.8 -- it's a real simplification, but a partial one, so I'd like to scope it as a follow-up rather than rework this PR mid-review.

What it buys: `untyped_storage().filename` can replace the registry/weakrefs for recognizing spilled tensors (views inherit it too). And you're right that with a `SpilledTensor(torch.Tensor)` subclass, the pickler override is scoped by exact type. One catch I found though: the subclass propagates through ops but the file-backing doesn't (`spilled + 1` is Spilled-typed with `filename=None`), so the reducer still has to branch on `filename` internally and fall back to torch's shipping for derived tensors. Handleable, just not free.

What stays either way, for both `from_file` and `np.memmap`:
1. Both create the file sparse -- probed: a 4 MiB `from_file` tensor had 4 KiB actually allocated. Holes are filled lazily on write, and if the disk fills, the write dies as SIGBUS (a signal, uncatchable, no traceback). We've seen exactly this failure shape in a run: a probe over-allocated a 64 MiB `/dev/shm` mount, the allocation succeeded, and the first write died as `Bus error (core dumped)`, exit 135, no traceback. So the `posix_fallocate` reservation stays regardless of which mapping API sits under it.

2. Sending a `from_file` tensor across processes is where it breaks down. Torch's shipping recipe for CPU tensors is "move the bytes into shared RAM and hand those over" -- probed: a from_file tensor either fails to pickle outright (default strategy) or gets its bytes copied into a fresh `/dev/shm` segment (file_system strategy). Either way the spill is undone, so the custom-pickling override in your proposal would be required to do all the critial work.

Net: registry/weakrefs and the explicit handle can go; the reservation, the spill-dir lifecycle, and the page-cache release -- the bulk of this file -- remain. I'd land this PR as is and do the subclass rework as its own PR so it gets its own review -- ok?

gigl/utils/share_memory.py
dsaini2-sc: Yes to both, intentionally. The single try makes "create file, reserve blocks, map it" a transaction: fail at any step and the handler unlinks the partial file (which already holds real reserved disk blocks, so leaking it costs GiBs until the next run's sweep) and the caller gets RAM instead. And None over raising is the contract: disk-backing is an optimization -- the job runs fine with the buffer in RAM, just more exposed under memory pressure -- so failure degrades instead of killing, and callers fall back to `torch.empty` in one expression rather than each wrapping this in try/except. (Contrast `allocate_preshared` below, which does raise -- there the caller is about to feed an unguarded copy, so proceeding without a safe home means SIGBUS later.)

gigl/utils/share_memory.py
dsaini2-sc: There is -- it's the follow-up we just agreed on in the other thread. With `SpilledTensor` as a `torch.Tensor` subclass and a reducer registered for that exact type, pickling ships (path, dtype, shape) and re-maps on the other side, so sending one across processes keeps the spill and nobody has to be careful. This NOTE describes the world until that lands; I'll drop it in the follow-up.

gigl/utils/share_memory.py
dsaini2-sc: No -- `.numpy()` on a CPU tensor is a zero-copy view over the same memory (it raises rather than copies for the cases where a view is impossible, e.g. GPU or requires_grad). The only byte movement here is the assignment itself, streaming the slice straight into the file mapping, and the loop bounds that to ~1 GiB in flight at a time.

gigl/utils/share_memory.py
dsaini2-sc: Fair -- it gave the evidence without the question. Purpose: the free-space check below only works if it measures the mount torch actually puts shared storages on (over-allocating tmpfs succeeds, then SIGBUSes on first write). Torch's docs read as if `file_system` uses `TMPDIR`; measured, both strategies land in `/dev/shm` and `TMPDIR` only gets a tiny socket dir. The docstring records that so a future reader of the torch docs doesn't "fix" this back to `TMPDIR`. Added a lead sentence saying so.

gigl/utils/share_memory.py
dsaini2-sc: Yes, planned in a followup PR :)

gigl/utils/share_memory.py
dsaini2-sc: It'd be simpler in signature but it would cost the log line its content: the returned string is the reason, and the caller interpolates it into the placement warning ("must live on disk -- /dev/shm has X free, short of Y..."). That decision is the most consequential one a run makes (a silently disk-backed scatter destination looks exactly like a hung job -- we lost an hour to that once), so the why goes into the log verbatim.

gigl/utils/share_memory.py
dsaini2-sc: As discussed offline, `filename` replaces the registry/weakrefs, the handle dataclasses are gone (spill returns a plain tensor), and IPC is one reducer that ships file-backed tensors by path and delegates everything else to torch's own -- gated on the spill env var so the module changes nothing for processes that never opted in. Three guards the rework surfaced: a rebuild validates file size first (from_file recreates a missing file as zeros otherwise), quantized dtypes are refused (their quantizer can't be rebuilt from raw bytes), and named tensors delegate to torch's explicit refusal. Reservation, spill-dir lifecycle, and page-cache release are unchanged. Net -85 lines, tests 33 -> 38.

kmontemayor2-sc: Thanks!

kmontemayor2-sc (approved):
Thanks for the interations here!

yliu2-sc (approved):

PR #758, opened 2026-08-21
Migrate markdown formatting from mdformat to dprint

kmontemayor2-sc:
I found that dprint is much much faster (and supports more file types, we could do e.x. yaml formatting in the future too)

Replace mdformat/.mdformat.toml with dprint (dprint.json), matching the previous mdformat style as closely as dprint allows: lineWidth=120 with textWrap=always (mdformat wrap=120), asterisk emphasis, dash bullets, and table formatting built in. The mdformat exclusion list (MD_EXCLUDE_DIRS) moves into dprint.json excludes; the formatted markdown file set is byte-identical between the old find expression and dprint's resolution. dprint also picks up toml (pyproject reformats) and cmake (gigl-core CMakeLists reformats; cmake commands are case-insensitive).

Remaining one-time normalizations dprint applies that mdformat did not (no config knobs exist for these): thematic breaks ______ -> ---, task-list continuation lines aligned under content, blank lines around headings/lists, table column padding, and common-indent stripping inside code fences. Output is idempotent; make check_format_md passes.

The dprint CLI is provided by the dprint-py wheel (a thin PyPI repackaging of the official dprint binary) so the existing 'uv run' pattern keeps working.

**Scope of work done**


mkolodner-sc (approved):
Thanks Kyle!

mkolodner-sc (commented):
Just curious how much faster dprint is relative to mdformat?

yliu2-sc (approved):

kmontemayor2-sc:
> Just curious how much faster dprint is relative to mdformat?

I went from like 10s to 1s

PR #760, opened 2026-08-27
Make sampling-loader bring-up and steady-state RPC timeouts configurable

dsaini2-sc:
**Scope of work done**

Constructing a distributed loader is a cluster-wide rendezvous: every sampling worker on every rank must complete GLT's `init_rpc`, which gathers all `num_workers x world_size` workers, and only then does each rank's parent pass its local init barrier. Nothing synchronizes ranks between the end of dataset building and the first loader construction, so any skew accumulated during partitioning or graph assembly is spent inside that gather -- against a hard-coded 600s tolerance, with the parent's barrier wait unbounded behind it.

This PR makes those tolerances configurable and bounds the wait, via three env knobs in `gigl/distributed/constants.py`:

- `GIGL_SAMPLING_RPC_TIMEOUT_SECONDS` (default 600, GLT's own default): the bring-up gather tolerance, consumed at the `rpc_timeout` site in `base_dist_loader.py`. Large colocated topologies whose ranks legitimately arrive more than 600s apart get a knob instead of a dead job.
- `GIGL_SAMPLING_STEADY_STATE_RPC_TIMEOUT_SECONDS` (default 600): GLT's `rpc_timeout` is not init-only -- it becomes the agent's default request timeout for every steady-state sampling and feature-collection RPC, so raising it for bring-up would also make a genuinely stuck steady-state request take that long to surface. Each worker therefore narrows the agent's timeout to this value immediately after passing its init barrier, exactly the boundary between the two regimes. With both values equal the reset is skipped. (`torch.distributed.rpc._set_rpc_timeout` is the same internal torch uses to implement `rpc_sync`'s `timeout` argument; there is no public per-agent setter after `init_rpc`, and per-call timeouts are unavailable because the steady-state calls are made inside GLT.)
- `GIGL_SAMPLING_WORKER_INIT_TIMEOUT_SECONDS` (default 1800): the parent's `barrier.wait()` becomes bounded. **This is the one behavior change without any configuration:** a worker that loses the gather dies before the barrier, and today its parent blocks forever while healthy peers sit in collectives until the training process group timeout. Such a job now fails in 30 minutes with a `RuntimeError` naming the dead workers and their exit codes. The default deliberately sits above the gather tolerance so the barrier reports the death rather than timing out first and blaming the wrong thing. The barrier is waited once with the full timeout, never polled -- a timed-out wait breaks a multiprocessing barrier permanently. On timeout the producer diagnoses first, then tears itself down via GLT's bounded `shutdown()`, then raises; worker-side sampler/RPC teardown moved into a `finally` so a broken barrier cannot leak either.

Invalid env values (non-integer, non-positive) warn and fall back to the default.

***Where is the documentation for this feature?***: docstrings and rationale comments in `gigl/distributed/constants.py` and `dist_sampling_producer.py`, plus the CHANGELOG entry

***Did you add automated tests or write a test plan?***

Yes: new `tests/unit/distributed/sampling_timeouts_test.py` (8 tests: env parsing with defaults/overrides/invalid fallback; the narrowing helper at the torch RPC boundary -- mismatch narrows, equality is a no-op, override applies; the barrier bound against a real multiprocessing barrier, including the raise naming a dead worker's exit code). The full `distributed_neighborloader_test.py` suite (36 tests, real spawned loaders exercising both barriers and the narrowing path) passes on this change.

### DIFF
--- a/gigl/distributed/constants.py
+++ b/gigl/distributed/constants.py
@@ -1,3 +1,10 @@
+import os
+from typing import Final
+
+from gigl.common.logger import Logger
+
+logger = Logger()
+
 # TODO (mkolodner-sc): Set these ports dynamically while ensuring no overlap
 # Ports for various purposes, we need to make sure they do not overlap.
 # Note that [master_port_for_inference, master_port_for_inference + num_inference_processes).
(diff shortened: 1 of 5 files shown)


zfan3-sc (approved):

kmontemayor2-sc:
> Nothing synchronizes ranks between the end of dataset building and the first loader construction

QQ Can't we add some barrier [here](https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/dataset_factory.py#L242-L247) and have the same result?

gigl/distributed/constants.py
```diff
@@ -1,7 +1,94 @@
...
+    except ValueError:
+        logger.warning(f"{name}={raw!r} is not an integer; using {default}")
+        return default
+    if parsed <= 0:
+        logger.warning(f"{name}={parsed} is not positive; using {default}")
+        return default
+    return parsed
+
+
+def sampling_rpc_timeout_seconds() -> int:
```
kmontemayor2-sc: nit can we name this `sampling_rpc_init_timeout_seconds`?

gigl/distributed/constants.py
```diff
@@ -1,7 +1,94 @@
...
+    "GIGL_SAMPLING_STEADY_STATE_RPC_TIMEOUT_SECONDS"
+)
+DEFAULT_SAMPLING_STEADY_STATE_RPC_TIMEOUT_SECONDS: Final[int] = 600
+
+SAMPLING_WORKER_INIT_TIMEOUT_ENV: Final[str] = (
+    "GIGL_SAMPLING_WORKER_INIT_TIMEOUT_SECONDS"
+)
+# Above the RPC tolerance: a worker that loses the gather is already dead, so the barrier should
+# outlive the gather and report that death rather than time out first and blame the wrong thing.
+DEFAULT_SAMPLING_WORKER_INIT_TIMEOUT_SECONDS: Final[int] = 1800
```
kmontemayor2-sc: I have a minor preference for not adding so many env vars if we don't need - it makes it harder to let people know which to use / tune (especially if we suspect they're all a bit interlinked).

I understand wanting the steady state sampling rpc to have a different timeout than the init ones, but if we have them both at 600 to start maybe it's fine?

One suggestion for DEFAULT_SAMPLING_WORKER_INIT_TIMEOUT_SECONDS is to just have `_wait_for_workers_at_barrier` use 2x (or 3x) `DEFAULT_SAMPLING_STEADY_STATE_RPC_TIMEOUT_SECONDS` and add a comment saying that we double it since we also need to wait for the inner one.

gigl/distributed/constants.py
```diff
@@ -1,7 +1,94 @@
...
+#
+# Two independent mechanisms bound that wait, and they must be set together:
+#   * the RPC gather tolerance (below) -- exceeded, a worker dies
+#   * ``DistSamplingProducer``'s init barrier -- unbounded, its parent would wedge forever
+# A dead worker whose parent wedges is the worst case: the healthy peers block in collectives
+# until the training process group's own timeout fires, which is now hours rather than minutes.
+# --------------------------------------------------------------------------------------------
+SAMPLING_RPC_TIMEOUT_ENV: Final[str] = "GIGL_SAMPLING_RPC_TIMEOUT_SECONDS"
+# 600 s is GLT's own default and is preserved so this change alters no existing behaviour.
+DEFAULT_SAMPLING_RPC_TIMEOUT_SECONDS: Final[int] = 600
```
kmontemayor2-sc: IMO fine to remove this?

dsaini2-sc:
> > Nothing synchronizes ranks between the end of dataset building and the first loader construction
> 
> QQ Can't we add some barrier [here](https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/dataset_factory.py#L242-L247) and have the same result?

There is already a barrier just after the mentioned [here](https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/dataset_factory.py#L242-L247). The point of this PR is to be more robust about drift that can happen after: caller code between `build_dataset()` and the loader constructor (checkpoint download/load, device setup), plus the worker spawn inside the loader itself.

gigl/distributed/constants.py
dsaini2-sc: If you mean the env var itself: this one is the keychange of the PR -- we would want custom tolerance as large colocated datasets demand more constriction time, and after your other comment it's now the only env var, with the steady-state reset and the worker-init bound both derived from it.

gigl/distributed/constants.py
dsaini2-sc: resolved

gigl/distributed/constants.py
dsaini2-sc: sounds good, done

gigl/distributed/constants.py
```diff
@@ -1,7 +1,85 @@
...
+# diagnosable stall for a slow one. So the worker resets the agent's timeout back to this value
+# (GLT's default) immediately after it passes the init barrier, which is exactly the boundary
+# between the two regimes. Deliberately not configurable: steady-state sampling keeps today's
+# behaviour no matter how far bring-up tolerance is raised.
+SAMPLING_STEADY_STATE_RPC_TIMEOUT_SECONDS: Final[int] = 600
+
+# The init barrier must outlive the RPC gather: a worker that loses the gather is already dead,
+# and the barrier should report that death rather than time out first and blame the wrong thing.
+# 3x leaves margin for post-gather worker setup, and 3 x 600 reproduces the previous 1800 s bound.
+SAMPLING_WORKER_INIT_TIMEOUT_MULTIPLIER: Final[int] = 3
```
kmontemayor2-sc, line 49: And reason to not just have the single RPC timeout if they're the same now? Do we have reasons to make them different?

If we have them the same we can also remove _narrow_rpc_timeout_after_init?

kmontemayor2-sc (approved):

gigl/distributed/constants.py
dsaini2-sc: One number can't serve both timeouts: Bring-up is a one-time cluster-wide gather that takes tens of minutes on large colocated datasets -- its timeout must be generous or slow-but-healthy jobs die. A steady-state sampling RPC healthily completes in milliseconds-to-seconds -- its timeout is pure failure-detection latency, and you want it small so a stuck request surfaces fast.

PR #761, opened 2026-08-27
Apply and verify local patches on the pinned GraphLearn-Torch build

dsaini2-sc:
**Scope of work done**

GiGL already builds GraphLearn-Torch from a pinned commit (`88ff111ac`) in `gigl/scripts/install_glt.sh`, precisely because needed fixes (GLT PRs 151/153/154) were merged upstream but never released. This PR extends that mechanism one step: local patch files applied on top of the pin before the wheel builds, plus a post-install gate proving the patches are live in the installed wheel.

Three patches in `gigl/scripts/patches/`, applied in order:

- `0001` (74 lines): replaces `at::_unique`'s distinct-count in `InitCPUGraphFromCSR` with an exact seen-bitmap count. `_unique` is sort-based and transiently allocates ~3x the size of `indices` (measured 3.00x at 200M edges), which OOM-kills CPU graph init at billion-edge scale. The bitmap is exact, two sequential passes, ~max_col/8 bytes; it falls back to `_unique` when the id domain is sparse relative to the edge count, rejects negative ids (which would index outside the bitmap), and rejects non-contiguous storage. Never worse than upstream.
- `0002` (491 lines): int32 CSR column-id support in the CPU samplers. When node ids fit in int32 the CSC column array halves. Exactly one of `col_idx_`/`col_idx32_` is set; samplers taught the layout sample byte-identically to int64 and still emit int64 ids; samplers not taught it (weighted, CUDA init) reject int32 loudly via `TORCH_CHECK` instead of reading a null pointer. An int64-only deployment is untouched.
- `0003` (131 lines): shared-memory queue unpin/cleanup on teardown -- the SysV segment is detached and removed (and `cudaHostUnregister` called for pinned mappings) from the shared_ptr deleter, so repeated loader construction/teardown cycles stop leaking segments and pinned registrations. An in-flight zero-copy message keeps its mapping alive until released.

The apply loop uses `patch -p1 --forward` with an explicit failure exit, so a re-run or an upstream change that absorbs a patch stops the build loudly rather than shipping an unpatched wheel. `gigl/scripts/verify_glt_patches.py` then runs after `pip install` and fails the build unless every patched behavior is observable in the installed wheel -- applying is a property of the source tree, shipping is a property of the compiled `.so`, and the gap between them (stale build dir, second wheel on the path) is silent.

`tests/unit/utils/glt_int32_indices_test.py` re-establishes the int32 parity guarantees against whatever wheel is installed; on a wheel without the patches every test skips (with a self-explaining skip reason) so existing CI stays green.

One pre-existing bug fixed along the way, because it undermines the gate: `post_install.py` returned `install_glt.sh`'s exit status from `main()` but never passed it to `sys.exit`, so when invoked as a file -- which is how image builds run it -- a failed GLT build/install/verification still exited 0 and produced a successful Docker layer. The `__main__` guard is now `sys.exit(main())`, and a `None` status (the runner's failure-to-execute path) maps to 1.

These patches carry no GiGL-side consumers yet: nothing in `gigl/` builds an int32 topology today. Landing the wheel change first keeps the follow-up PR (int32 edge-index plumbing) pure-Python and separately revertable.

Where is the documentation for this feature?: rationale comments in `install_glt.sh`, the patch files themselves, the `verify_glt_patches.py` module docstring, plus the CHANGELOG entry

Yes, and the change was exercised end-to-end locally: the three patches apply cleanly in sequence to a fresh clone of the pinned GLT commit; a `WITH_CUDA=ON` wheel built from that tree passes all 12 checks in `verify_glt_patches.py` (including the pinned-memory teardown path, on a T4 host); the new `glt_int32_indices_test.py` runs 11/11 green against the patched wheel and skips 11/11 against an unpatched one; the full `distributed_neighborloader_test.py` suite passes against the patched wheel, confirming the default int64 path is unchanged. The new `post_install_test.py` runs the real file in a subprocess with a shimmed `bash` and asserts the child's exit code (7 and 0) becomes the wrapper's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Apply local patches to the pinned GraphLearn-Torch build in `install_glt.sh` (`gigl/scripts/patches/`): an exact
+  bitmap distinct-count in CPU graph init replacing `at::_unique`'s ~3x transient allocation, int32 CSR column-id
+  support in the CPU samplers, and shared-memory queue unpin/cleanup on teardown. `verify_glt_patches.py` gates the
+  build on the patches being live in the installed wheel by @dsaini2 in https://github.com/Snapchat/GiGL/pull/PENDING
+
+### Fixed
+
+- `gigl/scripts/post_install.py` now propagates `install_glt.sh`'s exit status as its own process exit code; previously
+  a failed GLT build/install exited 0 when the file was invoked directly, as image builds do by @dsaini2 in
+  https://github.com/Snapchat/GiGL/pull/PENDING
+
 ### Removed
 
 - Remove the deprecated `RESOURCE_CONFIG_PATH` environment variable; use `GIGL_RESOURCE_CONFIG_URI` instead by
--- a/gigl/scripts/install_glt.sh
+++ b/gigl/scripts/install_glt.sh
@@ -44,11 +44,25 @@ then
     # * https://github.com/alibaba/graphlearn-for-pytorch/pull/153
     # * https://github.com/alibaba/graphlearn-for-pytorch/pull/151
     # Thus, checking out a specific commit instead of a tagged version.
+    # Resolve this script's directory BEFORE cd'ing, so the patch path survives the cd below.
+    GIGL_SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
     git clone https://github.com/alibaba/graphlearn-for-pytorch.git \
         && cd graphlearn-for-pytorch \
         && git checkout 88ff111ac0d9e45c6c9d2d18cfc5883dca07e9f9 \
         && git submodule update --init \
         && bash install_dependencies.sh
+    # Local patches applied on top of the pinned commit, in order. 0001 replaces the
+    # at::_unique distinct-count in InitCPUGraphFromCSR with an exact bitmap count: _unique's
+    # sort allocates ~3x the size of `indices` transiently (measured 3.00x at 200M edges), which
+    # is ~94 GiB at a 4.2B-edge partition and OOM-kills the graph init on hosts whose budget
+    # assumes init holds no copy of the topology. `patch` rather than `git apply` so failure
+    # modes are the familiar reject files; `--forward` + explicit failure so a re-run or an
+    # upstream change that absorbs the fix stops the build LOUDLY instead of shipping an
+    # unpatched wheel that dies 3 hours into a 16-GPU job.
+    for glt_patch in "${GIGL_SCRIPTS_DIR}"/patches/*.patch; do
+        echo "Applying ${glt_patch}"
+        patch -p1 --forward < "${glt_patch}" || { echo "FATAL: ${glt_patch} did not apply"; exit 1; }
+    done
     if has_cuda_driver;
     then
         echo "Will use CUDA for GLT..."
@@ -72,6 +86,13 @@ then
     uv pip install dist/*.whl \
         && cd .. \
         && rm -rf graphlearn-for-pytorch
+    # Applying a patch and SHIPPING it are different guarantees: the loop above proves the source
+    # tree changed, this proves the installed .so behaves. A stale build dir or a second wheel on
+    # the path would otherwise produce an image that silently OOMs (0001) or rejects the int32
+    # topology the trainer builds (0002), hours into a multi-GPU job.
+    echo "Verifying the GLT patches took effect in the installed wheel"
+    python "${GIGL_SCRIPTS_DIR}/verify_glt_patches.py" \
+        || { echo "FATAL: GLT patches did not take effect in the installed wheel"; exit 1; }
 else
     echo "Skipping install of GraphLearn-Torch on Mac"
 fi
--- a/gigl/scripts/patches/0001-glt-cpu-graph-col-count-bitmap.patch
+++ b/gigl/scripts/patches/0001-glt-cpu-graph-col-count-bitmap.patch
@@ -1,0 +1,74 @@
+--- a/graphlearn_torch/csrc/cpu/graph.cc
++++ b/graphlearn_torch/csrc/cpu/graph.cc
+@@ -16,6 +16,10 @@
+ #include "graphlearn_torch/include/graph.h"
+ #include "graphlearn_torch/include/common.h"
+
++#include <algorithm>
++#include <cstdint>
++#include <vector>
++
+ namespace graphlearn_torch {
+
+ void Graph::InitCPUGraphFromCSR(
+@@ -30,7 +34,59 @@
+   col_idx_ = indices.data_ptr<int64_t>();
+   row_count_ = indptr.size(0) - 1;
+   edge_count_ = indices.size(0);
+-  col_count_ = std::get<0>(at::_unique(indices)).size(0);
++  // Exact distinct-count of the column ids, replacing at::_unique. _unique is sort-based and
++  // allocates ~3x the size of `indices` in transient anonymous memory (measured 3.00x at 200M
++  // edges) -- ~94 GiB at a 4.2B-edge partition, which OOM-kills billion-edge CPU deployments
++  // whose host budget assumes graph init holds no second copy of the topology.
++  //
++  // A seen-bitmap over [0, max_col] gives the identical count in two sequential passes using
++  // max_col/8 + 1 bytes (~115 MiB at 1e9 nodes). It is only VALID for non-negative ids in
++  // contiguous storage, and only CHEAPER when the id domain is dense relative to the edge count.
++  // Both preconditions are checked rather than assumed: an unchecked negative id would cast to a
++  // huge uint64_t and write outside the bitmap, and a sparse-but-huge id would size the bitmap
++  // from the id domain instead of the edge count. The sparse case falls back to _unique, so this
++  // is never worse than upstream; the invalid cases raise instead of corrupting memory.
++  //
++  // Note on strides: `col_idx_` is flat storage and every sampler already reads it that way, so
++  // requiring contiguity here makes col_count_ agree with what sampling actually traverses --
++  // upstream's _unique(indices) followed the view's strides and could disagree with its own
++  // samplers.
++  TORCH_CHECK(indices.is_contiguous(),
++              "InitCPUGraphFromCSR requires contiguous indices: col_idx_ is read as flat "
++              "storage by the samplers, so a strided view would sample the wrong columns");
++  int64_t max_col = -1;
++  for (int64_t i = 0; i < edge_count_; ++i) {
++    const int64_t col = col_idx_[i];
++    TORCH_CHECK(col >= 0,
++                "InitCPUGraphFromCSR requires non-negative column ids (they are node ids), got ",
++                col, " at position ", i);
++    max_col = std::max(max_col, col);
++  }
++  if (max_col < 0) {
++    // No edges. at::_unique of an empty tensor is also empty.
++    col_count_ = 0;
++  } else {
++    const uint64_t bitmap_bytes = (static_cast<uint64_t>(max_col) >> 3) + 1;
++    if (bitmap_bytes >
++        static_cast<uint64_t>(edge_count_) * static_cast<uint64_t>(sizeof(int64_t))) {
++      // Domain so sparse that the bitmap would exceed what `indices` itself occupies; the sort
++      // is the cheaper peak here.
++      col_count_ = std::get<0>(at::_unique(indices)).size(0);
++    } else {
++      std::vector<uint8_t> seen(bitmap_bytes, 0);
++      int64_t distinct_cols = 0;
++      for (int64_t i = 0; i < edge_count_; ++i) {
++        const uint64_t col = static_cast<uint64_t>(col_idx_[i]);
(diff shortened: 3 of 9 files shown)


gigl/scripts/patches/0001-glt-cpu-graph-col-count-bitmap.patch
```diff
@@ -0,0 +1,74 @@
...
+ void Graph::InitCPUGraphFromCSR(
+@@ -30,7 +34,59 @@
+   col_idx_ = indices.data_ptr<int64_t>();
+   row_count_ = indptr.size(0) - 1;
+   edge_count_ = indices.size(0);
+-  col_count_ = std::get<0>(at::_unique(indices)).size(0);
++  // Exact distinct-count of the column ids, replacing at::_unique. _unique is sort-based and
++  // allocates ~3x the size of `indices` in transient anonymous memory (measured 3.00x at 200M
++  // edges) -- ~94 GiB at a 4.2B-edge partition, which OOM-kills billion-edge CPU deployments
++  // whose host budget assumes graph init holds no second copy of the topology.
```
kmontemayor2-sc, line 22: BTW, have we measured this change in isolation? I remember I did some experimentation here before and I found that it didn't really move the needle.

gigl/scripts/patches/0001-glt-cpu-graph-col-count-bitmap.patch
kmontemayor2-sc, line 1: Can we add some description at the top of each patch for what it's doing?

gigl/scripts/patches/0002-glt-int32-csr-indices.patch
```diff
@@ -0,0 +1,491 @@
...
+-  for (int64_t i = 0; i < edge_count_; ++i) {
+-    const int64_t col = col_idx_[i];
+-    TORCH_CHECK(col >= 0,
+-                "InitCPUGraphFromCSR requires non-negative column ids (they are node ids), got ",
+-                col, " at position ", i);
+-    max_col = std::max(max_col, col);
+-  }
+-  if (max_col < 0) {
+-    // No edges. at::_unique of an empty tensor is also empty.
+-    col_count_ = 0;
```
kmontemayor2-sc, line 147: Do you think it's preferable to have just one patch here? We're overriding the 0001 patch here right? 

Might make it more maintable / readable?

WDYT?

gigl/scripts/patches/0002-glt-int32-csr-indices.patch
```diff
@@ -0,0 +1,491 @@
...
++template <typename ColT>
++void UniformSampleImpl(const ColT* col_begin,
++                       const ColT* col_end,
++                       const int32_t req_num,
++                       int64_t* out_nbrs) {
++  // with replacement
++  const auto cap = col_end - col_begin;
++  if (req_num < cap) {
++    uint32_t seed = RandomSeedManager::getInstance().getSeed();
++    thread_local static std::mt19937 engine(seed);
```
kmontemayor2-sc, line 233: If we're doing this patch can we also apply https://github.com/alibaba/graphlearn-for-pytorch/pull/167? It's a pretty big sampling speedup.

gigl/scripts/install_glt.sh
```diff
@@ -44,11 +44,25 @@ then
...
+    # sort allocates ~3x the size of `indices` transiently (measured 3.00x at 200M edges), which
+    # is ~94 GiB at a 4.2B-edge partition and OOM-kills the graph init on hosts whose budget
+    # assumes init holds no copy of the topology. `patch` rather than `git apply` so failure
+    # modes are the familiar reject files; `--forward` + explicit failure so a re-run or an
+    # upstream change that absorbs the fix stops the build LOUDLY instead of shipping an
+    # unpatched wheel that dies 3 hours into a 16-GPU job.
+    for glt_patch in "${GIGL_SCRIPTS_DIR}"/patches/*.patch; do
+        echo "Applying ${glt_patch}"
+        patch -p1 --forward < "${glt_patch}" || { echo "FATAL: ${glt_patch} did not apply"; exit 1; }
+    done
```
kmontemayor2-sc, line 65: Can you add a TODO to migrate the patches to our own fork of GLT once we get that up?

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Apply local patches to the pinned GraphLearn-Torch build in `install_glt.sh` (`gigl/scripts/patches/`): an exact
+  bitmap distinct-count in CPU graph init replacing `at::_unique`'s ~3x transient allocation, int32 CSR column-id
+  support in the CPU samplers, and shared-memory queue unpin/cleanup on teardown. `verify_glt_patches.py` gates the
+  build on the patches being live in the installed wheel by @dsaini2 in https://github.com/Snapchat/GiGL/pull/PENDING
+
+### Fixed
+
+- `gigl/scripts/post_install.py` now propagates `install_glt.sh`'s exit status as its own process exit code; previously
+  a failed GLT build/install exited 0 when the file was invoked directly, as image builds do by @dsaini2 in
+  https://github.com/Snapchat/GiGL/pull/PENDING
```
kmontemayor2-sc, line 20: nit update pending?

gigl/scripts/verify_glt_patches.py
```diff
@@ -0,0 +1,285 @@
...
+        if leaked > 0
+        else f"20 cycles clean ({'pinned' if cuda_usable else 'unpinned; no CUDA device'})",
+    )
+
+    # Dequeue is zero-copy: the received tensors view the segment through ShmData. Destroying
+    # BOTH queue objects while the message is still held must leave the data readable -- this
+    # ordering is exactly why the unpin lives in the shared_ptr deleter, not in ~ShmQueue.
+    producer = pywrap.SampleQueue(8, 1 << 20)
+    producer.send({"ids": torch.arange(64, dtype=torch.int64)})
+    consumer = pickle.loads(pickle.dumps(producer))
```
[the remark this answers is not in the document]
dsaini2-sc: This is a false positive in this context: the pickle round-trip is same-process, on an object this script just constructed itself -- no external input is ever deserialized. It isn't incidental serialization either: pickling is how GLT hands a SampleQueue across processes (its shipped IPC path), and this check exists to exercise exactly that path -- the zero-copy shm mapping must survive queue teardown while a dequeued message is in flight. A JSON stand-in wouldn't test the behavior the patch changes. The script also only runs inside image builds, never on untrusted data.

gigl/scripts/patches/0002-glt-int32-csr-indices.patch
kmontemayor2-sc, line 1: There's a robot review that this is unreachable - I assume we're planning on wiring this in as a follow up ?

```
GLT's public
  Topology.__init__/Dataset.init_graph upcast edge indices to int64 (data.type(torch.int64) in GLT's python/data/graph.py:53), and
  DistDataset._initialize_graph() goes through that path. Only the tests and verify_glt_patches.py reach the int32 C++ code, via
  Topology.__new__. So today, production training silently gets zero benefit from patch 0002. Codex confirmed this empirically through the
  installed wheel. Action: if wiring int32 topology into DistDataset is a planned follow-up, say so in the CHANGELOG/patch header; otherwise this
  patch is dead weight. When you do wire it, add a test that goes through the real Dataset.init_graph path — the silent upcast means a mis-wired
  integration would pass every current gate.
```

kmontemayor2-sc (commented):
Thanks Deepak!

I would prefer if we can leave some TODOs here so we can migrate to our own fork of GLT once that's up

gigl/scripts/patches/0001-glt-cpu-graph-col-count-bitmap.patch
kmontemayor2-sc: To be clear, I meant "in isolation, at prod scale"

zfan3-sc (approved):
Giving a 2nd stamp to unblock merging; Please merge after addressing Kyle's comments. Thanks for the work!

dsaini2-sc:
> Thanks Deepak!
> 
> I would prefer if we can leave some TODOs here so we can migrate to our own fork of GLT once that's up

Added TODOs -- at the apply loop in install_glt.sh, and in verify_glt_patches.py

gigl/scripts/patches/0001-glt-cpu-graph-col-count-bitmap.patch
dsaini2-sc: Measured the mechanism in isolation: at 200M edges, `at::_unique`'s transient alloc during graph init is 3.00x the indices bytes; the bitmap is ~max_col/8 bytes.

This only lowers peak RSS during init. It matters only when the host is provisioned near the dataset size; there the unpatched init OOMs, which is why an isolated prod-scale A/B would just reproduce the OOM.

gigl/scripts/patches/0001-glt-cpu-graph-col-count-bitmap.patch
dsaini2-sc: Added a description block at the top of each patch (patch ignores text before the first ---)

gigl/scripts/patches/0002-glt-int32-csr-indices.patch
dsaini2-sc: I'd keep them separate. 0002 doesn't override 0001 -- it extends it (same region, so the hunks stack textually, but nothing is reverted)

gigl/scripts/patches/0002-glt-int32-csr-indices.patch
dsaini2-sc: Folded it in: 0002 already replaces both UniformSample bodies with dtype-generic helpers, so the one-time-seed construction from GLT#167 now lives there -- both int64 and int32 paths get it.

gigl/scripts/install_glt.sh
dsaini2-sc: Done

CHANGELOG.md
dsaini2-sc: Done, set to #761.

gigl/scripts/patches/0001-glt-csr-col-count-and-int32-indices.patch
dsaini2-sc: Yes -- the follow-up PR wires it: int32 edge-index plumbing on the GiGL side (dtype selection + building the topology without the int64 upcast), pure Python since the wheel support lands here first. That PR will also carry the robot's suggested test through the real Dataset.init_graph path -- it can only run once the plumbing exists. Until then the patch is inert by design: int64 deployments are untouched.
We plan on this python side change in the next PR.

gigl/scripts/patches/0002-glt-int32-csr-indices.patch
mkolodner-sc: I'd also have a slight preference here as well for handling these as one patch. The duplicity between the two here introduces potential conflict in the future if we ever modify patch 0001 for example without also updating 0002, which I can imagine even an agent missing in some cases. Having one source of truth here and a single CSR optimization patch file will make things a bit easier to manage in the future if we need to apply additional patches on top of this.

Is there a reason we might prefer to keep these separate otherwise?

mkolodner-sc (commented):
Thanks a lot Deepak! Generally this LGTM, I share one concern with @kmontemayor2-sc about unifying the 0001 and 0002 patches if possible to make future maintenance easier. If possible would appreciate this being addressed, should be good to stamp otherwise :)

gigl/scripts/verify_glt_patches.py
```diff
@@ -0,0 +1,289 @@
...
+import torch
+from graphlearn_torch.data import Graph, Topology
+
+# 0001 replaces at::_unique with a bitmap count; a column id beyond the id domain the bitmap can
+# describe must fall back rather than allocate, and a negative id must raise rather than write out
+# of bounds. Both are patch-only behaviours: upstream accepts them silently.
+_NEGATIVE_COLUMN_ID = -1
+
+
+def _cpu_graph(indptr: torch.Tensor, indices: torch.Tensor) -> Graph:
```
mkolodner-sc: nit: this is similar to what we have in `glt_int32_indices_test.py` for `_build_cpu_graph`, would be nice if we could create a shared utility for this and leverage that in both locations instead

dsaini2-sc:
> about unifying the 0001 and 0002 patches if possible to make future maintenance easier

@mkolodner-sc, Agreed -- merged them: the int32 parts carried the bitmap patch's lines as context, so they weren't independently editable anyway. One CSR patch file now (header explains the merge); kept 0003 separate since it touches disjoint files.

mkolodner-sc (approved):
A nit but should we rename the 0003 patch to 0002 for consistency?

Otherwise LGTM, thanks for the work here and investigation :)

dsaini2-sc:
> A nit but should we rename the 0003 patch to 0002 for consistency?
> 
> Otherwise LGTM, thanks for the work here and investigation :)

Done in the fixture `build_cpu_csr_graph` in `verify_glt_patches.py` and the test imports it. Put it there rather than in the test suite since that module has to stay importable during an image build, before test deps exist.

gigl/scripts/verify_glt_patches.py
mkolodner-sc: Maybe we could add this function to 

https://github.com/Snapchat/GiGL/blob/main/tests/test_assets/distributed/utils.py

since it might feel unusual to import a function from `gigl.scripts`, WDYT?

dsaini2-sc: sg, done!

PR #767, opened 2026-09-09
Add ppr_relation_features token input to the PPR Graph Transformer path

mkolodner-sc:
## What
Adds a reserved `ppr_relation_features` anchor-relative token input to the PPR
Graph Transformer path, and consolidates the reserved-feature validation shared
by the transform and the encoder.

## Details
- `transforms/graph_transformer.py`
  - New `PPR_RELATION_FEATURES_NAME` token input: the columns of the PPR
    `edge_attr` beyond the scalar weight (col 0) are gathered per (anchor,
    neighbor) into a `(batch, max_seq_len, k)` feature tensor in
    `_build_sequence_layout_from_ppr_edges` and exposed through the anchor
    feature composers.
  - Neighbors stay ordered by descending PPR weight; relation-feature rows are
    reordered by the same permutation so they remain aligned with their edges.
  - Relaxes `_validate_ppr_sequence_input` so the batch may also carry the
    original relation edge types (which feed the relation-message channel); only
    the `"ppr"` edges drive the sequence.
- New `_validate_reserved_anchor_feature_usage` helper centralizes the
  placement/method checks for `ppr_weight` and `ppr_relation_features`, replacing
  the duplicated blocks in `heterodata_to_graph_transformer_input` and
  `GraphTransformerEncoder`.

## Testing
- `ruff check` / `ruff format` clean; module compiles.
- Unit-level: validator rejects misuse (pairwise/scalar bias, non-ppr method);
  PPR sequences are weight-ordered with relation features aligned to their edges;
  truncation to `max_seq_len` keeps the top-weight neighbors.


yliu2-sc (commented):
Maybe some tests needs to be updated too
```
The new PPR validation now raises “requires at least one PPR edge type” when there are no PPR stores, but tests/unit/transforms/graph_transformer_test.py:575 still expects “contain only PPR edges”. That should fail the transform test suite. The fix is to update/remove that old assertion and add coverage for the new mixed PPR + sampled-edge behavior and ppr_relation_features.
```

gigl/transforms/graph_transformer.py
```diff
@@ -300,12 +339,23 @@ def heterodata_to_graph_transformer_input(
...
+        # Build the PPR sequence from PPR edges ONLY. When include_sampled_edges is
+        # on, the batch also carries the original relation edge types; those are
+        # consumed separately by _lookup_pairwise_relation_indices for the
+        # relation-message channel. Homogenizing the full batch would pad the
+        # original edges with zero-weight edge_attr and leak them into the
+        # sequence, so restrict to the "ppr" edge types here. Node set/order is
+        # unchanged (edge_type_subgraph keeps all nodes), so anchor_indices and
+        # num_nodes computed from homo_data stay valid.
+        ppr_edge_types = [et for et in data.edge_types if et[1] == "ppr"]
+        ppr_homo_data = data.edge_type_subgraph(ppr_edge_types).to_homogeneous()
```
yliu2-sc, line 351: codex: PPR-only homogenization can silently corrupt node indices. edge_type_subgraph() drops node types not participating in PPR edges, then to_homogeneous() renumbers the remaining types. The code still uses offsets from the full graph. I reproduced this with an original-only node type preceding user: a valid user -> item PPR edge disappeared, yielding only the anchor and valid_mask=[[True, False]]. The PPR graph must preserve every node store in the original order, or its indices must be remapped to the full graph.

yliu2-sc (approved):
thanks for the work!

svij-sc (approved):

PR #768, opened 2026-09-16
Build CSR/CSC with a memory-lean counting sort instead of coo_to_csr

dsaini2-sc:
**Scope of work done**

GiGL's graph build converts each edge type's COO to CSR/CSC through `graphlearn_torch.utils.coo_to_csr`, which delegates to `torch_sparse.SparseStorage`. That holds seven full-size int64 arrays at the peak -- row, col, a composite sort key, a permutation, `index_sort`'s discarded sorted values, and both gathered outputs -- measured at **7.25x one int64 array** on 400M edges. At billion-edge scale the conversion, not the resident graph, OOMs and it sits upstream of every other memory lever: feature spilling, dropping edge ids and narrowing dtypes cannot help with it.

This PR adds a replacement for that conversion and uses it on the dataset's graph-build path.

- `gigl/utils/csr.py` (new): `build_csr_from_coo`, a two-pass counting sort. Degrees give the exact output layout up front, so the output is allocated once and written in place, and the input is consumed in chunks, which bounds transients by chunk size rather than by edge count. Peak becomes `row + col + indices + 2 x (num_rows + 1) + O(chunk + max_degree)`, about 2.0x one array unless a single row is larger than a sort block. `indptr` is always int64; `indices` narrows to int32 when the installed graphlearn_torch can read it and every column id is verified to fit.
- `dist_dataset.py`: `_initialize_graph` builds one `Graph` per edge type, largest first and releasing each COO before starting the next, when there are no edge ids and no edge weights. Anything carrying either falls back to GLT's `init_graph` untouched -- permuting ids and weights alongside the columns is exactly the expensive part. `_build_topology_without_edge_ids` populates a `Topology` directly, since `Topology.__init__` fabricates `torch.arange(num_edges)` edge ids whenever they are absent and hands them to the conversion to be permuted: two full int64 arrays indexing nothing. `init_graph` also casts the whole edge-index dict to int64 before its per-edge-type loop, so an int32 input would exist in both widths for every edge type at once; the per-edge-type path avoids that too.

A spilled CSR needs no special `Topology` subclass to survive the boundaries GLT puts it through: `Graph.__init__`'s `share_memory_()` is a no-op on a `from_file(shared=True)` storage, and `Graph.share_ipc` pickles through `ForkingPickler`, where `gigl.utils.share_memory`'s registered reducer ships the path rather than the bytes. A plain `Topology` holding a disk-backed CSR returned from a spawned child process still mapped and with correct values.

***Where is the documentation for this feature?*** module and function docstrings in `gigl/utils/csr.py` and the new helpers in `dist_dataset.py`, plus the CHANGELOG entry

***Did you add automated tests or write a test plan?***

Yes. `tests/unit/utils/csr_test.py` (34 tests) compares against `coo_to_csr` directly rather than against hand-written expectations, since the consumer is a compiled extension that segfaults rather than raises on a malformed topology: dense and mostly-empty inputs, a row larger than a sort block, duplicate edges, id ranges far beyond the edge count, CSC as CSR-of-the-transpose, chunk sizes coprime with the row count so rows split across chunks, the empty-edge-type case, the error paths, when the int32 narrowing does and does not happen, and agreement between the banded and direct scatters. `tests/unit/distributed/compact_topology_test.py` (14 tests) covers the topology bypass, the fallback gating, the row-dimension sizing, and sampled-neighbourhood parity against a `Graph` built by GLT's own `Topology` for both layouts -- at full fanout, so the comparison is exact rather than distributional. `distributed_neighborloader_test.py`, which spawns real loaders through this path, passes in full on the change: 36 tests, one pre-existing skip.


gigl/distributed/dist_dataset.py
```diff
@@ -720,6 +848,40 @@ def _initialize_graph(
...
+            edge_features_registered=edge_features_registered,
+        ):
+            if isinstance(partitioned_edge_index, Mapping):
+                assert isinstance(edge_index, dict)
+                streaming_input = edge_index
+                # Drop the GraphPartitionData wrappers so this dict is the only remaining
+                # referrer to each COO tensor; otherwise popping an entry inside the build frees
+                # nothing and converting one edge type at a time buys nothing.
+                if isinstance(partitioned_edge_index, MutableMapping):
+                    partitioned_edge_index.clear()
```
zfan3-sc, line 863: Can we add a test to make sure the previous edge type's COO is completely gone before we process the next? We may have a potential dangling reference [here](https://github.com/Snapchat/GiGL/blob/c5a17243299ca05f06efae057a83879f56c035a4/gigl/distributed/dist_dataset.py#L1251)

gigl/utils/csr.py
```diff
@@ -0,0 +1,445 @@
...
+    ``sort_within_row=False`` to skip this pass entirely if that is not affordable -- GLT's
+    samplers do not require columns to be ordered within a row, only ``coo_to_csr`` parity does.
+    """
+    num_rows = indptr.numel() - 1
+    row_start = 0
+    while row_start < num_rows:
+        row_end = int(
+            torch.searchsorted(indptr, indptr[row_start] + block_edges).item()
+        )
+        row_end = min(max(row_end, row_start + 1), num_rows)
```
zfan3-sc, line 295: nit: Shall we also bound row_end by `row_start + max_rows_per_block`?

gigl/utils/csr.py
```diff
@@ -0,0 +1,445 @@
...
+        RuntimeError: If the compiled graph accepts int32 columns but samples them incorrectly.
+    """
+    try:
+        from graphlearn_torch import py_graphlearn_torch as pywrap
+        from graphlearn_torch.data import Graph, Topology
+    except ImportError as error:
+        logger.warning(
+            f"CSR indices will stay int64: graphlearn_torch is not importable ({error})"
+        )
+        return False
```
mkolodner-sc: Why do we need this here? We should expect graphlearn_torch to be importable at this point, right?

gigl/utils/csr.py
```diff
@@ -0,0 +1,445 @@
...
+    separating this expected rejection from a genuinely broken install.
+    """
+    if not isinstance(error, RuntimeError):
+        return False
+    message = str(error)
+    return "scalar type" in message and "Long" in message and "Int" in message
+
+
+@functools.lru_cache(maxsize=1)
+def glt_accepts_int32_indices() -> bool:
```
mkolodner-sc: Could we remove this function? Given that the patch changes have landed which means that used GLT versions should accept the int32 indices, it'd be great if we could simplify a lot of this somewhat hacky logic here and instead assume that int32 indices are available.

Keeping version-specific logic around like this will make our codebase increasingly harder to maintain.

gigl/distributed/dist_dataset.py
```diff
@@ -38,6 +40,125 @@
 _EntityType = TypeVar("_EntityType", NodeType, EdgeType)
 
 
+def _reference_topology_attributes() -> frozenset[str]:
```
mkolodner-sc: Hmm, a lot of the logic here to avoid having to call `Toplogy.__init__` feels a bit hacky. Generally, I'd prefer if we could avoid initializing toy/dummy tensors due to GLT constraints, as they hurt readability and long-term maintainability. 

Instead of this, can we define our own Topology subclass which doesn't call the GLT coo_to_csr automatically? This could simplify a lot of the work-around logic needed here.

gigl/distributed/dist_dataset.py
```diff
@@ -38,6 +40,125 @@
...
+    """
+    topology = Topology.__new__(Topology)
+    # For both layouts GLT stores the pointer array in _indptr and the neighbour list in _indices;
+    # only the argument order of its constructor differs.
+    topology._layout = layout
+    topology._indptr = indptr
+    topology._indices = indices
+    topology._edge_ids = None
+    topology._edge_weights = None
+    expected = _reference_topology_attributes()
```
mkolodner-sc: Generally, I'd prefer if we kept actual <-> expected comparisons scoped to test files, rather than in the core business logic.

gigl/distributed/dist_dataset.py
```diff
@@ -38,6 +40,125 @@
...
+    Bypassing ``__init__`` would silently miss a field if graphlearn_torch added one, and the
+    consumer is a compiled extension that segfaults rather than raises.
+    """
+    reference = Topology(
+        edge_index=torch.tensor([[0, 1], [1, 0]], dtype=torch.int64), layout="CSR"
+    )
+    return frozenset(reference.__dict__)
+
+
+def _can_build_topology_directly(
```
mkolodner-sc: Could we rename this function to be more grounded in whether it contains per-edge metadata? It's not immediately obvious from the name what conditions result in us being able to build topology directly, or what building topology directly necessarily entails.

gigl/distributed/dist_dataset.py
```diff
@@ -38,6 +40,125 @@
...
+        if any(ids is not None for ids in edge_ids.values()):
+            logger.info(
+                "Edge ids are materialized; using GLT's init_graph for the topology"
+            )
+            return False
+        return True
+    return edge_ids is None
+
+
+def _num_nodes_for_row_dimension(
```
mkolodner-sc: Is this function necessary? It seems like a fairly thin wrapper over the existing `get_total_ids` function and is only called once. Maybe we can just directly include this logic in `_build_graph_per_edge_type` to keep the amount of indirection a bit smaller here.

gigl/distributed/dist_dataset.py
```diff
@@ -38,6 +40,125 @@
...
+def _can_build_topology_directly(
+    edge_ids: Union[Optional[torch.Tensor], dict[EdgeType, Optional[torch.Tensor]]],
+    edge_weights: Optional[Union[torch.Tensor, dict[EdgeType, torch.Tensor]]],
+    edge_features_registered: bool,
+) -> bool:
+    """Whether the memory-lean per-edge-type graph build applies.
+
+    It handles topology only. Edge ids and edge weights have to be permuted alongside the columns
+    during the CSR conversion, which is the expensive part, so anything carrying them falls back
+    to ``glt.data.Dataset.init_graph``.
```
mkolodner-sc: For my learning, do you see a way for these learnings to be easily extendable to these use cases (edge weight, edge features) in the future?

gigl/distributed/dist_dataset.py
```diff
@@ -735,6 +897,87 @@ def _initialize_graph(
         else:
             logger.info("Initialized homogeneous graph to dataset")
```
mkolodner-sc: If we call partitioned_edge_index.clear(), does this logic also need to be updated?

gigl/utils/csr.py
```diff
@@ -0,0 +1,445 @@
...
+DEFAULT_SORT_BLOCK_EDGES = 1 << 25
+# Destination window for the banded scatter, used only when `indices` could not be placed in
+# memory. Small enough that a band's pages stay cached while being written, large enough that the
+# number of input passes stays low.
+DEFAULT_BAND_BYTES = 4 * 2**30
+
+_INT32_MAX_COLUMN_ID = 2**31 - 1
+
+
+def _is_int32_dtype_rejection(error: BaseException) -> bool:
```
mkolodner-sc: Similarly, would prefer to remove this versioning check functions if possible since GiGL should be in control of the version used here for consumers.

gigl/utils/csr.py
```diff
@@ -0,0 +1,445 @@
...
+        raise RuntimeError(
+            f"The installed graphlearn_torch accepts an int32 CSR column array but samples it "
+            f"incorrectly: expected {expected_neighbors} with counts {expected_counts}, got "
+            f"{neighbors.tolist()} with {counts.tolist()}. This wheel would return plausible but "
+            f"wrong neighbourhoods."
+        )
+    return True
+
+
+def _degrees(keys: torch.Tensor, num_rows: int, chunk_size: int) -> torch.Tensor:
```
mkolodner-sc: There is another function in GiGL which also computes degrees from an `indptr`: https://github.com/Snapchat/GiGL/blob/main/gigl/distributed/utils/degree.py#L168-L170. It might be worth renaming this to specify what it's being constructed from to distinguish this logic from that.

mkolodner-sc: Also nit: would prefer to avoid very generic names like "keys" if possible

gigl/utils/csr.py
```diff
@@ -0,0 +1,445 @@
...
+
+def _place_chunk(
+    row_chunk: torch.Tensor,
+    col_chunk: torch.Tensor,
+    cursor: torch.Tensor,
+    indices: torch.Tensor,
+) -> None:
+    """Write one chunk's columns to their CSR destinations and advance ``cursor``.
+
+    ``row_chunk`` must already be int64 and indexed relative to ``cursor``'s first row.
```
mkolodner-sc: Can we include descriptions for what each of the variables is doing here? It's not immediately obvious what role cursor and indices play here --  an example in the docstring could be helpful

mkolodner-sc (commented):
Thanks Deepak! Did an initial pass here and left some comments

Issue #408, opened 2025-12-03
Type fixes w/ #405


svij-sc:
W/ updating pytorch version we need to address some typecheck issues that are not addressed as part of this change

Issue #405, opened 2025-12-02
Update deps / python to latest supported version - use UV for dep management


svij-sc:
Python3.9 is reaching has reached end of life on October 31.
This poses some concerns incase there are any security issues w/ python or any of the deps we use as they will likely not publish any patches, etc.
So a need to update python version and deps exist - which is quite heavy handed change as we make use of libraries that are not all available on pypi.

1. Thus, the dependency management is complicated and involves us indexing wheels directly:
https://github.com/Snapchat/GiGL/blob/a48144bb1e1f677886e290056e0e6e6a988c95bb/python/pyproject.toml#L72-L83

2. Secondly, to be most secure, this involves us maintaining seperate frozen requirements for each host we support:
https://github.com/Snapchat/GiGL/tree/main/requirements

3. Finally, there are some other complications of trying to package and host wheels in corp/private registries: https://go/gigl-issue/1434

**Proposal to make this easier**
Taking these issues into account, we can make our dependencies easier to manage if we adopt `uv` as part of upgrading our deps / python version: https://docs.astral.sh/uv/

Specifically, the way it helps us is:
1. `uv` allows us to more easily maintain custom indexes for packages: https://docs.astral.sh/uv/concepts/indexes/
2. `uv` maintains a lockfile for us that contains information for all hosts / all dependency group resolutions: https://docs.astral.sh/uv/guides/projects/#uvlock
3. The customizability of custom auth paths/certs allows practitioners to more easy package host wheels in corp/private registries: https://docs.astral.sh/uv/concepts/authentication/

Beyond our issues above, `uv` further improves the developer experience by:
1. Installing and managing our python environment for us
2. It replaces our use of multiple other tools: pip, pip-tools, twine, conda, etc.
3. Its support for [workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/) allows us to manage GiGL as part of other larger codebases.
4. Container like sandboxing

svij-sc:
I have done some work to update us to UV.

**Helpful guidance:**
1. `uv` helps us manage our python environment, so Instead of doing `python -m ...`; we will need to do `uv run python -m ...`; subsequently tools like black can now be executed as `uv run black` and dont need to be installed at a global level or inside some unmanaged python env. As long as you are in the gigl directory, uv will know how to work.
2. We eliminate the need for using `pip` to manage/install deps; we also eliminate the need to maintain different requirements files, there is just one now `uv.lock`

**Pull requests:**
[...]

PR #769, opened 2026-09-18
Fix ty suppressions from #405 with real type fixes (#408)

lucianlavric:
## Summary

#405 introduced `ty: ignore` suppressions (tracked in #408) where the type fixes were not trivial. This removes every #408-linked suppression by fixing the underlying types instead: declared buffer types instead of `__getattr__` unions, fail-fast narrowing instead of ignores, and one real mismatch — `infer_task_inputs` declared `decoder: LinkPredictionDecoder` but assigns the bound `decode` method; it is now typed as the callable it actually is.

## Approach

- Declare attribute types for registered buffers (`LightGCN._layer_weights`, the early-stop test's `_DummyModel.foo`) so ty resolves them as `Tensor` rather than the `Tensor | Module` union from `nn.Module.__getattr__`.
- Make the `GnnModel` protocol `runtime_checkable` and narrow through it in `InferencerV1`; keep the dynamic `graph_backend` attach in the GraphSAGE spec explicit via `setattr`.
- Narrow with fail-fast asserts where the runtime type is guaranteed by construction: the model in `infer_task_inputs` (`LinkPredictionGNN`), and `ModuleDict` values in `NodeAnchorBasedLinkPredictionTasks._get_all_tasks`.
- `FeatureInteraction.reset_parameters`: bind via `getattr` so `callable()` narrows the optional method.
- KDD tutorial: `from torch.multiprocessing.spawn import spawn` — the dotted call resolves to the same-named submodule, which type checkers reject as non-callable.

## User-facing changes

None intended. Invalid model types now fail fast with descriptive assert messages instead of failing deeper in the call.

## Verification

- `make type_check`: no new diagnostics versus main (the same 48 pre-existing macOS-environment diagnostics appear on both — uninstallable-on-macOS imports and two Linux-only `os`/`psutil` APIs).
- `make check_format_py`: clean.
- Full `make unit_test_py` is not runnable on macOS (`graphlearn_torch` ships in the Docker/dev images only; the darwin env lacks `tensorflow_metadata`/`torchrec`), unchanged from main — relying on PR CI for the test gate.
- Runtime patterns validated standalone with pure torch: buffer registration and `+=` behavior unchanged with the declared annotation; `runtime_checkable` `GnnModel` isinstance correctly passes after `setattr` and fails on modules without the attribute; the explicitly imported `spawn` is callable.


gigl/src/common/modeling_task_specs/utils/infer.py
```diff
@@ -1,14 +1,14 @@
...
 
 import torch
 import torch.nn as nn
 from jaxtyping import Float, Int64
 from torch_geometric.data import Data
 from torch_geometric.data.hetero_data import HeteroData
 
-from gigl.src.common.models.layers.decoder import LinkPredictionDecoder
 from gigl.src.common.models.layers.loss import ModelResultType
+from gigl.src.common.models.pyg.link_prediction import LinkPredictionGNN
```
kmontemayor2-sc: This import introduces a circular dependency:

```
utils/infer.py
  → models/pyg/link_prediction.py
  → models/layers/task.py
  → utils/infer.py
```

The same import succeeds at the PR’s base commit. This blocks NodeAnchorBasedLinkPredictionModelingTaskSpec from loading.
Please avoid importing the concrete model here at runtime. A small structural Protocol describing the required decode and tasks.result_types members would both avoid the cycle and preserve the existing model contract. A TYPE_CHECKING-only import plus a cast would also avoid the cycle, although it would not provide runtime validation.

gigl/src/common/modeling_task_specs/utils/infer.py
```diff
@@ -132,16 +132,19 @@ def infer_task_inputs(
...
-        decoder = model.decode  # https://github.com/Snapchat/GiGL/issues/408  # ty: ignore[invalid-assignment] TODO(ty-torch-union-inference): fix ty Tensor/Module union inference regressions.
-        batch_result_types = (
-            model.tasks.result_types  # ty: ignore[unresolved-attribute] TODO(ty-torch-union-inference): fix ty Tensor/Module union inference regressions.
-        )  # https://github.com/Snapchat/GiGL/issues/408  # ty: ignore[invalid-assignment] TODO(ty-torch-union-inference): fix ty Tensor/Module union inference regressions.
+    base_model = (
+        model.module
+        if isinstance(model, torch.nn.parallel.DistributedDataParallel)
+        else model
+    )
+    assert isinstance(base_model, LinkPredictionGNN), (
```
kmontemayor2-sc: This changes the function from accepting any compatible nn.Module to requiring one specific legacy LinkPredictionGNN class.
Previously, a custom model or wrapper worked as long as it exposed decode and tasks.result_types, which matches the function’s generic nn.Module | DDP signature. The new check rejects compatible custom implementations, and the required concrete class is itself marked deprecated.
Could we type this against a structural protocol instead? If runtime validation is required, please raise an explicit TypeError rather than using assert, since the assertion disappears under python -O.

gigl/src/common/modeling_task_specs/utils/infer.py
```diff
@@ -132,16 +132,19 @@ def infer_task_inputs(
     input_batch = InputBatch(main_batch=main_batch, random_neg_batch=random_neg_batch)
 
     batch_result_types: Set[ModelResultType]
-    decoder: LinkPredictionDecoder
+    # The model's bound decode method: (query_embeddings, candidate_embeddings) -> scores.
+    decoder: Callable[[torch.Tensor, torch.Tensor], torch.Tensor]
```
kmontemayor2-sc: This declaration is safe because decoder is assigned unconditionally below, but I would initialize it where it is declared rather than leave a declaration-only local:

```
decoder: Callable[
    [torch.Tensor, torch.Tensor], torch.Tensor
] = base_model.decode
```

Alternatively, if the narrowed model type exposes the correct signature, decoder = base_model.decode should be sufficient.
I would not add a None or no-op default here: the decoder is required and model-specific, so such a default would weaken the invariant and could hide an invalid model.

gigl/nn/models.py
```diff
@@ -185,6 +185,9 @@ def __init__(
                 )
 
         # Register layer weights as a buffer so it moves with the model to different devices
+        # Declared type lets ty resolve the buffer as a Tensor instead of the
+        # Tensor | Module union produced by nn.Module.__getattr__.
+        self._layer_weights: torch.Tensor
```
kmontemayor2-sc: I have a preference against declaration only statements like this, can we avoid these?

kmontemayor2-sc (changes requested):
Thanks for removing the ty suppressions, and sorry for the late review 😅 . I found one blocking runtime regression in the current implementation: the new LinkPredictionGNN import creates a circular dependency that prevents the node-anchor link-prediction modeling spec from importing.

